Issue #897, opened 2022-12-01
`driver` crate design


sistemd:
In this issue I would like to leave a high-level idea of what the design of the `driver` crate _could_ look like, just so I don't forget it. It's brainstormed and definitely not 100% complete and things will need to be changed on the fly, but hopefully it presents a good starting point. Below is the module structure and the types and responsibilities of each module.

- `api`: contains all the API [DTOs](https://en.wikipedia.org/wiki/Data_transfer_object) and the implementation of the REST API endpoints.
- `logic`: the core logic module
  - `solution`: `Solution`, `Auction`, `Order`, the logic for picking the best solution, the logic for scoring solutions (not exposed), `SolutionId`, `AuctionId`. Maybe can have a `competition` submodule, maybe not. Uses the external top-level `solver` crate.
  - `quote`: `Order`, `Quote`, the logic for quoting, building the single-trade `Auction` (hidden), solving it and building `Quote` (hidden). This module uses the `solution` module since quoting boils down to solving a special type of `Auction`.
  - `settlement`: `Settlement`, the logic for creating (encoding) the settlement and broadcasting it. `Publisher` trait for different "strategies" for publishing the settlement (flashbots, mempool, etc.).
  - `liquidity`: logic for liquidity collection. I still need to understand this better and scope it out properly.
- `solver`: the `Solver` type and its appropriate implementation in terms of the HTTP transport. This is a total external detail. It should be trivial to e.g. replace HTTP with RPC or offer both, if we need to in the future. The interface of this module should be tailored to the `logic` module. I think it's good enough to provide a `solve` method which accepts a `solution::Auction` + `solution::SolutionId` and returns a `solution::Solution`.
- `eth`: Ethereum network access layer, probably implemented via `web3`. Implements a trait if we want to mock it but I think we can get away without mocking this and just use a local `hardhat` instead :crossed_fingers:. Needs to expose a `publish` method (or similar) and possibly some other stuff needed by `liquidity`.
- `publisher`: implements the `settlement::Publisher` trait. Has submodules for each impl, e.g. `mempool`, `flashbots`, `gelato`, etc.

The important points are: `logic` is separated from externals (`api`, `solver`, `eth`). The main modules are modeled around the concepts of our domain (`solution`, `quote`, `settlement`, etc.), rather than processes or [transaction scripts](https://www.martinfowler.com/eaaCatalog/transactionScript.html).

This can probably be closed when we close https://github.com/cowprotocol/services/issues/892 or when it becomes outdated.

Issue #895, opened 2022-12-01
`/solve` endpoint for `driver`


sistemd:
The `/solve` endpoint is used to calculate the solution to an auction. This endpoint should calculate the objective value (score) for each solution and return it in the response.

#### Input

```js
{
  "id": u64, // The auction ID
  "block": u64, // The current block
  "orders": [Order],
  "prices": Map<token address, big number>,
}
```

Where `Order` is defined as all of the fields on [`Order`](https://github.com/cowprotocol/services/blob/ed3529408a2ba2a252da974bfc93fe70b8103298/crates/model/src/order.rs#L43), with only the following taken from `metadata`:

```js
  "full_fee_amount": big number, // For objective value (score) computation
  "is_mature": bool, // Used by the solvers
  "class": "market" | "limit" | "liquidity", // For objective value (score) computation
  "surplus_fee": big number, // For effective limit price computation
  "executed_amount": big number // For partially fillable orders
  //  The following two fields are needed for settlement encoding
  "interactions": [
    {
      "to": string, // Address to call
      "value": string, // Value to use for the EVM call
      "data": string, // Hex-encoded call data (for the eth_call RPC, this is the input field)
    }
  ],
  "accessList": [
    {
      "address": string, // The smart contract address
      "storage_keys": [string], // Addressed storage keys
    }
  ]
```

#### Output

The response from the `/solve` endpoint should be the objective value (score) of the best solution. If the solver lies about its score, its rewards will be slashed by the DAO.

```js
{
  "score": rational,
}
```

Related: https://github.com/cowprotocol/services/issues/894.

vkgnosis:
It's not clear to me how the id is chosen. I get the main point of the id idea. We can store it in the database before calling /execute so we can later associate the transaction with the auction.

That relies on IDs being unique. But if solvers choose the id then how can this be guaranteed? And how it can be guaranteed that another id wasn't randomly part of some calldata. The longer the id is the more expensive it is to include it.

MartinquaXD:
Why is it necessary that the id is already part of the `/solve` request? Wouldn't it be enough to generate a single random id in the `autopilot` when the winner is chosen and put that in the `/execute` request instead?

vkgnosis:
If autopilot can choose the id then we can even guarantee that it is unique. We can use an auto incrementing integer in Postgres like what we do for auction ids. Actually, we can just reuse the auction id.

vkgnosis:
[...]

PR #909, opened 2022-12-03
Driver `/solve` functionality part 1

sistemd:
Progress on https://github.com/cowprotocol/services/issues/895. Implements part of the design outlined in https://github.com/cowprotocol/services/issues/897.

Implements an interface for talking to the external solvers. Sketches out a skeleton for the core logic and some of the externals.

### Test Plan

Will write API-level tests later down the line.

### DIFF
--- a/crates/driver/src/solver/dto/auction.rs
+++ b/crates/driver/src/solver/dto/auction.rs
@@ -37,7 +37,6 @@ struct Order {
     cost: TokenAmount,
     is_liquidity_order: bool,
     is_mature: bool,
-    #[serde(default)]
     mandatory: bool,
     has_atomic_execution: bool,
     reward: f64,
--- a/crates/driver/src/solver/dto/solution.rs
+++ b/crates/driver/src/solver/dto/solution.rs
@@ -135,7 +135,6 @@ enum ExecutionPlan {
 struct Coordinates {
     sequence: u32,
     position: u32,
-    #[serde(default)]
     internal: bool,
 }
 


crates/driver/Cargo.toml
```diff
@@ -15,26 +15,34 @@ name = "driver"
 path = "src/main.rs"
 
 [dependencies]
+num = "0.4"
```
nlordell, line 18: I noticed at first that these don't use `{ workspace = true }` dependencies (which I personally like).

At first, I was going to suggest to change them back, but I realize that this is probably intentional as we intend to move it into a separate repo. Is this correct?

crates/driver/src/solver/dto/auction.rs
```diff
@@ -0,0 +1,73 @@
...
+    tokens: BTreeMap<H160, Token>,
+    orders: BTreeMap<usize, Order>,
+    amms: BTreeMap<H160, Amm>,
+    metadata: Option<Metadata>,
+}
+
+#[derive(Debug, Serialize)]
+struct Order {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    id: Option<String>,
```
nlordell, line 29: All orders should have a UID (note that this is different than the `LimitOrder::id` field that we used to have in the `solver` - as this represents a protocol order which has an actual mandatory and unique ID).

crates/driver/src/solver/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    pub async fn solve(&self, auction: Auction) -> Result<Solution, Error> {
+        // TODO Ask about all the `config` stuff in DefaultHttpSolverApi, what is every
+        // field for exactly?
+        let mut url = self.url.join("solve").unwrap();
+        let time_limit = auction.deadline.solver_time_limit()?;
+        url.query_pairs_mut()
+            .append_pair("auction_id", &auction.id.0.to_string())
+            .append_pair("instance_name", &self.instance_name(auction.id))
```
nlordell, line 59: For example, this is literally telling the solver what to name the instance file. I think you could get it do crazy things by specifying a path (`../../etc/passwd` 😛).

crates/driver/src/util/serialize.rs
```diff
@@ -0,0 +1,30 @@
+use serde::{Deserialize, Serialize};
+
+/// Serializes [`ethereum_types::U256`] as a decimal string.
```
nlordell, line 3: I saw these change to `serde_as` in the other PR (which I think is more correct).

nlordell (approved):
Code looks good.

I think there are a few assumptions related to https://github.com/cowprotocol/services/issues/895#issuecomment-1337498190 in this PR as well. I already brought this up in #910, and I think this PR is good WRT driver <> solver communication (which is its focus anyway).

crates/driver/src/solver/dto/solution.rs
```diff
@@ -0,0 +1,172 @@
...
+    Coordinates(Coordinates),
+    #[serde(deserialize_with = "execution_plan_internal")]
+    Internal,
+}
+
+#[derive(Debug, Deserialize)]
+struct Coordinates {
+    sequence: u32,
+    position: u32,
+    #[serde(default)]
```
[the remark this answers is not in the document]
nlordell: Nice catch!

crates/driver/src/solver/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+            "application/json".parse().unwrap(),
+        );
+        headers.insert(reqwest::header::ACCEPT, "application/json".parse().unwrap());
+        // TODO(#907) Also add an auth header
+        Self {
+            url,
+            client: reqwest::ClientBuilder::new()
+                .default_headers(headers)
+                .build()
+                .unwrap(),
```
MartinquaXD, line 46: The client should probably passed into the `Solver` to be able to configure things like timeouts at a centralized place.

PR #874, opened 2022-11-30
Model driver communication types in Rust

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/862 .

Models the types used by the communication between autopilot an driver.

These models could be shared between driver and autopilot but we wanted stronger separation so I've added them to autopilot instead. Can change this if requested.

### Test Plan

This code is not used yet. Not sure if we need serialization tests but if we do I want to add them after discussion.

PR #910, opened 2022-12-04
Driver `/solve` functionality part 2

sistemd:

Please review https://github.com/cowprotocol/services/pull/909 first.

Implements the `/solve` endpoint of the driver. Cleans up some of the serialization code a bit. There is still much "core logic" functionality to fill in, most notably the objective value calculation.

Also, with this PR I have to ask for some forgiveness: I went ahead and used `axum` instead of `warp`. If anybody strongly opposes this change, I will undo it and use `warp` instead. I'm aware there will be some metrics stuff that needs to be done over, however I was planning on revisiting some of the metrics stuff anyway, so I don't think there's much waste.

But more to the point, I'm honestly not too happy with `axum` either. There are also confusing errors (e.g. if you don't use `State` as the first parameter to your handler, you get a confusing error message; the signature of [`with_state`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.with_state) is ridiculous, unintuitive, and not type-safe; for some odd reason `with_state` _must_ be called _after_ registering all the handlers; getting the `RequestBodyLimitLayer` middleware to work is also not very intuitive since it relies on `tower` so that's an additional framework that the user has to understand). But I do think it's a fewfold improvement over `warp`.

### Test Plan

Will write API-level tests most likely in the next PR or the one right after.

### DIFF
--- a/crates/driver/src/api/solve/dto/auction.rs
+++ b/crates/driver/src/api/solve/dto/auction.rs
@@ -29,7 +29,7 @@ pub struct Auction {
 #[derive(Debug, Deserialize)]
 struct Order {
     #[serde_as(as = "serialize::Hex")]
-    uid: Vec<u8>,
+    uid: [u8; 56],
     from: H160,
     sell_token: H160,
     buy_token: H160,


crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -0,0 +1,137 @@
...
+    orders: Vec<Order>,
+    deadline: chrono::DateTime<chrono::Utc>,
+    #[serde_as(as = "HashMap<_, DisplayFromStr>")]
+    prices: HashMap<H160, BigUint>,
+}
+
+// TODO Should we rename kind to side and class to kind in this interface?
+#[serde_as]
+#[derive(Debug, Deserialize)]
+struct Order {
```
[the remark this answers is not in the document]
vkgnosis: Right now we control both sides of the interface so we can easily make changes and make this as good as possible. I'm fine with this change. More changes have been discussed in https://github.com/cowprotocol/services/pull/874#discussion_r1035873673 . When we make changes here we should also update `crates/driver/openapi.yml`.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -0,0 +1,137 @@
...
+    buy_token_balance: BuyTokenBalance,
+    #[serde_as(as = "serialize::U256")]
+    #[serde(default)]
+    full_fee_amount: U256,
+    is_mature: bool,
+    #[serde(flatten)]
+    class: OrderClass,
+    #[serde_as(as = "serialize::U256")]
+    // TODO Is this correct?
+    executed_amount: U256,
```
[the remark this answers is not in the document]
vkgnosis: Seems correct to me.

vkgnosis (approved):

crates/driver/src/api/mod.rs
```diff
@@ -1,42 +1,56 @@
...
-) -> JoinHandle<()> {
-    let filter = handle_all_routes(drivers).boxed();
-    tracing::info!(%address, "serving driver");
-    let (_, server) = warp::serve(filter).bind_with_graceful_shutdown(address, shutdown_receiver);
-    task::spawn(server)
+pub mod execute;
+pub mod info;
+pub mod solve;
+
+const REQUEST_BODY_LIMIT: usize = 32 * 1024;
```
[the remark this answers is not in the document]
vkgnosis: I think at the moment we don't have a limit because the solver binary fetches the current auction instead of having a route like /solve .

Considering this will only be called by a trusted party (autopilot) we might not need a limit.

nlordell:
> But I do think it's a fewfold improvement over warp.

To me main motivation for using `axum` is that it seems more actively developed.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[serde_as(as = "HashMap<_, DisplayFromStr>")]
+    prices: HashMap<H160, BigUint>,
+}
+
+// TODO Should we rename kind to side and class to kind in this interface?
+#[serde_as]
+#[derive(Debug, Deserialize)]
+struct Order {
+    #[serde_as(as = "serialize::Hex")]
+    id: Vec<u8>,
```
nlordell, line 32: naming-nit: I would call this `uid` - as it is a unique identifier for the order (two orders have the same UID if and only if they are they represent the same on-chain order) and because that is the domain-specific term we use elsewhere in the protocol.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -0,0 +1,137 @@
...
+    buy_token: H160,
+    #[serde(default)]
+    receiver: Option<H160>,
+    #[serde_as(as = "serialize::U256")]
+    sell_amount: U256,
+    #[serde_as(as = "serialize::U256")]
+    buy_amount: U256,
+    valid_to: u32,
+    #[serde_as(as = "serialize::Hex")]
+    app_data: Vec<u8>,
```
nlordell, line 44: app-data is strictly a `[u8; 32]`.

crates/driver/src/api/mod.rs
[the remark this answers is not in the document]
nlordell: For `/solve` I think it would have to be in the MBs. For other requests, the 32KB should be fine.

crates/driver/src/api/mod.rs
```diff
@@ -1,42 +1,56 @@
...
+        tower::ServiceBuilder::new()
+            .layer(tower_http::limit::RequestBodyLimitLayer::new(
+                REQUEST_BODY_LIMIT,
+            ))
+            .layer(tower_http::trace::TraceLayer::new_for_http()),
+    );
+
+    // Add routes.
+    let app = solve::route(app);
+    let app = info::route(app);
```
nlordell, line 30: This is slightly different than what I had imagined and realised I didn't notice it when reading the issue https://github.com/cowprotocol/services/issues/895#issuecomment-1337498190.

I would change this a little. Basically, I would `solve::route(app, solver_name)` to mount one endpoint per solver in order to avoid doing competition in the `driver`. See the comment in the issue for more justification.

nlordell (commented):
Changes look good to me. Also see the long-term appeal of the proposed structure and like how organized things are!

I made a comment about the `driver` doing competition internally. I was under the impression that all competition should happen at the `autopilot` and that multiple solvers sharing a single driver was to share internal indexing processes (like indexing liquidity), but that each solver should still be represented (and not pre-filtered by the driver).

crates/driver/src/api/solve/dto/auction.rs
nlordell: > IIRC auction ids are not unique?

Correct, they can collide across deployments.

MartinquaXD (approved):

nlordell (approved):
Looks good. Just some small nits on the auction DTO.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -0,0 +1,136 @@
...
+    #[serde_as(as = "HashMap<_, DisplayFromStr>")]
+    prices: HashMap<H160, BigUint>,
+}
+
+// TODO Should we rename kind to side and class to kind in this interface?
+#[serde_as]
+#[derive(Debug, Deserialize)]
+struct Order {
+    #[serde_as(as = "serialize::Hex")]
+    uid: Vec<u8>,
```
MartinquaXD, line 32: `OrderUids` have a fixed length so `[u8; 56]` would be slightly more accurate.

Issue #898, opened 2022-12-01
Stricter guidelines


sistemd:
As a follow-up to https://github.com/cowprotocol/services/pull/764, there needs to be a tighter and more concrete set of guidelines.

For a start, I would focus on:
- Preferring E2E tests, avoiding mocking, and including unit tests only in cases where it's impossible to test at a higher level of abstraction.
- Forbidding API model reuse, especially in the core logic.
- Putting more attention on having a clean separation between the core logic and external parts of the system, including the database

PR #911, opened 2022-12-05
Stricter domain driven style guidelines in `driver`

sistemd:
Fixes https://github.com/cowprotocol/services/issues/898.

It's hard to include enough context and motivation while still keeping the guidelines brief. I hope I struck a good balance. Please let me know if I should include more or less reasoning. I think this also gives a good idea of why `driver` is currently structured like it is.

### DIFF
--- a/crates/driver/GUIDELINES.md
+++ b/crates/driver/GUIDELINES.md
@@ -1,0 +1,53 @@
+# Guidelines
+
+High-level guidelines for writing code in a domain-driven style.
+
+## Tests
+
+Prefer to test at the _highest level of abstraction_. This means that the best tests are those
+that interact with the code at the public API layer. E.g., if the code exposes a REST API,
+the tests should be hitting the REST API. Mocking should be brought to a minimum. The
+only components that should be mocked are those components which are _external_ to our system.
+Example: quoting or settlement scoring logic are not external to the system. However, a remote
+solver or an external API provider are external to our system.
+
+There are a few reasons why this is a good approach:
+- The only way to know that the boundary integrations work (e.g. the boundary between your API
+layer and some other layer) is to include that boundary in the tests.
+- Code needs to be refactorable. When tests touch implementation details, those implementation
+details become harder to change. Furthermore, because tests need to change along with the
+implementation details, tests are now useless because you may have made a mistake while changing
+them. Developers are less inclined to refactor and code starts to rot.
+
+It's OK to test implementation details if it's an edge case that's impossible to test from the
+highest layer, or if it's so difficult to test from the highest layer that testing it that way
+becomes impractical.
+
+## Core Logic
+
+This is also referred to as "domain logic". The core logic sets your software
+apart from all other software and brings value to your business. For example, almost
+every software system today has some form of persistence (DB, writing to files, etc.), some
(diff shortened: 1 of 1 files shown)


crates/driver/GUIDELINES.md
```diff
@@ -0,0 +1,53 @@
...
+that interact with the code at the public API layer. E.g., if the code exposes a REST API,
+the tests should be hitting the REST API. Mocking should be brought to a minimum. The
+only components that should be mocked are those components which are _external_ to our system.
+Example: quoting or settlement scoring logic are not external to the system. However, a remote
+solver or an external API provider are external to our system.
+
+There are a few reasons why this is a good approach:
+- The only way to know that the boundary integrations work (e.g. the boundary between your API
+layer and some other layer) is to include that boundary in the tests.
+- Code needs to be refactorable. When tests touch implementation details, those implementation
```
nlordell, line 17: This is the biggest practical win of the testing guidelines IMO.

crates/driver/GUIDELINES.md
```diff
@@ -0,0 +1,53 @@
...
+details become harder to change. Furthermore, because tests need to change along with the
+implementation details, tests are now useless because you may have made a mistake while changing
+them. Developers are less inclined to refactor and code starts to rot.
+
+It's OK to test implementation details if it's an edge case that's impossible to test from the
+highest layer, or if it's so difficult to test from the highest layer that testing it that way
+becomes impractical.
```
nlordell, line 24: Do you think cases like checking algorithms are a correct use of unit tests? We have a path-finding algorithm in the baseline solver, as well as basic algorithms around token input/output calculations for AMMs (which we want to check matches the on-chain version).

In general those tests ☝️ are limited enough in scope that they shouldn't require mocking (and if it does, then it could be a code-smell IMO) and should not be an issue with refactoring as:
- If they are removed, than the tests are removed
- If their interface changes in some fundamental way, then we likely need to generate new test data for these cases anyway

crates/driver/GUIDELINES.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Core Logic
+
+This is also referred to as "domain logic". The core logic sets your software
+apart from all other software and brings value to your business. For example, almost
+every software system today has some form of persistence (DB, writing to files, etc.), some
+form of UI (web UI, CLI, REST API), some external components that it consumes, etc. What
+sets these software systems apart are the _domain problems_ which they solve. For example, cowswap
```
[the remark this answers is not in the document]
nlordell: (mostly joking, just a tiny nit).

nlordell (approved):
Nice.

The scope here is nice and tight, so its easy to follow and the concepts/guidelines and reasoning is well explained in my opinion 👍.

crates/driver/GUIDELINES.md
vkgnosis: IMO local unit tests are fine and I want to keep using them. You can still local unit test an algorithm through its public api. The problem with the mocks shows up when you're using a mock of an interface of a different module and then when someone wants to change that module they have to fix all your tests.

vkgnosis (approved):

MartinquaXD:
The motivation makes sense to me. Let's see how practical and confidence inspiring testing at the highest level of abstraction will be.

PR #913, opened 2022-12-05
Check Solver Balance Before Submission

sunce86:
This PR checks the balance of the solver submission address and disqualifies a solution if it's cost is “comparable” to the cost of submitting the solution. (see also discussion [here](https://cowservices.slack.com/archives/C0375NV72SC/p1669150193874849))

There were two options:
1. To check the balance for all the settlements with a single batch call and skip those not passing the check.
2. Once the settlements are ranked, check the balance for each settlement sequentially until one settlement passes the check.

I've chosen the (1) because I would rather spend a few more RPC calls than potentially having to call them sequentially and slow down the ranking process.


vkgnosis:
Usually simulation would fail if the solver was out of ETH, right? Is the problem that if you just barely have enough then the simulation might succeed but when settling on chain it will fail?

vkgnosis:
This is unlikely to happen because we usually make sure solvers have enough funds. I wonder if it would be better to only check the winning solution's solver and if it fails the check get the next winner and so on.

crates/solver/src/settlement_rater.rs
```diff
@@ -221,3 +241,43 @@ impl SettlementRating for SettlementRater {
         ))
     }
 }
+
+async fn solver_balances(web3: &Web3, addresses: HashSet<H160>) -> Result<HashMap<H160, U256>> {
```
MartinquaXD, line 245: I have a few concerns with this function:
1) `join_all` works with iterator so you don't have to `.collect()`
2) using array indexing is safe in this case but IMO should be replaced with iterators if possible
3) if one of the `.balance()` calls fails the whole call fails causing all the settlements to be discarded. Alternatively you could never return an error and simply set the balance to 0 if a call fails.


Small suggestions which would address all those concerns:
```rust
async fn solver_balances(web3: &Web3, addresses: HashSet<H160>) -> HashMap<H160, U256> {
    let futures = addresses.into_iter().map(|address| async move {
        (
            address,
            web3.eth().balance(address, None).await.unwrap_or_default(),
        )
    });

    std::iter::FromIterator::from_iter(futures::future::join_all(futures).await.into_iter())
}
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -182,6 +182,26 @@ impl SettlementRanker {
                 )),
             );
         }
+
+        // Even though some settlements passed the simulation, we require from solvers to have
+        // a bit more ETH balance then needed at this moment, to cover the potential increase of
+        // cost of sending transaction onchain, because of the sudden gas price increase
+        let rated_settlements = rated_settlements
+            .into_iter()
+            .filter(|(solver, settlement, _)| {
```
MartinquaXD, line 191: Instead of calling `into_iter().filter().collect()` you could also call `rated_settlements.retain()`. Might look a little nicer.

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -26,6 +26,15 @@ pub struct RatedSettlement {
     pub scaled_unsubsidized_fee: BigRational, // In wei.
     pub gas_estimate: U256,                   // In gas units.
     pub gas_price: BigRational,               // In wei per gas unit.
+    pub solver_balance: U256,                 // In wei
+}
+
+impl RatedSettlement {
+    pub fn is_payable(&self) -> bool {
+        let cost = self.gas_estimate.to_big_rational() * &self.gas_price;
+        let multiplier = BigRational::new(3.into(), 1.into());
```
MartinquaXD, line 35: Was the 3x factor chosen because it would have successfully resolved the issues during the FTX crash or is this a rough guess?

MartinquaXD (approved):
I'm fine with this PR to catch a small number of possible reverts during very volatile market periods. I'm also not very concerned with adding a handful of eth calls to get a few balances. **IF** the latency of this approach is actually a problem we could also simply query the balances in parallel with simulating the settlements. Then this PR would probably add no overhead whatsoever.

nlordell:
We already simulate with a gas price. We could, in theory, simulate with a gas price * the multiplier introduced in this PR. Wouldn't that amount to the same thing but require one less RPC call?

sunce86:
> We already simulate with a gas price. We could, in theory, simulate with a gas price * the multiplier introduced in this PR. Wouldn't that amount to the same thing but require one less RPC call?

Closing since https://github.com/cowprotocol/services/pull/924 is merged.

PR #917, opened 2022-12-06
Use simpler Settlement type in SingleOrderSolver

vkgnosis:
In single_order_solver.rs we add a custom `struct SingleOrderSettlement` which single order solvers return instead of the normal `struct Settlement`. This struct is much simpler than the full Settlement. By using it we can share more code between single order solvers.

The motivation for this is that I want to fix two bugs in the settlement merging logic. This is currently implemented in terms of full settlement which makes it more complicated than it needs to be. Merging single order solver settlements is easier than merging full settlements and this is the only place we want to merge settlements.

### Test Plan

no logic change, some unit tests adjusted


nlordell (approved):
Makes perfect sense.

PR #919, opened 2022-12-06
Remove Approval::AllowanceSufficient in AllowanceManager refactor

vkgnosis:
While working with other code I found it wrong that "no allowance needed" is a separate enum case that also implements Interaction but is encoded as nothing so it disappears later. This feels weird because we still drag it around until the settlement is encoded.

### Test Plan

Ci with existing tests


vkgnosis:
Will merge it then. If it turns out this was useful then I'm fine with adding it back too.

nlordell:
> Will merge it then.

Please do! I added it **hoping** it would be useful but it has empirically proven not to be (not sure if this was clear enough in my comment).

PR #920, opened 2022-12-06
Add logging options to refunder

fedgiac:
Parameter to customize the refunder at a logger level.
They were taken from the other services and isolated to a macro.

### Test Plan

```
$ cargo run --bin refunder -- --ethflow-contract 0x0000000000000000000000000000000000000000 --refunder-pk 0x0  2>/dev/null
2022-12-06T14:08:49.918Z  INFO refunder: running refunder with validated arguments:
[...]
$ cargo run --bin refunder -- --ethflow-contract 0x0000000000000000000000000000000000000000 --refunder-pk 0x0 --log-stderr-threshold 'trace' 2>/dev/null
[no output]
```

### DIFF
--- a/crates/autopilot/src/main.rs
+++ b/crates/autopilot/src/main.rs
@@ -4,8 +4,8 @@ use clap::Parser;
 async fn main() {
     let args = autopilot::arguments::Arguments::parse();
     shared::tracing::initialize(
-        args.shared.log_filter.as_str(),
-        args.shared.log_stderr_threshold,
+        args.shared.logging.log_filter.as_str(),
+        args.shared.logging.log_stderr_threshold,
     );
     shared::exit_process_on_panic::set_panic_hook();
     tracing::info!("running autopilot with validated arguments:\n{}", args);
--- a/crates/driver/src/arguments.rs
+++ b/crates/driver/src/arguments.rs
@@ -7,6 +7,7 @@ use {
         ethrpc,
         gas_price_estimation::GasEstimatorType,
         http_client,
+        logging_args_with_default_filter,
         sources::{balancer_v2::BalancerFactoryKind, BaselineSource},
         tenderly_api,
     },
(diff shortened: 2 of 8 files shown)


vkgnosis (approved):
I'm ok with this but I'm wondering if it wouldn't be better to copy paste the logging arguments instead of a macro. Feels easier to read. Downside is if we change default arguments we need to do it in more places but it seems like we want different logging defaults already. So maybe there isn't much point in the reuse.

fedgiac:
I agree with you but overall I think it's more valuable to keep the various configurations in sync since we copy from one to the other and a non-existing flag is just ignored. (The refunder sketch configs actually had `LOG_STDERR_THRESHOLD` set and I was surprised it wasn't working.)
Anyway I don't have a strong opinion on this, I'm also ok with getting rid of the macro.

vkgnosis:
> I agree with you but overall I think it's more valuable to keep the various configurations in sync since we copy from one to the other and a non-existing flag is just ignored.

But they aren't in sync anyway, right? Because we have different defaults on the log levels between binaries.

> The refunder sketch configs actually had LOG_STDERR_THRESHOLD set and I was surprised it wasn't working

Why was it not working?

fedgiac:
> But they aren't in sync anyway, right? Because we have different defaults on the log levels between binaries.

Here I'm thinking of copying the config variables from the services to a new deployment. Things like `LOG_FILTER` would probably be copied to a new deployment config as it looks like a very basic feature (at least this is what happened) and while it's clear that `autopilot=warn` should be changed to `newservice=warn`, it's not clear if you instead should use `LOG_FILTERS` instead of `LOG_FILTER`. Then, if nothing is set, we still want the right defaults in the code.

> Why was it not working?

Because even if they were specified in the configs, the actual behavior in the code was hardcoded so they had no effect. The changes of this PR don't fix that, but at least you shouldn't have to worry about consistency with other deployments and from where to copy the code.

nlordell:
Personally, I like hygienic macros as a tool for generating boiler-plate code like we have here.

I don't think it adds much complexity and keeps consistency for the logging fields while making it easy for specifying different defaults per service.

I don't feel super strongly about it, but in general I like the change.

nlordell:
Also this passes the "can't be replaced by a function" test that you should ask when introducing macros.

vkgnosis:
I'm fine with merging. No strong opinion.

PR #903, opened 2022-12-02
Track solvable orders metrics grouped by order class

MartinquaXD:
Instead of just knowing how many orders are in the current auction or got filtered out it would be nice to know how many orders of each class there are.

This could help us see how popular the new limit order feature is and if we are filtering out lots and lots of limit orders because they are out of the market price range.

[related slack thread](https://cowservices.slack.com/archives/C049P351GKF/p1669924566478809?thread_ts=1669912430.193089&cid=C049P351GKF)

PR #921, opened 2022-12-06
Order Filtering Metrics

nlordell:
Follow up to #903 

This PR extends the metrics from the aforementioned PR to track per-class metrics per-filter.

This means, we should have a really good idea of the number of orders are getting filtered at each filtering stage in the auction building process.

This is accomplished by wrapping the solvable orders in a `OrderFilter` type which counts orders after every filtering step and writes metrics once the auction preparation is complete.

### Test Plan

Adjusted unit tests to new `OrderFilter` interface.

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -2,7 +2,8 @@ use crate::{database::Postgres, risk_adjusted_rewards};
 use anyhow::{Context as _, Result};
 use bigdecimal::BigDecimal;
 use chrono::Utc;
-use futures::StreamExt;
+use futures::{FutureExt as _, StreamExt as _};
+use itertools::Itertools;
 use model::{
     auction::Auction,
     order::{Order, OrderClass},
@@ -21,7 +22,9 @@ use shared::{
 };
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
+    future::Future,
     iter::FromIterator,
+    mem,
     sync::{Arc, Mutex, Weak},
     time::Duration,
 };
@@ -43,7 +46,7 @@ pub struct Metrics {
     auction_solvable_orders: IntGaugeVec,
 
     /// Auction filtered orders grouped by class.
-    #[metric(labels("class"))]
+    #[metric(labels("class", "reason"))]
     auction_filtered_orders: IntGaugeVec,
 
     /// Auction errored price estimates.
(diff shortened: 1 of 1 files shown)


nlordell:
> number_of_classes * number_of_filters?

Exactly. `(number_of_classes = 3) * (number_of_filters = 7) = 21` metrics.

vkgnosis (commented):
This feels over engineered to me. Why not keep the filtering functions as they are and only use something like OrderFilter to keep track of the counts in between? I don't like that we reimplement fitler, try_filter and so on and that the filtering functions become restricted this way.

I'm also not sure if the extra metrics are useful. Do we need to separately count the filtering for all the classes? For example, banned users usually get filtered when they create orders already so I find it hard to imagine that metric being useful. If we wanted to debug why a specific order gets filtered out, we'd need to approach this another way.

nlordell:
> This feels over engineered to me

What specifically? The nice part about `OrderFilter` is that it is not possible to filter orders without recording metrics without breaking the interface. Additionally, we make sure we don't update the gauge unless we successfully prepare the order (guaranteed by the `finish` method)

> I don't like that we reimplement fitler, try_filter and that the filtering functions become restricted this way.

These are just wrappers around a very generic `fn(Vec<Order>) -> Vec<Order>` mapping that stores metrics for recording, so it shouldn't be restrictive at all.

> I'm also not sure if the extra metrics are useful

I think I went a bit over-zealous on the metrics. Do you think reducing them so that they are per-filter reason (instead of `class * filter_reason`) be acceptable?

vkgnosis:
The metrics are definitely *acceptable*. I'm just wondering if we expect to use all of them. If don't expect to use a metric then we might not want to include it because that takes code, space in prometheus, etc. If you expect to use them then I don't mind having them.

>  it is not possible to filter orders without recording metrics

I feel this guarantee is not needed and doesn't justify the extra code and the more awkward interface (restricts how you can access the vector by design). If we loosened this guarantee and manually called OrderFilter::update_counters between filters and then OrderFilter::finish at the end the code would be simpler. 

This is not a strong feeling. The code looks fine and I'm fine with merging. I just weakly feel it could be less code.

nlordell:
>  If we loosened this guarantee and manually called OrderFilter::update_counters between filters and then OrderFilter::finish at the end the code would be simpler.

OK - will propose an alterante change with this in mind.

nlordell:
@vkgnosis https://github.com/cowprotocol/services/pull/926

PR #923, opened 2022-12-06
Alerter Track All Settleable Orders

nlordell:
This PR changes the order alerts to not only alert when an order is too long without trading, but also count all settle-able orders by class.

This allows us to keep better track if certain aspects of the orderbook are getting clogged up.

Note that this also changes slightly the behaviour of the alerter:
1. It now queries the status of all open orders against 0x. This makes it way "chattier". To mitigate this concern, I made the update interval configurable so we can tweak it if we start getting rate-limited.
2. The logic for alerting is slightly different. Specifically, before, the check for settle-able orders would find the **first** one that 0x reported as being "in the money" and then decide to flip the "alert" bit depending on whether or not it was waiting for a certain threshold. This made this logic dependent on the ordering of open orders from the orderbook (if for example, the orderbook ordered them by most recent, and a new settle-able order was being added to the book every second or so, then AFAIU the code wouldn't alert as the loop would exit after finding the first settle-able order and checking if it has been sette-able for too long). I changed the logic to flip the alert bit if **any** order has been open for too long if there are no recent trades.

### Test Plan

Run the alerter locally and see the metrics:
```
% curl localhost:9588/metrics
# HELP gp_v2_alerter_matchable_orders_count Gauge for counting the number of matchable but open trades in the
# TYPE gp_v2_alerter_matchable_orders_count gauge
gp_v2_alerter_matchable_orders_count{class="Limit"} 1
gp_v2_alerter_matchable_orders_count{class="Liquidity"} 0
gp_v2_alerter_matchable_orders_count{class="Market"} 0
# HELP gp_v2_alerter_no_trades_but_matchable_order Set to 0 or 1 depending on whether our alert condition is satisfied
# TYPE gp_v2_alerter_no_trades_but_matchable_order gauge
gp_v2_alerter_no_trades_but_matchable_order 0

% # after the trade happens...
% curl localhost:9588/metrics
# HELP gp_v2_alerter_matchable_orders_count Gauge for counting the number of matchable but open trades in the
# TYPE gp_v2_alerter_matchable_orders_count gauge
gp_v2_alerter_matchable_orders_count{class="Limit"} 0
gp_v2_alerter_matchable_orders_count{class="Liquidity"} 0
gp_v2_alerter_matchable_orders_count{class="Market"} 0
# HELP gp_v2_alerter_no_trades_but_matchable_order Set to 0 or 1 depending on whether our alert condition is satisfied
# TYPE gp_v2_alerter_no_trades_but_matchable_order gauge
gp_v2_alerter_no_trades_but_matchable_order 0
```


crates/alerter/src/main.rs
```diff
@@ -25,11 +29,32 @@ struct Order {
...
     status: OrderStatus,
     creation_date: DateTime<Utc>,
     partially_fillable: bool,
     is_liquidity_order: bool,
+    #[serde(flatten)]
+    class: OrderClass,
+}
+
+impl Order {
+    fn effective_sell_amount(&self) -> U256 {
```
nlordell, line 44: This is needed to not false-ly flag limit orders that are in the money but not tradable because of the surplus fee.

crates/alerter/src/main.rs
```diff
@@ -196,10 +230,16 @@ impl Alerter {
...
+
         tracing::debug!("found {} open orders", orders.len());
+
         std::mem::swap(&mut self.open_orders, &mut orders);
         // Keep only orders that were open last update and are not open this update.
-        orders.retain(|order| !self.open_orders.contains(order));
+        orders.retain(|(order, _)| {
+            self.open_orders
+                .iter()
+                .all(|(open_order, _)| open_order.uid != order.uid)
```
nlordell, line 241: Seems unrelated, but it isn't. We use order UIDs to identify the two, otherwise changes in surplus fee can cause an order to be considered a "closed" order, even if that is not the case.

MartinquaXD: Seems unintuitive enough to warrant an inline comment.

crates/alerter/src/main.rs
```diff
@@ -211,22 +251,34 @@ impl Alerter {
...
-            order.uid,
-        );
-    }
-
     pub async fn update(&mut self) -> Result<()> {
         self.update_open_orders().await?;
+        let matchable_orders = self.matchable_orders().await?;
+
+        let counts = matchable_orders.iter().fold(
+            OrderClass::VARIANTS
```
MartinquaXD, line 264: I had to do the same thing recently and didn't think about `.fold()`. I still don't love it but at least it's self contained. 👍

crates/alerter/src/main.rs
```diff
@@ -236,43 +288,76 @@ impl Alerter {
...
+        } else {
+            self.metrics.no_trades_but_matchable_order.set(0);
+        }
+
+        Ok(())
+    }
+
+    async fn matchable_orders(&mut self) -> Result<Vec<(Order, Duration)>> {
+        let mut result = Vec::new();
+        for (order, solvable_since) in &mut self.open_orders {
```
MartinquaXD, line 320: With the expectation that the auction will be growing soon this might need to be a bit more sophisticated than querying zeroex for every order one after the other. But obviously out of scope for this PR.

MartinquaXD (approved):

crates/alerter/src/main.rs
nlordell: Good point. I also think we should MUX out to multiple DEX aggregators just in case.

vkgnosis:
This changes the goal of the alerter a lot. Previously, getting rate limited by 0x was not a concern because we wouldn't enter the code that had to talk to 0x often because we already knew something got traded. Now we're expected to always check the tradeability of all orders. And as Martin said this could make determining `matchable_orders` slow.

I'm not super how I would use the metric either.

Side note: I'm wondering now if we couldn't replace the alerter with some extra code in the single order solver itself. They already check random orders against their external source (like 0x) so they could do what the alerter was doing before this PR.

crates/alerter/src/main.rs
vkgnosis: I don't know if this will work in practice. If we always check all orders then this function can take a really long time, and it errors if any individual request errors.

nlordell:
Cleanup

PR #924, opened 2022-12-06
Check Solver Balance Before Submission V2

sunce86:
Another version of https://github.com/cowprotocol/services/pull/913 as proposed [here](https://github.com/cowprotocol/services/pull/913#issuecomment-1339027508)

This PR bumps the gas price inside `SettlementRanker` only, because we don't want to have the bumped value stored in solver competition (right?).

Obviously much simpler and does not require additional node calls to check the balance of solver accounts.

On the other side, if simulation fails because of low solver balance, we don't send the specialized message to notify solvers (as in https://github.com/cowprotocol/services/pull/913), but we send a regular `simulation failure` where the reason of failing is hidden (luckily, we also send the used gas price to solvers so they can reproduce the transaction and figure out the reason of failure in tenderly).

https://github.com/cowprotocol/services/pull/913 seems like a proper way of doing this check. This PR, however, is much simpler change but feels like misusing the gas price. Also, https://github.com/cowprotocol/services/pull/913 seems like more future proof, because we might want to remove this gas price multiplier one day (for whatever reason I can't think of right now).

That said, I don't know which one to merge. I'll let the team decide.

### DIFF
--- a/crates/solver/src/settlement_ranker.rs
+++ b/crates/solver/src/settlement_ranker.rs
@@ -20,6 +20,12 @@ use std::{cmp::Ordering, sync::Arc, time::Duration};
 
 type SolverResult = (Arc<dyn Solver>, Result<Vec<Settlement>, SolverRunError>);
 
+// We require from solvers to have a bit more ETH balance then needed
+// at the moment of simulating the transaction, to cover the potential increase
+// of the cost of sending transaction onchain, because of the sudden gas price increase.
+// To simulate this sudden increase of gas price during simulation, we artificially multiply
+// the gas price with this factor.
+const SOLVER_BALANCE_MULTIPLIER: f64 = 3.;
 pub struct SettlementRanker {
     pub metrics: Arc<dyn SolverMetrics>,
     pub settlement_rater: Arc<dyn SettlementRating>,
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
I like this version. Less overhead and to me this check makes sense as part of the simulation. Simulation is for checking that the transaction would succeed in which we want to take into account rising gas prices.

It might be possible to still extract the reason for the simulation failure. We would need to check the node's error string and see if it distinguishes failing because the submitting account runs out of eth.

nlordell (approved):

nlordell:
> It might be possible to still extract the reason for the simulation failure. We would need to check the node's error string and see if it distinguishes failing because the submitting account runs out of eth.

That would be a nice improvement!

sunce86:
> It might be possible to still extract the reason for the simulation failure.
Indeed, however I would avoid doing this at all costs, since messages can change and adding a new submission strategy requires updating this part of code.

vkgnosis:
> Indeed, however I would avoid doing this at all costs, since messages can change and adding a new submission strategy requires updating this part of code.

"at all costs" feels too strong to me. We do a similar thing to distinguish reverts from other errors in ethcontract. It's not great but it's not terrible either. Considering that missclassification is not critical it seems fine to have this as a heuristic. But it might not be worth the time to implement it.

PR #878, opened 2022-11-30
Remove Custom Price Order Sell Price from UCP

sunce86:
Fixes https://github.com/cowprotocol/services/issues/848 and potentially https://github.com/cowprotocol/services/issues/849

Custom price orders (liquidity orders and limit orders) now store both `sell_price` and `buy_price` outside of the clearing price vector (as suggested by @MartinquaXD, thanks for clarifying the issue).

PR #929, opened 2022-12-07
Refactor LimitOrder id for searchers

sunce86:
This PR updates the `LimitOrderUid`.

This change is motivated by need to know class of order inside `LimitOrder` struct. (bug fix related to different treatment of different orders will be in a follow up PR). 

### Test Plan

Unit tests passing.


crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -217,17 +217,12 @@ fn match_prepared_and_settled_orders(
...
-                            !internalizable,
-                            "liquidity orders are not allowed to be internalizable"
-                        )
-                    }
+            if prepared.is_liquidity_order() {
+                if let Some(internalizable) = settled.exec_plan.as_ref().map(|plan| plan.internal) {
+                    ensure!(
+                        !internalizable,
+                        "liquidity orders are not allowed to be internalizable"
+                    )
```
sunce86, line 225: One bug fixed here (I know I could have fixed it by using field `is_liquidity_order` but I wanted to kill that field).

crates/solver/src/liquidity.rs
```diff
@@ -90,40 +90,50 @@ pub enum Exchange {
...
+    Limit(OrderUid),
+    Liquidity(LiquidityOrderUid),
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(Derivative))]
+#[cfg_attr(test, derivative(PartialEq))]
+pub enum LiquidityOrderUid {
+    /// Originates from the auction model of solvable orders
+    User(OrderUid),
```
sunce86, line 108: Maybe some other name? `WhitelistedUser` maybe?

sunce86:
Seems like I missed to classify `foreign_liquidity_orders` somehow. Looking into it.

crates/solver/src/liquidity.rs
```diff
@@ -90,40 +90,50 @@ pub enum Exchange {
 }
 
 /// Used to differentiate between different types of orders that can be sent to solvers.
-/// User orders containing OrderUid are the orders from the orderbook.
+/// User orders (market + limit) containing OrderUid are the orders from the orderbook.
 #[derive(Debug, Clone)]
 #[cfg_attr(test, derive(Derivative))]
 #[cfg_attr(test, derivative(PartialEq))]
 pub enum LimitOrderUid {
```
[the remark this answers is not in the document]
nlordell: AFAIU this is used as an ID for `LimitOrder`s in the `solver`. With that in mind, I think that "ID" should appear somewhere in the name.

Extremely nit-picky, but I would actually call it `*Id` (since, AFAIK, it isn't required for uniquely identifying orders, but just some debugging metadata):
```rust
enum LimitOrderId {}
enum LiquidityOrderId {}
```

crates/solver/src/liquidity.rs
nlordell: I think `CowProtocol` or just `Protocol` to indicate it is a protocol limit order would be good.

I also think `User` is a bit general.

nlordell (approved):
Very nice change!

A couple tiny naming nits, but feel free to ignore.

crates/solver/src/liquidity.rs
sunce86: I've changed it to ProtocolOrForeign, added comment why

crates/solver/src/liquidity.rs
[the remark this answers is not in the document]
sunce86: renamed

sunce86:
I failed to classify `foreign_liquidity_orders` as a separate enum variant, added comment why. Currently not important to differentiate between protocol and foreign liquidity order in `solver` but would be nice to split it once driver solver colocation is done.
Re-requesting review.

crates/solver/src/liquidity/order_converter.rs
```diff
@@ -55,8 +54,15 @@ impl OrderConverter {
             _ => (order.data.sell_amount, order.data.fee_amount),
         };
 
+        let id = match order.metadata.class {
+            OrderClass::Market => LimitOrderId::Market(order.metadata.uid),
+            OrderClass::Liquidity => {
+                LimitOrderId::Liquidity(LiquidityOrderId::ProtocolOrForeign(order.metadata.uid))
+            }
```
sunce86, line 61: Here I don't know how to make a difference between protocol and foreign liquidity order.
Either add this info to `Order` (which I don't want to do right now because `Order` is shared everywhere) or add a specialized implementation of function `normalize_limit_order` just for foreign liquidity orders which seems like an overkill right now.

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -337,13 +332,13 @@ fn match_settled_prices(
...
-        }
-    });
+    let executed_tokens = executed_limit_orders
+        .iter()
+        .flat_map(|order| match order.order.id {
+            LimitOrderId::Market(_) => vec![order.order.buy_token, order.order.sell_token],
+            LimitOrderId::Limit(_) => vec![],
+            LimitOrderId::Liquidity(_) => vec![],
+        });
     for token in executed_tokens {
```
sunce86, line 342: Second bug fixed. This is a follow up to https://github.com/cowprotocol/services/pull/878 where I missed to update this line. In https://github.com/cowprotocol/services/pull/878 we decided to decouple prices for limit and liquidity orders from UCP.

MartinquaXD: I think we still need to keep around the uniform clearing prices for tokens traded by `Limit` orders.

sunce86: Why do you think that?
All changes in https://github.com/cowprotocol/services/pull/878 were made around the idea of both liquidity and limit orders having custom prices therefore being removed from the UCP.

crates/solver/src/liquidity/order_converter.rs
MartinquaXD: What needs to be handled differently across normal and foreign liquidity orders?

sunce86: In this function I just need to set the variant correctly.
Generally in the code there is no different handling anywhere (at least at the moment).
I just like having this classification as it helps with code readability (for example, before this PR I was not 100% what kind of liquidity orders are used in our app).

crates/solver/src/solver/http_solver/settlement.rs
MartinquaXD: The [code](https://github.com/cowprotocol/services/blob/main/crates/solver/src/settlement/settlement_encoder.rs#L303-L310) that later computes the price adjusted for the `surplus_fee` in the `SettlementEncoder` still needs to know the uniform clearing price of the tokens traded in the limit order.
If we don't collect the traded tokens of the limit order here we might end up dropping their uniform clearing prices in `match_settled_prices()`.

MartinquaXD: Also it would be nice to know the original UCP for the limit order tokens in the `settle` transaction to prove that our custom prices still respect our promise of UCP (they just got adjusted with different `surplus_fee` amounts).

crates/solver/src/liquidity/order_converter.rs
MartinquaXD: Makes sense. 👍 
I see 2 more options:
1. pass another argument to the existing `normalize_limit_order()` function
2. manually update the `id` field of the normalized foreign liquidity order in `/http_solver/settlement.rs`

I don't really like any of those options TBH but I think I would go for `2`.

crates/solver/src/liquidity.rs
```diff
@@ -90,40 +90,57 @@ pub enum Exchange {
...
+///
+/// (1) and (2) are gathered when the auction is cut and they are sent to searchers
+/// (3) are received from searchers as part of the solution.
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(Derivative))]
+#[cfg_attr(test, derivative(PartialEq))]
+pub enum LiquidityOrderId {
+    /// TODO: Split into different variants once we have a DTO of order model for `driver` in driver solver colocation
+    /// TODO: The only reason why is together now is because function `normalize_limit_order` can't diferentiate between these two
+    ProtocolOrForeign(OrderUid),
```
nlordell, line 116: I would still call this `Protocol` - foreign liquidity orders are still protocol orders, they just don't come from our orderbook.

I think it is fine to punt this until a differentiation is needed (which I can't think of a reason we will need that right now. Maybe there is a solver reward/competition related reason though).

crates/solver/src/solver/http_solver/settlement.rs
nlordell: Yes, we definitely need to keep the UCPs in for limit orders.

As an example - a settlement that trades exactly 1 limit order should have 4 prices:
- UCP sell token price
- UCP buy token price
- limit order sell token price
- limit order buy token price

This is a soft requirement so that we can verify UCP after the fact.

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -337,13 +332,13 @@ fn match_settled_prices(
...
-            vec![order.order.sell_token]
-        } else {
-            vec![order.order.buy_token, order.order.sell_token]
-        }
-    });
+    let executed_tokens = executed_limit_orders
+        .iter()
+        .flat_map(|order| match order.order.id {
+            LimitOrderId::Market(_) => vec![order.order.buy_token, order.order.sell_token],
+            LimitOrderId::Limit(_) => vec![],
```
nlordell, line 339:
```suggestion
            LimitOrderId::Market(_) | LimitOrderId::Limit(_) => vec![order.order.buy_token, order.order.sell_token],
```

crates/solver/src/liquidity/order_converter.rs
nlordell: Personally, I think (since the differentiation isn't important yet) that this can be left out for now until we need it.

When we do, I slightly prefer option 2 as well.

crates/solver/src/solver/http_solver/settlement.rs
sunce86: Indeed. Added tokens for limit orders.

crates/solver/src/liquidity.rs
sunce86: I am ok with `Protocol` also since there is a comment explaining. Changed.

crates/solver/src/liquidity/order_converter.rs
sunce86: Let's keep it as is (for now) since I don't find any of the options being a good improvement :/

MartinquaXD (approved):

PR #843, opened 2022-11-23
Settlement Simulation Works Without Internalizing Interactions

sunce86:
Tackles points 3/4 from https://github.com/cowprotocol/services/issues/635

Dependent on https://github.com/cowprotocol/services/pull/791

Now SettlementRanker does simulation of settlements **without** internalizing interactions.
After simulation, ranking is done **with** internalizing interactions.

If settlement simulation fails on either without or with internalizing interactions, flag (`InternalizationStrategy`) and call data built with that flag will be forwarded to solvers via `notify` endpoint so solvers will know what exactly failed.

~Open to tweaking the implementation since after reading about the same task from multiple sources, I am not sure whether we should also do the ranking without internalizations and what exactly needs to be in the solver competition struct.~
Decided to go with the original idea from above.

### Test Plan

Updated unit tests.

PR #931, opened 2022-12-08
Mark interaction as internalizable instead of dropping it

MartinquaXD:
The optimization pipeline which should improve the objective value of a settlement by applying a few gas optimizations can cause solutions that are not viable to pass the simulation by internalizing AMM interactions that are overly optimistic.

Check out this [trade](https://explorer.cow.fi/orders/0x79d6e25eb8dfe279df8e6248c01fd9b337b14a2842af1719deb556e09483ca031a0af84ef9f0a5e60967c319b03dca9a9221bde563920171) selling `USDC` for `AMPL`. `Paraswap` currently gives too optimistic prices for `AMPL`. That means the settlement provided by `paraswap` would likely have reverted. Because the order was selling `USDC` (a trusted token) the optimization pipeline was allowed to internalize the erroneous AMM interaction which resulted in a working settlement with an unbeatable price (because we pay with our buffers).

I saw 2 ways to tackle this issue:
1. Don't apply optimizations if the original settlement would already revert
2. Instead of right out dropping the interaction we keep the original interaction but mark it as internalizable

I decided to go with `2` because together with #843 it should prevent the same problems while offering better debuggability than `1`.

### Test Plan
?

### DIFF
--- a/crates/solver/src/settlement/settlement_encoder.rs
+++ b/crates/solver/src/settlement/settlement_encoder.rs
@@ -139,7 +139,13 @@ impl SettlementEncoder {
             tokens: self.tokens.clone(),
             clearing_prices: self.clearing_prices.clone(),
             trades: self.trades.clone(),
-            execution_plan: Vec::new(),
+            execution_plan: self
+                .execution_plan
+                .iter()
+                // Instead of simply dropping the executions we mark all the interactions as
+                // internalizable.
+                .map(|(execution, _)| (execution.clone(), true))
+                .collect(),
             pre_interactions: self.pre_interactions.clone(),
             unwraps: self.unwraps.clone(),
         }


nlordell (approved):

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -139,7 +139,13 @@ impl SettlementEncoder {
...
             trades: self.trades.clone(),
-            execution_plan: Vec::new(),
+            execution_plan: self
+                .execution_plan
+                .iter()
+                // Instead of simply dropping the executions we mark all the interactions as
+                // internalizable.
+                .map(|(execution, _)| (execution.clone(), true))
+                .collect(),
             pre_interactions: self.pre_interactions.clone(),
```
sunce86, line 149: What is actually the difference here if you mark all the interactions internalizable and then get them dropped in the `finish`?

MartinquaXD: The difference is where the interactions gets dropped. If we only mark the interaction as internalizable we will simulate the original and optimized solution during the settlement ranking which will give us better insight in the actual solution. But if the interaction already got dropped there, we could end up with a solution where the original version would revert but the optimized one would succeed erroneously (see example in the description).

Another reason to use the flag instead of dropping the interaction is that this is how external solvers are supposed to do it exactly because we simulate to determine whether those interactions are even allowed to be internalized.

sunce86: Ah yes, I forgot you wrapped all the solvers in `OptimizingSolver`. In my head it was still the idea optimization happening just at the end of the auction process before submission.

sunce86 (approved):

Issue #925, opened 2022-12-07
The `solver` isn't simulating with access lists everywhere it needs to


sistemd:
The changes introduced in https://github.com/cowprotocol/services/pull/846 don't work in production. Simulating the winning solution works, but it seems that the settlement submission loop tries to do some other simulation and access list creation, which fails:

```
DEBUG auction{id=4895127 run=11}:submission{name=PublicMempool i=0}: solver::settlement_submission::submitter: access list not created, reason: contract call reverted with message: None
```

Logs: https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/r/s/i9SAI

PR #932, opened 2022-12-08
Add additional logging to access list simulation

sistemd:
Trying to figure out what's going on in https://github.com/cowprotocol/services/issues/925.

So `submit_with_increasing_gas_prices_until_simulation_fails` _does_ call the appropriate `estimate_settlement_access_list` function, which should do the partial access list estimation. However this `estimate_settlement_access_list` ends up with a tx reverted error. I added some logs to try to pinpoint what is failing (the partial access list estimation, or the final access list estimation). Would be good to merge this quickly and do another test on staging.

### DIFF
--- a/crates/solver/src/settlement_access_list.rs
+++ b/crates/solver/src/settlement_access_list.rs
@@ -20,7 +20,7 @@ use web3::{
     BatchTransport, Transport,
 };
 
-use crate::settlement::Settlement;
+use crate::{settlement::Settlement, settlement_simulation::tenderly_link};
 
 #[async_trait::async_trait]
 pub trait AccessListEstimating: Send + Sync {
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):

nlordell (approved):

crates/solver/src/settlement_access_list.rs
```diff
@@ -66,6 +66,8 @@ pub async fn estimate_settlement_access_list(
     settlement: &Settlement,
     tx: &TransactionBuilder<DynTransport>,
 ) -> Result<AccessList> {
+    tracing::debug!(?settlement, "generating access list for settlement");
```
nlordell, line 69: If we make these `trace` logs, we can temporarily enable them by configuration and then disable them once more once they've served their purpose.

sunce86:
Maybe you want to call `tenderly_link` somewhere so that you can reproduce the transaction in tenderly?

crates/solver/src/settlement_access_list.rs
nlordell: I'm not sure how Tenderly links support access lists. Just something to keep in mind.

PR #933, opened 2022-12-08
Check if http solver solution contains duplicate exec plans

sunce86:
This PR checks if a solution received from http searcher contains only interactions with unique execution plan coordinates.
For example, it is not allowed for a solution to contain two custom interactions with the same coordinates:

> {
>    sequence: 0,
>    position: 0,
> }

@harisang noticed that one of the searchers is sending this kind of solutions, so suggested to reject those and notify searchers about it.
While at it, I've introduced a `ConversionError` error enum, which should be used in future to detect different kind of validation errors for searchers solutions (which I think it can be very useful, especially for newjoiner searchers).

### Test Plan

Added unit tests.

### Release notes

We have a new error variant for `notify` endpoint.


sunce86:
Merging with 1 approval so it can be part of the next week release.

Issue #812, opened 2022-11-18
Allow solvers to specify if their solution should be submitted via flashbots


MartinquaXD:
Eventually we want the solver-driver co-location to enable solvers to submit their solutions however they want but work on this was paused.
However, solver maintainers are already [asking](https://cowservices.slack.com/archives/C0375NV72SC/p1668749776267779) for a way to signal that a solution should only be submitted via flashbots to avoid getting MEV attacked in the public mempool.
Until we actually finalize the solver-driver co-location adding a flag `submit_with_flashbots` in the solver response seems like a reasonable compromise.
In case we still want to take advantage of other private mempools (e.g. eden network) a more generically named flag `enforce_private_submission` might make more sense.

sunce86:
So, it seems to me that the best way is to add something like:

```
/// Settlement can override submission strategy
#[derive(Debug, Clone, Default)]
pub enum SettlementSubmitter {
    #[default]
    ProtocolDefault,
    Flashbots,
}

#[derive(Debug, Clone, Default)]
pub struct Settlement {
    pub encoder: SettlementEncoder,
    pub submitter: SettlementSubmitter,
}
```

Settlement is built using the solution received from http searcher, while `SolutionSubmitter` also works with the `Settlement`. Once we enter the submission, we would combine the default enabled strategies of protocol with the preferences of the winning settlement being submitted. 

WDYT

MartinquaXD:
The idea looks good to me. I would just call it `SubmissionStrategy`. I believe that naming would fit better with the existing terminology around solution submission.

nlordell:
Sounds very reasonable. I would call it "SubmissionPreference".

The other question is, "what happens if they choose a submitter that isn't configured?". I think we should reject the solution, the alternative would be to fallback to `ProtocolDefault`.

MartinquaXD:
> The other question is, "what happens if they choose a submitter that isn't configured?". I think we should reject the solution, the alternative would be to fallback to ProtocolDefault.

I guess we could let them control that with another flag like `required: bool`. If they specify `flashbots` is `required` but we currently don't have it enabled for some reason we would discard the solution.
We could also send the currently configured strategies in the metadata of the `/solve` request and they could pick and choose from those. Seems a bit overkill for this case but I just wanted to mention it.

poolpitako:
> Sounds very reasonable. I would call it "SubmissionPreference".
> 
[...]

PR #936, opened 2022-12-09
Searchers can provide preferred submission strategy

sunce86:
Fixes https://github.com/cowprotocol/services/issues/812
The simplest possible solution.

Additionally, we can improve this by:
1. Sending protocol strategies to http searchers and matching them with strategies received back from searchers .
2. Give option to searchers to define Vec<Strategy> instead of single strategy.

The case where searchers send solution and want to submit with flashbots or nothing would be best solved with (1) IMO.

### DIFF
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -185,6 +185,8 @@ pub struct SettledBatchAuctionModel {
     pub approvals: Vec<ApprovalModel>,
     #[serde(default)]
     pub interaction_data: Vec<InteractionData>,
+    #[serde(default)]
+    pub submitter: SubmissionPreference,
     pub metadata: Option<SettledBatchAuctionMetadataModel>,
 }
 
@@ -390,6 +392,14 @@ pub enum InternalizationStrategy {
     EncodeAllInteractions,
     #[serde(rename = "Enabled")]
     SkipInternalizableInteraction,
+}
+
+/// Searchers can override submission strategy of the protocol
+#[derive(Debug, Clone, Default, Deserialize, PartialEq, Eq, Serialize)]
+pub enum SubmissionPreference {
+    #[default]
+    ProtocolDefault,
+    Flashbots,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


crates/solver/src/settlement_submission.rs
```diff
@@ -176,9 +176,24 @@ impl SolutionSubmitter {
...
         }
 
+        // combine protocol enabled strategies with settlement prefered strategies
+        let strategies = match settlement.submitter {
+            SubmissionPreference::ProtocolDefault => self.transaction_strategies.as_slice(),
+            SubmissionPreference::Flashbots => {
+                let position = self
+                    .transaction_strategies
+                    .iter()
+                    .position(|strategy| matches!(strategy, TransactionStrategy::Flashbots(_)));
```
nlordell, line 186: I think this simple approach is a great start!

One issue I foresee is that this requires Flashbots to be enabled as a strategy - which affects all of our settlements and not just the ones that have `SubmissionPreference::Flashbots`. We currently have this disabled, but is this something we plan on enabling generally?

sunce86: For now, this is a more safe/strict solution, meaning solvers can express their preference only for strategies that are currently enabled by the protocol.
But we can definitely change this logic and give searchers more power - to override whatever protocol decided. However I was not sure we want this, because if we do and Flashbots are slow, it would affect our protocol throughput (at least until driver solver colocation is done where we have defined deadlines, penalties and so on)

nlordell: What I was trying to suggest is that I don't think we can enable Flashbots without also routing all settlements through Flashbots (i.e. I'm not sure how we configure the solver so only settlements with `SubmissionPreference::Flashbots` use Flashbots).

The only way I could think of solving it would be have both a `--supported-submission-strategies` and a `--default-submission-strategies` flag.

crates/shared/src/http_solver/model.rs
```diff
@@ -392,6 +394,14 @@ pub enum InternalizationStrategy {
     SkipInternalizableInteraction,
 }
 
+/// Searchers can override submission strategy of the protocol
+#[derive(Debug, Clone, Default, Deserialize, PartialEq, Eq, Serialize)]
+pub enum SubmissionPreference {
+    #[default]
+    ProtocolDefault,
+    Flashbots,
```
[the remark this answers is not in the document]
sunce86: Added.

crates/solver/src/settlement_submission.rs
nlordell: (to be clear, I think this can be done in a **follow up PR** - just think this might be needed to achieve our desired configuration).

MartinquaXD (approved):

crates/solver/src/settlement.rs
```diff
@@ -420,7 +434,10 @@ impl Settlement {
     /// See SettlementEncoder::merge
     pub fn merge(self, other: Self) -> Result<Self> {
         let merged = self.encoder.merge(other.encoder)?;
-        Ok(Self { encoder: merged })
+        Ok(Self {
```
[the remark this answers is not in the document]
sunce86: Makes sense. Fixed.

PR #906, opened 2022-12-02
Add start block for eth-flow sync

fedgiac:
The eth flow autopilot component takes a fairly long time to sync (»10min) as it starts from block 0.
This PR makes it start from the deployment block of the settlement contract. Most of the code was already there and I just moved it around.

### Test Plan

No unit test as all components rely on the node.
First, run the autopilot with `export ETHFLOW_CONTRACT=0xde2B7709e70CAd3148465fdd8489a27670B68532`, `export LOG_FILTER=trace` and redirect output to `grep -i "$ETHFLOW_CONTRACT"`. As expected, `fromBlock` is ~0x600000.

Then, also set `export SKIP_EVENT_SYNC=true` and run the same tests as before. No tracing log appears until new blocks are found and indexed.

### Release notes

It still takes some time to index all events from scratch, on my machine it takes ~7min.

PR #937, opened 2022-12-10
Add indexing starting block unless other events are available

fedgiac:
PR #906 introduces an issue when starting the services: regardless of the currently available events, the autopilot now reindexes all eth-flow events at every restart.

This PR introduces a new way to create an `EventHandler` that only prevents indexing old events but does not trigger a reindexing if newer events are available.

I also sneaked in a flag for specifying the start block of the eth-flow, if you prefer I can do it in another PR.

### Test Plan

New unit test. (However it's ignored by default because it requires a node and relies on events being generated in the settlement contracts at the time of running.)

### DIFF
--- a/crates/shared/src/event_handling.rs
+++ b/crates/shared/src/event_handling.rs
@@ -33,7 +33,7 @@ const MAX_PARALLEL_RPC_CALLS: usize = 128;
 /// 3. If this range is too big, split it into two subranges, one to update the deep history blocks, second one
 /// to update the latest blocks (last X canonical blocks)
 /// 4. Do the history update, and if successful, update `last_handled_blocks` to make sure the data is consistent.
-/// 5. If history update is successful, procceed with latest update, and if successful update `last_handled_blocks`.  
+/// 5. If history update is successful, procceed with latest update, and if successful update `last_handled_blocks`.
 pub struct EventHandler<C, S>
 where
     C: EventRetrieving,
@@ -78,6 +78,7 @@ pub trait EventRetrieving {
     fn get_events(&self) -> AllEventsBuilder<DynTransport, Self::Event>;
 }
 
+#[derive(Debug)]
 struct EventRange {
     /// Optional block number range for fetching reorg safe history
     history_range: Option<RangeInclusive<u64>>,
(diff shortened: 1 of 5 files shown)


nlordell (approved):
LGTM.

crates/shared/src/event_handling.rs
```diff
@@ -111,6 +112,29 @@ where
...
+        contract: C,
+        store: S,
+        skip_blocks_before: BlockNumberHash,
+    ) -> Result<Self> {
+        let last_handled_block = store.last_event_block().await?;
+        Ok(Self::new(
+            block_retriever,
+            contract,
+            store,
+            if last_handled_block >= skip_blocks_before.0 {
```
MartinquaXD, line 130: I believe if you would move this logic into `determine_ethflow_indexing_start()` you wouldn't need any new way to instantiate the `EventUpdater` or `EventHandler` (`new_skip_blocks_before()`).

fedgiac: Imho this is something we might need in the future for other contracts and not custom configuration for the eth-flow. On the other hand, if you prefer I can move this logic and it should work.

MartinquaXD: I'm just a bit hesitant to introduce this kind of specialization which could also be handled with the general case.
You could also consider dropping the changes to `EventHandler::new_skip_blocks_before()` and make the block selection logic part of `EventUpdated::new_skip_blocks_before()`. WDYT?

josojo (approved):
Sorry, that the issue only came up now and that I did not realize this before in staging.

One alternative solution could have been:
Simply handing over the deployment transaction as an env and then letting the code find its automatic starting date for the indexing. Since the deployment tx blocknumber <= last updated tx, we could also not run in any issues.

But your solution is even more powerful. I think its good.

crates/shared/src/event_handling.rs
josojo: >Imho this is something we might need in the future for other contracts and not custom configuration for the eth-flow.

do you have a concrete example in mind? 
Otherwise, I would also slightly lean putting the logic into determine_ethflow_indexing_start()

(but this is only a slight preference. should not block merging)

MartinquaXD (approved):

PR #938, opened 2022-12-12
Separate Configuration For Limit Order Update Interval

nlordell:
This PR adds a new `--surplus-fee-update-interval` flag.

Currently the surplus fee update interval is determined by the surplus fee expiry configuration, but it makes sense to have this be configured by a separate knob. Specifically, we may want to update more often in order to have limit orders get included in the auction sooner, without needing to make the limit order surplus fee expiry shorter.

The default value is set to 60s, so 30s shorter than the current default configuration.

### Test Plan

Run the autopilot and see that the the configuration option is set:
```
% cargo run -p autopilot
...
surplus_fee_update_interval: 60s
...
```

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -107,6 +107,15 @@ pub struct Arguments {
     )]
     pub max_surplus_fee_age: Duration,
 
+    /// The interval between successive limit order surplus fee updating loops in seconds.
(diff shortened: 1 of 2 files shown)


MartinquaXD (approved):
LGTM.
Since the quoter only quotes outdated orders it's probably fine to decrease the update interval even further.

PR #942, opened 2022-12-12
[refunder] Gas bump on priority fee

josojo:
Previously I implemented the gas bumping on the max_fee_per_gas of an EIP1559 tx - silly me. 
But actually, one needs to bump the tip of a tx and not the max_fee_per_gas, in order to replace a tx.

### Test Plan
unit tests


vkgnosis (approved):

nlordell (approved):

sunce86:
Are you sure about this?
AFAIR, most of the clients require both `max_fee_per_gas` and `max_priority_fee_per_gas` to be bumped. Example:
https://github.com/ethereum/go-ethereum/blob/c4a662176ec11b9d5718904ccefee753637ab377/core/txpool/list.go#L298-L303

josojo:
Thank you, Dusan for the rescue!

I was just following here: https://docs.alchemy.com/docs/retrying-an-eip-1559-transaction, but for sure the client code does not lie. I will implement a bump in both then

crates/refunder/src/submitter.rs
```diff
@@ -143,37 +120,52 @@ impl Submitter {
...
+    new_priority_fee_per_gas = f64::min(new_priority_fee_per_gas, new_max_fee_per_gas);
     let gas_price = GasPrice::Eip1559 {
-        max_fee_per_gas,
-        // The max_priority_fee_per_gas should normally allways be equal to MAX_PRIORITY_FEE_TIP
-        // Just in case the max_fee_per_gas is even lower than the constant, we need to adapt it
-        max_priority_fee_per_gas: U256::min(max_fee_per_gas, U256::from(MAX_PRIORITY_FEE_TIP)),
+        max_fee_per_gas: U256::from_f64_lossy(new_max_fee_per_gas),
+        max_priority_fee_per_gas: U256::from_f64_lossy(new_priority_fee_per_gas),
     };
     Ok(gas_price)
```
sunce86, line 170:
```suggestion
    fn calculate_submission_gas_price(
    gas_price_of_last_submission: Option<GasPrice1559>,
    web3_gas_estimation: GasPrice1559,
    newest_nonce: U256,
    nonce_of_last_submission: Option<U256>,
) -> Result<GasPrice1559> {
    let mut new_gas_price = web3_gas_estimation.bump(GAS_PRICE_BUFFER_FACTOR);
    new_gas_price.max_priority_fee_per_gas =
        (START_PRIORITY_FEE_TIP as f64).min(new_gas_price.max_fee_per_gas); // limit to max fee

    // If tx from the previous submission was not mined,
    // we incease the tip and max_gas_fee for miners
    // in order to avoid "tx underpriced errors"
    if Some(newest_nonce) == nonce_of_last_submission {
        let gas_price_of_last_submission = gas_price_of_last_submission
            .map(|gas_price| gas_price.bump(GAS_PRICE_BUMP))
            .unwrap_or_default();

        new_gas_price.max_fee_per_gas = new_gas_price
            .max_fee_per_gas
            .max(gas_price_of_last_submission.max_fee_per_gas);
        new_gas_price.max_priority_fee_per_gas = new_gas_price
            .max_priority_fee_per_gas
            .max(gas_price_of_last_submission.max_priority_fee_per_gas);
    }

    if new_gas_price.max_fee_per_gas > MAX_GAS_PRICE as f64 {
        tracing::warn!(
            "Refunding txs are likely not mined in time, as the current gas price {:?} \
                            is higher than MAX_GAS_PRICE specified {:?}",
            new_gas_price.max_fee_per_gas,
            MAX_GAS_PRICE
        );
        new_gas_price.limit_cap(MAX_GAS_PRICE as f64);
    }

    Ok(new_gas_price)
}
```

sunce86: nit: How about `calculate_submission_gas_price` working with `GasPrice1559` only. Also `gas_parameters_of_last_tx` could be changed to `Option<GasPrice1559>`. I think this struct has a way better interface than `ethcontract::GasPrice`.

Then, you would just convert into `ethcontract::GasPrice` once you need to send the transaction, using `into_gas_price` function, but you would probably need to move it to `shared` crate instead of `solver` crate where it is now.

crates/refunder/src/submitter.rs
```diff
@@ -22,28 +22,27 @@ use shared::{
...
-// Max priority fee that the refunder is willing to pay. (=2 Gwei)
-const MAX_PRIORITY_FEE_TIP: u64 = 2_000_000_000;
-// In order to resubmit a new tx with the same nonce, the gas price needs to
-// be increased by at least 10 percent. We increase it by 12 percent
+// Starting priority fee that the refunder is willing to pay. (=2 Gwei)
+const START_PRIORITY_FEE_TIP: u64 = 2_000_000_000;
+
+// In order to resubmit a new tx with the same nonce, the gas tip and
+// max_fee_per_gas needs to be increased by at least 10 percent.
 const GAS_PRICE_BUMP: f64 = 1.12;
```
sunce86, line 38: Just for the consistency, I would change this factor to `1.125` since we are using the same factor in `solver` crate for the same thing.

sunce86 (approved):
Looks good. Just a small nit to use `GasPrice1559` only if you want to pretty format the code.

crates/refunder/src/submitter.rs
josojo: to be honest, I don't like GasPrice1559 as I don't need `base_fee_per_gas`. but seeing how nicely this lib can be used to simplify the code, it was a must-go. Thanks for the re-write suggestions!

PR #943, opened 2022-12-12
Do not block onchain order parsing on unknown events

fedgiac:
(A rehashing of #934, if you are familiar with it just read the section **This PR** and keep in mind that the e2e test is copied from there.)

### Context

We pick up orders from the eth-flow contract by interpreting it as an instance of a generic wrapper contract `CoWSwapOnchainOrders`. This generic contracts defines events that should be interpreted as the creation of valid orders, directly from the blockchain.
However, the eth-flow contract can also emit other events that are not part of the `CoWSwapOnchainOrders` interface (namely order refunds). In the current implementation, our `CoWSwapOnchainOrders` contract instance (representing the contract at the eth-flow address) tries to interpret these events as an event that belongs to `CoWSwapOnchainOrders` itself and, when failing, returns an error that stalls the process of picking up events from the blockchain.

### This PR

Instead of deriving an event reader directly from the contract, we build a custom implementation that filters out events that aren't part of the contract ABI. This makes for a more localized and efficient implementation of #934.

(It looks different but it's somewhat the same as what suggested by Nick [here](https://github.com/cowprotocol/services/pull/934#issuecomment-1344306685).)

### Test Plan

A new e2e test is introduced to simulate the bad behavior.

### DIFF
--- a/crates/autopilot/src/database/onchain_order_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events.rs
@@ -1,4 +1,5 @@
 pub mod ethflow_events;
+pub mod event_retriever;
 
 use super::{
     events::{bytes_to_order_uid, meta_to_event_index},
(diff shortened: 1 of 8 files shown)


josojo (approved):
Super nice! I really like the solution.

vkgnosis (approved):

nlordell (commented):
Late to the party but changes look great!

PR #944, opened 2022-12-12
[Refunder] Ignore uids with zero owner for refunding

josojo:
The refunder checks the owner of a Uid in the earthflow contract. Normally, this owner is a non-zero value, as the 0 address can not place orders, and invalided orders have a non-zero uid.

But since we migrated the contracts, and the refunder tried to refund non-existing orders, this owner field has become zero for some orders. Hence, I decided to filter these cases out and throw a warning. 

(But this still is no perfect solution for contract migrations)

### DIFF
--- a/crates/refunder/src/refund_service.rs
+++ b/crates/refunder/src/refund_service.rs
@@ -175,9 +175,6 @@ impl RefundService {
                 invalid_uids
             );
         }
-        // Invalid uids will also be marked as refunded, as otherwise, the service will revisit
-        // these invalid uids every loop.
-        refunded_uids.append(&mut invalid_uids);
         let result = SplittedOrderUids {
             refunded: refunded_uids,
             to_be_refunded: to_be_refunded_uids,
--- a/crates/refunder/src/refund_service.rs
+++ b/crates/refunder/src/refund_service.rs
@@ -29,6 +29,13 @@ pub struct RefundService {
     pub submitter: Submitter,
 }
 
+#[derive(Debug, Eq, PartialEq)]
+enum RefundStatus {
+    Refunded,
+    NotYetRefunded,
+    Invalid,
+}
+
 struct SplittedOrderUids {
     refunded: Vec<OrderUid>,
     to_be_refunded: Vec<OrderUid>,
@@ -63,7 +70,7 @@ impl RefundService {
         let refundable_order_uids = self.get_refundable_ethflow_orders_from_db().await?;
 
         let order_uids_per_status = self
-            .identify_already_refunded_order_uids_via_web3_calls(refundable_order_uids)
+            .identify_uids_refunding_status_via_web3_calls(refundable_order_uids)
             .await?;
 
         self.update_already_refunded_orders_in_db(order_uids_per_status.refunded)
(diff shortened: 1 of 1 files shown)


crates/refunder/src/refund_service.rs
```diff
@@ -136,25 +143,42 @@ impl RefundService {
...
                 }
             })
             .collect::<Vec<_>>();
 
         batch.execute_all(MAX_BATCH_SIZE).await;
         let uid_with_latest_refundablility = futures::future::join_all(futures).await;
-        type TupleWithRefundStatus = (Vec<(OrderUid, bool)>, Vec<(OrderUid, bool)>);
-        let (refunded_uids, to_be_refunded_uids): TupleWithRefundStatus =
+        type TupleWithRefundStatus = (Vec<(OrderUid, RefundStatus)>, Vec<(OrderUid, RefundStatus)>);
+        let (refunded_uids, uid_with_latest_refundablility): TupleWithRefundStatus =
```
vkgnosis, line 160: Instead of partitioning multiple types I think it would be cleaner to create the three vectors you want then iterate once and add the results to the correct vec.

vkgnosis (approved):

crates/refunder/src/refund_service.rs
```diff
@@ -136,25 +143,42 @@ impl RefundService {
...
         let refunded_uids: Vec<OrderUid> = refunded_uids.into_iter().map(|(uid, _)| uid).collect();
         let to_be_refunded_uids: Vec<OrderUid> = to_be_refunded_uids
             .into_iter()
             .map(|(uid, _)| uid)
             .collect();
+        if !invalid_uids.is_empty() {
+            // In exceptional cases, e.g. if the refunder tries to refund orders from a previous contract,
+            // the order_owners could be zero
+            tracing::warn!(
+                "Following ethflow uids have a invalid owner (=0x0..0) stored onchain: {:?}",
```
fedgiac, line 178: Same nit: it sounds like having a zero owner is in general unexpected behavior, but this is fine and the real issue is that we try to invalidate some with zero owners. My suggestion:

```suggestion
                "Trying to invalidate orders that weren't created in the current contract. Uids: {:?}",
```

fedgiac (approved):
The solution looks good to make the refunder work, but wouldn't we have constant warnings at every loop if any bad uid is encountered at some point? My suggestion would be to also return these UIDs and mark them as refunded in the database (it's a bit cheating, but in principle if this happen we can monitor the warnings and manually trigger a refunds for these cases).

nlordell (approved):
LGTM.

Agree with @vkgnosis's suggestion of avoiding multiple `partition`s.

crates/refunder/src/refund_service.rs
nlordell: What you can do instead of looping is to use a `Itertools::group_by` on the refund status. Then you would get a `HashMap<RefundStatus, Vec<OrderUid>>`

josojo:
>My suggestion would be to also return these UIDs and mark them as refunded in the database

Hm. I don't like the API to return false information. This can easily end with a shoot into the foot. My plan was to manually delete these old orders.

PR #949, opened 2022-12-13
Allow multiple `Interaction`s per `Trade`

### LINKED ISSUE #866: Support multiple interactions when asking for a "binding" quote.
Solver might require multiple interactions to settle a trade efficiently. To allow for that the `Trade` type needs to hold a `Vec<Interaction>` instead of a single one. 
Then we could turn a settlement generated by a solver into a `Trade` to use it with the trade verifying API.
This requires slight changes when we turn `0x`, `1Inch` or `Paraswap` `/swap` requests into `Trade`s as well as a tiny change in `Trade::encode()`.

MartinquaXD:
Fixes #866 

Now we allow a `Trade` to include more than 1 `Interaction`. This is useful for solvers which might need more than one `Interaction` to settle a trade optimally for a `/quote` request.

To avoid unachievable quotes that settle with internal buffers of the contract we require at least 1 "main" interaction. However, this could easily be circumvented. The correct solution would be to solve that within the trade simulation logic itself.

### Test Plan
Updated existing unit test to cover encoding of multiple interactions.

### DIFF
--- a/crates/shared/src/price_estimation/trade_finder.rs
+++ b/crates/shared/src/price_estimation/trade_finder.rs
@@ -425,11 +425,11 @@ mod tests {
             out_amount: 2_000_000_u128.into(),
             gas_estimate: 133_700,
             approval: None,
-            interaction: Interaction {
+            interactions: vec![Interaction {
                 target: H160([0x7; 20]),
                 value: 0_u64.into(),
                 data: vec![1, 2, 3, 4],
-            },
+            }],
         };
         let sell_token_code = bytes!("05060708");
 
@@ -564,6 +564,7 @@ mod tests {
         };
         let trade = Trade {
             out_amount: 1_000_000_u128.into(),
+            interactions: vec![Default::default()],
             ..Default::default()
         };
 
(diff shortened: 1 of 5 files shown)


vkgnosis (approved):

sunce86 (approved):

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -425,11 +425,11 @@ mod tests {
             out_amount: 2_000_000_u128.into(),
             gas_estimate: 133_700,
             approval: None,
```
nlordell, line 427: One issue I foresee is that multiple approvals may be needed for each interaction. I think I'm OK with having it like this for now (as in theory, the `driver` would encode the `approval` as an interaction) **but**, in that case, I think that the JSON model that we parse from the driver should disallow approvals being sent.

nlordell (approved):
Looks good, just see an issue with the approvals. Maybe the "driver trade finder" should include something like `ensure!(trade.approval.is_none(), "driver trades do not allow approvals")`.

nlordell:
Superseded by #1504

PR #951, opened 2022-12-13
Handle base tokens in `LiquidityCollector`

### LINKED ISSUE #900: Move decision which `TokenPair`s to get liquidity for into the `LiquidityCollector`
At the moment all the individual components that implement liquidity fetching (balancer, uniswap, zeroex, ...) take a list of base tokens. When they get asked to fetch liquidity for a `&[TokenPair]` they compute a set of relevant pairs based on the base tokens and collect liquidity for those (e.g. [univ2](https://github.com/cowprotocol/services/blob/main/crates/solver/src/liquidity/uniswap_v2.rs#L80-L84)).
This is identical in all the components and can be moved into the `LiquidityCollector` to do it once and at a single place.

MartinquaXD:
Fixes #900 

This is just a small refactor to remove a bit of duplicated logic.
Instead of collecting relevant token pairs based on the `BaseTokens` in each `LiquidityCollecting` implementation we can simply do it once in the `LiquidityCollector`.

### Test Plan
Existing unit tests still pass

### DIFF
--- a/crates/driver/src/auction_converter.rs
+++ b/crates/driver/src/auction_converter.rs
@@ -11,6 +11,7 @@ use {
         solver::Auction,
     },
     std::{
+        collections::HashSet,
         sync::{
             atomic::{AtomicU64, Ordering},
             Arc,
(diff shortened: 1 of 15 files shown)


sunce86 (approved):
Nice!

vkgnosis (approved):

PR #953, opened 2022-12-13
Consider that access list makes tx possible in the first place

### LINKED ISSUE #925: The `solver` isn't simulating with access lists everywhere it needs to

MartinquaXD:
Fixes #925 

The problem was that the code in the `Submitter` didn't expect that a failing transaction could be made executable by adding an access list. Instead of `try_join` we now always simulate both transactions (with and without access list) and return the access list correctly based on which one was successful.

### Test Plan
Ran `orderbook`, `autopilot` and `solver` locally and settled an order that sells `COW` for `ETH` and sends it to my `goerli` test safe ([tx](https://goerli.etherscan.io/tx/0x5371bdc380eb27281f4f9fb92338b7308e26c067a1fe3762cf942aa93a88e7aa)).

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -526,27 +526,41 @@ impl<'a> Submitter<'a> {
             tx,
         )
         .await?;
-        let (gas_before_access_list, gas_after_access_list) = futures::try_join!(
+        let (without_access_list, with_access_list) = futures::join!(
+            // This call will fail for orders paying ETH to SC wallets
             tx.clone().estimate_gas(),
             tx.clone().access_list(access_list.clone()).estimate_gas()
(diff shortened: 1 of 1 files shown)


sunce86 (approved):
Ah nice catch!

vkgnosis (approved):

PR #851, opened 2022-11-24
Update eth-flow contract to rc3

fedgiac:
Updates the eth-flow contract abi to match the [new release](https://github.com/cowprotocol/ethflowcontract/pull/41).

I needed to adjust the refunder test as it relied on the ability to create expired orders, which is not possible anymore.
Initially I wanted to make the time mockable in the refunder, but I'd also need to do it in the autopilot because of their mutual interactions so this is instead the best I came up with.

Note that time in a testing node can only be moved forward and never backward. In this PR I initialize the node to start in the past so that I can get some more flexibility when running the test.

### Test plan

Existing e2e tests.

PR #954, opened 2022-12-13
[refunder] refundable order calculation should depend on blocktime

josojo:
The refunder showed some tricky patterns:
It was using the CPU time to check which orders are refundable. Then it simulated the refund tx on the last block, but in the last block, the orders were not yet refundable, and hence, the simulation resulted in a very low gas limit estimation. 
If then the tx would hit on chain, the tx would use a lot more gas - as it would actually do the refunding. See [here](https://goerli.etherscan.io/address/0x2c53223ec0d536defd79a422b6dedf324b398bbd) for the failed tx.

This PR fixes this, as the refunder will only use the block-time and not the CPU time for the calculation of the refundable orders.

One nice side effect of this change is that we no longer need to do the complicated "node starting point from the past" introduce here: https://github.com/cowprotocol/services/pull/851

Testplan:
e2e tests

### DIFF
--- a/crates/e2e/tests/e2e/refunder.rs
+++ b/crates/e2e/tests/e2e/refunder.rs
@@ -101,6 +101,11 @@ async fn refunder_tx(web3: Web3) {
         ))
         .await
         .expect("Must be able to set block timestamp");
+    // mine next block to push time forward
+    web3.api::<TestNodeApi<_>>()
+        .mine_pending_block()
+        .await
+        .expect("Unable to mine next block");
 
     // Create the refund service and execute the refund tx
     let pg_pool = PgPool::connect_lazy("postgresql://").expect("failed to create database");
--- a/crates/e2e/tests/e2e/refunder.rs
+++ b/crates/e2e/tests/e2e/refunder.rs
@@ -13,8 +13,8 @@ use model::quote::{
 };
 use refunder::refund_service::RefundService;
 use shared::{
-    current_block::blockchain_time, ethrpc::Web3, http_client::HttpClientFactory,
-    maintenance::Maintaining,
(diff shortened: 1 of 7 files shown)


crates/e2e/tests/e2e/refunder.rs
```diff
@@ -90,22 +93,19 @@ async fn refunder_tx(web3: Web3) {
     // Run autopilot indexing loop
     services.maintenance.run_maintenance().await.unwrap();
 
-    // The blockchain is in the past for this test. We validate this assumption here, if this assertion fails then the
-    // original test conditions should be restored.
-    let now = chrono::offset::Utc::now().timestamp();
-    assert!(
-        (valid_to as i64) < now,
-        "Order should be expired from the point of view of the refunder."
-    );
```
fedgiac, line 99: I'm happy to see this go! 
Consider also changing `test_node_in_the_past` to `test` at the start of the test. (We should probably revert the new test infrastructure from #851.)

crates/e2e/tests/e2e/refunder.rs
```diff
@@ -90,30 +94,27 @@ async fn refunder_tx(web3: Web3) {
...
+        .expect("Unable to mine next block");
 
     // Create the refund service and execute the refund tx
     let pg_pool = PgPool::connect_lazy("postgresql://").expect("failed to create database");
     let mut refunder = RefundService::new(
         pg_pool,
         web3,
         contracts.ethflow.clone(),
-        i64::MIN, // Needs to be negative, as valid to was chosen to be in the past
+        validity_duration as i64 - 10,
```
fedgiac, line 117: I'm also very happy to see this go!

fedgiac (approved):
Nice!

PR #958, opened 2022-12-14
Driver `/solve` functionality part 4

sistemd:
Progress on #895.

Implements mapping at the `boundary`.

Currently, to facilitate progress, the domain objects are not what I want them to be ultimately. The domain model is somewhat [anemic](https://en.wikipedia.org/wiki/Anemic_domain_model), but this will be fixed as some of the interfaces become better-defined and as I move the scoring logic into the `driver`. (It will be a while before this; I will do this at some point as a refactoring _after_ I already have automated tests for the `driver`).

The next step is calling `SettlementRater` from the `score` method.

### DIFF
--- a/crates/driver/src/boundary/settlement.rs
+++ b/crates/driver/src/boundary/settlement.rs
@@ -128,7 +128,7 @@ fn into_boundary_order(order: &competition::auction::Order) -> Order {
                 competition::auction::order::Kind::Liquidity => OrderClass::Liquidity,
                 competition::auction::order::Kind::Limit { surplus_fee } => {
                     OrderClass::Limit(LimitOrderClass {
-                        surplus_fee: surplus_fee.into(),
+                        surplus_fee: Some(surplus_fee.into()),
                         ..Default::default()
                     })
                 }
@@ -146,6 +146,7 @@ fn into_boundary_order(order: &competition::auction::Order) -> Order {
             settlement_contract: Default::default(),
             ethflow_data: Default::default(),
             onchain_user: Default::default(),
+            onchain_order_data: Default::default(),
             is_liquidity_order: order.kind.is_liquidity(),
         },
         signature: Default::default(),
@@ -306,7 +307,7 @@ fn into_boundary_solution(solution: competition::Solution) -> SettledBatchAuctio
                     .collect(),
                 // TODO I have no clue why there's an exec plan here? This is a single interaction,
                 // what sort of ordering is needed? I don't know.
-                exec_plan: None,
+                exec_plan: Default::default(),
                 cost: None,
             })
             .collect(),


crates/driver/src/boundary/settlement.rs
```diff
@@ -0,0 +1,336 @@
...
+    let limit_orders = auction
+        .orders
+        .iter()
+        .filter_map(|order| {
+            let boundary_order = into_boundary_order(order);
+            match order_converter.normalize_limit_order(boundary_order) {
+                Ok(order) => {
+                    // TODO I don't think this matters?
+                    // The reward doesn't matter for objective value computation, right?
+                    // order.reward = auction.rewards.get(&uid).copied().unwrap_or(0.);
```
vkgnosis, line 56: Reward doesn't matter for objective value.

crates/driver/src/logic/eth/mod.rs
```diff
@@ -54,8 +58,82 @@ impl From<H160> for Address {
...
+
+/// Block number.
+#[derive(Debug, Clone, Copy)]
+pub struct BlockNo(pub u64);
+
+/// An onchain transaction which interacts with a smart contract.
+#[derive(Debug)]
+pub struct Interaction {
+    pub target: Address,
+    pub value: U256,
```
vkgnosis, line 137:
```suggestion
    pub value: Ether,
```
If we have this type then it makes sense to use here too.

crates/driver/src/logic/eth/mod.rs
```diff
@@ -54,8 +58,82 @@ impl From<H160> for Address {
...
+#[derive(Debug, Clone, Copy)]
+pub struct Asset {
+    pub amount: U256,
+    pub token: Token,
+}
+
+// TODO Do some web browsing to improve this doc comment.
+/// An amount of native Ether tokens.
+#[derive(Debug, Clone, Copy)]
+pub struct Ether(pub U256);
```
vkgnosis, line 113: Should clarify that this is measured in wei (ether atoms). Technically 1e18 wei = 1 ether.

crates/driver/src/blockchain/mod.rs
```diff
@@ -44,10 +47,33 @@ impl Ethereum {
...
             .allowance(owner.0, spender.address.0)
             .call()
             .await?;
         Ok(eth::Allowance { spender, amount }.into())
     }
 }
+
+// TODO This could probably do some caching, I might want to open an issue for
+// this?
+pub struct Contracts<'a>(&'a Ethereum);
```
vkgnosis, line 60: It would be nice if the methods that return the contract addresses weren't async. The contract address is determined by the chain id and hard coded into the contract deployment info. There is no network request involved there.

crates/driver/src/logic/competition/auction/order.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub kind: Kind,
+    pub app_data: competition::AppData,
+    /// A partial order doesn't require the full amount to be traded. E.g. only
+    /// 10% of the requested amount may be traded, if this leads to the most
+    /// optimal solution.
+    pub partial: bool,
+    /// The autopilot marks orders as mature after a certain time period. The
+    /// solvers can use heuristics on this field to optimize solution sizes.
+    pub mature: bool,
+    pub executed: U256,
```
MartinquaXD, line 29: Would be nice to document that the executed amount refers to `sell` for sell orders and `buy` for buy orders.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -42,17 +45,20 @@ struct Order {
...
-    #[serde_as(as = "serialize::U256")]
-    fee_amount: U256,
     kind: OrderKind,
     partially_fillable: bool,
+
+    #[serde_as(as = "serialize::U256")]
+    fee_amount: U256,
+    #[serde_as(as = "serialize::U256")]
+    #[serde(default)]
+    full_fee_amount: U256,
```
nlordell, line 55: Should this be:
```suggestion
    scoring_fee_amount: U256,
```

Or something to indicate that this is the fee amount that we use for scoring?

crates/driver/src/blockchain/mod.rs
nlordell: Oh yeah! That is a nice idea, we can have `Contracts(Ethereum, Network)` and then the contract accessors would be sync and not require a `Result`:
```rust
let address = contracts::GPv2Settlement::raw_contract().get(&self.1).unwrap().address;
contracts::GPv2Settlement::at(&self.0.web3, address)
```

crates/driver/src/logic/competition/auction/order.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// The onchain calls necessary to fulfill this order. These are set by the
+    /// user and included in the settlement transaction.
+    pub interactions: Vec<eth::Interaction>,
+
+    pub sell_source: SellSource,
+    pub buy_destination: BuyDestination,
+
+    // TODO This is a placeholder, there should probably be a proper submodule for signing stuff
+    pub signature: i32,
```
nlordell, line 38: This is passed in. You shouldn't need to sign or verify signatures in the `driver` at al - you can just assume its some "bytes" that need to be encoded as-is.

crates/driver/src/api/solve/dto/auction.rs
nlordell: Oops! This is the dto 🤦

crates/driver/src/logic/competition/solution/mod.rs
```diff
@@ -1,40 +1,145 @@
...
+    pub sell: U256,
+    pub buy: U256,
+    // TODO Revisit after #831
+    pub plan: Option<ExecutionPlan>,
+    pub fee: Option<eth::Asset>,
+
+    // TODO This dictates where tokens are moved from, I guess I should document this and talk to
+    // Nick about whether we should change it
+    pub sell_source: SellSource,
+    pub buy_destination: BuyDestination,
```
nlordell, line 87: Shouldn't be needed in the solution. This is an inherent property to orders, but handled internally by the Smart Contract. The driver just needs to make sure this is encoded in the settlement calldata, but this has no bearing on actually computing the solution.

crates/driver/src/logic/competition/solution/mod.rs
```diff
@@ -1,40 +1,145 @@
...
+pub enum BuyDestination {
+    /// Direct ERC20 allowances to the Vault relayer contract
+    Erc20,
+    /// ERC20 allowances to the Vault with GPv2 relayer approval
+    Internal,
+}
+
+/// An order which was executed as part of this solution.
+#[derive(Debug)]
+pub struct Order {
```
nlordell, line 77: I think this can be simplified a bit. The only information that we need from a solver is "how much was executed". Everything else is an inherent property to the order that doesn't need to be echoed here.

vkgnosis (approved):

nlordell (approved):

crates/driver/src/logic/competition/solution/mod.rs
nlordell: My point is that this is included in the order model between `protocol` and `driver` but is irrelevant for the order model between `driver` and `solver/searcher`.

nlordell: This changes with #962, so I don't want to block the merge on it.

PR #960, opened 2022-12-14
Improve flanky refunder test

josojo:
The refunder test is flanky. E.g. [here](https://github.com/cowprotocol/services/actions/runs/3695869171/jobs/6258798439) it failed on main branch.

The reason is probably that the time of the blockchain proceeds too much between these commands: 
```
let valid_to = timestamp_of_current_block_in_seconds(&web3).await.unwrap() + validity_duration;
    // Accounting for slippage is necesary for the order to be picked up by the refunder
    let ethflow_order =
        ExtendedEthFlowOrder::from_quote(&quote_response, valid_to).include_slippage_bps(9999);
    ethflow_order
        .mine_order_creation(&user, &contracts.ethflow)
        .await;
```
such that the validity_duration - 10 - this is the time between mining and valid_to -  was too strict:
```
RefundService::new(..., validity_duration -10, ...)
 ```
 
I fixed it by being less strict about it.

### Test Plan
None

### DIFF
--- a/crates/e2e/tests/e2e/refunder.rs
+++ b/crates/e2e/tests/e2e/refunder.rs
@@ -81,7 +81,7 @@ async fn refunder_tx(web3: Web3) {
     assert_eq!(quoting.status(), 200);
     let quote_response = quoting.json::<OrderQuoteResponse>().await.unwrap();
 
-    let validity_duration = 60;
+    let validity_duration = 600;
     let valid_to = timestamp_of_current_block_in_seconds(&web3).await.unwrap() + validity_duration;
     // Accounting for slippage is necesary for the order to be picked up by the refunder
     let ethflow_order =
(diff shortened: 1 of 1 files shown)


fedgiac:
I think the issue might be related to the fact that the logic of the refunder is not independent of system time (not because of its own code but because of the dependency of it on the autopilot (and in turn the database). So orders too much in the past are not picked up. (Not sure why, conjecture: the quote is deleted based on system time and so the order isn't picked up.)
What I think happened [here](https://github.com/cowprotocol/services/actions/runs/3695869171/jobs/6258798439) is that the node started at the time of running `Setup Hardhat`, then a few other CI steps happened that took 3.5 minutes, meaning that the node was too much in the past for the run to be successful (compare also with [this successful run](https://github.com/cowprotocol/services/actions/runs/3696239611/jobs/6259671948) that took 30 seconds less).
So I suppose this change fixes the issue as long as our CI takes no more than `600-60` seconds compared to now.

fedgiac (approved):
Anyway, as a stop gap solution this works.

josojo:
>So orders too much in the past are not picked up
I think they are picked up, but with the system time and not with the blocktime?

Probably the proper fix is to make the autopilot indexing process also independent of the system time. It's an issue in our overview issue. But I will merge it as I think it improves the flakiness significantly

PR #964, opened 2022-12-15
Fixing conflict in reorg handling by autopilot for ethflow

josojo:
Currently, we see on staging the following error:
```
WARN maintenance{block=16189813}: shared::maintenance: Service Maintenance Error: insert_orders failedCaused by:
    0: error returned from database: current transaction is aborted, commands ignored until end of transaction block
    1: current transaction is aborted, commands ignored until end of transaction block
 ```
 
 I think this is due to the fact that a first insert_order returns an error - duplication of uid - which we ignore. Then we try to proceed with another query. But since it's a db-transaction, and not a single query, it keeps on telling that the last command produced already an error. 
 This PR makes sure that the first insertion does not even fail.
 (see here for more context: https://stackoverflow.com/questions/10399727/psqlexception-current-transaction-is-aborted-commands-ignored-until-end-of-tra)

### Test Plan
test only

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -157,16 +157,14 @@ pub async fn insert_orders_and_ignore_conflicts(
     orders: &[Order],
 ) -> Result<(), sqlx::Error> {
     for order in orders {
-        match insert_order(ex, order).await {
-            Ok(_) => (),
-            Err(err) if is_duplicate_record_error(&err) => (),
-            Err(err) => return Err(err),
-        }
+        insert_order(ex, order).await?;
     }
     Ok(())
 }
 
 pub async fn insert_order(ex: &mut PgConnection, order: &Order) -> Result<(), sqlx::Error> {
+    // Since each order has a unique UID even after a reorg onchain placed orders have the same
+    // data. Hence, we can disregard any conflicts.
     const QUERY: &str = r#"
 INSERT INTO orders (
     uid,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by fedgiac** [comment]
Some preliminary comments: note that this changes conflicts with the idea of using the block timestamp for `creation_timestamp` (as it should be updated on reorgs).
For the other parameters it looks reasonable except maybe these that I don't know `surplus_fee` `surplus_fee_timestamp`.

File: crates/database/src/orders.rs [hunk 1]
```diff
@@ -850,15 +849,16 @@ mod tests {
...
-    async fn postgres_insert_same_order_twice_fails() {
+    async fn postgres_insert_same_order_twice_results_in_only_one_order() {
         let mut db = PgConnection::connect("postgresql://").await.unwrap();
         let mut db = db.begin().await.unwrap();
         crate::clear_DANGER_(&mut db).await.unwrap();
 
         let order = Order::default();
         insert_order(&mut db, &order).await.unwrap();
-        let err = insert_order(&mut db, &order).await.unwrap_err();
-        assert!(is_duplicate_record_error(&err));
```
nlordell, line 861: I think we currently rely on the current behaviour in `POST /orders` (and maybe a few other places as well).

I think we may need to specialize the query for:
- Orders placed through the API
- On-chain orders

josojo: Federico and I concluded the same. On its way!

vkgnosis (approved):

nlordell (approved):
LGTM.

Nice reuse of the query and SQLx code!

crates/database/src/orders.rs
```diff
@@ -195,7 +190,25 @@ INSERT INTO orders (
...
-    sqlx::query(QUERY)
+
+pub async fn insert_order_and_ignore_conflicts(
+    ex: &mut PgConnection,
+    order: &Order,
+) -> Result<(), sqlx::Error> {
+    // To be used only for the ethflow contract order placement, where reorgs force us to update
+    // orders
+    // Since each order has a unique UID even after a reorg onchain placed orders have the same
+    // data. Hence, we can disregard any conflicts.
```
fedgiac, line 201: I'd move this comment to where the function is used, since the name is clear enough.

fedgiac (approved):

PR #965, opened 2022-12-15
Prevent limit order fee updater from getting stuck

vkgnosis:
This is a theoretical issue I noticed while thinking about limit orders as part of oncall.

Currently, the limit order fee updater retrieves orders with outdated fees in an arbitrary ordering because the sql query does not include an ORDER BY clause. This is problematic because it is possible that the fee estimate for an order keeps failing forever, for example because the quote is out of price. If this happened for all of the 100 orders that the query returns, then the updater would never make progress.

This PR fixes this issue by slightly changing the semantics of the surplus_fee and surplus_fee_timestamp columns. When updating the fee fails, surplus_fee is set to null and surplus_fee_timestamp is updated to the current time. With this change we can return the limit orders with outdated fees ordered by surplus_fee_timestamp, giving us the most outdated orders first.

As part of this I am categorizing errors into preventable and unpreventable to improve logging and metrics.

---

The diff is hard to read because I had to move a lot of code around even though the logic hasn't changed that much. It is more understandable side-by-side instead of inline.

### Test Plan

adjusted the db unit tests, there is an e2e test for limit orders and will watch logs afterwards

### DIFF
--- a/crates/autopilot/src/database/orders.rs
+++ b/crates/autopilot/src/database/orders.rs
@@ -93,6 +93,11 @@ impl Postgres {
                 }),
             ),
             FeeUpdate::Failure { timestamp } => (
+                // Note that the surplus fee must be removed so that the order does not count as
+                // solvable. In order to be solvable the timestamp must be recent and the fee must
+                // be set. We don't reset the timestamp because it indicates the last update time
+                // (regardless of error or success). This is needed so that we can query the least
+                // recently updated limit orders. See #965 .
                 database::orders::FeeUpdate {
                     surplus_fee: None,
                     surplus_fee_timestamp: *timestamp,
--- a/crates/autopilot/src/database/orders.rs
+++ b/crates/autopilot/src/database/orders.rs
@@ -1,6 +1,6 @@
 use super::Postgres;
-use anyhow::Result;
-use chrono::{Duration, Utc};
+use anyhow::{Context, Result};
+use chrono::{DateTime, Duration, Utc};
 use database::byte_array::ByteArray;
 use ethcontract::U256;
 use futures::{StreamExt, TryStreamExt};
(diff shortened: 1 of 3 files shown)


nlordell (approved):
LGTM! Nice catch.

crates/database/src/orders.rs
```diff
@@ -600,9 +603,11 @@ pub fn limit_orders_with_outdated_fees(
         ORDERS_FROM,
         " WHERE o.valid_to >= $1 ",
         "AND o.class = 'limit' ",
-        "AND o.surplus_fee_timestamp < $2 ",
+        "AND COALESCE(o.surplus_fee_timestamp, 'epoch') < $2 ",
+        "ORDER BY o.surplus_fee_timestamp ASC NULLS FIRST",
```
MartinquaXD, line 607: I'm not 100% up to date on this but do we need `surplus_fee_timestamp` to be nullable. If it's not nullable we could create a btree index for `surplus_fee_timestamp` to speed up finding the most outdated fees.

crates/autopilot/src/database/orders.rs
```diff
@@ -62,38 +70,52 @@ impl Postgres {
...
+                    gas_amount: quote.fee_parameters.gas_amount,
+                    gas_price: quote.fee_parameters.gas_price,
+                    sell_token_price: quote.fee_parameters.sell_token_price,
+                    sell_amount: u256_to_big_decimal(&quote.sell_amount),
+                    buy_amount: u256_to_big_decimal(&quote.buy_amount),
+                }),
+            ),
+            FeeUpdate::Failure { timestamp } => (
+                database::orders::FeeUpdate {
+                    surplus_fee: None,
```
MartinquaXD, line 97: This seems a bit risky. With the timestamp now being updates `surplus_fee == None` is the only thing preventing this limit order from being part of the solvable orders.
Maybe we could add a comment here saying that `surplus_fee` absolutely has to stay `None` in this case.

crates/database/src/orders.rs
vkgnosis: It is currently nullable. We could make it not nullable by using a placeholder timestamp like epoch to indicate there has never been an update. This doesn't matter for btree indexes though because they work fine with null values too. However, an index on surplus_fee does not help here because we also need to determine that the order is still valid (valid_to > now). We have an index on that and after applying the first index, the second one is no longer useful.

crates/autopilot/src/database/orders.rs
vkgnosis: Added comment

PR #966, opened 2022-12-15
Add maintainers names to metrics

### LINKED ISSUE #819: Add `ServiceMaintaining` Names
In order to increase debuggability and observability of our various "maintenance" components, we should add a name to them so that we can better track their success and failure rates with metrics.

_Originally posted by @sunce86 in https://github.com/cowprotocol/services/pull/811#discussion_r1026402054_

sunce86:
Fixes #819 

For each failure, added the names of all the maintainers that failed, so we can better observe the system and make better alerts.
Also, @josojo brought this up again since we did not detect [this](https://github.com/cowprotocol/services/pull/964) failure on time (because we don't have appropriate metrics).

One consequence of this change is that:
1. `.reset()` is called inside `run_maintenance_for_blocks()` (this is the same as before this PR).
2. but `failure`.inc() is now called inside `run_maintenance()`.
Not super happy with this but could not find cleaner solution.

### Test Plan

UT passing.

### DIFF
--- a/crates/autopilot/src/database/quotes.rs
+++ b/crates/autopilot/src/database/quotes.rs
@@ -1,6 +1,6 @@
 use super::Postgres;
 use anyhow::{Context, Result};
-use shared::maintenance::{Maintainer, Maintaining};
+use shared::maintenance::Maintaining;
 use sqlx::types::chrono::{DateTime, Utc};
 
 impl Postgres {
@@ -24,7 +24,7 @@ impl Maintaining for Postgres {
             .context("fee measurement maintenance error")
     }
 
-    fn name(&self) -> Maintainer {
-        Maintainer::Postgres
+    fn name(&self) -> &str {
+        "Postgres"
     }
 }
--- a/crates/autopilot/src/database/quotes.rs
+++ b/crates/autopilot/src/database/quotes.rs
@@ -1,6 +1,6 @@
 use super::Postgres;
 use anyhow::{Context, Result};
-use shared::maintenance::Maintaining;
+use shared::maintenance::{Maintainer, Maintaining};
 use sqlx::types::chrono::{DateTime, Utc};
 
 impl Postgres {
@@ -22,5 +22,9 @@ impl Maintaining for Postgres {
         self.remove_expired_quotes(Utc::now())
             .await
             .context("fee measurement maintenance error")
+    }
+
+    fn name(&self) -> Maintainer {
+        Maintainer::Postgres
     }
 }
--- a/crates/shared/src/maintenance.rs
+++ b/crates/shared/src/maintenance.rs
@@ -46,11 +46,12 @@ impl Maintaining for ServiceMaintenance {
                     self.maintainers[i].name(),
                     err
                 );
-                no_error = false;
                 self.metrics
                     .runs
                     .with_label_values(&["failure", self.maintainers[i].name()])
                     .inc();
+
+                no_error = false;
             }
         }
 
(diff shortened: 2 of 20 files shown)


josojo:
Super nice Dusan. Thanks a lot. 

I think Valentin's work is also very nice, but I like the added feature that not just logic is improved, but also the metrics. This could be used specifically to have an alert for the ethflow-event updater.
(very sorry Valentin, that Dusan and I discussed it via dm and not in the chat. Next time we will do it in the channel again)

Going one step further would also help: Not just distinguish the errors by the maintenance service, but also by the error type: Web3 connection error, db-error, other error. Then frequent errors like timeout of web3 could easily be disregarded. But this would require introducing a more involved PR with a even bigger refactor.

crates/shared/src/maintenance.rs
```diff
@@ -34,19 +63,33 @@ impl Maintaining for ServiceMaintenance {
             if let Err(err) = result {
                 tracing::warn!("Service Maintenance Error: {:?}", err);
```
josojo, line 64: Could you also add the name of service in this error message?

josojo (approved):

crates/shared/src/maintenance.rs
```diff
@@ -2,20 +2,48 @@ use crate::current_block::{self, BlockInfo, CurrentBlockStream};
...
+    Postgres,
+    BalancerPoolFetcher,
+    UniswapV2PoolFetcher,
+    UniswapV3PoolFetcher,
+}
+
+impl Maintainer {
+    fn label(&self) -> &'static str {
+        match self {
+            Maintainer::ServiceMaintenance => "ServiceMaintenance",
```
MartinquaXD, line 24: `strum` also offers a macro to generate this boiler plate code.

MartinquaXD (approved):
LGTM.

crates/autopilot/src/database/quotes.rs
```diff
@@ -23,4 +23,8 @@ impl Maintaining for Postgres {
             .await
             .context("fee measurement maintenance error")
     }
+
+    fn name(&self) -> Maintainer {
```
nlordell, line 27: nit: I would actually make this a `&'static str`. My reasoning is that `Maintaining` implementations can also be defined by consumers of the `shared` crate, so conceptually, it feels weird that the dependency knows about implementation details of its dependents.

nlordell (commented):
Very welcome change!

Just one nit about the return type of `name`.

crates/autopilot/src/database/quotes.rs
sunce86: Indeed. Initially I set it to be `&str` but then I was worried because `&str` does not enforce name to be unique for different components across the system. New `Maintaining` implementation would need to check other implementations to see which names are occupied.
I don't have a strong opinion here since both ways are not perfect.

crates/shared/src/maintenance.rs
sunce86: Ah yes, forgot that. Thanks

crates/shared/src/maintenance.rs
sunce86: Didn't know. Fixed.

crates/shared/src/maintenance.rs
sunce86: Reopening since fix is not correct.

sunce86: Now it's fixed properly.

crates/autopilot/src/database/quotes.rs
sunce86: Switched back to `&str`

nlordell (approved):
LGTM, but I #967 does something very similar, and we should probably choose which one we want to use.

PR #967, opened 2022-12-15
Add more result context

vkgnosis:
While oncall we had maintenance fail with an unhelpful error message like "Bad Gateway". These contexts are supposed to help with that. Some of the messages could be better but I didn't want to invest the time to understand the surrounding code more and this is already enough to make it easier to trace the execution from the error message.

I moved the maintainer impl names into the ServiceMaintenance struct instead of the Maintenance trait because it we really only need the names in ServiceMaintenance and I didn't want to force it on all the implementations. Not completely clear what is better.

### Test Plan

next time these errors happen should have better message

### DIFF
--- a/crates/autopilot/src/lib.rs
+++ b/crates/autopilot/src/lib.rs
@@ -399,8 +399,11 @@ pub async fn main(args: arguments::Arguments) -> ! {
         block_retriever.clone(),
         skip_event_sync_start,
     ));
-    let mut maintainers: Vec<Arc<dyn Maintaining>> =
-        vec![pool_fetcher.clone(), event_updater, Arc::new(db.clone())];
+    let mut maintainers: Vec<(Arc<dyn Maintaining>, &'static str)> = vec![
+        (pool_fetcher.clone(), "pool_fetcher"),
+        (event_updater, "event_updater"),
+        (Arc::new(db.clone()), "db"),
+    ];
 
     let gas_price_estimator = Arc::new(InstrumentedGasEstimator::new(
         shared::gas_price_estimation::create_priority_estimator(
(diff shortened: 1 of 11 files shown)


nlordell (approved):
This is an alternative to #966 right?

One advantage of this implementation is that, if we add N different instances of `MySuperMaintenanceImplemantion`, then we can specify different names on adding them.

The advantage of #966 is that the name is determined by the implementer (which I kind of like).

I ever so slightly prefer #966 (with the suggested change of changing the name to a `&str`), but I like this solution as well.

Curious about what others think.

vkgnosis:
I made this PR as a reaction to a bad error message I saw during on call. At the time I didn't know about Dusan's PR.

I somewhat prefer assigning names in ServiceMaintenance rather than making them part of the trait. Maintenance implementations do not care about what their name. What cares about the name is ServiceMaintenance because it has metrics and it prints log messages about its child Maintining impls.

sunce86:
#966 purpose is to add metrics for each individual maintainer (so we have more control with alerts), which is not covered in this PR as far as I see (nor it is easy to add right?).

vkgnosis:
It would be easy to add these metrics into ServiceMaintenance. All we need is a new prometheus counter with a label for the name. ServiceMaintenance has all the names.

Another perspective is that the individual maintenance implementations should be responsible for their own metrics. I'm unsure how much sense it makes to force the same metrics strategy on all of them. I can imagine that the requirements could be pretty different.

vkgnosis:
I'm not happy with the whole maintenance infrastructure. We should define more clearly what it is for. The current state too ad-hoc.

nlordell:
> We should define more clearly what it is for

To me, it was always code that needed to run on each new block.

nlordell:
> Maintenance implementations do not care about what their name. What cares about the name is ServiceMaintenance because it has metrics and it prints log messages about its child Maintining impls.

This is a very valid point. Now I'm slightly leaning towards this implementation.

Either way, I don't feel strongly about either and am curious about what others think.

nlordell:
Since #966 we merged, this can be closed right?

PR #968, opened 2022-12-15
Use the block_time as creation timestamp for the ethflow orders

josojo:
testing only.

In the event, that we have an issue with autopilot order indexing, and we would index orders delayed, there could be a bigger divergence between the order creation onchain and the order indexing CPU time. In order to reflect better the real-time of mining a transaction, we introduce this change. This helps then also later the refunder to make fairer judgments on the actual order_validity_duration for its refunding considerations.

Testplan:
unit tests.


crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -276,6 +282,8 @@ impl<T: Send + Sync + Clone, W: Send + Sync> OnchainOrderParser<T, W> {
         Vec<(database::events::EventIndex, OnchainOrderPlacement)>,
         Vec<Order>,
     )> {
+        let block_number_timestamp_hashmap =
+            get_block_numbers_of_events(&self.web3, &events).await?;
```
vkgnosis, line 286: Another way to implement this would be to create the hashmap here and fill in the block numbers you need on demand. So in the loop below you would check if the hashmap already has the block number in which case you do nothing, or if it doesn't you would retrieve it and store it.

Downside of that approach is no parallelism. I think you need block numbers for all events anyway so the current solution is better.

vkgnosis (approved):

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -325,6 +341,33 @@ impl<T: Send + Sync + Clone, W: Send + Sync> OnchainOrderParser<T, W> {
...
+    event_block_numbers.dedup();
+    let futures = event_block_numbers
+        .into_iter()
+        .map(|block_number| async move {
+            let timestamp =
+                timestamp_of_block_in_seconds(web3, U64::from(block_number).into()).await?;
+            Ok((block_number, timestamp))
+        });
+    let block_number_timestamp_pair: Vec<anyhow::Result<(u64, u32)>> =
+        stream::iter(futures).buffer_unordered(10).collect().await;
```
nlordell, line 367: You shouldn't need to `buffer_unordered` here, the `BufferedTransport` implementation we have already does buffering with configurable amounts.

I think it would be fine to just `join_all`.

nlordell (approved):

crates/autopilot/src/database/onchain_order_events.rs
vkgnosis: It still has value IMO. The number of events into the function is unbounded. Creating 10000 futures at the same time still uses a lot of memory. I feel we should always bound these kind of tasks and not use join_all if there is no obund.

nlordell: Good point.

PR #723, opened 2022-11-04
[Ethflow] Openapi changes that are still planned - do not merge

josojo:
@anxolin asked for an interface definition of the ethflow changes to the API that will still happen. I summarized them in this PR.

PR will not be merged.

Issue #676, opened 2022-10-27
[onchain orders] Invalid onchain orders


josojo:
Some orders that are parsed via the onchain order indexer might be "illegal".
E.g. it could be that the valid_to is too high or they might use "bad tokens" or their quote might no longer be valid.

These illegal orders should always be prevented, as users spend gas for orders that we will not settle. Hence, we should try to flag them already in the quoting process as illegal. 
But since users might not go through the quoting procedure, they could still create illegal orders via on-chain parsing, or it could happen simply because the quote is no longer valid.

Now, we have two requirements:
- We should still show the users that we indexed their orders and did not forget about it. Hence, illegal orders should be created and put into the database
- Showing this information is quite important as this will be the entry point for the users of the refunds
- Illegal orders should not affect any db-query performance and should not appear in any auction or any metric about "non-settled" orders.

Solution proposal:
- The order field: order_status gets a new enum entry: `Invalid order`. 
- On db, level, we give each onchain_order a new optional enum entry in the `onchain_order_placed` table: `reason_for_invalid_order_placement` with the enum fields: `valid_to`, `bad_token`,  `no quote`, 'invalid_order_class' ...

The logic would then be that illegal orders are being parsed but get the flag `reason_for_invalid_order_placement="reason"`.
Whenever an order has this field set(!=null), we would derive the order status as `Invalid order`, but still serve it via API, such that users can see their order as being parsed.
Orders with `valid_at_order_placement!=null` are excluded from the solvable_orders.

wdyt?

MartinquaXD:
The idea looks reasonable to me. Just a few questions:

> But since users might not go through the quoting procedure, they could still create illegal orders via on-chain parsing.

This can only happen if the user interacts with the smart contract themselves instead of going through our frontend, right?
Would refunding those illegal orders work out of the box or is there some additional work to be done?

josojo:
> Would refunding those illegal orders work out of the box or is there some additional work to be done?

I think we should not refund illegal orders. If DAO use this functionality of placing orders without a quote, they should check beforehand how they are placing the orders. They can always refund orders themselves.

>Do we really want to support orders that weren't created through the frontend for this?
I think long-term the DAO -treasury management use case would need this. But yes, for ethflow orders, it's not needed right now.

>Intuitively I think we can index illegal orders into the onchain orders table but we shouldn't create an entry in the full orders table.
Agreed.

vkgnosis:
[...]

PR #969, opened 2022-12-16
BD layout and operations for onchain order error classification

josojo:
The first step of: https://github.com/cowprotocol/services/issues/676

The intention is to still index orders that had some errors during indexing. This is needed, as we still need the order data to later initiate any refunds and the users should still see the orders in their order overview.

The following errors could happen:

```
QuoteIDNotFound: This can not happen during launch.
ValidToTooFarInFuture: We said that we don't need to constrict a valid, as onchain orders are expensive to place. But still we might wanna do this in the future. During launch, the front-end will enforce valid_to < 180 minutes
NonAcceptedOrderClass: It is currently possible to create limit orders from on-chain events even if they are disabled at the API level. We can use this error, in case we run into this situation.
NotAllowedBuyToken: If a buy token turns into a bad token between quoting and order placement, we can deal with it here.
```

### API changes
It incorporates these API changes:
https://github.com/cowprotocol/services/pull/723/files#diff-ea5f6887f8360abaf702f3dcece07c3b2e87022458254f48bb2fb66adf15b95fR612


crates/model/src/order.rs
```diff
@@ -493,6 +493,27 @@ pub struct EthflowData {
     pub is_refunded: bool,
 }
 
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Derivative, Deserialize, Serialize)]
+#[derivative(Debug)]
+#[serde(rename_all = "camelCase")]
```
vkgnosis, line 499:
```suggestion
#[derive(Debug, Eq, PartialEq, Clone, Derivative, Deserialize, Serialize)]
```
I think we don't need the extra attributes. The struct below needs camelCase but not the rest either.

vkgnosis (approved):

crates/model/src/order.rs
nlordell: `camelCase` has an effect on how the enum names get serialized/deserialized (`QuoteIdNotFound` vs `quoteIdNotFound`) so I would argue needing `serde(rename_all = ...)` depends on what the expected output is.

Agree that `serde_as` and `derivative` aren't used here and can be dropped.

nlordell (approved):
LGTM, some nits with the `derive` statements.

fedgiac (approved):
Looks good to me.

fedgiac (approved):
Nice!

PR #970, opened 2022-12-16
Index refund events of ethflow and store the tx hash

### LINKED ISSUE #803: [Ethflow] Refund tx in API
The eth-flow smart contract was modified to emit two different events when an order is invalidated onchain: [`OrderInvalidation`](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapOnchainOrders.sol#L49) and [`OrderRefund`](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapEthFlow.sol#L16).
It used to be only the former: this means that the autopilot right now is only picking up events for `OrderCancellation` and not for `OrderRefund`.

This was done to be able to distinguish which orders have been refunded after expiration and which orders were instead refunded directly by the user (possibly before expiration). Listening to the new events allows us to:
1) Present this information neatly to the user;
2) Determine in a reorg-robust way which orders need to be refunded by the refunder service.

Tentative plan:
- Update the database scheme for `ethflow_orders` (change bool `is_refunded` to tx_hash with name `automated_refund_tx` - default to null).
- Implement the db-operations for the table `refund_tx` updatable. This will be used for inserting new txs by the autopilot and it will be used for updating the tx_hash during reorgs.
- Currently, the autopilot uses `CowSwapOnchainOrder` ABI to listen to the events OrderInvalidation and OrderPlacement. The new OrderRefund event is not part of the order `CowSwapOnchainOrder` ABI and we need to switch to the `Ethflow` contract abi.
- Then the current event updater can be used to also process the event of OrderRefund. In the meta-data of an event, there is the tx_hash that was used to emit the event OrderRefund. This tx_hash should be taken out of the meta-event data and should be inserted into the database for the UID emitted in the event.
- Reorgs should be handled automatically by the event updater if the function replace_events is properly adapted.

<details> 
<summary>
Discarded option:
</summary>
The refunder should for each OrderUid that it refunded store the tx_hash of its refund tx.

That means:
- refunder waits for the refund tx to be mined
- refunder receives the events from the refund tx
- refunder inspects which orderUids have really been refunded
(challenge: if a tx is not mined and we retry with another tx and higher gas prices, then we still the old tx could be mined and we would have to keep track of it.)

</details>

MartinquaXD:
Fixes #803

This PR implements the required changes to move from an `is_refunded` flag for ethflow orders to storing the transaction hash that refunded the order.

- added new components that listen for and index only refund events from the ethflow. This seemed like the right approach because the other ethflow events are already indexed by the generic `CoWSwapOnChainOrderContract` interface and indexing the refund events as part of that didn't seem possible.

- added a new table `ethflow_refunds`
    - `order_uid` as a private key to quickly associate a refund with the corresponding order
    - `tx_hash` to have the data we are interested in
    - `bock_number` to be able to deal with reorgs
    
- I split `EthOrderPlacement` into 2 types (one for creating the order, and one for reading the order which contains the `refund_tx`)

- I added `refundTxHash` to the `openapi` spec (`isRefunded` was kept for backwards compatibility reasons but only implemented with custom serialization logic)

- added a DB migration script to create placeholder entries in `ethflow_refunds` for all ethflow orders with `is_refunded = true`. Those placeholders can later be replaced with the correct values with a manual update.

### Test Plan
updated existing unit and e2e tests


MartinquaXD:
The e2e test `local_node_refunder_tx` only fails in the CI pipeline with this kind of error:
```
test refunder::local_node_refunder_tx ... 2022-12-16T11:18:01.098Z  INFO orderbook: serving order book address=127.0.0.1:8080
2022-12-16T11:18:01.142Z  INFO orderbook::api::request_summary: 127.0.0.1:60389 "POST /api/v1/quote/ HTTP/1.1" 200 "-" "cowprotocol-services/2.0.0" 41.4245ms
2022-12-16T11:18:01.283Z  WARN shared::maintenance: Service Maintenance Error: unexpected response format

Caused by:
    invalid type: null, expected struct Block
error: test failed, to rerun pass `-p e2e --test e2e`

Caused by:
  process didn't exit successfully: `/Users/martin/work/backend/main/target/debug/deps/e2e-c004f996b06a7b0e local_node --ignored --test-threads 1` (exit status: 1)
```

I believe this might be related to the recently added time travel logic for ethflow e2e tests. Maybe the order of the tests now matters. Will keep looking.

josojo (commented):
Wow, looks really really good. Thank you so much ❤️

crates/orderbook/openapi.yml
```diff
@@ -604,19 +604,20 @@ components:
         Provides the additional data for ethflow orders
       type: object
       properties:
-        isRefunded:
```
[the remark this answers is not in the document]
MartinquaXD: I added it back in [f68265a](https://github.com/cowprotocol/services/pull/970/commits/f68265afcb08abc31ef118dae1fbc27a86e10d86) again. Technically the field is redundant and only needed during serialization so I'll try to look how this could get implemented with `serde`. But to not delay possible testing I already implemented the basic solution.

MartinquaXD:
I pass the ownership of this PR on to @josojo now as I will be on vacation for the rest of the year.

crates/model/src/order.rs
```diff
@@ -486,11 +486,37 @@ pub struct CancellationPayload {
...
 #[serde(rename_all = "camelCase")]
 pub struct EthflowData {
     pub user_valid_to: i64,
-    pub is_refunded: bool,
+    pub refund_tx_hash: Option<H256>,
+}
+
+// We still want to have the `is_refunded` field in the JSON response to stay backwards compatible.
+// However, `refund_tx_hash` already contains all the information we need so we rather have a
+// custom `Serialize` implementation than redundant fields that could possibly contradict each other.
```
vkgnosis, line 498: Do we really need backward compatibility for an unreleased feature? Or is this used by the frontend?

MartinquaXD: That's a good point. My assumption was that it was needed to not disrupt the frontend. In that case it could make sense to keep `isRefunded` until the frontend completely switches over to `refundTxHash` but drop `isRefunded` from the `openapi.yaml` so nobody else decides to depend on that flag in the meantime.

josojo: Yes, its used by the front-end and we discussed it today in a call, that we want to have "backwards compatible" for the next week. We will then delete it in the week after that. 

Note that `isRefunded` will never really be published, we will take it with this PR (my next commit ) out of the openapi spec.

vkgnosis (approved):

fedgiac (approved):
This is a really nice PR!

nlordell (approved):
LGTM.

PR #971, opened 2022-12-16
Smooth out limit order fee updates

vkgnosis:
The current update mechanism works in the following way:

1.  Get all outdated fee limit orders.
2. Update the fee for every order. Do this for N orders in parallel.
3. Sleep 60 seconds.
4. Repeat.

The new update mechanism works in the following way:
1. Get the N most outdated fee limit orders.
2. Update the fee these N orders in parallel.
3. If there were no outdated orders sleep 10 seconds. If there were outdated do not sleep.
4. Repeat

The new update mechanism has the following advantages:
- When there are too many limit orders to keep up with, it has more throughput because there is no sleep between updates.
- When we can keep up with all limit orders, the average time an order is outdated for is lower.
- A single update is shorter which prevents wasting work on orders that become invalid during the update loop.

Other changes:
- Because the new update mechanism is more efficient I've lowered the parallelism factor from 10 to 5. We might want to make this configurable in a later PR.
- Removed the configuration option for the wait between loops. I see no reason to change this with the new mechanism.
- Change the limit order metrics update interval to 10 seconds.

### Test Plan

tested by e2e test and I will observe metrics after merging

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -113,15 +113,6 @@ pub struct Arguments {
     )]
     pub max_surplus_fee_age: Duration,
 
-    /// The interval between successive limit order surplus fee updating loops in seconds.
-    #[clap(
-        long,
-        env,
-        default_value = "60",
-        value_parser = shared::arguments::duration_from_seconds,
-    )]
-    pub surplus_fee_update_interval: Duration,
-
     #[clap(long, env)]
     pub cip_14_beta: Option<f64>,
     #[clap(long, env)]
@@ -184,11 +175,6 @@ impl std::fmt::Display for Arguments {
         writeln!(f, "banned_users: {:?}", self.banned_users)?;
         writeln!(f, "max_auction_age: {:?}", self.max_auction_age)?;
         writeln!(f, "max_surplus_fee_age: {:?}", self.max_surplus_fee_age)?;
-        writeln!(
-            f,
-            "surplus_fee_update_interval: {:?}",
-            self.surplus_fee_update_interval
-        )?;
         display_option(f, "cip_14_beta", &self.cip_14_beta)?;
         display_option(f, "cip_14_alpha1", &self.cip_14_alpha1)?;
         display_option(f, "cip_14_alpha2", &self.cip_14_alpha2)?;
--- a/crates/autopilot/src/limit_orders/quoter.rs
+++ b/crates/autopilot/src/limit_orders/quoter.rs
@@ -3,8 +3,7 @@ use crate::database::{
     Postgres,
 };
 use anyhow::Result;
-use chrono::{Duration, Utc};
-use futures::future::join_all;
+use futures::StreamExt;
 use model::{
     order::{Order, OrderKind, OrderUid},
     quote::{OrderQuoteSide, QuoteSigningScheme, SellAmount},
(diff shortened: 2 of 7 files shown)


crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -33,40 +31,43 @@ pub struct LimitOrderQuoter {
...
-                    .instrument(tracing::debug_span!(
-                        "surplus_fee",
-                        order =% order.metadata.uid
-                    ))
-                }))
-                .await;
-            }
+            let sleep = match self.update().await {
+                Ok(0) => Duration::from_secs_f32(10.),
+                Ok(_) => Duration::from_secs_f32(0.),
```
nlordell, line 41: Shouldn't this branch only happen if `orders.len() < PARALLELISM`. I don't think it makes much of a difference (since it will just trigger another DB lookup, and would return right away in case there are no new orders), but this surprised me a bit when trying to understand the code compared to the PR comment.

Maybe, it makes more sense for `update` to return a `bool` indicating whether or not there are "more updates to process". This would, IMO make the call here a little clear WRT the intention of each of the methods.

Very nitpicky though - not blocking merge from my point of view.

crates/autopilot/src/arguments.rs
```diff
@@ -113,15 +113,6 @@ pub struct Arguments {
...
     pub max_surplus_fee_age: Duration,
 
-    /// The interval between successive limit order surplus fee updating loops in seconds.
-    #[clap(
-        long,
-        env,
-        default_value = "60",
-        value_parser = shared::arguments::duration_from_seconds,
-    )]
-    pub surplus_fee_update_interval: Duration,
```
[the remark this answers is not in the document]
nlordell: **Edit**: I see your PR comment about not ever needing to change it. I think since it was already there it wouldn't hurt to have kept it, but I also see your point that it isn't clear if we will really change it ever.

nlordell (approved):
LGTM. Very nice change that should make the limit order price updating more streamlined 💪.

crates/autopilot/src/limit_orders/quoter.rs
vkgnosis: `update` doesn't exactly know whether there are more updates to process. It knows that if it gets less than N orders, then at that moment there would be no more outdated orders. By the time `update` has completed there might now be outdated orders again.

If we have 0 outdated orders then we definitely want to sleep to prevent busy looping on the DB. If we had at least 1 order then `update` internally spent some time outside of the DB query so in that sense it has already slept.

Changing `Ok(0)` to `Ok(n) if N < PARALLELISM` would be reasonable too. Both versions look fine to me.

I also see the argument in favor of just returning a bool but then there is still the decision of which comparison should determine the bool. Like you saiud this is a small detail. I'll think about it more and make a decision.

crates/autopilot/src/arguments.rs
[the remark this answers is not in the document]
vkgnosis: I prefer not having unnecessary configuration. It makes the arguments overall harder to understand and there is more code moving the config around. Regardless of whether the argument was already there, we should independently judge whether it was useful.

With the new loop I don't see no reason to configure this argument because it no longer affects throughout. It did in the previous version which made it important there.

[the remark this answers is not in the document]
nlordell: Makes sense to me.

crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -33,40 +31,43 @@ pub struct LimitOrderQuoter {
...
+
+    /// Returns how many orders were updated.
+    async fn update(&self) -> Result<usize> {
+        const PARALLELISM: usize = 5;
+        let orders = self
+            .database
+            .limit_orders_with_outdated_fees(self.limit_order_age, PARALLELISM)
+            .await?;
+        futures::stream::iter(&orders)
+            .for_each_concurrent(PARALLELISM, |order| {
```
MartinquaXD, line 59: Is there any advantage of using a stream compared to `join_all` here?
Not that it's a problem I just think usually streams are a bit harder to work with and to understand (I might be alone on this, though).

MartinquaXD (approved):

sunce86 (approved):

crates/autopilot/src/limit_orders/quoter.rs
vkgnosis: I felt the stream semantics were clearer than join_all. for_each_concurrente feels clear to me and it shows that I don't care about the individual futures because they never leave the function.

nlordell: I also personally prefer `for_each_concurrent()` over `for _ in chunked() { join_all }`.

PR #972, opened 2022-12-16
Driver `/solve` functionality part 5

sistemd:
Almost finished with https://github.com/cowprotocol/services/issues/895.

Implements converting a `boundary::Settlement` into an Ethereum transaction, and also computing the objective value (scoring) the settlement. These should be the last steps for the `boundary` module.

This should also be the last step for implementing the scoring.

Next steps:
- Update the domain types to match the ideas discussed in #962.
- Implement the Tenderly solution simulation.
- Update the DTOs according to #962.
- Implement the mapping between the DTOs and the domain objects, which should be pretty trivial.

### DIFF
--- a/crates/driver/src/blockchain/mod.rs
+++ b/crates/driver/src/blockchain/mod.rs
@@ -7,21 +7,12 @@ use {
 
 pub mod contracts;
 
-const MAX_BATCH_SIZE: usize = 100;
-
-#[derive(Debug, Error)]
-pub enum Error {
-    #[error("method error: {0:?}")]
-    Method(#[from] ethcontract::errors::MethodError),
-    #[error("deploy error: {0:?}")]
-    Deploy(#[from] ethcontract::errors::DeployError),
-}
-
 /// The Ethereum blockchain.
 #[derive(Debug)]
 pub struct Ethereum {
     web3: Web3<DynTransport>,
     chain_id: eth::ChainId,
+    network: eth::NetworkName,
 }
 
 impl Ethereum {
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):

crates/driver/src/blockchain/mod.rs
```diff
@@ -55,25 +46,32 @@ impl Ethereum {
...
     /// The settlement contract.
-    pub async fn settlement(&self) -> Result<eth::Contract, Error> {
-        Ok(contracts::GPv2Settlement::deployed(&self.0.web3)
-            .await?
-            .address()
-            .into())
+    pub fn settlement(&self) -> contracts::GPv2Settlement {
+        let address = contracts::GPv2Settlement::raw_contract()
+            .networks
+            .get(&self.0.network.0)
```
nlordell, line 56: Note that, for historical reasons, the `networks` is a `HashMap<NetworkVersion, Deployment>` and **not** a `HashMap<NetworkName, Deployment>`. specifically, it indexes contracts based on the result of the RPC call `net_version`, which is typically a decimal number string.

In fact, for all networks we support, `net_version == to_string(eth_chainId)`.

crates/driver/src/blockchain/mod.rs
```diff
@@ -7,21 +7,12 @@ use {
...
-    #[error("deploy error: {0:?}")]
-    Deploy(#[from] ethcontract::errors::DeployError),
-}
-
 /// The Ethereum blockchain.
 #[derive(Debug)]
 pub struct Ethereum {
     web3: Web3<DynTransport>,
     chain_id: eth::ChainId,
+    network: eth::NetworkName,
```
nlordell, line 15: This should probably be a:
```suggestion
    network: eth::NetworkVersion,
```

Since this is not a descriptive name, but the result of a `net_version` RPC call. We can also consider moving all our code to use `eth::ChainId` exclusively instead (my personal preference, since chain ID is the newer standard to identify different networks and is specified "better" IMO).

crates/driver/src/simulator/tenderly/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Debug)]
+pub(super) struct Tenderly {
+    client: reqwest::Client,
+    config: Config,
+}
+
+#[derive(Debug)]
+pub struct Config {
+    pub network_id: eth::NetworkName,
```
nlordell, line 13: I think tenderly uses chain IDs and not networks names or versions. Probably run into this in the other PR, so I think its fine to leave this in here.

nlordell (commented):
Code looks good. I think network name and network versions are getting conflated in a few areas. I think we should move to using `ChainId` everywhere as:
- It has a more rigid specification, and is unique across blockchains.
- `impl ChainId { fn name(&self) -> &'static str; }` is possible, so we don't need to pass around two network identifiers but use a single one.

PR #975, opened 2022-12-17
Update `driver` domain objects

sistemd:
Update the domain objects to reflect the discussions from https://github.com/cowprotocol/services/pull/962. Update the boundary mapping as well.

I asked some questions in that PR, I will move this out of draft once those are resolved.

### DIFF
--- a/crates/driver/src/boundary/settlement.rs
+++ b/crates/driver/src/boundary/settlement.rs
@@ -398,13 +398,13 @@ async fn to_boundary_solution(
                             token: output.token.into(),
                         })
                         .collect(),
-                    exec_plan: ExecutionPlan {
+                    exec_plan: Some(ExecutionPlan {
                         coordinates: ExecutionPlanCoordinatesModel {
                             sequence: 0,
                             position: index.try_into().unwrap(),
                         },
                         internal: interaction.internalize,
-                    },
+                    }),
                     cost: None,
                 }),
                 competition::solution::Interaction::Liquidity(_) => None,


crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+#[derive(Debug)]
+pub enum Kind {
+    /// Order intended to be immediately executed. This is the "regular" type of
+    /// order.
+    Market,
+    /// Order intended to be executed possibly far into the future, when the
+    /// price is such that the order can be executed.
+    Limit { surplus_fee: eth::Ether },
```
nlordell, line 108: In theory, this is `eth::Asset` denominated in the `order.sell.token`.

crates/driver/src/logic/competition/solution/mod.rs
```diff
@@ -1,131 +1,117 @@
...
-    pub prices: HashMap<eth::Token, eth::Ether>,
-    pub approvals: Approvals,
-    pub interactions: Vec<SafeInteraction>,
+    /// Trades settled by this solution.
+    pub trades: Vec<Trade>,
+    /// Token prices for this solution.
+    pub prices: HashMap<eth::TokenAddress, competition::Price>,
+    /// List of interactions with smart contracts needed to execute this
+    /// solution on the blockchain.
+    pub interactions: Vec<Interaction>,
```
nlordell, line 38: In theory, interactions are split into pre-, intra-, and post- interactions. While this doesn't matter for the `driver` <> `solver/searcher` DTO, I think this matters for the Solution domain object (since we already make use of pre-interactions for example for EthFlow contracts).

crates/driver/src/solver/mod.rs
```diff
@@ -90,9 +90,12 @@ impl Solver {
     pub async fn solve(&self, auction: &Auction) -> Result<Solution, Error> {
         let mut url = self.config.url.join("solve").unwrap();
         let time_limit = auction.deadline.solver_time_limit()?;
+        if let Some(id) = auction.id {
+            url.query_pairs_mut()
+                .append_pair("auction_id", &id.0.to_string())
+                .append_pair("instance_name", &self.instance_name(id));
+        }
```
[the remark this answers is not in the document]
nlordell: Yeah, I think we should get rid of it.

crates/driver/src/solver/mod.rs
```diff
@@ -90,9 +90,12 @@ impl Solver {
...
         url.query_pairs_mut()
-            .append_pair("auction_id", &auction.id.0.to_string())
-            .append_pair("instance_name", &self.instance_name(auction.id))
             .append_pair("time_limit", &time_limit.as_secs().to_string())
             .append_pair("max_nr_exec_orders", MAX_NR_EXEC_ORDERS);
```
nlordell, line 100: I think these could also be removed.

crates/driver/src/logic/eth/eip712.rs
```diff
@@ -9,7 +9,7 @@
 pub struct DomainSeparator(pub [u8; 32]);
 
 impl DomainSeparator {
-    pub fn new(chain_id: super::ChainId, verifying_contract: super::Contract) -> Self {
+    pub fn new(chain_id: super::ChainId, verifying_contract: super::ContractAddress) -> Self {
```
nlordell, line 12: As a side note, this is not a general `DomainSeparator`, but a CoW Protocol specific one.

crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// A just-in-time order. JIT orders are added at solving time by the solver to
+/// generate a more optimal solution for the auction. Very similar to a regular
+/// [`Order`].
+#[derive(Debug)]
+pub struct Jit {
+    pub from: eth::Address,
+    pub sell: eth::Asset,
+    pub buy: eth::Asset,
+    pub fee: eth::Ether,
```
nlordell, line 137: This is also in sell token.

crates/driver/src/logic/competition/order/signature.rs
```diff
@@ -0,0 +1,17 @@
+/// Signature over the order data.
+#[derive(Debug)]
+pub struct Signature {
```
nlordell, line 3: For the domain object, it might make sense for the `from` address to also be part of the signature (since, AFAIK, it is only ever used in the context of encoding a signature by the driver).

crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub receiver: Option<eth::Address>,
+    pub valid_to: util::Timestamp,
+    pub app_data: AppData,
+    pub side: Side,
+    pub partially_fillable: bool,
+    pub executed_buy_amount: U256,
+    pub executed_sell_amount: U256,
```
nlordell, line 144: I think these aren't part of JIT orders, but JIT trades (these are implied by the `trade::Jit::executed` field).

crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Debug)]
+pub struct Order {
+    pub uid: Uid,
+    pub from: eth::Address,
+    pub receiver: Option<eth::Address>,
+    pub valid_to: util::Timestamp,
+    pub sell: eth::Asset,
+    pub buy: eth::Asset,
+    pub side: Side,
+    pub fee: eth::Asset,
```
nlordell, line 20: I think that the order domain object needs two fees:
- The fee that is signed by the user (used for building settlement calldata)
- The "full fee amount" that is used for settlement competition.

crates/driver/src/boundary/settlement.rs
```diff
@@ -238,117 +234,109 @@ fn into_boundary_order(order: &competition::auction::Order) -> Order {
...
-                        // TODO @nlordell is this right...?
-                        // We're getting rid of this and it's only needed for price estimation, so
-                        // shouldn't be used here?
+                        execution: vec![ExecutedAmmModel {
+                            sell_token: interaction.output.token.into(),
+                            buy_token: interaction.input.token.into(),
+                            exec_sell_amount: interaction.output.amount,
+                            exec_buy_amount: interaction.input.amount,
+                            // TODO I don't know what I should do here
+                            exec_plan: Default::default(),
```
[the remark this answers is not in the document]
nlordell: The `exec_plan` was used for ordering interactions. You can actually get away with
using `solution.interactions.iter().enumerate()` and then generating the exec plan based on the index of the interaction.

Same for `interaction_data` field 👇.

crates/driver/src/boundary/settlement.rs
```diff
@@ -238,117 +234,109 @@ fn into_boundary_order(order: &competition::auction::Order) -> Order {
...
+            .iter()
+            .enumerate()
+            .filter_map(|(index, trade)| match trade {
+                competition::solution::Trade::Fulfillment(fulfillment) => Some((
                     index,
                     ExecutedOrderModel {
-                        exec_sell_amount: order.sell,
-                        exec_buy_amount: order.buy,
+                        exec_sell_amount: fulfillment.order.sell.amount,
+                        exec_buy_amount: fulfillment.order.buy.amount,
```
nlordell, line 251: This needs to be computed from the `fulfillment.executed` value, and not the order amounts.

crates/driver/src/boundary/settlement.rs
```diff
@@ -167,41 +165,39 @@ impl Settlement {
...
+fn to_boundary_order(order: &competition::Order) -> Order {
     Order {
         data: OrderData {
             sell_token: order.sell.token.into(),
             buy_token: order.buy.token.into(),
             sell_amount: order.sell.amount,
             buy_amount: order.buy.amount,
-            fee_amount: order.fee.user.into(),
+            // TODO Not sure if this was an oversight by Nick or if this needs to change
+            fee_amount: Default::default(),
```
nlordell, line 176: My understanding is that the `competition::Order` domain object needs to contain both the `fee_amount` that is signed by the user and the `full_fee_amount` which is used for the objective value computation (and gets used by the DTO between `driver` and `solver/searcher` here).

nlordell (commented):
Code looks very nice.

Added some comments in line.

crates/driver/src/boundary/settlement.rs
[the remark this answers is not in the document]
nlordell: 1. In theory, this needs to be part of the domain object and in the `driver` <> `solver/searcher` DTO (I included it here).
2. Yes this is fine. It uses sorting order, so in theory you could use a step of 42 and it would still be OK.

crates/driver/src/logic/competition/solution/mod.rs
nlordell: Yes, it is needed when building the settlement calldata. The `searcher/solver` doesn't care about this because their interactions always go in the middle, but when including things like EthFlow pre-interactions when building settlement data for the driver to submit on-chain, it would need to care about this.

crates/driver/src/logic/competition/order/signature.rs
nlordell: My point was that `from` isn't really order data, but data inherent to the signature:
- for `eip712` and `ethsign`, it is the recovered address
- for `eip1271` and `presign`, it needs to be included in the signature bytes 

The `from` is actually not part of order data from the smart contract point of view.

I think of the `from` more as a `signer` anyway - which would fit a little better with the `Signature` domain object.

This wasn't a strong suggestion, just thrown out there since I know that when encoding the signature for settlement calldata, the `signer` address is needed. I think it is fine as is though.

crates/driver/src/boundary/settlement.rs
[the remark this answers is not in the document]
nlordell: > I included it here

I thought I linked it correctly but apparently not:

https://github.com/cowprotocol/services/blob/0f04e6cdd6439c42148abb814a00f3298c0db370/crates/searchers/openapi.yml#L638

crates/driver/src/logic/competition/order/mod.rs
nlordell: Looks great.

crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        matches!(self.partial, Partial::Yes { .. })
+    }
+}
+
+#[derive(Debug)]
+pub enum Partial {
+    /// A partially order doesn't require the full amount to be traded.
+    /// E.g. only 10% of the requested amount may be traded, if this leads
+    /// to the most optimal solution.
+    Yes {
```
nlordell, line 53: Nice.

crates/driver/src/logic/competition/order/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+/// A just-in-time order. JIT orders are added at solving time by the solver to
+/// generate a more optimal solution for the auction. Very similar to a regular
+/// [`Order`].
+#[derive(Debug)]
+pub struct Jit {
+    pub from: eth::Address,
+    pub sell: eth::Asset,
+    pub buy: eth::Asset,
+    ///  Fee denominated in the sell token.
+    // TODO The fee token should be validated, possibly with a newtype
```
nlordell, line 150: I like this idea. Reminds me of "parsing is validation" principle (something like https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

crates/driver/src/logic/eth/eip712.rs
```diff
@@ -8,16 +8,23 @@
 #[derive(Debug, Clone, Copy)]
 pub struct DomainSeparator(pub [u8; 32]);
 
+#[derive(Debug)]
+pub struct DomainFields {
```
nlordell, line 12: Technically, all the fields are optional. For example the `SolverTrampoline` contract used for Gelato submissions only uses 2 of them:
https://github.com/cowprotocol/solver-trampoline/blob/0d65e071bd014d38c6c4af3a57364754568b98e3/contracts/SolverTrampoline.sol#L39-L43

I actually like it as-is for now, and we can change it when we introduce ☝️

nlordell (approved):
LGTM.

The only potential issue I see is there is no representation of a "pre-interaction" in the solution domain object. **That being said**, the only source of pre-interactions at the moment are from user orders (like the EthFlow order), and the order domain object already tracks the pre-interactions, so maybe its actually better this way (in fact, its growing on me the more I think of it, as the only valid use of a pre-interaction is to make "preparation" calls for user orders, and the domain object as is better captures this).

crates/driver/src/boundary/settlement.rs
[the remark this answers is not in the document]
sunce86: I think you should probably add `ExecutionPlan` to the `competition::solution::Interaction::Liquidity` and then just map it here (because exec plan is set to be mandatory for amms)

(if I did not completely missed the context of this function `to_boundary_solution`)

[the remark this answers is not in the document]
sunce86: and the same for `competition::solution::Interaction::Custom`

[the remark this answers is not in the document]
nlordell: Yes, we are getting rid of them. There is no need for them since the interactions are already ordered in a `Vec` now.

nlordell (approved):

nlordell:
> Let me know if you want me to undo the commit linked above or if I should keep it.

Conceptually, an **encoded** settlement has pre-/intra-/post-interactions. The question is:
- _What should the driver/solver DTO have?_ I think, it should only have `intra-` interactions (so a single `Vec<dto::Interaction>` is all that is needed. The rationale is that there is currently no real reason for a solver to include any other interaction type. Potentially, in the future, solver engines may want to add some backrunning to their own solutions, in which case a `post-` interactions might be needed. All order setup is handled by the `driver`, so `pre-` interactions are not needed in the DTO
- _What should the driver domain object have?_ Again, like in the case for the DTO, there is no real reason for the domain object to differentiate the different interaction types before encoding. Any `pre-` interactions that we need to encode are already attached to the order and IMO do not need to be duplicated in the `Solution` domain object, so this makes me lean towards `interaction: Vec<Interaction>` for now.

PR #976, opened 2022-12-17
Onchain order placement error handling

josojo:
The second step of: https://github.com/cowprotocol/services/issues/676.

Takes the error during order placement indexing and stores them together with the orders in the database. 

### Note
This PR is important as it handles the error QuoteNotFound, which can happen, if the autopilot is down for an extended time and the quotes get deleted after a restart. In such a scenario, before this PR the order placements would be discarded - as no quote can be found -  and then could not be refunded. With this PR, it creates the orders in the database with the error, such that they can be refunded by the refunder later.

### Test Plan

Unit tests.


crates/shared/src/order_validation.rs
```diff
@@ -109,6 +109,25 @@ pub enum ValidationError {
...
+        _ => OnchainOrderPlacementError::Other,
+    }
+}
+
+impl ValidationError {
+    pub fn from(error: OnchainOrderPlacementError) -> Self {
+        match error {
+            OnchainOrderPlacementError::QuoteNotFound => ValidationError::QuoteNotFound,
+            _ => ValidationError::Other(anyhow!("non classified onchain placement error")),
+        }
```
fedgiac, line 127: Why not also `InvalidQuote` and `InsufficientFee`?

fedgiac (commented):
Overall it looks good to me but I want to check which errors can be thrown where more in detail later.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -437,25 +449,37 @@ async fn parse_general_onchain_order_placement_data(
...
 
-            let quote = get_quote(quoter, order_data, signing_scheme, &event, &quote_id).await?;
+            let quote_result =
+                get_quote(quoter, order_data, signing_scheme, &event, &quote_id).await;
+            let mut placement_error = None;
+            let mut quote = None;
+            match quote_result.clone() {
+                Ok(found_quote) => quote = Some(found_quote),
+                Err(err) => placement_error = Some(err),
+            }
```
nlordell, line 459: uber-nano-nit: A slightly more idiomatic way of writing this would be:
```suggestion
            let (quote, placement_error) = match quote_result.clone() {
                Ok(found_quote) => (Some(found_quote), None),
                Err(err) => (None, Some(err)),
            };
```

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -516,36 +540,39 @@ async fn get_quote(
...
 ) -> Result<(OnchainOrderPlacement, Order)> {
-    let full_fee_amount = quote.data.fee_parameters.unsubsidized();
-
-    let is_outside_market_price =
+    let full_fee_amount = if let Some(ref quote) = quote {
+        quote.data.fee_parameters.unsubsidized()
+    } else {
+        // Since the order will anways have an order_placement_error,
+        // we can set an "arbitrary amount"
+        order_data.fee_amount
```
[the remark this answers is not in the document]
josojo: Rust is so nice. Every time I am writing ugly rust code, there was a structural issue. :) 
With your suggestion, the code looks so much more beautiful.

nlordell (approved):
LGTM.

vkgnosis (approved):

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -197,9 +197,15 @@ impl<T: Sync + Send + Clone, W: Sync + Send + Clone> EventStoring<ContractEvent>
         .await
         .context("append_onchain_orders failed")?;
 
-        insert_quotes(&mut transaction, quotes.as_slice())
-            .await
-            .context("appending quotes for onchain orders failed")?;
+        // It's okay to not store quotes for orders is with placement_errors as
+        // only orders that will be used in auctions need a quote stored
+        // to calcualte the rewards.
```
fedgiac, line 202: Nit, I find the comment a bit confusing. Note that this comment is duplicated and also the other should be changed, in case!

```suggestion
        // We only need to insert quotes for orders that will be included in an
        // auction (they are needed to compute solver rewards). If placement
        // failed, then the quote is not needed.
```

fedgiac (approved):
Looks good!

PR #977, opened 2022-12-17
Silent node connection issues

josojo:
We got quite some warnings about node gateway connections: 
https://cowservices.slack.com/archives/C0371SB243E/p1671308516493619

I think its perfectly fine, if we just log them, but don't error.

### DIFF
--- a/crates/refunder/src/refund_service.rs
+++ b/crates/refunder/src/refund_service.rs
@@ -139,7 +139,7 @@ impl RefundService {
                     let order_owner = match order.await {
                         Ok(order) => Some(order.0),
                         Err(err) => {
-                            tracing::error!(
+                            tracing::debug!(
                                 "Error while getting the current\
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
Depends on your semantics whether you want to consider this an error. You probably want a metric so you can get an alert if this process fails too often in a row?

crates/refunder/src/refund_service.rs
```diff
@@ -139,7 +139,7 @@ impl RefundService {
                     let order_owner = match order.await {
                         Ok(order) => Some(order.0),
                         Err(err) => {
-                            tracing::error!(
+                            tracing::debug!(
```
nlordell, line 142: Should this be a `warn!`?

nlordell (approved):
Agree that metrics to capture this and be able to set up alerts when there are extended periods of errors would be a good improvement.

PR #978, opened 2022-12-18
Implement Tenderly simulation in `driver`

sistemd:

Implement Tenderly simulation in `driver`. Add a `solver` field to the `Solution`.

After this PR, all that's needed is to update the DTOs in `/solve` and implement the DTO mapping, and the `/solve` flow should be ready for testing.

### DIFF
--- a/crates/driver/src/blockchain/mod.rs
+++ b/crates/driver/src/blockchain/mod.rs
@@ -12,7 +12,6 @@ pub mod contracts;
 pub struct Ethereum {
     web3: Web3<DynTransport>,
     chain_id: eth::ChainId,
-    network: eth::Network,
 }
 
 impl Ethereum {
@@ -20,10 +19,6 @@ impl Ethereum {
     /// URL.
     pub fn eth_rpc(_url: Url) -> Self {
         todo!()
-    }
-
-    pub fn network(&self) -> &eth::Network {
-        &self.network
     }
 
     pub fn chain_id(&self) -> eth::ChainId {
@@ -65,7 +60,7 @@ impl Contracts<'_> {
     pub fn settlement(&self) -> contracts::GPv2Settlement {
         let address = contracts::GPv2Settlement::raw_contract()
             .networks
-            .get(&self.0.network.0)
+            .get(self.0.chain_id().network_id())
             .unwrap()
             .address;
         contracts::GPv2Settlement::at(&self.0.web3, address)
@@ -75,7 +70,7 @@ impl Contracts<'_> {
     pub fn weth(&self) -> contracts::WETH9 {
         let address = contracts::WETH9::raw_contract()
             .networks
-            .get(&self.0.network.0)
+            .get(self.0.chain_id().network_id())
             .unwrap()
             .address;
         contracts::WETH9::at(&self.0.web3, address)
--- a/crates/driver/src/logic/eth/mod.rs
+++ b/crates/driver/src/logic/eth/mod.rs
@@ -18,23 +18,17 @@ pub use {
 // TODO It might make sense to re-export H160 and U256 from here and not
 // reference primitive_types directly anywhere, it's probably the best idea
 
-// TODO Constructing this type should probably do some validation, or maybe this
-// should be an enum with a Display implementation
-/// Name of an Ethereum network, e.g. mainnet or testnet.
-#[derive(Debug, Clone)]
-pub struct Network(pub String);
-
-impl std::fmt::Display for Network {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(&self.0)
-    }
-}
-
 /// Chain ID as defined by EIP-155.
 ///
 /// https://eips.ethereum.org/EIPS/eip-155
 #[derive(Debug, Clone, Copy)]
 pub struct ChainId(pub u64);
+
+impl ChainId {
+    pub fn network_id(&self) -> &'static str {
+        todo!()
+    }
+}
 
 /// Gas amount.
 #[derive(Debug, Default, Clone, Copy)]
--- a/crates/driver/src/simulator/tenderly/mod.rs
+++ b/crates/driver/src/simulator/tenderly/mod.rs
@@ -12,7 +12,7 @@ pub(super) struct Tenderly {
 pub struct Config {
     /// The URL of the Tenderly API.
     pub url: reqwest::Url,
-    pub network: eth::Network,
+    pub chain_id: eth::ChainId,
     /// Save the transaction on Tenderly for later inspection, e.g. via the
     /// dashboard.
     pub save: bool,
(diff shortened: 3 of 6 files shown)


crates/driver/src/blockchain/mod.rs
```diff
@@ -47,6 +53,12 @@ impl Ethereum {
             .await?;
         Ok(eth::Allowance { spender, amount }.into())
     }
+
+    /// Check if a smart contract is deployed to the given address.
+    pub async fn is_contract(&self, address: eth::Address) -> Result<bool, Error> {
+        let code = self.web3.eth().code(address.into(), None).await?;
+        Ok(!code.0.is_empty())
+    }
 }
```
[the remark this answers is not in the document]
sunce86: I also see that we have a similar function `shared::is_smart_contract`. I assume you don't want to reuse it to avoid having dependency?

crates/driver/src/blockchain/mod.rs
```diff
@@ -12,22 +12,28 @@ pub mod contracts;
 pub struct Ethereum {
     web3: Web3<DynTransport>,
```
sunce86, line 13: Should we change the struct name since we support other chains, not only Ethereum? `Blockchain` or similar...

crates/driver/src/logic/competition/solution/mod.rs
```diff
@@ -107,6 +109,66 @@ impl Solution {
...
+            from: self.solver.account(),
+            to: trade.order.receiver(),
+            value: 1.into(),
+            input: Vec::new(),
+        };
+        simulator
+            .access_list(&tx, eth::AccessList::default())
+            .await
+            .map_err(Into::into)
+    }
```
sunce86, line 171: Is there a specific reason for access list not being a part of the `eth::Tx`? At the end it always end up as a part of the transaction, and we would avoid forwarding access list as a separate parameter to `simulator.access_list` function etc.

crates/driver/src/simulator/ethereum.rs
```diff
@@ -1,10 +1,14 @@
...
-pub(super) struct Ethereum;
+pub(super) struct Ethereum(crate::Ethereum);
 
 impl Ethereum {
-    pub async fn simulate(&self, _tx: &eth::Tx, _access_list: &eth::AccessList) -> eth::Simulation {
+    pub async fn simulate(
+        &self,
+        _tx: &eth::Tx,
+        _access_list: &eth::AccessList,
+    ) -> super::Simulation {
```
sunce86, line 11: As mentioned, I don't see a reason for access list to be a special parameter here. It's just another tx parameter as all others.

crates/driver/src/boundary/settlement.rs
```diff
@@ -57,8 +56,8 @@ pub struct Settlement {
 impl Settlement {
     pub async fn encode(
         eth: &Ethereum,
-        solver: &Solver,
         solution: &competition::Solution,
+        // TODO I think it's possible to remove this parameter, do this in a follow-up
```
nlordell, line 60: I think it is fine for the `competition::Solution` domain object to include enough information for encoding a settlement.

crates/driver/src/logic/competition/solution/mod.rs
nlordell: I think conceptually (at least since EIP-2930), access lists are part of transactions.

crates/driver/src/simulator/tenderly/mod.rs
```diff
@@ -19,26 +21,65 @@ pub struct Config {
...
+        tx: &eth::Tx,
+        access_list: &eth::AccessList,
+        generate_access_list: GenerateAccessList,
+    ) -> Result<super::Simulation, Error> {
+        let url = self.config.url.join("simulate").unwrap();
+        let res: dto::Response = self
+            .client
+            .post(url)
+            .json(&dto::Request {
+                network_id: self.config.network.to_string(),
```
nlordell, line 51: This should be a "network ID" or "network version" (result of `net_version` RPC call) and not a descriptive network name ("testnet"/"mainnet").

Personally, I would be for using `eth::ChainId` everywhere, and implementing a `eth::ChainId::network_id() -> &'static str` method (since a network ID/version is fixed for a chain ID).

nlordell (commented):
Code looks good.

I think there is an issue with `Network` here as well. Also, conceptually I agree with @sunce86's point that an access list is part of a transaction (EIP-2930 extended Ethereum transactions to add access lists); but I don't feel strongly about it).

crates/driver/src/blockchain/mod.rs
nlordell: For context, there was some discussion about this naming in a previous PR. `Blockchain` was a candidate, but we felt like `Ethereum` better captured the concept of an "EVM-compatible blockchain". For example, Bitcoin is a blockchain, but can't be used here.

nlordell (approved):
Awesome!

crates/driver/src/simulator/mod.rs
```diff
@@ -1,63 +1,53 @@
...
-                .simulate(tx, access_list, tenderly::Speed::Fast)
-                .await
-                .map_err(Into::into),
-            Self::Ethereum(ethereum) => Ok(ethereum.simulate(tx, access_list).await),
-        }
-    }
+#[derive(Debug)]
+struct Simulation {
+    gas: eth::Gas,
+    access_list: eth::AccessList,
```
sunce86, line 50: Should this be maybe:
`access_list: Option<eth::AccessList>`

crates/driver/src/logic/eth/mod.rs
```diff
@@ -187,8 +248,18 @@ pub struct Interaction {
 /// An onchain transaction.
 #[derive(Debug)]
 pub struct Tx {
-    pub from: Address,
+    pub from: Account,
     pub to: Address,
     pub value: Ether,
     pub input: Vec<u8>,
+    pub access_list: AccessList,
```
sunce86, line 255: Same, why not `Option<AccessList>` since empty access list is still a valid list.

sunce86 (approved):
Lg.
Just one nit about adding `Option` to access lists but does not prevent merging.

crates/driver/src/logic/eth/mod.rs
sunce86: I think we will send the tx onchain without access list if
1. Access list estimators failed for any reason (we should not abort the submission process if we don't have access list)
2. Access list estimators disabled by configuration (in cases of high percent of failed/reverted transactions onchain)

On the other hand, I _think_ that nothing will be different if you send a tx with empty access list compared to tx without access list, at least for the node clients (although I did not try it). But we do have 3rd party submitters (infura, nodereal) and I can only guess how they treat it.

nlordell: They are equivalent AFAIU - because `None` get encoded as an empty access list in EIP-1559 transactions anyway.

PR #784, opened 2022-11-15
Quote limit orders when they are created

sistemd:
First follow up PR for https://github.com/cowprotocol/services/pull/775#issuecomment-1315104202.

Quotes limit orders when they are created. Does not yet update their `surplus_fee` and `surplus_fee_timestamp` in the immediate API response.

### Test Plan

Automated tests.

PR #775, opened 2022-11-14
Limit orders e2e full flow

sistemd:
Progress on https://github.com/cowprotocol/services/issues/769.

Adds a test case exercising adding and settling a single limit order. Makes the optionality of `surplus_fee` and `surplus_fee_timestamp` explicit in the API. If anybody disagrees with this API change we can revert the commit - originally I thought this was a bug (it's not), but either way I think things are more explicit this way.

I also originally fixed the bug that actually ended up being fixed (in a better way) by Nick in https://github.com/cowprotocol/services/pull/772, so this PR is based on his PR.

There is still one outstanding issue: the test case works as-is, but if we try adding another seemingly completely valid limit order, it fails with

```
VM Exception while processing transaction: reverted with reason string 'ERC20: transfer amount exceeds balance'
```

This needs to be fixed. I would also like to add a test case for having mixed limit and market orders. Both of these should be addressed in follow-up PRs.

(If anybody has seen that message before and has some tips about what it might be, please let me know. So far the working assumption is that something is wrong with the test setup and the settlement contract doesn't have enough liquidity or something.)

### Release notes

`surplus_fee` and `surplus_fee_timestamp` are now nullable. Since the frontend already integrated when they were non-nullable, I'm not sure if this is a good idea. Would be happy to revert that change if anybody asks.

Edit: As @nlordell and I agreed, disregard the above paragraph, we will make the surplus fee and timestamp non-nullable.

PR #981, opened 2022-12-19
Fix handling of limit orders without surplus_fee

vkgnosis:
This fixes the panic that we saw in the last hour on staging.

The root cause is a mismatch on the expectations on the surplus_fee field for limit orders. In #775 we want back and forth on whether it should be nullable. In the end we decided against nullable becausae we would quote limit orders on order creation. That however was decided against in #784 . This left some parts of the code expecting the field to be non nullable and other parts expecting it to be nullable. 

In this PR make sure we can handle the fields being null.

Note that this issue wasn't introduced by #965 . That just made it more visible. We already had orders without surplus_fee in the database which would trigger a panic if anyone tried to fetch them from the /orders endpoint.

### Test Plan

CI, adjusted unit tests, manual test

### DIFF
--- a/crates/shared/src/db_order_conversions.rs
+++ b/crates/shared/src/db_order_conversions.rs
@@ -123,16 +123,11 @@ pub fn order_class_from(order: &FullOrderDb) -> OrderClass {
         DbOrderClass::Market => OrderClass::Market,
         DbOrderClass::Liquidity => OrderClass::Liquidity,
         DbOrderClass::Limit => OrderClass::Limit(LimitOrderClass {
-            surplus_fee: big_decimal_to_u256(
-                order
-                    .surplus_fee
-                    .as_ref()
-                    .expect("limit orders must have surplus fee set"),
-            )
-            .unwrap(),
-            surplus_fee_timestamp: order
-                .surplus_fee_timestamp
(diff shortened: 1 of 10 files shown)


crates/shared/src/db_order_conversions.rs
```diff
@@ -123,16 +123,11 @@ pub fn order_class_from(order: &FullOrderDb) -> OrderClass {
...
-            )
-            .unwrap(),
-            surplus_fee_timestamp: order
-                .surplus_fee_timestamp
-                .expect("limit orders must have surplus fee timestamp set"),
+            surplus_fee: order
+                .surplus_fee
+                .as_ref()
+                .map(|fee| big_decimal_to_u256(fee).unwrap()),
+            surplus_fee_timestamp: order.surplus_fee_timestamp,
```
vkgnosis, line 130: This is the main change that fixes the panic. This function is not just used for solvable orders so it needs to handle the null case.

josojo (approved):
Don't have too much background, but LGTM

sunce86 (commented):
Lg

nlordell:
> That however was decided against in https://github.com/cowprotocol/services/pull/784

Looks like that PR was closed and not merged.

PR #982, opened 2022-12-19
Reapply #965

vkgnosis:
We reverted this PR because of a panic. The panic has been independently fixed so now the PR can be applied again. There are no changes to the version that has already been reviewed.

### Test Plan

CI


josojo (approved):

nlordell:
I think this PR suffers from the same issue where the update loop expects non-null `surplus_fee_timestamp`. See #983

vkgnosis:
I think it does not because I change the sql query.

nlordell (approved):
You are right, this does not have the same issue.

PR #983, opened 2022-12-19
Revert "Fix handling of limit orders without surplus_fee"

nlordell:
Reverts cowprotocol/services#981

Reverting #981 as the condition that determines whether or not to include an order in the list of pending surplus fee updates requires a non-null surplus fee timestamp, and that PR adds orders with NULL `surplus_fee_timestamp` which cause limit orders to never get updated. This was also not caught by our metrics because they have similar conditions on `surplus_fee_timestamp`.

Reverting for now to unblock barn limit orders.

### Release Notes

The reason this PR is blocked in "Draft" mode is that we get panics from orders with NULL `surplus_fee` that were already added to the orderbook. A correct fix also requires a DB migration to fix the orders that were added with unexpected values since #965 and #981:

```sql
UPDATE orders
SET surplus_fee = 0
WHERE surplus_fee IS NULL
AND class='limit';

UPDATE orders
SET surplus_fee_timestamp = '1970-01-01 00:00:00'::timestamp
WHERE surplus_fee_timestamp IS NULL
AND class='limit';
```

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -566,9 +566,8 @@ async fn filter_unsupported_tokens(
 }
 
 fn filter_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
-    // Unwrap because solvable orders always have a surplus fee.
     orders.retain(|order| match &order.metadata.class {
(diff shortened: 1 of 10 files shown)


nlordell:
Closing as superceeded by #982

PR #985, opened 2022-12-20
Fixes non updating of limit orders

josojo:
Reenables the e2e tests and fixes non-settlement of limit orders after Valentin's fix:
https://github.com/cowprotocol/services/pull/981

Another unit test could be added by someone else


fedgiac:
I'd propose to change [this line](https://github.com/cowprotocol/services/blob/5371879b6661bd9744709a60cb38233b5a7cc4f2/.github/workflows/pull-request.yaml#L102-L105) from:

```
- run: cargo test local_node -- --ignored --test-threads 1 --nocapture
```

to:

```
- run: cargo test --workspace --exclude e2e local_node -- --ignored --test-threads 1 --nocapture
- run: cargo test --package e2e local_node -- --ignored --test-threads 1 --nocapture
```

(The first line isn't actually needed since all local node tests are e2e tests, but it's better to have it just in case we ever write one.)

vkgnosis:
Let's remove the limit order query change from this. It is part of https://github.com/cowprotocol/services/pull/982 which I can merge.

vkgnosis:
> ```
> - run: cargo test --workspace --exclude e2e local_node -- --ignored --test-threads 1 --nocapture
> - run: cargo test --package e2e local_node -- --ignored --test-threads 1 --nocapture
> ```

I think that causes an unfortunate recompilation because of the use of --workspace and --package which makes CI take much longer (based on when I was experimenting with reducing CI time before). And it seems like your change just changes the order in which the tests are run, so it doesn't really matter.

fedgiac:
Makes sense, then let's not do it. (`local_node` wasn't supposed to appear in the second line, I fixed the original message)

vkgnosis:
The e2e changes from here have been included in https://github.com/cowprotocol/services/pull/982 and that changes the query too so this PR is no longer needed.

PR #989, opened 2022-12-21
Add Swapr Mainnet Liquidity

nlordell:
This PR adds Swapr liquidity on Mainnet.

Swapr is UniV2 clone from the DXdao which we already supported on Gnosis Chain, so adding support on Mainnet was super easy.

### Test Plan

Added a unit test to check that the pool address matches an existing pool on Mainnet.

Also ran the solver locally with just Swapr liquidity and [generated a settlement simulation](https://dashboard.tenderly.co/gp-v2/staging/simulator/3338c8c5-a982-49c2-a02e-6f42466a43da). Note that the transferred amounts are exact to the atom, so it looks like the pool computation is working as expected (Swapr has different fees per pool).


vkgnosis (approved):

sunce86 (approved):
Easy money!

PR #990, opened 2022-12-21
Fix not filtering out already solved orders for limit order quoting

vkgnosis:
The queries dealing with limit orders are not taking into account all the conditions that can cause an order to be unsolvable. Most importantly the "already settled" condition was missing.

In this PR I factor out the solvable orders query so that it can be reused by the solvable_orders function and the limit order related queries. This ensures that both work on the same conditions. Reusing this query is becoming hairy. If these queries diverge more I would stop reusing and copy paste instead.

After merging we should look at the speed of the changed queries to see if there is problematic performance.

### Test Plan

CI, existing and adjusted unit tests

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -551,13 +551,11 @@ pub fn user_orders<'a>(
         .fetch(ex)
 }
 
-pub fn solvable_orders(
-    ex: &mut PgConnection,
-    min_valid_to: i64,
-    min_surplus_fee_timestamp: DateTime<Utc>,
-) -> BoxStream<'_, Result<FullOrder, sqlx::Error>> {
-    #[rustfmt::skip]
-    const QUERY: &str = const_format::concatcp!(
+/// The base solvable orders query used in specialized queries. Parametrized by valid_to.
+///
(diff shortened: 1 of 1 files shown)


crates/database/src/orders.rs
```diff
@@ -551,13 +551,11 @@ pub fn user_orders<'a>(
...
-    min_valid_to: i64,
-    min_surplus_fee_timestamp: DateTime<Utc>,
-) -> BoxStream<'_, Result<FullOrder, sqlx::Error>> {
-    #[rustfmt::skip]
-    const QUERY: &str = const_format::concatcp!(
+/// The base solvable orders query used in specialized queries. Parameterized by valid_to.
+///
+/// Does not take limit order surplus fee into account.
+#[rustfmt::skip]
+const SOLVABLE_ORDERS: &str = const_format::concatcp!(
```
nlordell, line 558: naming-nit: I would call this "open orders" maybe, since `(surplus_fee IS NOT NULL)` is required for it to be solvable.

nlordell (approved):
Changes look good to me, just a naming nit.

PR #991, opened 2022-12-21
Include Uninternalized Settlement In Solver Competition Data

nlordell:
Fixes https://github.com/cowprotocol/services/issues/604

This PR adds a new `uniniternalized_call_data` field to the settlement competition debug JSON blob that gets uploaded to the database.

The motivation behind this change is that, currently, the only way for external solvers to inspect internalized data from their competition is to ask us to look it up in the logs. This is not ideal.

For this change, I made the `uniniternalized_call_data` **only get added if it different from `call_data`**. The reason for this is database size considerations.

### Test Plan

Rust compiler. Serialization test.

### DIFF
--- a/crates/solver/src/settlement.rs
+++ b/crates/solver/src/settlement.rs
@@ -453,6 +453,17 @@ impl Settlement {
     pub fn encode(self, internalization_strategy: InternalizationStrategy) -> EncodedSettlement {
         self.encoder.finish(internalization_strategy)
     }
+
+    pub fn encode_uninternalized_if_different(self) -> Option<EncodedSettlement> {
+        if self.encoder.contains_internalized_interactions() {
+            Some(
+                self.encoder
+                    .finish(InternalizationStrategy::EncodeAllInteractions),
+            )
+        } else {
+            None
+        }
+    }
 }
 
 // The difference between what you were willing to sell (executed_amount * limit_price)
--- a/crates/model/src/solver_competition.rs
+++ b/crates/model/src/solver_competition.rs
@@ -1,5 +1,6 @@
 use crate::{
     auction::AuctionId,
+    bytes_hex::BytesHex,
     order::OrderUid,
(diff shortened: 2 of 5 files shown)


vkgnosis:
> The motivation behind this change is that, currently, the only way for external solvers to inspect internalized data from their competition is to ask us to look it up in the logs. This is not ideal.

They are asking for their own internalized data? Don't they already have this information?

crates/solver/src/settlement.rs
```diff
@@ -453,6 +453,17 @@ impl Settlement {
...
+    pub fn encode_uninternalized_if_different(self) -> Option<EncodedSettlement> {
+        if self.encoder.contains_internalized_interactions() {
+            Some(
+                self.encoder
+                    .finish(InternalizationStrategy::EncodeAllInteractions),
+            )
+        } else {
+            None
+        }
+    }
```
sunce86, line 466:
```suggestion
        self.encoder.contains_internalized_interactions().then_some(
            self.encoder
                .finish(InternalizationStrategy::EncodeAllInteractions),
        )
```

nit: Found a cool one-liner but feel free to ignore since I'm not sure if it improves the readibility.

sunce86 (approved):
Lg.
I think we should merge this PR, since protocol, at some point in time, might decide to do partial internalizations or some other improvement strategy that is in the interest of protocol.

Fixes https://github.com/cowprotocol/services/issues/604

nlordell:
> They are asking for their own internalized data? Don't they already have this information?

No, they want to inspect internalized data from the competing solvers.

crates/solver/src/settlement.rs
nlordell: 😍 - that is awesome and didn't know it was a thing!

harisang:
Nice! Definitely in favor of this.

vkgnosis (approved):

PR #928, opened 2022-12-07
Make execution_plan mandatory for custom interactions

sunce86:
Follow up PR for https://github.com/cowprotocol/services/pull/791

https://github.com/cowprotocol/services/pull/791 was merged so we can safely simulate settlements both with and without internalized interactions. However, custom interactions, although planned, were not made mandatory.

This PR makes custom interactions mandatory (this time really) and it will be merged once it's confirmed all solvers send execution plans.

So far, seems like we are waiting for `ichi`.

PR #994, opened 2022-12-23
Revert "Make execution_plan mandatory for custom interactions"

nlordell:
Reverts cowprotocol/services#928

We need to revert this as solvers were not aware this would also affect price finding (i.e. quoting) and not just solving.

This is a temporary revert to not break price estimation for next release - this should be re-applied as soon as the issues are addressed by the affected external solver teams.

cc @sunce86


harisang:
Adding here that this is most likely a temporary revert, as it would be good to eventually always have an execution plan.

harisang (approved):

PR #995, opened 2022-12-23
Make `execution_plan` Mandatory for Custom Interactions Again

nlordell:
This PR re-applies https://github.com/cowprotocol/services/pull/928.

We needed to **temporarily** disable it because it would break price estimation. This should be merged once the HTTP solver price estimation is changed to include execution plans.


harisang (approved):

sunce86 (approved):

harisang:
Was this merged by accident? It was not supposed to as far as I understand cc @sunce86

sunce86:
Yes, by accident. I will revert this change. Thanks for pinging.

Issue #894, opened 2022-12-01
`solver`-`driver` interface


sistemd:
The driver needs to talk to the solver. For this, we will need to implement the REST API that is to be defined here: https://github.com/cowprotocol/services/issues/831. The interface that we end up landing on in that issue is unlikely to be too far off from what we have today with [`HttpSolverApi`](https://github.com/cowprotocol/services/blob/3a1ad995a305edcaae0fad3a904dbb69cc2b4b49/crates/shared/src/http_solver.rs#L19).

For now, we will adopt the working assumption that these interfaces will be similar, so for now we can expect the solver API to conform to something very similar to [`HttpSolverApi`](https://github.com/cowprotocol/services/blob/3a1ad995a305edcaae0fad3a904dbb69cc2b4b49/crates/shared/src/http_solver.rs#L19).

This is a tracking issue to use as a reminder that once https://github.com/cowprotocol/services/issues/831 has been resolved, the solver REST interface might need to be updated in the `driver`.

PR #999, opened 2022-12-24
Solver DTOs for the `driver`

sistemd:
Fixes https://github.com/cowprotocol/services/issues/894.

Implement solver DTOs and DTO mapping in the driver.

Next step is implementing the `/solve` REST API DTOs, and then I can start writing some automated tests for the `/solve` endpoint.

### DIFF
--- a/crates/driver/src/solver/dto/auction.rs
+++ b/crates/driver/src/solver/dto/auction.rs
@@ -211,8 +211,8 @@ struct ConcentratedLiquidityPool {
     #[serde_as(as = "serialize::U256")]
     liquidity: U256,
     tick: i32,
-    #[serde_as(as = "HashMap<_, serialize::U256>")]
-    liquidity_net: HashMap<usize, U256>,
+    #[serde_as(as = "HashMap<serde_with::DisplayFromStr, serialize::U256>")]
+    liquidity_net: HashMap<i32, U256>,
     #[serde_as(as = "serde_with::DisplayFromStr")]
     fee: bigdecimal::BigDecimal,
 }


crates/driver/src/logic/competition/auction.rs
```diff
@@ -52,32 +52,30 @@ impl From<chrono::DateTime<chrono::Utc>> for Deadline {
 /// This type contains a [`std::time::Duration`] rather than
 /// [`chrono::Duration`] because [`std::time::Duration`] is guaranteed
 /// to be nonnegative, while [`chrono::Duration`] can be negative as well.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub struct SolverDeadline(std::time::Duration);
```
nlordell, line 56: I think in #962 I specified this as a date time and not a duration.

crates/driver/src/solver/dto/solution.rs
```diff
@@ -1,201 +1,324 @@
...
     serde::Deserialize,
     serde_with::serde_as,
     std::collections::HashMap,
 };
 
-impl From<Solution> for logic::competition::Solution {
-    fn from(_solution: Solution) -> Self {
-        todo!()
+impl Solution {
+    pub fn into(
```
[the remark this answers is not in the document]
nlordell: You could also implement `TryFrom<Solution> for logic::competition::Solution`.

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
+#[serde_as]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct ConstantProductPool {
+    id: String,
+    address: H160,
+    #[serde_as(as = "serialize::U256")]
+    gas_estimate: U256,
+    tokens: HashMap<H160, ConstantProductReserve>,
+    fee: f64,
```
nlordell, line 143: I think `fee: BigDecimal` or something similar as the OpenAPI spec specifies this as a `Decimal` type, which is an arbitrary precision decimal number encoded as a string (so `"13.37"` and not `13.37`).

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,240 @@
...
+    gas_estimate: U256,
+    tokens: Vec<H160>,
+    // TODO OpenAPI wants this to be a BigInt, why? Can this be negative?
+    #[serde_as(as = "serialize::U256")]
+    sqrt_price: U256,
+    #[serde_as(as = "serialize::U256")]
+    liquidity: U256,
+    // TODO OpenAPI wants this to be a BigInt, why? Can this be negative?
+    #[serde_as(as = "serialize::U256")]
+    tick: U256,
```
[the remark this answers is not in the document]
nlordell: Note that `tick`s fit in a `int24`:

https://github.com/Uniswap/v3-core/blob/05c10bf6d547d6121622ac51c457f93775e1df09/contracts/UniswapV3Pool.sol#L93-L96

So I will propose changing this to an `i32`.

crates/driver/src/solver/dto/solution.rs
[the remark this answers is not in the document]
nlordell: Oh yeah - duh.

I would say `into_domain_solution`, (even if it is a bit wordy). Aligns with the existing `to_boundry_*` naming that we already have.

crates/driver/src/logic/competition/auction.rs
nlordell: Oh missed that! Very nice.

crates/driver/src/solver/dto/solution.rs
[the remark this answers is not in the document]
nlordell: Also prefer 1.

crates/driver/src/util/serialize/string.rs
```diff
@@ -0,0 +1,49 @@
+use {
```
nlordell, line 1: This is `serde_with::DisplayFromStr`, I don't think we need to re-implement it.

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct ConstantProductPool {
+    id: String,
+    address: H160,
+    #[serde_as(as = "serialize::U256")]
+    gas_estimate: U256,
+    tokens: HashMap<H160, ConstantProductReserve>,
+    #[serde_as(as = "serialize::String")]
+    fee: num::BigRational,
```
nlordell, line 144: I think this needs to be a `bigdecimal::BigDecimal`. Rationals serialize differently:
```
assert_eq!(num::rational::Ratio::new(1, 10).to_string(), "1/10");
```

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
 #[derive(Debug, Serialize)]
-struct Amm {}
+#[serde(rename_all = "camelCase")]
+struct WeightedProductPool {
+    id: String,
+    address: H160,
+    #[serde_as(as = "serialize::U256")]
+    gas_estimate: U256,
+    tokens: HashMap<H160, WeightedProductReserve>,
+    fee: f64,
```
nlordell, line 163: This also needs to be a big decimal type. For example, the Uniswap V2 fee is 0.003 which is already cannot be exactly represented by a `f64`:
```rust
    assert_eq!(
        format!("{:.60}", 0.003),
        "0.003000000000000000062450045135165055398829281330108642578125",
    );
```

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
-struct Metadata {
-    environment: Option<String>,
-    auction_id: Option<i64>,
-    run_id: Option<u64>,
-    gas_price: Option<f64>,
-    native_token: Option<H160>,
+struct WeightedProductReserve {
+    #[serde_as(as = "serialize::U256")]
+    balance: U256,
+    weight: f64,
```
nlordell, line 171: Idem, needs to be a `bigdecimal::BigDecimal`.

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct StablePool {
+    id: String,
+    address: H160,
+    #[serde_as(as = "serialize::U256")]
+    gas_estimate: U256,
+    tokens: HashMap<H160, StableReserve>,
+    amplification_parameter: f64,
+    fee: f64,
```
nlordell, line 184: ☝️

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
+    gas_estimate: U256,
+    tokens: Vec<H160>,
+    #[serde_as(as = "serialize::U256")]
+    sqrt_price: U256,
+    #[serde_as(as = "serialize::U256")]
+    liquidity: U256,
+    tick: i32,
+    #[serde_as(as = "HashMap<_, serialize::U256>")]
+    liquidity_net: HashMap<usize, U256>,
+    fee: f64,
```
nlordell, line 213: Sorry for being annoying 🙈.

nlordell (approved):
LGTM. Approving to speed up the back and forth, but I see two small DTO issues:
- `num::BigRational` does not serialize in the expected format (we can implement serialization if you want to avoid `bigdecimal::BigDecimal`).
- Some `f64`s need to be converted to `BigDecimal` (or `BigRational` if you want to go that route).

crates/driver/src/solver/dto/auction.rs
```diff
@@ -1,81 +1,234 @@
...
+    #[serde_as(as = "serialize::U256")]
+    gas_estimate: U256,
+    tokens: Vec<H160>,
+    #[serde_as(as = "serialize::U256")]
+    sqrt_price: U256,
+    #[serde_as(as = "serialize::U256")]
+    liquidity: U256,
+    tick: i32,
+    #[serde_as(as = "HashMap<_, serialize::U256>")]
+    liquidity_net: HashMap<usize, U256>,
```
nlordell, line 212: These are a map of tick -> U256, so this should be:
```suggestion
    #[serde_as(as = "HashMap<serialize::String, serialize::U256>")]
    liquidity_net: HashMap<i32, U256>,
```

Might be worth adding a `Tick` domain type (not needed for the DTO here though).

PR #1000, opened 2022-12-26
Driver `/solve` API

sistemd:
Progress towards https://github.com/cowprotocol/services/issues/895, the only thing missing now are tests :tada: .

Implements the auction and solution DTOs for `/solve`.

The auction DTO is very similar to the auction DTO from the driver-solver interface, with a few additional fields. The solution DTO is very straightforward.


crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -1,84 +1,211 @@
...
+                            },
+                            data: order.signature,
+                            signer: order.owner.into(),
+                        },
+                        reward: order.reward,
+                    })
+                })
+                .try_collect()?,
+            // TODO #899
+            liquidity: Default::default(),
```
nlordell, line 123: I don't think that the protocol will ever send liquidity as part of the `Auction` DTO - I only think this exists in the driver <> solver `Auction` DTO (as it is populated by the driver, and doesn't come from the protocol). I think we can remove it here.

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -1,84 +1,211 @@
...
     valid_to: u32,
-    #[serde_as(as = "serialize::Hex")]
-    app_data: [u8; 32],
-    kind: OrderKind,
+    kind: Kind,
+    receiver: Option<H160>,
+    owner: H160,
     partially_fillable: bool,
-
+    // TODO Always zero if the order is not partially fillable, is that OK?
```
nlordell, line 172: 👍

crates/driver/src/api/solve/dto/auction.rs
```diff
@@ -87,56 +214,151 @@ enum SellTokenBalance {
...
+    available_balance: U256,
+    trusted: bool,
 }
 
-#[derive(Debug, Default, Deserialize)]
-struct Interactions {
-    pre: Vec<InteractionData>,
+#[derive(Debug, Deserialize)]
+#[serde(tag = "kind", rename_all = "lowercase")]
+enum Liquidity {
```
nlordell, line 256: Shouldn't this exist in the `solver::dto` and not in `api::solve::dto`? It is my understanding that the former is the DTO between the driver and solver, and that the latter is the DTO between the protocol and the driver.

crates/driver/src/solver/dto/auction.rs
```diff
@@ -91,6 +91,8 @@ struct Order {
     partially_fillable: bool,
     class: Class,
     reward: f64,
+    // TODO Was the intention to have the driver verify signatures, and the solvers blindly trust
+    // the orders and the driver auth? Or did we just forget to add the signature here?
```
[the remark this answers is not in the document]
nlordell: The solver engine doesn't need the order signature - as it is inconsequential to the settlement from a purely "mathematical price-finding" perspective.

> blindly trust the orders and the driver auth?

These signatures are only important for ensuring user orders are authorized. As such, even if we gave an order to the driver with a bad signature which was then relayed to the solver (with signature omitted) and then used in a solution - the worse that can happen is that the solution fails to simulate and/or revert.

Furthermore, as the driver is intended to be co-located with the solver, I don't think we need strong trust assumptions here, and I never imagined that solver engines would verify signatures for all the orders it receives.

crates/driver/src/api/solve/dto/auction.rs
nlordell: I see it was add in #999 in the solver DTO (where I think it should be) - so I would remove this from here.

crates/driver/src/api/solve/dto/auction.rs
nlordell: Same comment as ☝️, so closing this one for less noise.

vkgnosis (approved):

crates/driver/src/api/solve/dto/auction.rs
nlordell: I understood that this was the DTO that was being sent from the `autopilot` to the `driver` which doesn't have liquidity 🤦.

crates/driver/src/api/solve/dto/auction.rs
nlordell, line 256: This shouldn't be here though right?

There is no liquidity DTO from `autopilot` -> `driver`.

nlordell (approved):
LGTM. I don't think we need `Liquidity` DTO from `autopilot` -> `driver`.

PR #1001, opened 2022-12-26
Some CLI args for the driver

sistemd:
Progress on https://github.com/cowprotocol/services/issues/895, preparing the codebase to start adding tests.

Add some CLI args. Add local dev geth to docker-compose. I did this because only geth supports `eth_createAccessList` which I want to use during testing. Adds the Ethereum implementation for `Simulator` which uses `eth_createAccessList` and `eth_estimateGas`.

### DIFF
--- a/crates/driver/src/blockchain/mod.rs
+++ b/crates/driver/src/blockchain/mod.rs
@@ -19,10 +19,7 @@ impl Ethereum {
     /// Access the Ethereum blockchain through an RPC API hosted at the given
     /// URL.
     pub async fn ethrpc(url: &Url) -> Result<Self, web3::Error> {
-        // TODO Enable batching, reuse ethrpc? Put it in the boundary module?
-        // I feel like what we have in shared::ethrpc could be simplified if we use
-        // web3::transports::batch or something, but I haven't looked deep into it, just
-        // a gut feeling.
+        // TODO Probably move shared::ethrpc into its own crate and reuse it here
         let web3 = Web3::new(web3::transports::Http::new(url.as_str())?);
         let chain_id = web3.eth().chain_id().await?.into();
         let network_id = web3.net().version().await?.into();


crates/driver/src/blockchain/mod.rs
```diff
@@ -1,30 +1,46 @@
...
 impl Ethereum {
     /// Access the Ethereum blockchain through an RPC API hosted at the given
     /// URL.
-    pub fn eth_rpc(_url: Url) -> Self {
-        todo!()
+    pub async fn ethrpc(url: &Url) -> Result<Self, web3::Error> {
+        // TODO Enable batching, reuse ethrpc? Put it in the boundary module?
+        // I feel like what we have in shared::ethrpc could be simplified if we use
+        // web3::transports::batch or something, but I haven't looked deep into it, just
+        // a gut feeling.
```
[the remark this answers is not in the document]
nlordell: Metrics, logging and auto-batching are the three reasons we use it. It is basically a custom `Transport` implementation that we use. It can live alone in a `ethrpc` crate AFAIU.

crates/driver/src/cli.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn is_specified(&self) -> bool {
+        if self.tenderly_url.is_none()
+            && self.tenderly_api_key.is_none()
+            && self.tenderly_user.is_none()
+            && self.tenderly_project.is_none()
+        {
+            false
+        } else if self.tenderly_api_key.is_some()
+            && self.tenderly_user.is_none()
+            && self.tenderly_project.is_none()
```
nlordell, line 71:
```suggestion
            && self.tenderly_user.is_some()
            && self.tenderly_project.is_some()
```

crates/driver/src/main.rs
```diff
@@ -124,513 +68,43 @@ async fn run() {
...
-            IUniswapLikeRouter::at(&web3, router_address),
-            settlement_contract.clone(),
-            web3.clone(),
-            pool_cache.clone(),
-        )));
-    }
-    res
-}
+// TODO For solvers, I feel like we should have a YAML or JSON file and only
+// specify a path to it, otherwise we get into nightmare land. Opinions?
```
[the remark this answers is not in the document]
nlordell: The issue we ran into the past was K8s annoyances.

I agree 100% though (and think this is true for more than just solver configuration, I prefer generally configuring things with files).

crates/driver/src/settlement_proposal.rs
```diff
@@ -1,464 +0,0 @@
...
-/// competition.
-#[derive(Debug, Default, Clone)]
-pub struct SettlementProposal {
-    pub clearing_prices: HashMap<H160, U256>,
-    pub trades: Vec<TradedOrder>,
-    pub execution_plan: Vec<Arc<dyn InteractionProposal>>,
-    pub auction_id: u64,
-}
-
-impl SettlementProposal {
```
nlordell, line 146: I think this still needs to be ported into the driver. We wanted the driver to verify solutions before submitting them to prevent accidental bad behavious from the `solver engine`s.

crates/driver/src/driver.rs
```diff
@@ -1,393 +0,0 @@
...
-                    }
-                },
-                _ = &mut timeout => return current_solution
-            }
-        }
-    }
-
-    /// Validates that the `Settlement` satisfies expected fairness and
-    /// correctness properties.
-    async fn validate_settlement(&self, settlement: Settlement) -> Result<SimulationWithResult> {
```
nlordell, line 117: I think this still needs to be ported into the driver. We wanted the driver to verify solutions before submitting them to prevent accidental bad behaviour from the `solver engine`s.

nlordell (approved):
LGTM

crates/driver/src/driver.rs
nlordell: (Not in this PR of course).

vkgnosis (approved):

crates/driver/src/driver.rs
nlordell: I think simulation is one aspect of `validate_settlement`. We also do some other basic validations like token conservation and checking that internalized transactions are legal (i.e. in the list of accepted tokens). This is needed because not all rules are enforced on-chain, and breaking some of the off-chain rules results in slashing. We wanted to reference driver to protect against these cases.

Might be worth creating an issue to not forget to get back to this.

PR #1006, opened 2022-12-27
Load solver config from TOML file in `driver`

sistemd:
Load solver config from a TOML file.

### DIFF
--- a/crates/driver/src/infra/config/solvers.rs
+++ b/crates/driver/src/infra/config/solvers.rs
@@ -2,13 +2,16 @@ use {
     crate::{domain::eth, infra::solver, util::serialize},
     serde::Deserialize,
     serde_with::serde_as,
-    std::{fs, path::Path},
+    std::path::Path,
+    tokio::fs,
 };
 
-/// Load the solver configuration from a YAML file. Despite being an I/O
-/// operation, this is done at startup, so it doesn't need to be async.
-pub fn load(path: &Path) -> Vec<solver::Config> {
-    let data = fs::read(path).unwrap_or_else(|_| panic!("I/O error while reading {path:?}"));
+/// Load the solver configuration from a YAML file. Panics if the config is
+/// invalid or on I/O errors.
+pub async fn load(path: &Path) -> Vec<solver::Config> {
+    let data = fs::read(path)
+        .await
+        .unwrap_or_else(|_| panic!("I/O error while reading {path:?}"));
     let config: Config = toml::de::from_slice(&data)
         .unwrap_or_else(|_| panic!("YAML syntax error while reading {path:?}"));
     config
--- a/crates/driver/src/main.rs
+++ b/crates/driver/src/main.rs
@@ -37,7 +37,7 @@ async fn run() {
     let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
     let eth = ethereum(&args).await;
     let serve = Api {
-        solvers: solvers(&args),
+        solvers: solvers(&args).await,
         simulator: simulator(&args, &eth),
         eth,
         addr: args.bind_addr,
@@ -82,8 +82,9 @@ async fn ethereum(args: &cli::Args) -> Ethereum {
         .expect("initialize ethereum RPC API")
 }
 
-fn solvers(args: &cli::Args) -> Vec<Solver> {
+async fn solvers(args: &cli::Args) -> Vec<Solver> {
     config::solvers::load(&args.solvers_config)
+        .await
         .into_iter()
         .map(Solver::new)
         .collect()


crates/driver/src/infra/config/solvers.rs
```diff
@@ -0,0 +1,47 @@
...
+    serde_with::serde_as,
+    std::{fs, path::Path},
+};
+
+/// Load the solver configuration from a YAML file. Despite being an I/O
+/// operation, this is done at startup, so it doesn't need to be async.
+pub fn load(path: &Path) -> Vec<solver::Config> {
+    let data = fs::read(path).unwrap_or_else(|_| panic!("I/O error while reading {path:?}"));
+    let config: Config = serde_yaml::from_slice(&data)
+        .unwrap_or_else(|_| panic!("YAML syntax error while reading {path:?}"));
```
vkgnosis, line 13: Why panic instead of returning an error? The answer is probably that this function is only used at startup where we would panic anyway. Still, the overhead of converting this to result is probably very small and then I'd rather unwrap in main so it is clear that that is what's going on. If the panic is kept inside here then it should be documented.

crates/driver/solvers.example.yaml
```diff
@@ -0,0 +1,10 @@
+solvers:
```
vkgnosis, line 1: If we have an example yaml config then we might as well test it with the function that is supposed to load it to make sure the example and the code don't go out of sync.

crates/driver/Cargo.toml
```diff
@@ -23,10 +23,11 @@ hyper = "0.14"
 serde_with = "2.0"
 async-trait = "0.1"
 itertools = "0.10"
-url = "2.3"
+url = { version = "2.3", features = ["serde"] }
 ethabi = "18.0"
 rand = "0.8"
 web3 = "0.18"
+serde_yaml = "0.9"
```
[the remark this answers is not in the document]
nlordell: TOML makes nested objects harder to represent. YAML is a little nicer for that, but the downside being that it is sensitive to whitespace and generally a much more complicated format.

I would suggest:
- If the configuration is mostly flat, TOML because it is simpler
- If the configuration will be deeply nested, YAML is more convenient.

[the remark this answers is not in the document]
nlordell: My **personal** preference is TOML, but mostly because it is much simpler than YAML ([TOML spec](https://toml.io/en/v1.0.0) is less then half the size of the [YAML spec](https://yaml.org/spec/1.2.2/)) and has [fewer gotchas](https://noyaml.com/).

That being said, I do see the value of YAML for complicated configurations.

crates/driver/src/infra/config/solvers.rs
nlordell: > Why panic instead of returning an error?

For things like loading configurations, I think it is fine to panic within the function. My reasoning is that:
- You no longer need to pass the configuration around as any code can `Config::load()` and not need to panic internally.
- Invalid configuration is an error case where I don't see any code needing to handle the error case (if its wrong, there is nothing to do).

> If the panic is kept inside here then it should be documented.

Agreed

crates/driver/src/boundary/settlement.rs
```diff
@@ -409,6 +409,15 @@ fn to_token_amount(asset: &eth::Asset) -> TokenAmount {
     }
 }
 
+fn to_big_decimal(value: bigdecimal::BigDecimal) -> num::BigRational {
+    let (x, exp) = value.into_bigint_and_exponent();
+    let numerator_bytes = x.to_bytes_le();
+    let base = num::bigint::BigInt::from_bytes_le(numerator_bytes.0, &numerator_bytes.1);
+    let ten = num::BigRational::new(10.into(), 1.into());
+    let numerator = num::BigRational::new(base, 1.into());
+    numerator / ten.pow(exp.try_into().expect("should not overflow"))
```
vkgnosis, line 418:
```suggestion
    numerator / num::traits::Pow::pow(&ten, exp)
```
no unwrap

crates/driver/src/boundary/settlement.rs
```diff
@@ -409,6 +409,15 @@ fn to_token_amount(asset: &eth::Asset) -> TokenAmount {
     }
 }
 
+fn to_big_decimal(value: bigdecimal::BigDecimal) -> num::BigRational {
+    let (x, exp) = value.into_bigint_and_exponent();
```
vkgnosis, line 413: This feels like a function that could have a unit test because it's easy to switch up numerator and denominator or have an off by one error.

vkgnosis (approved):

crates/driver/solvers.example.toml
```diff
@@ -0,0 +1,12 @@
+[[solvers]]
+name = "mysolver" # Arbitrary name given to this solver, must be unique
+url = "http://localhost:1234/solve" # Endpoint of the solver
+absolute-slippage = 12 # Denominated in wei, optional
+relative-slippage = 0.1 # Percentage in the [0, 1] range
```
MartinquaXD, line 5: Should passing something outside the range of `[0, 1]` throw an error? This case is currently not checked.

MartinquaXD (approved):

crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -18,8 +17,11 @@ pub struct Ethereum {
 impl Ethereum {
     /// Access the Ethereum blockchain through an RPC API hosted at the given
     /// URL.
-    pub async fn ethrpc(url: &Url) -> Result<Self, web3::Error> {
-        // TODO Probably move shared::ethrpc into its own crate and reuse it here
+    pub async fn ethrpc(url: &url::Url) -> Result<Self, web3::Error> {
+        // TODO Enable batching, reuse ethrpc? Put it in the boundary module?
+        // I feel like what we have in shared::ethrpc could be simplified if we use
+        // web3::transports::batch or something, but I haven't looked deep into it, just
```
nlordell, line 23: We use `web3::transports::batch` under the hood. The issue is that it only allows manual batching AFAIK.

crates/driver/src/infra/config/solvers.rs
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/// Load the solver configuration from a YAML file. Panics if the config is
+/// invalid or on I/O errors.
+pub async fn load(path: &Path) -> Vec<solver::Config> {
+    let data = fs::read(path)
+        .await
+        .unwrap_or_else(|_| panic!("I/O error while reading {path:?}"));
+    let config: Config = toml::de::from_slice(&data)
+        .unwrap_or_else(|_| panic!("YAML syntax error while reading {path:?}"));
```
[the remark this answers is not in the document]
nlordell:
```suggestion
        .unwrap_or_else(|_| panic!("TOML syntax error while reading {path:?}"));
```

crates/driver/solvers.example.toml
```diff
@@ -0,0 +1,12 @@
...
+address = "0xE33bff104C4E1140A3f86f855AD5bDAb8Dc6CA8C" # The ETH address of this solver
```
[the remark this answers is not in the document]
nlordell: I like this, as we can use the following to specify a solver by address (for running in "DryRun" mode for example):
```
address = "0xE33bff104C4E1140A3f86f855AD5bDAb8Dc6CA8C"
```

and the following to specify a solver account by private key (for actually running).
```
privatekey = "0x..."
```

crates/driver/solvers.example.toml
```diff
@@ -0,0 +1,12 @@
+[[solvers]]
+name = "mysolver" # Arbitrary name given to this solver, must be unique
+url = "http://localhost:1234/solve" # Endpoint of the solver
```
nlordell, line 3: From the code, I think this is:
```suggestion
endpoint = "http://localhost:1234/solve" # Endpoint of the solver
```

nlordell (approved):
Code looks good. Can you `s/YAML/TOML/` in the PR title and description?

PR #1015, opened 2022-12-30
Linear Order Matching Algorithm in `alerter`

nlordell:
This PR refactors the limit order inclusion check to be linear instead of quadratic in the `alerter`. Given the orderbook size of ~500 orders, I do think this may start to pay off at this point.

### Test Plan

Run the alerter and see it do its thing.

### DIFF
--- a/crates/alerter/src/main.rs
+++ b/crates/alerter/src/main.rs
@@ -12,7 +12,10 @@ use model::{
 use primitive_types::{H160, U256};
 use prometheus::IntGauge;
 use reqwest::Client;
-use std::time::{Duration, Instant};
+use std::{
+    collections::HashMap,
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

fleupold (approved):
🎉

harisang:
Minor comment. I would change the name of the PR. "Non-quadratic" is a bit imprecise. I suspect you mean linear here, right?

Was it chosen this way to highlight that up till now we were doing quadratic?

nlordell:
> I suspect you mean linear here, right?

I do.

> Was it chosen this way to highlight that up till now we were doing quadratic?

It was from a "this is quadratic" comment in a different PR. Agree that the name isn't great here. Will change it.

PR #1022, opened 2023-01-02
Missing pools update is not maintenance error

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1020

UniswapV3PoolFetcher does two things in a single maintenance run:
1. Update events for already cached pools.
2. Fetch pools state checkpoint for non-cached pools (this is where the problematic subgraph call is used).

This PR lowers severity of error for task (2). In case of error during missing pools update, we just log the error.

### DIFF
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -452,7 +452,7 @@ impl Maintaining for UniswapV3PoolFetcher {
             self.checkpoint.update_missing_pools()
         );
         result1?;
-        // since failure in updating the missing pools is not critical for UniswapV3PoolFetcher maintenance 
+        // since failure in updating the missing pools is not critical for UniswapV3PoolFetcher maintenance
         // and future liquidity fetch calls, then there is no need to return error
         if let Err(err) = result2 {
             tracing::warn!(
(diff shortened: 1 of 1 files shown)


crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -451,7 +451,15 @@ impl Maintaining for UniswapV3PoolFetcher {
...
-        result1.and(result2)?;
+        result1?;
+        // since failure in updating the missing pools is not critical for UniswapV3PoolFetcher maintenance
+        // and future liquidity fetch calls, then there is no need to return error
+        if let Err(err) = result2 {
+            tracing::warn!(
+                "UniswapV3PoolFetcher failed to update missing pools: {}",
+                err
+            );
+        }
```
nlordell, line 462: If I remember correctly, pool fetching works by:
- If the pool is missing, queue an update
- Do an update
- Get the updated pool info.

Is there some code in the pool fetching that expects the missing pool to be there after the update?

sunce86: No. Failure to update missing pools is safe. The only consequence is that we would need to wait for the next maintenance run for it to be updated. Also, if we try to fetch liquidity for tokens contained in the missing pool before the next maintenance run it will be a cache miss therefore not critical.

vkgnosis: Do we get old data or do we get no data? Old data might be a problem.

sunce86: We get no data.

nlordell (approved):
LGTM! Thanks for solving this so quickly.

PR #1023, opened 2023-01-02
Update hardhat config for docker testing container

MartinquaXD:
While working on a patch I wanted to run e2e tests locally but they failed. After some investigation I saw that the hardhat config file we initialize the docker testing container with differs slightly from the hardhat config we use for our CI tests.
The difference was that only the CI config sets the `initialDate` to enable time travel shenanigans required for eth flow e2e tests.

Edit: Instead of keeping the github workflow file and the `hardhat.config.js` file in sync I made the github workflow file import the other file. That means all the hardhat changes required for our e2e test should be added to [hardhat.config.js](https://github.com/cowprotocol/services/blob/main/docker/hardhat/hardhat.config.js) from now on.

### Test Plan
Ran e2e test locally using the updated docker container.
e2e tests in CI still work.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -74,19 +74,7 @@ jobs:
       - name: Setup Hardhat
         working-directory: ${{ env.HARDHAT_PATH_ }}
         run: |
-          cat > hardhat.config.js <<EOF
-          module.exports = {
-            networks: {
-              hardhat: {
-                initialBaseFeePerGas: 0,
-                initialDate: "2000-01-01T00:00:00.000+00:00",
-                accounts: {
-                  accountsBalance: "1000000000000000000000000"
-                },
-              },
-            },
-          };
-          EOF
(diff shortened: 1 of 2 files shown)


nlordell:
Maybe it makes sense to use the `hardhat` Docker image that we build for local testing for the E2E test as well?

nlordell (approved):

vkgnosis (approved):

.github/workflows/pull-request.yaml
```diff
@@ -74,19 +74,7 @@ jobs:
...
-                initialBaseFeePerGas: 0,
-                initialDate: "2000-01-01T00:00:00.000+00:00",
-                accounts: {
-                  accountsBalance: "1000000000000000000000000"
-                },
-              },
-            },
-          };
-          EOF
+          cp ${{ github.workspace }}/docker/hardhat/hardhat.config.js > hardhat.config.js
```
nlordell, line 77: 🎉

PR #1024, opened 2023-01-02
Fetch missing native prices in background task

MartinquaXD:
This PR tries to optimize how we fetch native prices for the solvable orders.

Currently the approach is to:
1. collect all traded tokens in the solvable orders
2. query their native prices
3. filter out all solvable orders for which we don't have the native prices of the tokens

Step `2` already uses the `CachingNativePriceEstimator` so all the cached prices get returned immediately and we only issue price estimation requests for the uncached prices. In order to not delay the auction creation for too long in case we have too many uncached prices step `2` gets aborted after some time (3.5 seconds). Since estimating a single price can already take 2 seconds and we currently don't do parallel price estimates we can usually only get 2 new prices in the 3.5 seconds. This means by delaying the entire auction by up to 3.5s we might add only a single solvable order in the worst case.
Also note that the price estimation requests get aborted after 3.5 seconds so we can't make additional progress there.

To decrease the auction creation time and keep making progress in the background this PR proposes following change:
1. collect all traded tokens in the solvable orders
2. return the currently cached relevant native prices immediately
3. mark entries of missing prices in the `CachingNativePriceEstimator` such that the existing background task will fetch them during the next maintenance cycles
4. filter out all solvable orders for which we don't have the cached prices of the traded tokens

### Test Plan
Updated existing unit tests.

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -46,7 +46,11 @@ impl Inner {
     }
 
     /// Returns cached prices and uncached indices.
-    fn get_cached_prices(&self, tokens: &[H160], max_age: &Duration) -> (Vec<(usize, f64)>, Vec<usize>) {
+    fn get_cached_prices(
+        &self,
+        tokens: &[H160],
+        max_age: &Duration,
+    ) -> (Vec<(usize, f64)>, Vec<usize>) {
         if tokens.is_empty() {
             return Default::default();
         }
@@ -77,9 +81,7 @@ impl Inner {
                     // check if price is cached by now
                     let now = Instant::now();
                     let mut cache = self.cache.lock().unwrap();
-                    if let Some(price) =
-                        Self::get_cached_price(token, now, &mut cache, &max_age)
-                    {
+                    if let Some(price) = Self::get_cached_price(token, now, &mut cache, &max_age) {
                         return (index, Ok(price));
                     }
                 }
@@ -115,7 +117,7 @@ impl Inner {
 /// Wrapper around `Box<dyn PriceEstimating>` which caches successful price estimates for some time
 /// and supports updating the cache in the background.
 /// The size of the underlying cache is unbounded.
-pub struct CachingNativePriceEstimator{
+pub struct CachingNativePriceEstimator {
     inner: Arc<Inner>,
     max_age: Duration,
     metrics: &'static Metrics,
@@ -146,7 +148,11 @@ impl CachingNativePriceEstimator {
             max_age.saturating_sub(prefetch_time.unwrap_or(PREFETCH_TIME)),
         ));
         let metrics = Metrics::instance(global_metrics::get_metric_storage_registry()).unwrap();
-        Self{ inner, max_age, metrics }
+        Self {
+            inner,
+            max_age,
+            metrics,
+        }
     }
 
     /// Returns cached prices immediately and spawns a background task that fetches the missing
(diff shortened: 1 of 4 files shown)


nlordell:
Ah, this is a nice idea - so an auction can only ever "queue" native prices to be collected in a background task but doesn't actually compute them.

vkgnosis:
The design you outlined makes sense to me. Don't we already have an automatic native price updater component? It would make sense to merge this with the component you propose. Something that keeps track of which native prices are important and at some rate per second estimates then. Then like you said solvable_orders uses whatever prices are available now and can queue the others into the component.

nlordell:
> Don't we already have an automatic native price updater component?

I think it updates recent prices as part of "service maintenance". Agree that we should potentially consolidate these two together.

MartinquaXD:
> Don't we already have an automatic native price updater component? It would make sense to merge this with the component you propose.

Yes. The background task we have at the moment periodically checks for outdated prices and updates the n most recently requested ones. One way to integrate that would be to query the prices we are interested in and update the `requested_at` field (which would give them a higher priority in the next update cycle.

I believe this approach should only require slight modifications to this PR.

vkgnosis (approved):

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -41,11 +41,8 @@ impl Inner {
             Entry::Occupied(mut entry) => {
                 let entry = entry.get_mut();
                 entry.requested_at = now;
-                if now.saturating_duration_since(entry.updated_at) < *max_age {
-                    Some(entry.price)
-                } else {
-                    None
-                }
+                let is_recent = now.saturating_duration_since(entry.updated_at) < *max_age;
+                is_recent.then_some(entry.price)
```
nlordell, line 45: `then_some` is super cool!

nlordell (approved):
Code looks good.

Do we need a policy to remove tokens from the cache? My intuition is "not super important", since un-requested entries will just sit there and don't take up much space, but in theory the cache grows in an unbounded way.

MartinquaXD:
> Do we need a policy to remove tokens from the cache? My intuition is "not super important", since un-requested entries will just sit there and don't take up much space, but in theory the cache grows in an unbounded way.

Good thing you brought that up. I think it was not an issue before but this PR currently has a problem because of the way I made the background task fetch missing prices. Now we create entries for all tokens that have ever been requested instead of only for the tokens where we got a successful price estimate. That means theoretically a malicious actor could request native prices for all sorts of nonsensical tokens and blow up our cache.
We could get back to the status quo by only creating new entries for prices missing during auction building. There we know the tokens are legit because we already have orders for them in the system.
I think this edge case should be fixed before merging this PR but afterwards I would only think about a removal policy when it actually becomes a problem.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -30,12 +30,12 @@ struct Inner {
 
 impl Inner {
     // Returns a single cached price and updates its `requested_at` field.
-    // If there is no entry for the token yet an outdated cache entry gets created.
     fn get_cached_price(
         token: &H160,
         now: Instant,
         cache: &mut MutexGuard<HashMap<H160, CachedPrice>>,
         max_age: &Duration,
+        create_missing_entry: bool,
```
vkgnosis, line 38: It's going to be unclear later why we have this. Could you add a comment somewhere explaining the same reasoning as in the PR.

MartinquaXD: I already added it [here](https://github.com/cowprotocol/services/pull/1024/commits/50d6b585c6e24484b3259809441b38314dedb500#diff-50d12b254f11a969c7dc675e6f3181c39a3785e541dac0963f869d2d3ec9bacaR51-R52). Do you think it should rather be documented at the function signature?

vkgnosis: That's fine. In the diff I glanced over that this is extending the comment too.

Issue #1021, opened 2023-01-02
Fix "economic viability" for the Gnosis_0x solver


harisang:
The 0x solver is one of the worst solvers in terms of settling non viable trades, often picking on so-called arbitrage stable-to-stable trades.  (see also discussion [here](https://cowservices.slack.com/archives/C036JAGRQ04/p1672679574673589?thread_ts=1672154090.388719&cid=C036JAGRQ04)).

We propose that we modify the solver by forcing it to not submit any solution that has non-positive objective value.

MartinquaXD:
If this approach is correct for 0x it's probably the right thing to do in general. In that case this fix should be done for all solvers (even if they might not run into that problem frequently).

harisang:
> If this approach is correct for 0x it's probably the right thing to do in general. In that case this fix should be done for all solvers (even if they might not run into that problem frequently).

We are still a bit undecided on this, as we don't have a clear idea how much this would impact the user experience. The thing is that in times of high volatility, we still want to maintain a good throughput, and by enforcing this for multiple solvers we might risk low settlement rates.

It might sound contradictory to an extent, but I specifically targeted the 0x solver as it seems to be lagging a bit, picking consistently on arbitrage opportunities even under low gas costs, that by the time it attempts to extract those, these opportunities are gone. This is not so much the case with 1inch or Paraswap, that's why I propose to first start only with 0x to see how it goes.

PR #1025, opened 2023-01-02
Discard solution with negative objective value

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1021

Removes 0x solutions with negative objective value and notifies solver about it.

### DIFF
--- a/crates/solver/src/settlement_ranker.rs
+++ b/crates/solver/src/settlement_ranker.rs
@@ -164,10 +164,32 @@ impl SettlementRanker {
             );
         }
 
-        let (mut rated_settlements, errors) = self
+        let (rated_settlements, errors) = self
             .settlement_rater
             .rate_settlements(solver_settlements, external_prices, gas_price)
             .await?;
+
+        // Discard solutions with negative obj value
+        let mut rated_settlements = rated_settlements
+            .into_iter()
+            .filter(|(solver, rated_settlement, _)| {
+                // TODO for now, apply only for 0x solver
+                if solver.name() != "0x" {
+                    return true;
+                }
+
+                let objective_value = rated_settlement.objective_value();
+                let objective_value_positive = objective_value.ge(&Ratio::from_integer(0.into()));
+                if !objective_value_positive {
+                    tracing::debug!("Solution rejected for negative objective value. Solver={}, objective value={}", solver.name(), objective_value);
+                    solver.notify_auction_result(
+                        auction_id,
+                        AuctionResult::Rejected(SolverRejectionReason::NegativeObjectiveValue),
+                    );
+                }
+                objective_value_positive
+            })
+            .collect::<Vec<_>>();
 
         // Before sorting, make sure to shuffle the settlements. This is to make sure we don't give
         // preference to any specific solver when there is an objective value tie.
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -332,6 +332,9 @@ pub enum SolverRejectionReason {
     /// The solution contains non unique execution plans (duplicated coordinates)
     InvalidExecutionPlans,
 
+    /// Solution has negative objective value
+    NegativeObjectiveValue,
+
     /// The solution didn't pass simulation. Includes all data needed to re-create simulation locally
     SimulationFailure(TransactionWithError),
 }


crates/solver/src/settlement_ranker.rs
```diff
@@ -164,11 +164,33 @@ impl SettlementRanker {
...
-        let (mut rated_settlements, errors) = self
+        let (rated_settlements, errors) = self
             .settlement_rater
             .rate_settlements(solver_settlements, external_prices, gas_price)
             .await?;
 
+        // Discard solutions with negative obj value
+        let mut rated_settlements = rated_settlements
+            .into_iter()
+            .filter(|(solver, rated_settlement, _)| {
```
MartinquaXD, line 175: `.into_iter().filter().collect()` could be replaced with `.retain()`.

MartinquaXD (approved):

crates/solver/src/settlement_ranker.rs
```diff
@@ -164,11 +164,33 @@ impl SettlementRanker {
...
+                // TODO for now, apply only for 0x solver
+                if solver.name() != "0x" {
+                    return true;
+                }
+
+                let objective_value = rated_settlement.objective_value();
+                let objective_value_positive = objective_value.ge(&Ratio::from_integer(0.into()));
```
MartinquaXD, line 182: This could be slightly easier to read with `Ratio::zero()`.

```suggestion
                let objective_value_positive = objective_value.ge(&Ratio::zero());
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -164,11 +164,33 @@ impl SettlementRanker {
...
             .await?;
 
+        // Discard solutions with negative obj value
+        let mut rated_settlements = rated_settlements
+            .into_iter()
+            .filter(|(solver, rated_settlement, _)| {
+                // TODO for now, apply only for 0x solver
+                if solver.name() != "0x" {
+                    return true;
+                }
```
nlordell, line 179: This feels like a code smell - where we add solver-specific logic to the driver run loop.

Can we instead implement something on the solver level for this? A "DiscardNegativeObjectiveValueSolver" that only wraps the 0x solver?

vkgnosis: I wouldn't even create a wrapper. This should be done directly in the 0x code.

crates/solver/src/settlement_ranker.rs
sunce86: Ah yes. Thanks

crates/solver/src/settlement_ranker.rs
sunce86: Nice.

crates/solver/src/settlement_ranker.rs
sunce86: This is indeed a red flag, however acceptable (IMO) since we will either:
1. remove this line and therefore apply the code change to all solvers
2. revert this PR completely.

That said, would you be ok with merging this as is?

vkgnosis: I would like to have it directly in the 0x solver code and only there. Unless it turns out that doing that has other problems.

sunce86: I don't think we have a settlement simulation / gas price and everything else needed to compute the objective value in the solver code itself.

vkgnosis: Our Solver trait contains all that is needed to calculate objective value afaik.

nlordell: > This should be done directly in the 0x code.

I agree.

Originally, I had not suggested this as 0x is implemented as a single order solver though, so we would need wrap it I think:
```
struct ZeroExSolver(SingleOrderSolver<Inner>);
```

The reason being that merging solutions may be a net benefit (i.e. push a part of a solution from having a negative objective value to a positive one).

vkgnosis: It can still be done directly in the zeroex solver implementation at `crates/solver/src/solver/zeroex_solver.rs`. Not super clean but you can convert a single settlement into a full settlement and calculate the objective value. I would then not emit any single settlement that has negative object value. It shouldn't be part of the merging.

nlordell: The issue is gas metering on Ethereum means that the `... - gas_estimate` part can change depending on whether or not the partial settlement is alone or part of a larger merged settlement (and potentially make the objective become positive).

I.e. given:
- Settlement A: -X objective
- Settlement B: +Y objective

It is possible that:
- Settlement A+B: +Z objective
- where Z > Y

nlordell: And, to me at least, this isn't super contrived, and related to our value proposition (batching trades is more efficient then settling them on their own, and can push us into economical viability).

harisang: > I.e. given:
> 
>     * Settlement A: -X objective
> 
>     * Settlement B: +Y objective
> 
> 
> It is possible that:
> 
>     * Settlement A+B: +Z objective
> 
>     * where Z > Y

Interesting. Wasn't aware that this could happen! Could you briefly explain this or give a pointer so that I get an understanding of it?

harisang:
Also, as a comment, I understand that it is a bit tricky to get a very clean, 0x-contained, fix for this PR. In case this cannot be resolved easily, we could also experiment with turning off the 0x solver for a few days. I will do a brief post on slack, showcasing how many batches 0x currently wins, how large is the difference with the 2nd best solution etc, so that we are more confident towards turning it off, if needed.

crates/solver/src/settlement_ranker.rs
nlordell: Specifically, on Ethereum, reusing state (accessing contract code and storage for example) comes at a reduced cost the second time around. Additionally, we have a fixed overhead per settlement for executing it on-chain (i.e. calling the settlement contract).

With this in mind, If we consider a recent solution that was executed at 15 Gwei, and that the fixed overhead of a settlement is ~90K gas, that is about -0.00135Ξ to the objective value that will not be there when the settlements are combined (about $1.60). This is even before we start considering the significant discounts for accessing the same contract twice within a transaction (which is 2500 gas per state access - be it a storage slot or contract code).

PR #1026, opened 2023-01-03
Add metrics for refunder

fedgiac:
Closes #977.

Adds metrics and liveness checks to the refunder.
Refund loop errors have been demoted from `error!`s to `warn!`s.

### Test Plan

I ran the refunder locally with `cargo run --bin refunder -- --ethflow-contract 0xD02De8Da0B71E1B59489794F423FaBBa2AdC4d93 --refunder-pk 0x1111111111111111111111111111111111111111111111111111111111111111`.

For checking that everything worked I used:
```
curl --write-out '%{http_code}' localhost:9590/liveness
curl localhost:9590/metrics | grep refunder_main
```

I triggered refunder failures by disconnecting my computer from the internet.

<details><summary>Example output</summary>

```
# HELP refunder_main_refunding_loops Tracks the result of every refunding loops.
# TYPE refunder_main_refunding_loops counter
refunder_main_refunding_loops{result="error"} 20
refunder_main_refunding_loops{result="success"} 9
```
</details>

### DIFF
--- a/crates/refunder/src/lib.rs
+++ b/crates/refunder/src/lib.rs
@@ -13,8 +13,8 @@ use std::{
     time::{Duration, Instant},
 };
 
-const SLEEP_TIME_BETWEEN_LOOPS: Duration = Duration::from_secs(30);
-const SECONDS_FROM_LAST_LOOP_BEFORE_UNHEALTHY: Duration = Duration::from_secs(120);
+const LOOP_INTERVAL: Duration = Duration::from_secs(30);
+const DELAY_FROM_LAST_LOOP_BEFORE_UNHEALTHY: Duration = LOOP_INTERVAL.saturating_mul(4);
 
 pub async fn main(args: arguments::Arguments) {
     let pg_pool = PgPool::connect_lazy(args.db_url.as_str()).expect("failed to create database");
@@ -51,7 +51,7 @@ pub async fn main(args: arguments::Arguments) {
                 tracing::warn!("Error while refunding ethflow orders: {:?}", err)
             }
         }
-        tokio::time::sleep(SLEEP_TIME_BETWEEN_LOOPS).await;
+        tokio::time::sleep(LOOP_INTERVAL).await;
     }
 }
 
@@ -67,7 +67,7 @@ impl LivenessChecking for Liveness {
             .ok()
             .map(|last_successful_loop| {
                 Instant::now().duration_since(*last_successful_loop)
-                    < SECONDS_FROM_LAST_LOOP_BEFORE_UNHEALTHY
+                    < DELAY_FROM_LAST_LOOP_BEFORE_UNHEALTHY
             })
             .unwrap_or(false)
     }
--- a/crates/refunder/src/lib.rs
+++ b/crates/refunder/src/lib.rs
@@ -42,9 +42,7 @@ pub async fn main(args: arguments::Arguments) {
         match refunder.try_to_refund_all_eligble_orders().await {
             Ok(_) => {
                 track_refunding_loop_result("success");
-                *liveness.last_successful_loop.write().expect(
-                    "Lock is only written to at this point and the program should just crash on a panic",
-                ) = Instant::now()
+                *liveness.last_successful_loop.write().unwrap() = Instant::now()
             }
             Err(err) => {
                 track_refunding_loop_result("error");
@@ -62,14 +60,8 @@ struct Liveness {
 #[async_trait::async_trait]
 impl LivenessChecking for Liveness {
     async fn is_alive(&self) -> bool {
-        self.last_successful_loop
-            .read()
-            .ok()
-            .map(|last_successful_loop| {
-                Instant::now().duration_since(*last_successful_loop)
-                    < DELAY_FROM_LAST_LOOP_BEFORE_UNHEALTHY
-            })
-            .unwrap_or(false)
+        Instant::now().duration_since(*self.last_successful_loop.read().unwrap())
+            < DELAY_FROM_LAST_LOOP_BEFORE_UNHEALTHY
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/refunder/src/main.rs
```diff
@@ -9,5 +9,6 @@ async fn main() {
     );
     shared::exit_process_on_panic::set_panic_hook();
     tracing::info!("running refunder with validated arguments:\n{}", args);
+    global_metrics::setup_metrics_registry(Some("refunder".into()), None);
```
fedgiac, line 12: Note that the standard way to name metrics would have been `gp_v2_refunder`. I decided I could drop the prefix `gp_v2` here.

nlordell: This was important when we were under Gnosis infra to distinguish between the various projects. Agree that this can be dropped now.

I actually think we can drop the `gp_v2_*` prefix everywhere.

crates/refunder/src/lib.rs
```diff
@@ -6,14 +6,25 @@ pub mod submitter;
...
 use sqlx::PgPool;
-use std::time::Duration;
+use std::{
+    sync::{Arc, RwLock},
+    time::{Duration, Instant},
+};
 
-const SLEEP_TIME_BETWEEN_LOOPS: u64 = 30;
+const SLEEP_TIME_BETWEEN_LOOPS: u64 = 1;
+const SECONDS_FROM_LAST_LOOP_BEFORE_UNHEALTHY: u64 = 4 * SLEEP_TIME_BETWEEN_LOOPS;
```
vkgnosis, line 17: Not caused by this PR but these should be Duration.

nlordell, line 17: Ubernit: I would make these `const`s into `Duration`s:

```suggestion
const LOOP_INTERVAL: Duration = Duration::from_secs(1);
const DELAY_FROM_LAST_LOOP_BEFORE_UNHEALTHY: Duration = LOOP_INTERVAL.saturating_mul(4);
```

vkgnosis (approved):

crates/refunder/src/lib.rs
```diff
@@ -29,9 +40,53 @@ pub async fn main(args: arguments::Arguments) {
     loop {
         tracing::info!("Staring a new refunding loop");
         match refunder.try_to_refund_all_eligble_orders().await {
-            Ok(_) => (),
-            Err(err) => tracing::error!("Error while refunding ethflow orders: {:?}", err),
+            Ok(_) => {
+                track_refunding_loop_result("success");
+                *liveness.last_successful_loop.write().expect(
+                    "Lock is only written to at this point and the program should just crash on a panic",
```
nlordell, line 46: I would expect this to never panic because the lock should never be poisoned (i.e. there is no opportunity of a panic unwinding while a lock is held).

crates/refunder/src/lib.rs
```diff
@@ -29,9 +40,53 @@ pub async fn main(args: arguments::Arguments) {
...
+struct Liveness {
+    last_successful_loop: RwLock<Instant>,
+}
+
+#[async_trait::async_trait]
+impl LivenessChecking for Liveness {
+    async fn is_alive(&self) -> bool {
+        self.last_successful_loop
+            .read()
+            .ok()
```
nlordell, line 67: Would `unwrap` here, since there should be chance of poisoning the lock.

nlordell (approved):
LGTM. Some small nits:
- Should `unwrap` lock operations because the lock should never be poisoned (I would consider it a panic-worthy bug if the lock were to get poisoned)
- Would use `Duration` type for the constants (I feel like using `u64` is just some type erasure over what the constant actually represents - a duration; if you need the seconds as an integer, you can always, for example, `LOOP_INTERVAL.as_secs()`).

crates/refunder/src/lib.rs
vkgnosis: I would do a plain unwrap without explanation. This type of unwrap is pretty ocmmon.

crates/refunder/src/lib.rs
fedgiac: Thanks for pointing out `saturating_mul`. :slightly_smiling_face:

crates/refunder/src/lib.rs
```diff
@@ -6,14 +6,25 @@ pub mod submitter;
...
-const SLEEP_TIME_BETWEEN_LOOPS: u64 = 30;
+const SLEEP_TIME_BETWEEN_LOOPS: u64 = 1;
+const SECONDS_FROM_LAST_LOOP_BEFORE_UNHEALTHY: u64 = 4 * SLEEP_TIME_BETWEEN_LOOPS;
 
 pub async fn main(args: arguments::Arguments) {
     let pg_pool = PgPool::connect_lazy(args.db_url.as_str()).expect("failed to create database");
+
+    let liveness = Arc::new(Liveness {
+        // Program will be healthy at the start even if no loop was ran yet.
+        last_successful_loop: RwLock::new(Instant::now()),
```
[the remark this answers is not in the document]
fedgiac: I was unsure but overall it somewhat makes sense to be healthy at the start at this point, so I'd leave things like this.

crates/refunder/src/lib.rs
fedgiac: To me this wasn't obvious since `shared::metrics::serve_metrics` might in theory write into the `RwLock` and panic there. I considered passing to `Liveness` a "wrapped" `RwLock` that can only be read to be certain that it can't be poisoned, but I didn't find a way to do it nicely and I thought a manual implementation was overkill.

Anyway, I removed this line and replaced it with an `unwrap()`.

josojo (approved):
First metric for the refunder. Great. thanks

nlordell (approved):

Issue #334, opened 2022-07-01
Setup easy way to store large amount of data outside of logs


vkgnosis:
For example in https://github.com/cowprotocol/services/pull/325 there is some data we would like to access later and logging is the easiest way to implement this. But logging is not the best tool for this because we aren't interested in reading this text live and it puts more pressure on Kibana etc.

It would be better if we had some sort of network storage that we could upload files to. This would be one way. We don't need to interact with anything afterwards (no edit, delete, read). This should be easy to setup with an FTP server, or simple service we host that accepts files to put on disk through some POST route.

Traditionally we have used S3 for this. Marco felt it didn't work well for this job. I feel cloud services like this might be overkill because we don't need all the extra guarantees they give us. I would be fine with a self hosted ftp server somewhere. On the other hand since our other infrastructure is "cloudy" it can make sense to keep using it.

nlordell:
We can just host some FTP pod with a mounted volume and periodic cleaning that we expose with `kubectl port-forward`.

I think we should be able to do this without involving DevOps as well.

josojo:
I would prefer if this data is stored in a real database, such that we can efficiently run queries on it. From a data-analytics point of view, it would be a hassle to download single files and parse them, just to receive the prices that we need for price calculation. It would be more convenient to just download the "right" SQL rows fetched via a custom query.

cc: @bh2smith

vkgnosis:
That doesn't make sense to me with regards to the example I mentioned at the start of the post.

vkgnosis:
Using S3 is a good enough solution.

PR #1027, opened 2023-01-04
Enable s3 instance uploading

vkgnosis:
We want Quasimodo instances to be stored on AWS S3 for later debugging. See https://github.com/cowprotocol/services/issues/334 .

This currently happens through the obsolete MIP solver which in our cluster is configured to do nothing except upload the instance. However, this instance uploading is hard to maintain and brittle. Recently it started failing and filling up the disk space of the container. It is also unnecessary overhead to run the whole MIP image just to upload instances.

There were two main ideas for re-implementing the uploading in Rust:
1. Continue using a full solver pod to upload instances.
2. Integrate the uploading into the driver.

I went with 2. because the extra pod has too much organizational overhead. I do not want to maintain a whole binary, pod, staging configuration to upload instances. An even better solution might be to do the instance uploading in Quasimodo because it isn't really relevant to the driver. But that would be more work because it would be harder to use Rust there.

To interface with S3 I decided on the official AWS Rust libraries. They are technically in preview stage but this seems to be referring to breaking changes and not bugs. I intentionally use as little of the API. Breaking changes won't be difficult to handle.

I tried out manually  interfacing with S3's http rest api in order to reduce dependencies. This did not work well because AWS recently added signing requirements to requests which are too difficult to implement standalone. The AWS libraries we depend on now do not add many transitive dependencies anyway because a lot of the tokio and http libraries we already depend on.

Follow up tasks if this works on staging:
- Remove references to MIP solver in the Rust code.
- Remove references to MIP solver in the cluster.
- Clean up the old S3 bucket?

### Test Plan

manually ran the ignored s3 upload tests, can only test for real after merging and configuring staging

### DIFF
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -1,5 +1,6 @@
 use crate::{
     liquidity::slippage,
+    s3_instance_upload_arguments::S3UploadArguments,
     settlement_access_list::AccessListEstimatorType,
     solver::{single_order_solver, ExternalSolverArg, SolverAccountArg, SolverType},
 };
@@ -300,25 +301,8 @@ pub struct Arguments {
     #[clap(long, env, default_value = "0")]
     pub solution_comparison_decimal_cutoff: u16,
 
-    /// The s3_instance_upload_* arguments configure how quasimodo instances should be uploaded to
-    /// AWS S3. They must either all be set or all not set. If they are set then every instance sent to
-    /// Quasimodo as part of auction solving is also uploaded to S3.
-    #[clap(long, env)]
-    pub s3_instance_upload_region: Option<String>,
-
-    #[clap(long, env)]
-    pub s3_instance_upload_bucket: Option<String>,
-
-    /// Prepended to the auction id to form the final instance filename on S3. Something like
-    /// "staging/mainnet/quasimodo/". Should end with `/` if intended to be a folder.
-    #[clap(long, env)]
-    pub s3_instance_upload_filename_prefix: Option<String>,
-
-    #[clap(long, env)]
-    pub s3_instance_upload_access_key_id: Option<String>,
-
-    #[clap(long, env)]
-    pub s3_instance_upload_secret_access_key: Option<String>,
+    #[clap(flatten)]
+    pub s3_upload: S3UploadArguments,
 }
 
 impl std::fmt::Display for Arguments {
@@ -432,31 +416,6 @@ impl std::fmt::Display for Arguments {
             f,
             "token_list_restriction_for_price_checks: {:?}",
             self.token_list_restriction_for_price_checks
-        )?;
-        display_option(
-            f,
-            "s3_instance_upload_region",
-            &self.s3_instance_upload_region,
-        )?;
-        display_option(
-            f,
-            "s3_instance_upload_bucket",
-            &self.s3_instance_upload_bucket,
-        )?;
-        display_option(
-            f,
-            "s3_instance_upload_filename_prefix",
-            &self.s3_instance_upload_filename_prefix,
-        )?;
-        display_option(
-            f,
-            "s3_instance_upload_access_key_id",
-            &self.s3_instance_upload_access_key_id,
-        )?;
-        display_secret_option(
-            f,
-            "s3_instance_upload_secret_access_key",
-            &self.s3_instance_upload_secret_access_key,
         )?;
         Ok(())
     }
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -299,6 +299,26 @@ pub struct Arguments {
     /// value 0.0012 ETH and 0.0016 ETH equivalent.
     #[clap(long, env, default_value = "0")]
     pub solution_comparison_decimal_cutoff: u16,
+
+    /// The s3_instance_upload_* arguments configure how quasimodo instances should be uploaded to
+    /// AWS S3. They must either all be set or all not set. If they are set then every instance sent to
+    /// Quasimodo as part of auction solving is also uploaded to S3.
+    #[clap(long, env)]
+    pub s3_instance_upload_region: Option<String>,
+
+    #[clap(long, env)]
+    pub s3_instance_upload_bucket: Option<String>,
+
+    /// Prepended to the auction id to form the final instance filename on S3. Something like
+    /// "staging/mainnet/quasimodo/". Should end with `/` if intended to be a folder.
+    #[clap(long, env)]
+    pub s3_instance_upload_filename_prefix: Option<String>,
+
+    #[clap(long, env)]
+    pub s3_instance_upload_access_key_id: Option<String>,
+
+    #[clap(long, env)]
+    pub s3_instance_upload_secret_access_key: Option<String>,
 }
 
 impl std::fmt::Display for Arguments {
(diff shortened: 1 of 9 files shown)


crates/solver/src/arguments.rs
```diff
@@ -299,6 +299,23 @@ pub struct Arguments {
...
+    /// Prepended to the auction id to form the final instance filename on S3. Something like
+    /// "staging/mainnet/quasimodo/". Should end with `/` if intended to be a folder.
+    #[clap(long, env)]
+    pub s3_instance_upload_filename_prefix: Option<String>,
+
+    #[clap(long, env)]
+    pub s3_instance_upload_access_key_id: Option<String>,
+
+    #[clap(long, env)]
+    pub s3_instance_upload_secret_access_key: Option<String>,
```
MartinquaXD, line 321: Recently we started to group related arguments into a separate argument struct which overall seemed pretty nice. This could be done here as well.

crates/solver/src/arguments.rs
```diff
@@ -299,6 +299,26 @@ pub struct Arguments {
     /// value 0.0012 ETH and 0.0016 ETH equivalent.
     #[clap(long, env, default_value = "0")]
     pub solution_comparison_decimal_cutoff: u16,
+
+    /// The s3_instance_upload_* arguments configure how quasimodo instances should be uploaded to
+    /// AWS S3. They must either all be set or all not set. If they are set then every instance sent to
+    /// Quasimodo as part of auction solving is also uploaded to S3.
```
MartinquaXD, line 305: Not sure why we tie the s3 upload to quasimodo.
This could also be used to debug the naive or baseline solver. And overall I think keeping that data is valuable anyway. So unless the s3 storage is very expensive I would err on the side of caution and store all of the auctions.

crates/solver/src/solver.rs
```diff
@@ -306,39 +307,44 @@ pub fn create(
...
+                } else {
+                    http_instance_with_all_orders.clone()
+                },
+                filter_non_fee_connected_orders,
+                slippage_calculator,
+                market_makable_token_list.clone(),
+                *domain,
+                instance_uploader_config
+                    .map(crate::s3_instance_upload::S3InstanceUploader::new)
+                    .map(Arc::new),
```
MartinquaXD, line 345: I guess this could be moved out of `create_http_solver` such that we can only ever have a single instance of `S3InstanceUploader`. I know we are currently only creating a single instance but in a slightly weird way.

crates/solver/src/solver/http_solver.rs
```diff
@@ -516,6 +520,25 @@ impl Solver for HttpSolver {
             }
         };
 
+        // Upload the instance if configured. Happens in a task to not delay solving.
+        if let Some(uploader) = &self.instance_uploader {
```
MartinquaXD, line 524: `HttpSolver::solve()` is already very big. This functionality could be moved into a separate method.

crates/solver/src/arguments.rs
vkgnosis: I don't strongly feel either way but I slightly prefer this. The arguments are already grouped by having a similar name and being next to eachother. I don't see a benefit from a struct except when that allows us to reuse the same arguments struct.

crates/solver/src/arguments.rs
vkgnosis: The instance is meaningful for http solvers. And between those they all share the same instance. It's not useful to upload it multiple times. The reason I'm tying it to Quasimodo is that that is the main reason we keep the instances in the first place as we do now.

vkgnosis: It would make a difference if we disabled Quasimodo but uploading a generic instance without tying it to a solver would be a bigger refactor.

crates/solver/src/solver.rs
vkgnosis: The way it is implemented now, if we wanted multiple uploaders then they would need to be configured separately. But I can still pass in the uploader here directly instead of the config.

crates/solver/src/arguments.rs
MartinquaXD: Having a struct for this would also give us a nice place to put the verification logic that is currently free floating in `main()`.

vkgnosis: I like that argument. Will do.

crates/solver/src/s3_instance_upload.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.upload(key, value).await
+    }
+
+    fn filename(&self, now: DateTime<Utc>, auction: AuctionId) -> String {
+        let date = now.format("%Y-%m-%d");
+        format!("{}{date}/{auction}.json", self.filename_prefix)
+    }
+
+    async fn upload(&self, key: String, value: Vec<u8>) -> Result<()> {
+        self.client
```
MartinquaXD, line 56: Depending on how expensive storage in s3 is we could compress the file here (ideally dependent on a CLI flag).
Judging by a single test we could save:
* ~90% disk space with `gzip`
* ~91.6% with `brotli`
The downside would be that you would have to decompress the files to know their content not sure how much of an inconvenience would be.

MartinquaXD (approved):
Approving since having a generic upload (not tied to a specific http solver) would be a bigger refactor (that doesn't seem to be very urgent).

crates/solver/src/s3_instance_upload.rs
MartinquaXD: According to Anuj we should pay at most ~40$ per month at the moment. However, since we launched limit orders the S3 disk space obviously grows more rapidly than before:
Screenshot 2023-01-05 at 1 29 37 PM

vkgnosis: Compression is also on my radar. It makes using the instances slightly less convenient because you have to decompress them locally first but I also think saving 90% is worth it.

vkgnosis:
I'm going to merge this now because I want to start testing and running this on staging. Feel free to still comment afterwards which I can address in future PRs.

harisang:
Compression is something SeaSolver has already asked for it (they are already sending back their responses as gzip).  If you think it makes sense, we can definitely go ahead with it. It's not a big deal for solvers to implement it anyways.

MartinquaXD:
> Compression is something SeaSolver has already asked for it (they are already sending back their responses as gzip). If you think it makes sense, we can definitely go ahead with it. It's not a big deal for solvers to implement it anyways.

Compressing the `instance.json` in S3 can be done without compressing it for solvers so those would be 2 independent changes. The biggest (only?) problem would be for the solver team because whenever you want to inspect an `instance.json` from S3 you would have to download and manually decompress it first because we'd store it in a human unreadable format.

harisang:
Right.

> The biggest (only?) problem would be for the solver team because whenever you want to inspect an instance.json from S3 you would have to download and manually decompress it first because we'd store it in a human unreadable format.
I don't think that's a concern.

crates/solver/src/solver.rs
```diff
@@ -282,6 +283,7 @@ pub fn create(
     order_prioritization_config: &single_order_solver::Arguments,
     post_processing_pipeline: Arc<dyn PostProcessing>,
     domain: &DomainSeparator,
+    mut s3_instance_uploader_for_quasimodo: Option<S3InstanceUploader>,
```
nlordell, line 286: Quasimodo is disabled on Mainnet AFAIK. I think this uploader may not be used at all in production.

harisang: Wait, doesn't this pretend it is used for Quasimodo so it is just creating subinstances appropriate for Quasimodo? I.e., does it really require that the instance is sent to Quasimodo itself? I thought it would in any case upload, which tbh, it still very useful as we use Quasimodo for debugging and checking certain solutions, if though, as you said, Quasimodo is NOT used as a solver on mainnet anymore.

PR #1028, opened 2023-01-05
Track number of limit orders with valid `surplus_fee`

MartinquaXD:
While looking into ways to optimize `surplus_fee` quoting I found 1 metric somewhat confusing.

Limit order quoting was implemented in a way that updates the `surplus_fee` before the `autopilot` would consider it too old to be used in an auction.

However, the current metrics look a bit misleading:

Screenshot 2023-01-05 at 11 00 59

The upper graph shows spikes of unquoted orders whereas the other graphs combined suggest a constant number of orders with usable `surplus_fee`. After some investigation I realized the lower graph shows the number of orders the `LimitOrderQuoter` intends to quote; not the number of orders where the `surplus_fee` is too old to be considered.

To monitor the overall health/efficiency of the system I think it's more insightful to track the unusable limit orders instead.

In case you consider the original metric also helpful I can obviously update the PR to track both of them.

### DIFF
--- a/crates/autopilot/src/lib.rs
+++ b/crates/autopilot/src/lib.rs
@@ -62,9 +62,9 @@ use shared::{
 use std::{collections::HashSet, sync::Arc, time::Duration};
 use tracing::Instrument;
 
-/// To no never get to the state where a limit order can not be considered usable because the
+/// To never get to the state where a limit order can not be considered usable because the
 /// `surplus_fee` is too old the `surplus_fee` is valid for longer than its update interval.
-/// This factor controls how much longer its considered valid.
+/// This factor controls how much longer it's considered valid.
 /// If the `surplus_fee` gets updated every 5 minutes and the factor is 2 we consider limit orders
 /// valid where the `surplus_fee` was computed up to 10 minutes ago.
 const SURPLUS_FEE_EXPIRATION_FACTOR: u8 = 2;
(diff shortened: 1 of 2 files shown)


vkgnosis:
That's a good point. There is really "limit orders with quotes too old for auction" and "limit orders that are going to be quoted by the quoter" where the first is >= the second. I'd like to have both metrics because the second (the one we had before this PR) is relevant to understand the progress of the quoter.

MartinquaXD:
I'm tracking both metrics now and ran a manual test to see that the metrics show up:
```
# HELP gp_v2_autopilot_limit_orders Open limit orders.
# TYPE gp_v2_autopilot_limit_orders gauge
gp_v2_autopilot_limit_orders 0
# HELP gp_v2_autopilot_quoted_limit_orders Quoted limit orders.
# TYPE gp_v2_autopilot_quoted_limit_orders gauge
gp_v2_autopilot_quoted_limit_orders 0
# HELP gp_v2_autopilot_unquoted_limit_orders Limit orders awaiting a quote.
# TYPE gp_v2_autopilot_unquoted_limit_orders gauge
gp_v2_autopilot_unquoted_limit_orders 0
# HELP gp_v2_autopilot_unusable_limit_orders Limit orders with surplus_fee too old for the auction.
# TYPE gp_v2_autopilot_unusable_limit_orders gauge
gp_v2_autopilot_unusable_limit_orders 0
# HELP gp_v2_autopilot_usable_limit_orders Limit orders usable in the auction.
# TYPE gp_v2_autopilot_usable_limit_orders gauge
gp_v2_autopilot_usable_limit_orders 0
```

vkgnosis (approved):

nlordell (approved):

PR #1031, opened 2023-01-05
Print s3 arguments

vkgnosis:
After refactoring into a second argument struct I forgot to add the print here.

### Test Plan

not needed

### DIFF
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -417,6 +417,7 @@ impl std::fmt::Display for Arguments {
             "token_list_restriction_for_price_checks: {:?}",
             self.token_list_restriction_for_price_checks
         )?;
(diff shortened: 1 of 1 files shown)


crates/solver/src/arguments.rs
```diff
@@ -417,6 +417,7 @@ impl std::fmt::Display for Arguments {
             "token_list_restriction_for_price_checks: {:?}",
             self.token_list_restriction_for_price_checks
         )?;
+        writeln!(f, "{}", self.s3_upload)?;
```
MartinquaXD, line 420: Small nit: All the other argument structs that get flattened into this struct are at the top of this function. Makes some sense to keep them grouped together.

MartinquaXD (approved):

PR #1034, opened 2023-01-06
Refactor http solver instance creation

vkgnosis:
This PR is split into two commits that make reviewing easier:
1. *Move solver model creation*: Moves code from `http_solver.rs` to `instance_creation.rs` without logic changes.
2. *Refactor http solver instance caching*:  Rewrites the shared instance creation code. Contains logic changes.

---

After merging #1027 it turned out that the approach of uploading only Quasimodo instances is not going to work because
- Quasimodo doesn't run in all environments.
- Quasimodo instances contain less information than other http solver instances because non fee connected tokens are filtered out.

Instead, we want to always upload the "proper" instance. For every auction one such instance is created and it is shared between all solvers. This was handled by an ad hoc `InstanceCache` in `http_solver.rs`. This implementation was hard to work with. In this PR it becomes a less tightly coupled component in its own file `instance_cache.rs`.

In order to do this I first needed to refactor the instance creation code out of `http_solver.rs` because it should not be tied to single solver. It now lives in `instance_creation.rs`.

### Test Plan

Did not delete any tests. Added a new one. Will carefully observe after merging on staging.


nlordell (approved):
Very nice change!

crates/solver/src/solver/http_solver/instance_cache.rs
```diff
@@ -0,0 +1,187 @@
...
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+/// To `Driver` every http solver is presented as an individual `Solver` implementor. Internally
+/// http solvers share the same data that is needed to create the instance for the same auction. In
+/// order to waste less resources we create the instance once per auction in this component.
+pub struct SharedInstanceCreator {
+    creator: InstanceCreator,
+    uploader: Option<Arc<S3InstanceUploader>>,
+    last: Mutex<Option<Cache>>,
```
MartinquaXD, line 14: I don't expect it to matter performance wise but I think a `RwLock` might be a bit more appropriate here. There will only every be a single writer but possibly multiple parallel readers.

crates/solver/src/solver/http_solver/instance_cache.rs
```diff
@@ -0,0 +1,187 @@
...
+        Self {
+            creator,
+            uploader,
+            last: Default::default(),
+        }
+    }
+
+    /// The first call for a new run id creates the instance and stores it in a cache. Subsequent
+    /// calls copy the instance from the cache (and block until the first call completes).
+    pub async fn get_instance(&self, auction: Auction, instance: InstanceType) -> Instance {
```
MartinquaXD, line 51: I believe a solver only ever uses a reference to an `Instance`. In that case we could avoid copying those huge structs by storing and returning `Arc<Instance>` here.

MartinquaXD (approved):
Approving since we were fine so far without my suggestions.

crates/solver/src/solver/http_solver/instance_cache.rs
vkgnosis: There is going to be ~10 solvers requesting the instance. The first one locks the mutex and performs the instance creation which takes ~5 seconds. Then all of them read a copy of the instance. Copying the instance is not an expensive operation (especially with the Arc change below) so I'd expect the difference between Mutex and RwLock to be unnoticeable.

nlordell: I think a `RwLock` isn't a super obvious gain to me here, because all 10 solvers can potentially are contending for upgrading their read lock to a write lock for the initial update.

If there was a single writer, then I agree a `RwLock` would make more sense.

crates/solver/src/solver/http_solver/instance_cache.rs
vkgnosis: I like this. Will implement in separate PR.

nlordell: I like the idea. Its a pretty big data structure which takes a lot of memory, `Arc` seems worth it in my eyes.

crates/solver/src/solver/http_solver/instance_cache.rs
MartinquaXD: Oh, yeah `RwLock` doesn't make much sense indeed. A solver can't know whether a read locks is sufficient so they would all have to request a write lock anyway. But as Valentin mentioned with the `Arc` change contention should be minimal anyway.

PR #1036, opened 2023-01-06
Create http solver filtered and unfiltered instance together

vkgnosis:
Currently (since https://github.com/cowprotocol/services/pull/461) we do some duplicate work in `prepare_model` because the function either returns a filtered or an unfiltered version. Both versions use the same data from the node so it is wasteful to fetch it twice. With this PR we fetch the data only once.

### Test Plan

adjusted test still works

### Release notes

Optionally add notes, suggestions and warnings for the releaser. They will be included in the release changelog.

### DIFF
--- a/crates/solver/src/solver/http_solver/instance_creation.rs
+++ b/crates/solver/src/solver/http_solver/instance_creation.rs
@@ -25,6 +25,17 @@ use std::{
     sync::Arc,
 };
 
+pub struct Instances {
+    pub plain: Instance,
+    pub filtered: Instance,
+}
+
+#[derive(Clone)]
+pub struct Instance {
+    pub model: BatchAuctionModel,
+    pub context: SettlementContext,
+}
+
 pub struct InstanceCreator {
     pub native_token: H160,
     pub token_info_fetcher: Arc<dyn TokenInfoFetching>,
@@ -35,7 +46,7 @@ pub struct InstanceCreator {
 
 impl InstanceCreator {
     #[allow(clippy::too_many_arguments)]
-    pub async fn prepare_model(
+    pub async fn prepare_instances(
         &self,
         auction_id: AuctionId,
         run_id: u64,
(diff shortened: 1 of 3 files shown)


crates/solver/src/solver/http_solver/instance_creation.rs
```diff
@@ -101,33 +111,30 @@ impl InstanceCreator {
...
-            tokens.iter().copied().collect()
-        };
         let gas_model = GasModel {
             native_token: self.native_token,
             gas_price,
         };
 
+        // Some solvers require that there are no isolated islands of tokens unconnected to the
+        // native token.
+        let fee_connected_tokens_: HashSet<H160> =
```
nlordell, line 121: naming nit: 
```suggestion
        let fee_connected_tokens: HashSet<H160> =
```

nlordell (approved):
LGTM, nice savings too!

crates/solver/src/solver/http_solver/instance_creation.rs
```diff
@@ -101,33 +111,30 @@ impl InstanceCreator {
...
-            // know about.
-            tokens.iter().copied().collect()
-        };
         let gas_model = GasModel {
             native_token: self.native_token,
             gas_price,
         };
 
+        // Some solvers require that there are no isolated islands of tokens unconnected to the
+        // native token.
```
MartinquaXD, line 120: This comment currently suggests to me that the final filtered `Instance` should rather have its `tokens` updated than the `orders`. I had to `git blame` Nick's original change to make sure that we only care about filtering orders and not tokens or amms.

nlordell: I think only matters for orders.

vkgnosis: Changed comment to talk about orders.

PR #1043, opened 2023-01-09
Driver `/solve` tests

sistemd:
Fixes https://github.com/cowprotocol/services/issues/895.

Successfully calling `/solve` on geth with access list simulation. (I.e. I managed to figure out all the issues I was complaining about on the weekly sync.) :tada: 

What's done:

- A test for the `/solve` endpoint.
- A test that our local `example.solvers.toml` file is correct.
- Local testing environment based on dev-mode `geth`, I used `geth` because it implements `eth_createAccessList` so I don't have to mock access list simulation.
- A bunch of bugs fixed.
- Minor domain logic refactoring.
- A way to mock the current time in tests.
- A way to override contract addresses via CLI args.
- A way to disable access list simulation via CLI args.

What's missing:

- A way to reset the local geth env. Simply restarting the geth instance can facilitate this, but there needs to be a way to do this automatically at the start of each test.
- More test cases.
- Testing allowances?

I'm out of the shit for now, `/quote` and `/execute` should be fairly easy to implement. Then when I get to implementing liquidity collection I'll be in the shit again since it's a total unknown to me, it requires a complex boundary integration, and I don't know how long it will take or how to test it well.

Shoutout @MartinquaXD for serving as a sanity check at times when I was stuck and @nlordell for giving me the pointer to follow the `uniswapTrade` test from the contracts repo.

### Test Plan

Yyyyeeeeeeaaahhhhhh.

### DIFF
--- a/crates/driver/src/infra/api/mod.rs
+++ b/crates/driver/src/infra/api/mod.rs
@@ -56,7 +56,7 @@ impl Api {
                 solver,
                 shared: Arc::clone(&shared),
             });
-            app = app.nest(&format!("/{}", name), router);
+            app = app.nest(&format!("/{name}"), router);
         }
 
         // Start the server.
--- a/crates/driver/src/tests/setup/solver.rs
+++ b/crates/driver/src/tests/setup/solver.rs
@@ -1,6 +1,6 @@
-use {
-    std::{net::SocketAddr, sync::Arc},
-    tokio::sync::Mutex,
+use std::{
+    net::SocketAddr,
+    sync::{Arc, Mutex},
 };
 
 /// Configuration for mocking a solver.
@@ -37,7 +37,7 @@ pub async fn setup(config: Config) -> Solver {
             axum::routing::post(
                 |axum::extract::State(state): axum::extract::State<State>,
                  axum::extract::Json(req): axum::extract::Json<serde_json::Value>| async move {
-                    let mut state = state.0.lock().await;
+                    let mut state = state.0.lock().unwrap();
                     assert!(
                         !state.is_empty(),
                         "got another solve request, but didn't expect any more"


crates/driver/example.solvers.toml
```diff
@@ -1,12 +1,12 @@
 [[solvers]]
 name = "mysolver" # Arbitrary name given to this solver, must be unique
-url = "http://localhost:1234/solve" # Endpoint of the solver
-absolute-slippage = 12 # Denominated in wei, optional
-relative-slippage = 0.1 # Percentage in the [0, 1] range
+endpoint = "http://0.0.0.0:33219/solve"
+absolute-slippage = "12" # Denominated in wei, optional
+relative-slippage = "0.1" # Percentage in the [0, 1] range
```
nlordell, line 5: Not for this PR, but you could make this:
```suggestion
slippage = { relative = "12", absolute = "0.1" }
```

crates/driver/src/boundary/settlement.rs
```diff
@@ -224,7 +226,21 @@ fn to_boundary_order(order: &competition::Order) -> Order {
             onchain_order_data: Default::default(),
             is_liquidity_order: order.is_liquidity(),
         },
-        signature: Default::default(),
+        // TODO Different signing schemes imply different sizes of signature data, which indicates
```
nlordell, line 229: Correct, ECDSA signatures (EIP-712 and EthSign) both are exactly 65 bytes long, ERC-1271 signatures (sometimes called EIP-1271 - this is because it moved from EIP to ERC recently) have variable length bytes (from 0 to ∞).

crates/driver/src/boundary/settlement.rs
```diff
@@ -440,3 +450,9 @@ impl AllowanceManaging for AllowanceManager {
             .collect())
     }
 }
+
+fn to_array<const N: usize>(data: &[u8]) -> [u8; N] {
```
nlordell, line 454: This is in the `std` with [`&[u8]; TryInto<[u8; N]`](https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3C%26%5BT%5D%3E-for-%5BT%3B%20N%5D).

crates/driver/src/domain/competition/auction.rs
```diff
@@ -48,38 +71,23 @@ impl From<chrono::DateTime<chrono::Utc>> for Deadline {
...
-    fn time_buffer() -> chrono::Duration {
-        chrono::Duration::seconds(1)
+impl SolverDeadline {
+    pub fn timeout(&self, now: time::Now) -> Result<std::time::Duration, DeadlineExceeded> {
+        let timeout = self.0 - now.now();
+        if timeout <= chrono::Duration::zero() {
+            Err(DeadlineExceeded)
+        } else {
+            Ok(timeout.to_std().expect("already checked non-negative"))
+        }
```
nlordell, line 90: You can make this more succinct with:
```suggestion
        timeout.to_std().map_err(|_| DeadlineExceeded)
```

crates/driver/src/domain/eth/mod.rs
```diff
@@ -96,7 +96,7 @@ impl From<EffectiveGasPrice> for U256 {
 ///
 /// https://eips.ethereum.org/EIPS/eip-2930
 #[derive(Debug, Clone, Default)]
-pub struct AccessList(HashMap<Address, HashSet<StorageKey>>);
+pub struct AccessList(HashMap<ContractAddress, HashSet<StorageKey>>);
```
nlordell, line 99: This is incorrect. Access lists can also warm non-contract addresses (for example if you transfer ETH to an EOA, you can include its address (without storage slots) to warm up the address and pay less for the transfer).

crates/driver/src/domain/eth/mod.rs
```diff
@@ -197,17 +204,17 @@ impl From<ContractAddress> for Address {
 ///
 /// https://eips.ethereum.org/EIPS/eip-20
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct TokenAddress(pub H160);
+pub struct TokenAddress(pub ContractAddress);
```
nlordell, line 207: This is interesting - I would have kept `TokenAddress(pub H160)` and have a conversion function `TokenAddress -> ContractAddress`.

Not suggesting to change this, I'm more just curious about the rationale of wrapping multiple "marker" types.

crates/driver/src/infra/api/mod.rs
```diff
@@ -45,14 +56,24 @@ impl Api {
                 solver,
                 shared: Arc::clone(&shared),
             });
-            app = app.nest(&name.0, router);
+            app = app.nest(&format!("/{}", name.0), router);
```
nlordell, line 59: Should we make `name: Display` and change this to:
```suggestion
            app = app.nest(&format!("/{}", name), router);
```

crates/driver/src/infra/api/solve/dto/auction.rs
```diff
@@ -130,12 +132,16 @@ pub enum Error {
...
+    #[error("received an auction with an exceeded deadline")]
+    DeadlineExceeded,
 }
 
 #[serde_as]
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Auction {
+    // TODO This should not be optional. Also, this seems to indicate that there should be two
+    // different auction types, or maybe set the ID to Default::default() for quoting?
```
nlordell, line 144: I think we purposefully wanted to use the same API for solving and quoting to reduce the additional work on the `solver engine`.

crates/driver/src/infra/blockchain/contracts.rs
```diff
@@ -1 +1,49 @@
+use {crate::domain::eth, ethcontract::Web3};
+
 pub use crate::boundary::contracts::{GPv2Settlement, ERC20, WETH9};
+
+#[derive(Debug, Clone)]
+pub struct Contracts {
+    pub(super) settlement: contracts::GPv2Settlement,
+    pub(super) weth: contracts::WETH9,
```
nlordell, line 8: Is `pub(super)` needed here? Can't we use the accessor functions?

crates/driver/src/infra/config/cli.rs
```diff
@@ -1,13 +1,15 @@
...
 };
 
 #[derive(Debug, clap::Parser)]
 pub struct Args {
-    /// The address to bind the driver to.
+    /// The address to bind the driver to. "auto" to bind to 0.0.0.0 and any
+    /// free port.
     #[clap(long, env)]
-    pub bind_addr: SocketAddr,
+    pub bind_addr: String,
```
nlordell, line 12: You can implement `api::Addr: Display + FromStr` and then:
```suggestion
    pub bind_addr: api::Addr,
```

crates/driver/src/tests/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod boundary;
+mod cases;
+mod setup;
+
+fn hex_address(value: ethcontract::H160) -> String {
+    format!("0x{}", hex::encode(value.as_bytes()))
```
nlordell, line 6: This also works: 
```suggestion
    format!("0x{value:?}")
```

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+
+pub mod uniswap;
+
+pub use uniswap::Uniswap;
+
+pub const WEB3_URL: &str = "http://localhost:8546";
+
+pub fn web3() -> Web3<DynTransport> {
+    Web3::new(DynTransport::new(
+        web3::transports::Http::new(WEB3_URL).expect("node should be up, run docker-compose up -d"),
```
nlordell, line 11: I think this errors on `WEB3_URL` being an invalid URL and unable to parse.

crates/driver/src/tests/setup/driver.rs
```diff
@@ -0,0 +1,117 @@
...
+                        address,
+                        name,
+                        ..
+                    },
+                addr,
+            } = solver;
+            format!(
+                "[[solvers]]\nname = \"{name}\"\nendpoint = \"http://{addr}\"\nabsolute-slippage \
+                 = \"{absolute_slippage}\"\nrelative-slippage = \"{relative_slippage}\"\naddress \
+                 = \"{address}\"\n"
```
nlordell, line 112: This may be easier to read with with something like:
```suggestion
                "[[solvers]]\n\
                 name = \"{name}\"\n\
                 endpoint = \"http://{addr}\"\n\
                 absolute-slippage = \"{absolute_slippage}\"\n\
                 relative-slippage = \"{relative_slippage}\"\n\
                 address = \"{address}\"\n"
```

nlordell (approved):
🎉 

PR was a bit long, but most of the changes were just the test setup and expected inputs and outputs which I just skimmed through.

crates/driver/example.solvers.toml
nlordell: Neither do I. Let's keep it as is then.

crates/driver/src/domain/eth/mod.rs
nlordell: Makes sense.

crates/driver/src/infra/config/cli.rs
nlordell: Ah - good point. NVM then.

crates/driver/src/tests/mod.rs
nlordell: Good point. Let's leave it as is.

crates/driver/src/infra/api/mod.rs
nlordell: Why not `format!("{name}")`.

crates/driver/src/infra/api/solve/dto/auction.rs
```diff
@@ -10,12 +11,9 @@ use {
...
 
 impl Auction {
-    pub fn into_domain(self) -> Result<competition::Auction, Error> {
+    pub fn into_domain(self, now: infra::time::Now) -> Result<competition::Auction, Error> {
         Ok(competition::Auction {
-            id: match self.id {
-                Some(id) => Some(FromStr::from_str(&id).map_err(|_| Error::InvalidAuctionId)?),
-                None => None,
-            },
+            id: Some(FromStr::from_str(&self.id).map_err(|_| Error::InvalidAuctionId)?),
```
nlordell, line 16: Shouldn't the domain `Auction` object also have a mandatory ID? (Unrelated to this PR - feel free to address this in a follow up).

crates/driver/src/infra/api/solve/dto/auction.rs
```diff
@@ -169,7 +170,9 @@ struct Order {
     #[serde_as(as = "serialize::U256")]
     executed: eth::U256,
     interactions: Vec<Interaction>,
+    #[serde(default)]
     sell_token_balance: SellTokenBalance,
+    #[serde(default)]
     buy_token_balance: BuyTokenBalance,
```
MartinquaXD, line 176: Should these rather be optional?
Otherwise we might encounter an error while fetching the balances and a solver trusting those fields would assume this order can not be settled.

crates/driver/src/infra/blockchain/contracts.rs
```diff
@@ -1 +1,49 @@
...
+pub struct Contracts {
+    settlement: contracts::GPv2Settlement,
+    weth: contracts::WETH9,
+}
+
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Addresses {
+    pub settlement: Option<eth::ContractAddress>,
+    pub weth: Option<eth::ContractAddress>,
+}
```
MartinquaXD, line 15: Would be cool if we only ever had to initialize those addresses once. In that case we could make `settlement` and `weth` required because the service will not be useful when those addresses are unknown.
Just some idea.

crates/driver/src/infra/config/cli.rs
```diff
@@ -1,13 +1,15 @@
...
+    std::{path::PathBuf, str::FromStr},
     url::Url,
 };
 
 #[derive(Debug, clap::Parser)]
 pub struct Args {
-    /// The address to bind the driver to.
+    /// The address to bind the driver to. "auto" to bind to 0.0.0.0 and any
+    /// free port.
     #[clap(long, env)]
```
MartinquaXD, line 11: Should this have a default port?

crates/driver/src/tests/setup/solver.rs
```diff
@@ -0,0 +1,62 @@
...
+                 axum::extract::Json(req): axum::extract::Json<serde_json::Value>| async move {
+                    let mut state = state.0.lock().await;
+                    assert!(
+                        !state.is_empty(),
+                        "got another solve request, but didn't expect any more"
+                    );
+                    let Solve {
+                        req: expected_req,
+                        res,
+                    } = state.pop().unwrap();
```
MartinquaXD, line 48:
```suggestion
                    let Solve {
                        req: expected_req,
                        res,
                    } = state.pop().expect("expects another solve request");
```

crates/driver/src/tests/setup/solver.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Set up an HTTP server exposing a solver API and acting as a solver mock.
+pub async fn setup(config: Config) -> Solver {
+    let state = Arc::new(Mutex::new(config.solve.clone()));
+    let app = axum::Router::new()
+        .route(
+            "/",
+            axum::routing::post(
+                |axum::extract::State(state): axum::extract::State<State>,
+                 axum::extract::Json(req): axum::extract::Json<serde_json::Value>| async move {
+                    let mut state = state.0.lock().await;
```
MartinquaXD, line 40: An async aware `Mutex` is only needed if you want to hold the lock across `.await`s.

crates/driver/src/infra/time.rs
```diff
@@ -0,0 +1,17 @@
...
+    Real,
+    /// Always return the same time. Used for testing.
+    Fake(chrono::DateTime<chrono::Utc>),
+}
+
+impl Now {
+    pub fn now(&self) -> chrono::DateTime<chrono::Utc> {
+        match self {
+            Self::Real => chrono::Utc::now(),
+            Self::Fake(time) => time.to_owned(),
```
MartinquaXD, line 14: We might need a way to adjust this value during the test eventually for testing time sensitive stuff.

MartinquaXD (commented):
This is a lot to take in.
Might need to do another round.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -8,7 +8,8 @@ use crate::infra::{blockchain, Ethereum};
 
 /// Address used in place of an actual buy token address in an order which buys
 /// ETH.
-const BUY_ETH_ADDRESS: eth::TokenAddress = eth::TokenAddress(eth::H160([0xee; 20]));
+const BUY_ETH_ADDRESS: eth::TokenAddress =
+    eth::TokenAddress(eth::ContractAddress(eth::H160([0xee; 20])));
```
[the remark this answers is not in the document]
nlordell: This address only represents ETH as the `buy_token` from the SC point of view. So it depends on what we want this address to mean from the domain point of view.

Maybe it makes sense to:
```rust
pub enum TokenAddress {
    Erc20(ContractAddress),
    Eth,
}
```

crates/driver/src/infra/time.rs
nlordell: I think in that case you could introduce something like:
```rust
pub fn adjust(&self, new_time: chrono::DateTime<chrono::Utc>) {
    match self {
        Self::Fake(time /* : Mutex<chrono::DateTime<chrono::Utc> */) => *time.lock().unwrap() = new_time,
        Self::Read => panic!("attempt to adjust the system clock"),
    }
}
```

I would leave introducing that to when it is needed.

crates/driver/src/domain/competition/order/mod.rs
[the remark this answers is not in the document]
MartinquaXD: But wasn't this specific address picked because the `0xeee.eee` is already commonly used for `ETH`? In that case I would be fine with dropping it. Also I believe we are already using it in other contexts.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -130,10 +130,19 @@ impl Solution {
...
         // The solution is to do access list estimation in two steps: first, simulate
         // moving 1 wei into every smart contract to get a partial access list.
-        let partial_access_lists = try_join_all(
-            self.user_trades()
-                .map(|trade| async { self.partial_access_list(eth, simulator, trade).await }),
-        )
+        let partial_access_lists = try_join_all(self.user_trades().map(|trade| async {
+            if !trade.order.buys_eth() || !trade.order.pays_to_contract(eth).await? {
+                return Ok(Default::default());
+            }
```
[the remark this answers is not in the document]
MartinquaXD: Usually you can't make meaningful progress after an error. But in this case when this call fails for an order that doesn't pay out to a smart contract we could take our chances and assume it doesn't pay out to a smart contract and use a default access list there.
If we are right we save the `Settlement` from erroring out and if we are wrong the settlement will fail again when we try to estimate the gas (because we didn't generate the required access list after the error) so no harm done IMO.

crates/driver/src/infra/api/solve/dto/auction.rs
MartinquaXD: I'm specifically thinking about [this](https://cowservices.slack.com/archives/C036JAGRQ04/p1673260079140729) idea. If we decide to go the route of letting solvers decide which order to include based on `sell_token_balance` it would matter for them if the value is actually 0 or if we just encountered an error while computing the amount and defaulted to 0.
If, on the other hand, we make the driver responsible for querying the current balances for each order we might as well drop those fields from the order coming from the `autopilot` altogether.

crates/driver/src/infra/blockchain/contracts.rs
MartinquaXD: Ahh, I misunderstood the code. It's a struct to configure `Contracts` in case you want to set some specific values. 👌

crates/driver/src/tests/setup/solver.rs
vkgnosis: Purely technical side:
You definitely shouldn't use a sync mutex across an await point. When there is no await point it depends on how long one expects acquiring the lock to usually take. If you're just incrementing an integer or cloning struct then a sync mutex is going to be faster. When you're waiting for some long expensive computation then it should be an async mutex.

Non technical:
One reason I prefer the sync mutex if possible is that it is a simpler component. An async mutex is a sync mutex with extra stuff. I don't want to involve extra stuff when it's not needed.  Another reason is that it communicates the intent to some degree. If I see a sync mutex in async code then I know it's meant to be held for short durations.

It's true that there is some danger of forgetting what kind of mutex it is but usually I still prefer the sync mutex even when taking that into account. Often it is clear that the operation is fast and there is little danger of misusing the mutex.

Vaguely related, I also wouldn't mark a function that doesn't need to be async as async just in case. It's more efficient to do this but it also feels more correct. For an async function there is no danger of misuse because the compiler will help you so the downside is smaller. But, I think clippy already has the equivalent warning for a sync Mutex held across await point.

crates/driver/src/infra/time.rs
vkgnosis: Tokio has a concept of fake time btw https://docs.rs/tokio/latest/tokio/time/index.html .

vkgnosis (approved):

crates/driver/src/infra/api/solve/dto/auction.rs
nlordell: 👍

PR #1046, opened 2023-01-10
Introduce Order and order::NonLiquidity Domain Objects

nlordell:
This PR adds the `Order` domain model. Additionally, it introduces a `NonLiquidity` marker type that wraps an order to indicate that it is a user and non-liquidity order. This allows the Baseline solver to operate on `NonLiquidity` orders only so it never accidentally solver for a liquidity order.

### Test Plan

Rust compiler. An E2E test will be added for testing baseline solving.

### DIFF
--- a/crates/solvers/src/domain/order.rs
+++ b/crates/solvers/src/domain/order.rs
@@ -33,11 +33,11 @@ pub enum Class {
     Liquidity,
 }
 
-/// A user order marker type.
+/// An order that is guaranteed to not be a liquidity order.
 #[derive(Debug)]
-pub struct UserOrder<'a>(&'a Order);
+pub struct NonLiquidityOrder<'a>(&'a Order);
 
-impl<'a> UserOrder<'a> {
+impl<'a> NonLiquidityOrder<'a> {
     /// Wraps an order as a user order, returns `None` if the specified order is
     /// not a user order.
     pub fn new(order: &'a Order) -> Option<Self> {
--- a/crates/solvers/src/domain/order.rs
+++ b/crates/solvers/src/domain/order.rs
@@ -35,9 +35,9 @@ pub enum Class {
 
 /// An order that is guaranteed to not be a liquidity order.
 #[derive(Debug)]
-pub struct NonLiquidityOrder<'a>(&'a Order);
+pub struct NonLiquidity<'a>(&'a Order);
 
-impl<'a> NonLiquidityOrder<'a> {
+impl<'a> NonLiquidity<'a> {
     /// Wraps an order as a user order, returns `None` if the specified order is
     /// not a user order.
     pub fn new(order: &'a Order) -> Option<Self> {
--- a/crates/solvers/src/domain/order.rs
+++ b/crates/solvers/src/domain/order.rs
@@ -1,0 +1,54 @@
+//! The domain object representing a CoW Protocol order.
+
+use crate::domain::eth;
+
+/// A CoW Protocol order in the auction.
+#[derive(Debug, Clone)]
+pub struct Order {
+    pub uid: Uid,
+    pub sell: eth::Asset,
+    pub buy: eth::Asset,
+    pub side: Side,
+    pub class: Class,
+}
+
+/// UID of an order.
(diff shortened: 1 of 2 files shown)


crates/solvers/src/domain/order.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// An order with a fixed sell amount and a minimum buy amount.
+    Sell,
+}
+
+/// The order classification.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Class {
+    Market,
+    Limit,
+    Liquidity,
```
MartinquaXD, line 33: Might be nice to add some comments here describing the specifics of each type. I guess the most important aspect here in the context of solvers is that `Liquidity` orders should never get surplus.

crates/solvers/src/domain/order.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/// A user order marker type.
+#[derive(Debug)]
+pub struct UserOrder<'a>(&'a Order);
```
MartinquaXD, line 38: Is the deciding factor here that the order comes from a user or that it's surplus capturing? Because some users (in the future maybe all [[0](https://cowservices.slack.com/archives/C035N0YEB6J/p1673017823742939)]) can create liquidity orders.

MartinquaXD (approved):
Types look reasonable just some questions about details.

crates/solvers/src/domain/order.rs
vkgnosis, line 38: This seems like it might become awkward because the type can only hold a reference. This means you can't pass user order by value. But if you make the type own an Order then the only way to create a user order is to consume the original order. Alternatively you could add a ref user order and an owning user order but that feels awkward too. I don't know how big of a problem this is or how to solve it. Also not blocking the PR.

vkgnosis (approved):

crates/solvers/src/domain/order.rs
nlordell: I used a reference for now as that is what I currently need. We can always change to a:
```rust
pub struct UserOrder<'a>(Cow<'a, Order>);
```

If we need to support both owned and borrowed orders.

nlordell: You are right. `UserOrder` is not a good name... Basically I need a name for a non-liquidity order (which we have called "user order" in the past, where I got the name from).

Not sure what it should be though...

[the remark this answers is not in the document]
nlordell: In my model, `type Order = LiquidityOrder | UserOrder`; hence a `UserOrder` is a specialization of an `Order`. ATM, I don't need to distinguish liquidity orders from other orders, which is why I don't have a `LiquidityOrder` abstraction.

Concretely, I wanted to take a `[Order]` and map it to a `[UserOrder]` and only solve for those. The marker type makes it impossible at a type level to solve for `LiquidityOrder`s which the baseline solver was not intended to do.

MartinquaXD: SurplusCapturingOrder is a mouthful but I guess it would be correct.

nlordell: We want to add [surplus capturing liquidity orders](https://www.notion.so/cownation/Surplus-Capturing-Liquidity-Orders-Advanced-Wait-for-CoW-006c00cfe9b3487a8cb6254b08859fc3) which would not be a "user order", since they shouldn't ever trade against on-chain liquidity, and therefore should not be valid order input to the Baseline solver.

nlordell: I will rename this to `NonLiquidityOrder` so it is consistent with the `usize`/`NonZeroUsize` analogy.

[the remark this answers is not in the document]
nlordell: I will rename this to `NonLiquidityOrder` so it is consistent with the `usize`/`NonZeroUsize` analogy.

nlordell: ddbfdfea740c780fcc1223c0d03175862e25895a

PR #1048, opened 2023-01-10
Introduce Auction and Solution Domain Models

nlordell:
This PR introduces the auction and solver domain models and adds conversion boiler plate between the their equivalent DTOs.

### Test Plan

Rust Compiler, and a pinky promise for E2E test suite that covers the new logic.

### DIFF
--- a/crates/solvers/src/api/dto/auction.rs
+++ b/crates/solvers/src/api/dto/auction.rs
@@ -11,8 +11,7 @@ use serde_with::{serde_as, DisplayFromStr};
 use std::collections::HashMap;
 
 impl Auction {
-    /// Converts an object data transfer object into its domain object
-    /// representation.
+    /// Converts a data transfer object into its domain object representation.
     pub fn to_domain(&self) -> Result<auction::Auction, Error> {
         Ok(auction::Auction {
             orders: self
--- a/crates/solvers/src/api/dto/auction.rs
+++ b/crates/solvers/src/api/dto/auction.rs
@@ -1,9 +1,60 @@
-use crate::util::serialize;
+use crate::{
+    api::dto::Error,
+    domain::{auction, eth, liquidity, order},
+    util::{conv, serialize},
+};
 use bigdecimal::BigDecimal;
 use ethereum_types::{H160, U256};
+use itertools::Itertools as _;
 use serde::Deserialize;
 use serde_with::{serde_as, DisplayFromStr};
 use std::collections::HashMap;
+
+impl Auction {
+    /// Converts an object data transfer object into its domain object
+    /// representation.
+    pub fn to_domain(&self) -> Result<auction::Auction, Error> {
+        Ok(auction::Auction {
+            orders: self
+                .orders
+                .iter()
+                .map(|order| -> order::Order {
+                    order::Order {
+                        uid: order::Uid(order.uid),
+                        sell: eth::Asset {
+                            token: eth::TokenAddress(order.sell_token),
+                            amount: order.sell_amount,
+                        },
+                        buy: eth::Asset {
+                            token: eth::TokenAddress(order.buy_token),
+                            amount: order.buy_amount,
+                        },
+                        side: match order.kind {
+                            Kind::Buy => order::Side::Buy,
+                            Kind::Sell => order::Side::Sell,
+                        },
+                        class: match order.class {
+                            Class::Market => order::Class::Market,
+                            Class::Limit => order::Class::Limit,
+                            Class::Liquidity => order::Class::Liquidity,
+                        },
+                    }
+                })
+                .collect(),
+            liquidity: self
+                .liquidity
+                .iter()
+                .filter_map(|liquidity| match liquidity {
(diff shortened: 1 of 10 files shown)


crates/solvers/src/util/conv.rs
```diff
@@ -0,0 +1,32 @@
+//! Conversion utilities.
+
+use crate::domain::eth;
+use bigdecimal::BigDecimal;
+use ethereum_types::U256;
+
+/// Converts a `BigDecimal` value to a `eth::Rational` value.
+pub fn decimal_to_rational(d: &BigDecimal) -> Option<eth::Rational> {
```
vkgnosis, line 8: It is unclear under what conditions this function returns None. eth::Rational is a ratio of U256. Does this mean that all values >= eth::Rational::MIN and <= eth::Rational::MAX should return Some and all other values None?

I tried to convert `BigDecimal::from_str("1.1111111111111111111111111111111111111111111111111111111111111111111111111111111``")` which failed.

This feels like a function that should have a unit test to me too because the implementation is pretty difficult and the previous criticism of unit tests with regards to mocking and making refactoring harder doesn't apply.

nlordell: Will add some unit tests here.

crates/solvers/src/domain/solution.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::collections::HashMap;
+
+/// A domain model of a solution to an auction.
+pub struct Solution {
+    pub prices: ClearingPrices,
+    pub trades: Vec<Trade>,
+    pub interactions: Vec<Interaction>,
+}
+
+/// A set a uniform clearing prices. They are represented as a mapping of token
```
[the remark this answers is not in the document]
nlordell: ed2b187b297bc71bad0235e6979009948e6f8562

crates/solvers/src/domain/solution.rs
```diff
@@ -0,0 +1,48 @@
+use crate::domain::{eth, liquidity, order};
+use ethereum_types::U256;
+use std::collections::HashMap;
+
+/// A domain model of a solution to an auction.
```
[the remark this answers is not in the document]
nlordell: ed2b187b297bc71bad0235e6979009948e6f8562

crates/solvers/src/api/dto/auction.rs
```diff
@@ -1,10 +1,61 @@
...
+};
 use bigdecimal::BigDecimal;
 use ethereum_types::{H160, U256};
+use itertools::Itertools as _;
 use serde::Deserialize;
 use serde_with::{serde_as, DisplayFromStr};
 use std::collections::HashMap;
 
+impl Auction {
+    /// Converts an object data transfer object into its domain object
```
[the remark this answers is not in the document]
nlordell: ed2b187b297bc71bad0235e6979009948e6f8562

crates/solvers/src/util/conv.rs
nlordell: > tried to convert `BigDecimal::from_str("1.1111111111111111111111111111111111111111111111111111111111111111111111111111111")` which failed.

I thought it panic'ed - phew! This is expected to return `None` is it would overflow the numerator (and, incidentally, the denominator as well). Added a unit test and refactored the implementation to make sure it reduces the ratio before returning.

nlordell: 45836eb5a6db76e0b9f942fb01e4cac6a1b9b1f2

crates/solvers/src/util/conv.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::domain::eth;
+use bigdecimal::BigDecimal;
+use ethereum_types::U256;
+use num::{rational::Ratio, BigUint};
+
+/// Converts a `BigDecimal` value to a `eth::Rational` value. Returns `None` if
+/// the specified decimal value cannot be represented as a rational of `U256`
+/// integers.
+pub fn decimal_to_rational(d: &BigDecimal) -> Option<eth::Rational> {
```
vkgnosis, line 11: Are you sure you want this to fail if the value cannot be *exactly* represented? The function could round instead. 

Maybe it's fine but it might be unexpected that a value like from my previous example doesn't work. I mean a value that fits into the range of values that Rational can represent but not exactly. For example, when converting a u64 to f32 it is usually expected that the closest f32 value is chosen when it can't represent the original value exactly.

The current behavior only makes sense if it is guaranteed that the original values can be perfectly represented. Is this the case?

nlordell: > The current behavior only makes sense if it is guaranteed that the original values can be perfectly represented. Is this the case?

Yes. Rationals only appear for liquidity in ways that can be represented as U256 rationals:
- Uniswap fees (these are actually u24 rationals)
- Balancer fees and token weights (these are represented as x/1e18 where x needs to fit within a u256 because of its on-chain representation)

If we need an approximate rational, we can add that later, but for now I think this is the correct intended behaviour.

vkgnosis (approved):

PR #1050, opened 2023-01-11
Fix alerter getting rate limited by api

vkgnosis:
Fixes https://github.com/cowprotocol/services/issues/1017

Changes:
- Try to find closed market orders first.
- Locally rate limit the outgoing /api/orders requersts

### Test Plan

manually ran alerter

### DIFF
--- a/crates/alerter/src/main.rs
+++ b/crates/alerter/src/main.rs
@@ -189,6 +189,7 @@ struct Alerter {
     // hasn't been a trade for some time and that there is an order that has been matchable for some
     // time.
     no_trades_but_matchable_order: IntGauge,
+    api_get_order_min_interval: Duration,
 }
 
 struct AlertConfig {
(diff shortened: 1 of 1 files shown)


nlordell (approved):

crates/alerter/src/main.rs
```diff
@@ -236,19 +243,30 @@ impl Alerter {
...
-            let order = self.orderbook_api.order(closed_order).await?;
-            if order.status == OrderStatus::Fulfilled {
+        closed_orders.retain(|order| !self.open_orders.contains_key(&order.uid));
+        // We're trying to find an order that has been filled. Try market orders first because they
+        // are more likely to be.
+        closed_orders.sort_unstable_by_key(|order| match order.class {
+            OrderClass::Market => 0u8,
+            OrderClass::Limit(_) => 1,
+            OrderClass::Liquidity => 2,
+        });
```
MartinquaXD, line 255: Do we care about liquidity orders in the alerter at all? If not, they could also be filtered out in the `retain()` step.
As a side note for future reference: If we only care about bringing the market orders to the front `itertools::partition` could do that in linear time. But I see that it's currently not a dependency of the alerter and introducing that for this use case seems overkill.

MartinquaXD: Just saw a few lines up outside the diff that liquidity orders already get filtered.

MartinquaXD (approved):

PR #1052, opened 2023-01-11
Remove deprecated fee and price routes

vkgnosis:
While on call I noticed that we still have these routes even though they have been deprecated between 1.5 years and half a year ago.

We still get some requests to them but that is out of our control and it isn't reasonable to keep them for this reason.

### Test Plan

CI, metrics after merge


crates/e2e/tests/e2e/eth_integration.rs
```diff
@@ -75,28 +74,6 @@ async fn eth_integration(web3: Web3) {
     let http_factory = HttpClientFactory::default();
     let client = http_factory.create();
 
-    // Test fee endpoint
```
nlordell, line 78: Should we test `/quote` endpoint here instead?

nlordell (approved):
Yay!

crates/e2e/tests/e2e/eth_integration.rs
vkgnosis: I don't think it's really needed because this test is just checking for the status code and we already create a quote when handling a new order. But I could add it back.

nlordell: If we already "use" the `/quote` path elsewhere, I'm OK with removing this test here.

vkgnosis: I mean internally we create a quote but not explicitly in this test.

vkgnosis: Added a test back

Issue #1053, opened 2023-01-12
Introduce hard deadline for price estimates in the `autopilot`


MartinquaXD:
The native price cache and limit order quoting in the `autopilot` is very sensitive to individual price estimators slowing down the process.
This can lead to a majority of limit orders not being quoted because of a single slow price estimator [[0](https://cowservices.slack.com/archives/C0361CDD1FZ/p1673473223246509)].
We already have some rate limiting logic in place but that doesn't work great in this case. Instead it makes sense to introduce another CLI argument for the `autopilot` that configures a hard timeout after which slow price estimation request get cancelled.

This would make the limit order quoting process very predictable which also helps with tuning other parameters. Ideally we would also get some metrics to count how many requests got aborted for each price estimator. This would allow us to create an alert based on those metrics so we can look into the poor performance in a timely manner.

vkgnosis:
There is already a hard deadline for price estimates in the http request timeout. This might not be fine grained enough.

> This can lead to a majority of limit orders not being quoted because of a single slow price estimator.

True but it feels intrinsic with the way we have currently defined what a price estimate is. In the autopilot to get a price estimate we need to evaluate *all* price estimators to use the best result. This is problematic because as we have observed a single slow estimator makes every request slow. By using a deadline (as we already do through the http timeout), what we are really doing is softening this requirement so that we no longer require all price estimators. If that is what we want, we should do it more explicitly and configure the RacingCompetitionEstimator abstraction to require some subset amount rather than all.

vkgnosis:
Made a writeup of some thoughts related to this here https://github.com/cowprotocol/services/issues/1055 .

MartinquaXD:
> ... what we are really doing is softening this requirement so that we no longer require all price estimators. If that is what we want, we should do it more explicitly

This makes sense to me.

> configure the RacingCompetitionEstimator abstraction to require some subset amount rather

I think putting the timing requirement into the `RacingCompetitionEstimator` might make things a bit weird. For example what happens if we get the `n` estimates required by the `RacingCompetitionEstimator` to return early but there is still plenty of time left? Should we return early or wait for more prices?
In the first case we could get suboptimal prices.
In the second case a new flavor like a `TimedCompetitionEstimator` which only concerns itself with time could make more sense (the current implementation is already quite complicated).

PR #1054, opened 2023-01-12
Native price cache concurrent requests

MartinquaXD:
If we want to be able to keep all limit orders quoted at the same time we have to keep the native price cache up-to-date at all times. Unfortunately the number of unique tokens across all limit orders is too big (~200) for the native price cache background worker to keep up-to-date when we are updating prices purely sequentially.

This PR introduces a CLI argument (`--native-price-cache-concurrent-requests`) which allows the background worker to update prices concurrently.
This PR in conjunction with #1053 should allow us to perfectly tweak the parameters required to keep the native price cache entirely up-to-date on average.

### Test Plan
Updated existing unit tests
Added new unit test
Manual test to verify argument printing:
```
2023-01-12T10:22:46.540Z  INFO autopilot: running autopilot with validated arguments:
...
native_price_cache_max_update_size: 100
native_price_cache_concurrent_requests: 1
amount_to_estimate_prices_with: None
...
```
Manual test to verify help text:
```
      --native-price-cache-concurrent-requests <NATIVE_PRICE_CACHE_CONCURRENT_REQUESTS>
          How many price estimation requests can be executed concurrently in the maintenance task

          [env: NATIVE_PRICE_CACHE_CONCURRENT_REQUESTS=]
          [default: 1]
```

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -91,10 +91,12 @@ impl Inner {
         &'a self,
         tokens: &'a [H160],
         max_age: Duration,
+        parallelism: usize,
     ) -> futures::stream::BoxStream<'_, (usize, NativePriceEstimateResult)> {
-        futures::stream::iter(tokens)
+        let estimates = tokens
+            .iter()
             .enumerate()
-            .then(move |(index, token)| async move {
+            .map(move |(index, token)| async move {
                 {
                     // check if price is cached by now
                     let now = Instant::now();
@@ -128,7 +130,9 @@ impl Inner {
                 };
 
                 (index, result)
-            })
+            });
+        futures::stream::iter(estimates)
+            .buffered(parallelism)
             .boxed()
     }
 }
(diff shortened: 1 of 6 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -231,8 +237,10 @@ async fn update_recently_used_outdated_prices(
     update_interval: Duration,
     update_size: Option<usize>,
     max_age: Duration,
+    concurrent_requests: usize,
 ) {
     while let Some(inner) = inner.upgrade() {
+        tracing::error!("run background task");
```
nlordell, line 243: Left-over log used for debugging?

vkgnosis (approved):
It is good to have this configurable.

In general, it is not reasonable to have all native prices up to date because the work required for that scales with the number of unique tokens in the set of solvable orders. If we had a million tokens then it would be unreasonable to expect all of them to have an up to date native price at all times. So I don't know what the best value for this parameter is, but it is good to be able to configure it.

crates/shared/src/price_estimation/native_price_cache.rs
MartinquaXD: Yeah, this one slipped through the cracks.
Removed in [7351ee7](https://github.com/cowprotocol/services/pull/1054/commits/7351ee736d8209aa9ae17a80ee79c90c9250961c)

nlordell (approved):

Issue #896, opened 2022-12-01
`/execute` endpoint for `driver`


sistemd:
The purpose of `/execute` is to publish the settlement transaction on the blockchain. It needs to encode the solution into its onchain form. The reference implementation of this process can be found in the [`SettlementEncoder`](https://github.com/cowprotocol/services/blob/6988d10081b57c8a12c1ac0adff6bac13e93626c/crates/solver/src/settlement/settlement_encoder.rs#L32). The only additional detail concerning the `driver` is that the solution ID should be encoded as part of `calldata`. More details about this below.

The `driver` needs to support multiple submission strategies:
- [ ] Using the public mempool.
- [ ] Flashbots.
- [ ] Gelato.

Reference implementations of each strategy can be found today as implementations of the [`SettlementSubmitting`](https://github.com/cowprotocol/services/blob/b40e4883e719a6f235b94a26ac9efb2cbb8af594/crates/solver/src/settlement_submission/submitter.rs#L101) trait. The `driver` should be configured with CLI flags to determine which strategy should be used.

#### Input

TODO

#### Output

TODO

vkgnosis:
Continuing the discussion from https://github.com/cowprotocol/services/issues/895#issuecomment-1335375205 about the ID. If there is no security reason for doing the CSPRNG approach then I'd like to re-use the auction id. It's simpler and it intuitively feels more proper.

I'm not completely certain that there is no security issue yet. Still thinking about it.

MartinquaXD:
I believe simply using the auction id to associate the solution with a solver has a slight problem. Staging and prod have different auction id counters but use the same settlement contract, right? That would mean eventually we would end up with 2 settlements that have the same id associated with them (when the prod auction id catches up to the current staging id). Technically we would be able to know which settlement belonged to staging and which to prod based on the submission timing and solver address (so far solvers always had different addresses across staging and prod) but it seems cleaner to me to simply use random ids to avoid collisions.

vkgnosis:
Auction id is what identifies auctions. It is natural to use it here because our goal is to identify an auction. I don't think it's more likely that we overlook something there versus another approach of picking it.

With a random id there is also a chance of conflict (I didn't do the math for how many auctions). With the auction id the conflict is only between staging and prod which we have a solution with the reasonable (?) assumption that solver addresses are not reused between staging and prod. With a random id there can be conflicts in the same environment which is harder to handle.

nlordell:
[...]

PR #1056, opened 2023-01-12
Driver `/execute` endpoint

sistemd:
Half of https://github.com/cowprotocol/services/issues/896.

Add the `/execute` endpoint. Store settlements using an in-memory `DashMap` for now. We can think more about if persistence is needed for these, since the benefits seem somewhat minimal.

### Test Plan

Add an automated test in a follow-up.

### DIFF
--- a/crates/driver/src/infra/blockchain/mod.rs
+++ b/crates/driver/src/infra/blockchain/mod.rs
@@ -118,13 +118,6 @@ impl Ethereum {
             .map_err(Into::into)
     }
 
-    /// Broadcast a transaction.
-    pub async fn send_transaction(&self, tx: eth::Tx) -> Result<(), Error> {
-        let tx = Self::into_request(tx);
-        self.web3.eth().send_transaction(tx).await?;
-        Ok(())
-    }
-
     /// Necessary for the boundary integration.
     pub fn web3(&self) -> DynWeb3 {
         self.web3.clone()


crates/driver/src/domain/competition/mod.rs
```diff
@@ -7,24 +23,81 @@ pub use {
...
+        let settlement = solution::Settlement::encode(&self.eth, auction, &solution).await?;
+        let score = settlement.score(&self.eth, auction, gas).await?;
+        let solution_id = solution::Id::random();
+        self.settlements.insert(solution_id, settlement);
+        // Remove the settlement after it's expired.
+        let settlements = Arc::clone(&self.settlements);
+        tokio::spawn(async move {
+            tokio::time::sleep(Self::expiration_time()).await;
+            settlements.remove(&solution_id);
+        });
```
vkgnosis, line 69: There is only one competition at a time. This means that from the perspective of the driver it gets a call to /solve and optionally a call to /execute afterwards. Then it gets a call to /solve again and so on. Based on this, we don't need to get rid of after a fixed time. We can keep it until the next call to /solve. This means we don't need a hashmap and can just store the most recently returned solution.

At least that's how I think this process works. Might have to clarify it in the openapi spec.

vkgnosis: It is not necessary because we can replace the stored solution on the next /solve call. There is no need to also replace after timeout.

nlordell: I think then you limit your driver to being run per deployment. One advantage of the `DashMap` is that you can, in theory, as an external solver use the same `driver` + `solver` instance to both production and staging. Since the implementation is quite simple, I think I'm fine with the `DashMap` (although I probably would have used a `Mutex<HashMap>` to avoid the external dependency since performance is not a concern here - the locking operation happens 0.2 times per second).

vkgnosis: > I think then you limit your driver to being run per deployment. 

That's true. If we want to support running for multiple deployments then this doesn't work. 

> DashMap vs Mutex<HashMap>

I also prefer Mutex<HashMap> when performance doesn't matter like in this case. No dependency and much simpler component.

nlordell: > But aren't production and staging always running on different networks?

No, we run the Mainnet/Production and Mainnet/Staging for example. A small solver team may want to run a single driver instance to reduce on node RPC costs for example.

> So I think @vkgnosis might be correct here

👍 - I think @vkgnosis's suggestion is good and it simplifies the code a bit. We can always revisit if we get a request for using a single solver + driver instance for multiple deployments.

crates/driver/src/infra/mempool/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+        infra::blockchain::{self, Ethereum},
+    },
+    futures::future::join_all,
+};
+
+/// The mempool to use for publishing onchain transactions. The public mempool
+/// of an [`Ethereum`] node can be used, or one of the private mempools offered
+/// by various transaction relay services.
+#[derive(Debug, Clone)]
+pub struct Mempool(Inner);
```
nlordell, line 13: I'm not a fan of moving this into the domain just yet. Executing transactions on chain seems easy (just use the `eth_sendTransaction` RPC call right?), but it is actually super difficult to get right and made reliable.

I would definitely wrap our current settlement submission logic in a `boundary` module for transaction submitting. To name a couple of the things that we need to support:
- Re-broadcasting the transaction with higher gas prices if it does not get included
- Sending cancellation transactions if we go over the deadline of the solution stops submitting successfully in order to save on gas costs

crates/driver/src/domain/competition/mod.rs
```diff
@@ -7,24 +24,81 @@ pub use {
...
+/// expire after a certain amount of time, at which point trying to use them
+/// will return an `[Error::SolutionNotFound]`.
+#[derive(Debug)]
+pub struct Competition {
+    pub solver: Solver,
+    pub eth: Ethereum,
+    pub simulator: Simulator,
+    pub now: time::Now,
+    pub mempools: Vec<Mempool>,
+    pub settlements: Arc<DashMap<solution::Id, settlement::Simulated>>,
```
nlordell, line 42: I was under the impression that we had decided to simplify this and only keep a single settlement per driver that can be `/settle`-ed.

I think it is also fine this way, just wanted to point that out.

nlordell (approved):
LGTM.

crates/solver/src/settlement_submission.rs
```diff
@@ -309,10 +310,18 @@ pub enum SubmissionError {
...
     /// The submission is disabled
     Disabled(DisabledReason),
-    /// An error occured.
+    /// An error occurred.
     Other(anyhow::Error),
 }
 
+impl std::fmt::Display for SubmissionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{self:?}")
```
nlordell, line 319: Should we implement `Display` here properly?

crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -120,6 +117,29 @@ impl Ethereum {
             .map(Into::into)
             .map_err(Into::into)
     }
+
+    /// Broadcast a transaction.
+    pub async fn send_transaction(&self, tx: eth::Tx) -> Result<(), Error> {
+        let tx = Self::into_request(tx);
+        self.web3.eth().send_transaction(tx).await?;
+        Ok(())
+    }
```
nlordell, line 126: I think this may be dead code.

vkgnosis (approved):
Approving assuming we do the store only one settlement change that Nick also mentioned again.

vkgnosis:
Sorry I accidentally pushed to this branch while testing.

PR #1057, opened 2023-01-12
Do not treat http solver rate limit as generic error

vkgnosis:
Quasimodo has recently gotten the ability to signal when it can't keep up through the TOO_MANY_REQUESTS status code. Our code is not aware of this and treats it as a generic error. Generic errors sometimes end up logged as internal errors so that we can investigate them. This doesn't make sense for the rate limit.

This PR makes us aware of the rate limit and turns it into the already existing price estimation error of the same type.

Note that this PR does not make the Rust RateLimiter aware of this status code. We might want to integrate that this so that it counts towards our local rate limit too. This can be done in another PR.

### Test Plan

Didn't see a good way to unit test this. After merging the error logs about this status code should go away.

### DIFF
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -1,9 +1,9 @@
 use crate::http_client::response_body_with_size_limit;
 use ::model::auction::AuctionId;
-use anyhow::{ensure, Context, Result};
+use anyhow::{anyhow, Context};
 use reqwest::{
     header::{self, HeaderValue},
-    Client, Url,
+    Client, StatusCode, Url,
 };
 use serde_json::json;
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):

crates/shared/src/http_solver.rs
```diff
@@ -94,15 +102,15 @@ impl HttpSolverApi for DefaultHttpSolverApi {
...
+    ) -> Result<model::SettledBatchAuctionModel, Error> {
         // The timeout we give to the solver is one second less than
         // the deadline to make up for overhead from the network.
         // We use one second because the old MIP solver uses integer timeouts.
         let solver_timeout = timeout
             .checked_sub(Duration::from_secs(1))
             .context("no time left to send request")?;
 
-        let mut url = self.base.join("solve")?;
+        let mut url = self.base.join("solve").context("join base")?;
```
nlordell, line 113: This might even be `.unwrap()` worthy. AFAIK, this can only fail if `"solve"` is not a valid URL path segment.

nlordell (approved):
Nice change!

crates/shared/src/http_solver.rs
vkgnosis: I thought so too but wasn't completely sure so I kept the same behavior as before.

PR #1059, opened 2023-01-12
Let http solvers indicate they cannot keep up

vkgnosis:
Motivated by recent Quasimodo changes.

Note that for the driver API I only added the status code to /quote and not the other routes. The other routes should always be available.

### Test Plan

n/a

### DIFF
--- a/crates/driver/openapi.yml
+++ b/crates/driver/openapi.yml
@@ -23,6 +23,8 @@ paths:
                 $ref: "#/components/schemas/QuoteResponse"
         400:
           $ref: "#/components/responses/BadRequest"
+        429:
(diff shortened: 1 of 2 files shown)


MartinquaXD (approved):

nlordell (approved):

nlordell:
> /quote and not the other routes.

Should we also add this to the `/solve` route? I think it _may_ make sense there.

vkgnosis:
If we only use solve for full auctions then I don't think it should be self rate limited because these requests come in regular intervals and not concurrently.

vkgnosis:
I'm merging this now because either way we'll take it for /quote but feel free to continue to discuss here or in new issue.

PR #1060, opened 2023-01-12
Make api internal error logging more explicit

vkgnosis:
We are converting anyhow::Error to internal error ApiReply in many places. This happens through functions like `internal_error` and `convert_json_response` and the IntoWarpReply implementation on anyhow::Error. These functions end up internally `tracing::error!` logging the error which makes it hard to understand when these error logs happen and where they come from.

This PR turns most of the implicit error logs into explicit logs and adds more context in some places. This PR was motivated by seeing a bunch of rate limited price estimation error logs show up. This rate limit error no longer results in an error log as we already have metrics on price estimators.

### Test Plan

Did not manually test how all errors are printed. After merging should observe same or better api internal error server logs.


crates/shared/src/api.rs
```diff
@@ -272,21 +253,15 @@ impl IntoWarpReply for PriceEstimationError {
...
-            ),
-            Self::Other(err) => with_status(
-                internal_error(err.context("price_estimation")),
-                StatusCode::INTERNAL_SERVER_ERROR,
-            ),
+            Self::UnsupportedOrderType => {
+                tracing::error!("PriceEstimaton::UnsupportedOrderType");
+                internal_error_reply()
+            }
+            Self::RateLimited(_) => internal_error_reply(),
```
MartinquaXD, line 260: This error no longer gets logged. Is that intentional?

MartinquaXD (approved):

PR #1063, opened 2023-01-12
Introduce Baseline Solver E2E Test

nlordell:
Last PR from the stack starting at #1045 

This PR adds a simple E2E test for the Baseline solver. Specifically, the E2E test just covers the solver matching an order directly with a Uniswap V2 pool.

Adding new test cases should be very straight forward now:
- Duplicate the `direct_swap.rs` file
- Modify the auction and solution JSON for the new test case

I plan on introducing new E2E tests in order to get more coverage with the Baseline solver. Namely:
- Test Balancer V2 weighted pool logic
- Port existing Baseline solver unit tests to this repo

### Test Plan

🥄


crates/solvers/src/tests/baseline/direct_swap.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::num::NonZeroUsize;
+
+#[tokio::test]
+async fn test() {
+    let engine = tests::SolverEngine::new(cli::Command::Baseline(cli::baseline::Arguments {
+        weth: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+            .parse()
+            .unwrap(),
+        base_tokens: Vec::new(),
+        max_hops: NonZeroUsize::new(2).unwrap(),
```
MartinquaXD, line 15: Wasn't `max_hops == 1` supposed to be used for direct swaps?

MartinquaXD (approved):

crates/solvers/src/tests/baseline/direct_swap.rs
```diff
@@ -0,0 +1,105 @@
...
+    }))
+    .await;
+
+    let solution = engine
+        .solve(json!({
```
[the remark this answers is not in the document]
nlordell: That would be awesome!

vkgnosis (approved):

crates/solvers/src/tests/baseline/direct_swap.rs
MartinquaXD: We could test that `max_hops` gets respected as an upper limit by introducing pool data which would lead to a better settlement with 2 hops.

nlordell: For sure! That should be in a separate E2E test though right?

MartinquaXD: I have no problem with adding that in another PR. Just wanted to mention it because this would have caught the issue I pointed out.

nlordell: Started writing the E2E test for this, and I found a bug 🙈. Hops are **additional tokens** in a path, and not **additional segments**. So, by the baseline solver code, `A -> B` is a 0-hop path, `A -> B -> C` is a 1 hop path, etc.

PR #1067, opened 2023-01-14
Refactor objective value computation to be more reusable

vkgnosis:
The objective value computation was haphazardly added to RatedSettlement which made it hard to understand and reuse. This PR moves into a standalone module.

### Test Plan

refactored test still works


crates/solver/src/objective_value.rs
```diff
@@ -0,0 +1,153 @@
+use crate::settlement::{external_prices::ExternalPrices, Settlement};
+use num::BigRational;
+use number_conversions::u256_to_big_rational;
+use primitive_types::U256;
+
+pub struct Inputs {
```
nlordell, line 6: Not a huge fan of the name (I see where its coming from though, "the inputs to the objective value"). It feels a bit general.

Don't have any better ideas though.

nlordell (approved):

crates/solver/src/objective_value.rs
vkgnosis: The motivation for making this a struct instead of a function is that I wanted better distinction of the arguments that all have the same type. Agree the name that doesn't feel great but also don't have a better one. Referring to this as `objective_value::Inputs` is ok.

nlordell: > The motivation for making this a struct instead of a function is that I wanted better distinction of the arguments that all have the same type

I like the struct! Was really just referring to the name `Inputs`. I think the PR is a great fix!

Not blocking merging in any way, just pointed it out.

PR #1073, opened 2023-01-16
Better tracability with more info spans

MartinquaXD:
Ideally we would like to know what caused each log to appear. This means for any sort of request that we are using info spans containing the `request_id`.
But we have a bunch of background tasks which can produce a lot of logs which fly under the radar.
This PR adds an info span for every task we spawn (minus tests, request batching and the main tasks like the `orderbook`).

Note that logs originating from the initialization sequence didn't get any new tracing spans but I think this is fine.

### Test Plan
Ran autopilot locally to check logs (I just picked a bunch of updated logs in no particular order):
```
2023-01-16T12:02:36.063Z DEBUG ServiceMaintenance:maintenance{block=16419218}: shared::event_handling: latest blocks: Some((16419091, 0xbbbf69cb04d06485409046a10bdd55da81582849a7cbae8798ca34159ed06375)) - Some((16419218, 0xe0061b782368497a6aca7d309cadacfa868b3a5bf6ee625e4d0d46e7585c8d13))
2023-01-16T12:02:34.712Z DEBUG LimitOrderQuoter:surplus_fee{order_spec=OrderFeeSpecifier { sell_token: 0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce, buy_token: 0xb4a81261b16b92af0b9f7c4a83f1e885132d81e4, sell_amount: BigDecimal("3875000000000000000000000000") }}: shared::price_estimation::competition: winning price estimate query=Query { from: None, sell_token: 0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy } result=Ok(Estimate { out_amount: 148056793816200000000000000, gas: 356391 }) estimator="ZeroEx"
2023-01-16T12:02:36.803Z DEBUG SolvableOrdersCache: autopilot::solvable_orders: filtered order order=0x112cca73d7670158cb3947889ac7d84e5c47774a8c727106fb7654c49fbc64b0e63a13eedd01b624958acfe32145298788a7a7ba800186a0 reason=insufficient_balance
2023-01-16T12:02:37.311Z DEBUG CurrentBlockStream: shared::current_block: increased current block number delta=1.0 new_block=16419219
2023-01-16T12:02:31.504Z  WARN AutoUpdatingSolverTokenOwnerFinder: shared::bad_token::token_owner_finder::solvers::solver_finder: failed to update token list err=error decoding response body: invalid length 19, expected a (both 0x-prefixed or not) hex string with length of 40 at line 2 column 25...
```

### DIFF
--- a/crates/autopilot/src/limit_orders/metrics.rs
+++ b/crates/autopilot/src/limit_orders/metrics.rs
@@ -54,7 +54,7 @@ impl LimitOrderMetrics {
                     tokio::time::sleep(Duration::from_secs(10)).await;
                 }
             }
-            .instrument(tracing::info_span!("LimitOrderMetrics")),
+            .instrument(tracing::info_span!("limit_order_metrics")),
         );
     }
 }
--- a/crates/autopilot/src/limit_orders/metrics.rs
+++ b/crates/autopilot/src/limit_orders/metrics.rs
@@ -1,4 +1,5 @@
 use std::time::Duration;
+use tracing::Instrument;
 
 use prometheus::IntGauge;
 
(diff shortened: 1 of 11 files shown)


crates/autopilot/src/limit_orders/metrics.rs
```diff
@@ -14,43 +15,47 @@ pub struct LimitOrderMetrics {
...
+                    unquoted_limit_orders_gauge.set(awaiting_quote);
+                    quoted_limit_orders_gauge.set(quoted_limit_orders);
+                    usable_limit_orders_gauge.set(usable_limit_orders);
+                    unusable_limit_orders_gauge.set(unusable);
+
+                    tokio::time::sleep(Duration::from_secs(10)).await;
+                }
             }
-        });
+            .instrument(tracing::info_span!("LimitOrderMetrics")),
```
nlordell, line 57: naming-nit. We've been using `snake_case` for tracing spans. Might be worth coming up with a standard for the span names so they are consistent everywhere (doesn't have to be `snake_case`)

nlordell (approved):
Nice! Thanks for going over all of these.

Changes look good, just a comment about coming up with some consistent naming scheme for spans.

crates/autopilot/src/limit_orders/metrics.rs
MartinquaXD: I started out using `snake_case` and stumbled over [this](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/lib.rs#L561-L565) which used `CamelCase`. I guess I can just turn everything to snake case including the existing info span.

nlordell: Not necessarily set on `snake_case` (although, this would likely be my subjective preference). However, there might even be an argument for `PascalCase` or `SCREAMING_SNAKE_CASE` so that it stands out from the log module name. Up to you on what you think would be best in this case.

> I guess I can just turn everything to snake case including the existing info span.

Personally, I like this idea.

PR #1074, opened 2023-01-16
Demote 0x, Paraswap and 1Inch api logs to `trace!`

MartinquaXD:
The `0x`, `Paraswap` and `1Inch` api implementation can issue a lot of logs which essentially all get duplicated in the `CompetitionPriceEstimator`.
Also `1Inch` and `0x` can issue logs which are crazy long (`1Inch` liquidity sources and `0x` limit orders as liquidity).

All of this seems to be just noise when you are debugging something locally. I think one could even argue how useful those logs are in the actual deployment but I left them just to be safe.


nlordell (approved):
Love it!

vkgnosis (approved):
Nice. I wanted to do this change too.

PR #1075, opened 2023-01-16
Don't log rate limit reset unnecessarily

MartinquaXD:
We only need to log that we reset the rate limit when we were rate limited before.


nlordell (approved):
🙏

vkgnosis (approved):

Issue #893, opened 2022-12-01
`/quote` endpoint for `driver`


sistemd:
The `/quote` endpoint is used to provide price estimates to the protocol and to the end users. This provides our users with some reasonable expectations before they place an order.

The `/quote` endpoint should build a "fake" auction with a single order, and use this auction to call the `/solve` endpoint. Then, it should build the quote from the solution that it gets.

For some inspiration on what data should be expected and returned, check the quote endpoint of the orderbook.

Related to https://github.com/cowprotocol/services/issues/894.

PR #1077, opened 2023-01-16
Driver `/quote` test

sistemd:
Fixes https://github.com/cowprotocol/services/issues/893.

- Implements a small `dev-geth` crate which is ran as part of `docker-compose` and allows a `geth` process to be restarted, therefore clearing the testing environment.
- Updates the test setup code to restart `geth` before setting up the blockchain state.
- Updates the test setup code to shut down the driver at the end of each test.
- Rename `side` to `kind` in all APIs for consistency.

### Test Plan

:cowboy_hat_face:

### DIFF
--- a/crates/dev-geth/Cargo.toml
+++ b/crates/dev-geth/Cargo.toml
@@ -3,8 +3,6 @@ name = "dev-geth"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 tokio = { version = "1.24", features = ["full"] }
 axum = "0.6"


crates/dev-geth/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "dev-geth"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
nlordell, line 6: Nit: can we remove this comment from the auto-generated template `cargo init`?

crates/dev-geth/rustfmt.toml
```diff
@@ -0,0 +1,12 @@
+format_code_in_doc_comments = true
```
nlordell, line 1: nit: Should you use `ln -s` here to keep the rustfmt configurations in sync?

crates/dev-geth/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+//! `dev-geth` exposes an HTTP port to which a POST request can be made in order
+//! to restart geth and provide a clean testing environment. Note that
+//! restarting the program will wait until geth starts responding to RPC
+//! requests before returning 200 to the caller.
+//!
+//! geth is the only node with dev mode which implements the
+//! `eth_createAccessList` endpoint. When hardhat adds support for
+//! `eth_createAccessList`, this can be removed from our infrastructure.
+//!
+//! [geth]: https://geth.ethereum.org/
```
nlordell, line 13: Not suggesting that we change it here - but what would be really cool (and help us avoid having needing to run integration tests with `--threads 1`) would be to start multiple `geth` instances bound to different ports for each test and have this helper manage them.

nlordell (approved):

nlordell:
Boss!

crates/dev-geth/src/main.rs
vkgnosis: Kind of related: This doesn't need to be separately run in docker. It would be better to make this a normal struct that is instantiated by the test that needs geth. The only reason to make it a separate container is so that the same geth can be used in multiple tests, but if each test is restarting it anyway, or with Nic's suggestion has it's own node, then this should be integrated directly into the test.

vkgnosis (approved):
This isn't run by CI, right? Should it?

PR #1079, opened 2023-01-17
Rename driver's `/execute` endpoint to `/settle`

sistemd:
Thought this was a better name.

### Test Plan

Tests still passing. The `/execute` endpoint wasn't being tested yet anyway.


nlordell (approved):
I think `settle` is a better name!

vkgnosis (approved):
Does this need an openapi update too?

PR #1082, opened 2023-01-17
Driver submission CLI args

sistemd:
Follow-up to https://github.com/cowprotocol/services/pull/1056.

Adds CLI args for settlement submission.

### Test Plan

Tests still passing, my next PR will be testing the `/settle` endpoint.

### DIFF
--- a/crates/driver/src/boundary/mempool.rs
+++ b/crates/driver/src/boundary/mempool.rs
@@ -46,9 +46,8 @@ pub enum HighRisk {
     Disabled,
 }
 
-/// The mempool to use for publishing settlements onchain. The public mempool
-/// of an [`Ethereum`] node can be used, or one of the private mempools offered
-/// by various transaction relay services.
+// TODO Perhaps a better name for this in the future might be Relay
+/// The mempool to use for publishing settlements onchain.
 #[derive(Clone)]
 pub struct Mempool {
     config: Config,


crates/driver/src/infra/config/cli.rs
```diff
@@ -34,6 +36,42 @@ pub struct Args {
...
+    /// encoded string. This can't be specified along with --solver-private-key,
+    /// exactly one of them must be specified.
+    #[clap(long, env)]
+    pub solver_address: Option<eth::H160>,
+
+    /// The private key used to sign transactions. Expects a 32-byte hex encoded
+    /// string. This can't be specified along with --solver-address, exactly one
+    /// of them must be specified.
+    #[clap(long, env)]
+    pub solver_private_key: Option<String>,
```
vkgnosis, line 69: The natural Rust type of this would Vec<u8> combined with passing a function to clap to parse it from a hex string similar to what we do with Duation. Or a wrapper type that is parsed that way similar to H160 above. Could be a complete type that validates that it is a correct private key. Then you get better error messages from clap rather than failing when we validate it later.


Not required to merge the PR, just a note.

vkgnosis (approved):

crates/driver/src/boundary/mempool.rs
```diff
@@ -28,20 +29,41 @@ use {
...
+/// the mempool.
+#[derive(Debug, Clone)]
+pub enum GasPriceCalculation {
+    EthGasStation,
+    GasNow,
+    GnosisSafe,
+    Web3,
+    BlockNative { api_key: String },
+    Native,
+}
```
nlordell, line 42: This is quite complete! I do think that just hardcoding this to be `Native` should be good enough. We have multiple sources from pre-EIP-1559 gas estimates, but in the new driver, we probably don't need to support this.

crates/driver/src/infra/config/cli.rs
```diff
@@ -34,6 +36,42 @@ pub struct Args {
...
+        long,
+        env,
+        use_value_delimiter = true,
+        default_value = "https://rpc.flashbots.net"
+    )]
+    pub flashbots_api_urls: Vec<Url>,
+
+    /// The solver address used to sign transactions. Expects a 20-byte hex
+    /// encoded string. This can't be specified along with --solver-private-key,
+    /// exactly one of them must be specified.
```
nlordell, line 61: `clap` supports specifying this [with argument groups](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations).

crates/driver/src/infra/config/cli.rs
```diff
@@ -34,6 +36,42 @@ pub struct Args {
...
+    #[clap(long, env)]
+    pub solver_address: Option<eth::H160>,
```
nlordell, line 63: I think `solver_address` and `solver_private_key` belong in the solver configuration. Specifically, different solvers tied to the same driver need to submit transactions with different private keys/addresses.

nlordell (commented):
In general code looks good. Comments are:
- Solver accounts are tied to individual solver engines and not the driver (i.e. 1 driver <> N solver accounts)
- We can just always use `Native` gas estimator. This removes configuration options and is supported on all networks (including the dev-geth setup we use for E2E). The other estimators are legacy from a pre-EIP-1559 time.

crates/driver/src/infra/config/cli.rs
```diff
@@ -143,6 +181,77 @@ impl std::fmt::Display for ContractAddresses {
...
+    /// above regular gas price estimation.
+    #[clap(long, env, default_value = "3")]
+    pub submission_max_additional_flashbots_tip: f64,
+}
+
+#[derive(Debug, Clone, Copy, clap::ValueEnum)]
+pub enum Mempool {
+    Public,
+    Flashbots,
+}
```
MartinquaXD, line 247: Not sure if `Mempool` is the right name (with regard to `Flashbots` and `Gelato`).
Maybe `Relayer`? But this would also make the `Public` variant fit less.

nlordell: I think for `Flashbots` it is (it is just a private mempool). Gelato is a relayer, and isn't a `Mempool`.

nlordell: Gelato submits a "call" to their "relay network". A relayer picks up the call and is responsible for crafting and executing the transaction on the blockchain (the relayer _probably_ uses some mempool, but this is orthogonal to the "call" we submit to their "relay network").

crates/driver/src/main.rs
```diff
@@ -59,24 +60,111 @@ pub async fn run(
...
+                            mempool::HighRisk::Disabled
+                        } else {
+                            mempool::HighRisk::Enabled
+                        },
+                    )
+                    .await
+                    .unwrap()],
+                    cli::Mempool::Flashbots => {
+                        join_all(args.flashbots_api_urls.iter().map(|url| async {
+                            Mempool::flashbots(
```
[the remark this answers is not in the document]
MartinquaXD: When creating a flashbots instance a new `dyn GasPriceEstimating` also gets created although one could only have a single instance and copy that around where it's needed with the `Arc`. Having only one instance of `dyn GasPriceEstimating` used everywhere could reduce latency by sharing or caching requests.
Just wanted to bring it up in case that was not intended.

crates/driver/src/infra/config/cli.rs
MartinquaXD: > Do we even want to support Gelato?

Quite a bit of [work](https://github.com/cowprotocol/services/blob/main/crates/shared/src/gelato_api.rs) was done to work toward that. The nice thing about Gelato is that they would take care of all the trickiness involved in submitting stuff.

nlordell: > Do we even want to support Gelato?

Yes, eventually we want to support even other relayers (Infura ITX for example).

I do think it fits nicely with the way things are currently organized, so I wouldn't worry too much about optimizing the current abstractions for it until they are needed.

I think `Mempool` or `Txpool` makes sense as a name for now.

crates/driver/src/boundary/mempool.rs
```diff
@@ -30,18 +30,26 @@ use {
...
 
+#[derive(Debug, Clone, Copy)]
+pub enum HighRisk {
+    Enabled,
+    Disabled,
+}
+
+/// The mempool to use for publishing settlements onchain. The public mempool
+/// of an [`Ethereum`] node can be used, or one of the private mempools offered
+/// by various transaction relay services.
```
nlordell, line 52:
```suggestion
/// by various private networks.
```

I think flashbots is typically referred to as a private network.

crates/driver/src/infra/config/cli.rs
```diff
@@ -143,6 +181,77 @@ impl std::fmt::Display for ContractAddresses {
...
+    /// Don't submit transactions with high revert risk (i.e. transactions that
+    /// interact with on-chain AMMs) to the public mempool. This can be
+    /// enabled to avoid MEV when private transaction submission strategies
+    /// are available.
+    #[clap(long, env)]
+    pub submission_disable_high_risk_public_mempool_transactions: bool,
+
+    /// The maximum time to spend trying to settle a transaction
+    /// through the Ethereum network before going back to solving. Specified in
+    /// seconds.
```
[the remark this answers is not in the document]
nlordell: > I think how the new driver is currently implemented already supports solving a new auction while it still tries to bring the other one on-chain

I think there will be dragons if we try to `/execute` multiple settlements while a previous one is still pending.

Might actually make sense to allow `/solve`-ing, but reject `/execute` requests in case a previous settlement transaction is still "in flight".

crates/driver/src/main.rs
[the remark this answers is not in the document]
nlordell: I do think the `dyn GasPriceEstimating` should be shared if possible. Basically it polls `eth_getFeeHistory` and updates its internal state based on the resulting data, so having multiple instances of it are a bit wasteful.

MartinquaXD (approved):

PR #1085, opened 2023-01-18
Initialize Uniswap V2 Liquidity Fetcher

nlordell:
This PR initializes the boundary liquidity fetcher such that it can fetch Uniswap V2 compatible liquidity.

Note that the new domain representation for Uniswap V2 liquidity is a departure from the existing implementation in the `shared` and `solver` crates. Specifically, we now **dynamically** specify the liquidity fetching parameters (router address and init code). This allows N number of Uniswap V2 compatible exchanges to be configured for liquidity, meaning we no longer require special code for adding new Uniswap V2 clones and can do it 100% through configuration.

Currently, the fetcher is just initialized and not used. In the following 3 PRs, I will:
- [ ] Add logic for fetching Uniswap V2 liquidity and mapping them to domain objects
- [ ] Integrate it with the current driver solving infrastructure
- [ ] Add an E2E test.

### Test Plan

The Rust compiler for now. E2E test after we map boundary liquidity to `uniswap::v2::Pool`s.

### DIFF
--- a/crates/driver/src/domain/liquidity/uniswap/v2.rs
+++ b/crates/driver/src/domain/liquidity/uniswap/v2.rs
@@ -1,33 +1,12 @@
-use {crate::domain::eth, std::cmp::Ordering};
-
 /// The famous Uniswap V2 constant product pool, modelled by `x · y = k` [^1].
 ///
 /// Note that there are many Uniswap V2 clones with identical behaviour from a
 /// liquidity point of view, that are therefore modelled by the same type:
 /// - SushiSwap (Mainnet and Gnosis Chain)
+/// - Swapr (Mainnet and Gnosis Chain)
 /// - Honeyswap (Gnosis Chain)
 /// - Baoswap (Gnosis Chain)
 ///
 /// [^1]: <https://uniswap.org/whitepaper.pdf>
 #[derive(Clone, Debug)]
-pub struct Pool {
-    pub router: eth::ContractAddress,
-    pub reserves: Reserves,
-}
-
-/// A Uniswap V2 pool reserves. These reserves are orders by token address and
-/// are guaranteed to be for distict tokens.
-#[derive(Clone, Copy, Debug)]
-pub struct Reserves(eth::Asset, eth::Asset);
-
-impl Reserves {
-    /// Creates new Uniswap V2 token reserves, returns `None` if the specified
-    /// token addresses are equal.
-    pub fn new(a: eth::Asset, b: eth::Asset) -> Option<Self> {
-        match a.token.cmp(&b.token) {
-            Ordering::Less => Some(Self(a, b)),
-            Ordering::Equal => None,
-            Ordering::Greater => Some(Self(b, a)),
-        }
-    }
-}
+pub struct Pool {}


crates/driver/src/domain/liquidity/fetcher.rs
```diff
@@ -1,17 +0,0 @@
-use crate::domain::{competition::auction, liquidity};
```
nlordell, line 1: This technically moved to `crates/driver/src/domain/liquidity/fetcher/mod.rs`, GitHub just doesn't show it that way.

crates/driver/src/main.rs
```diff
@@ -4,33 +4,29 @@
 
 use {
     crate::{domain::competition, infra::api},
-    infra::blockchain,
```
nlordell, line 7: Just formatting changes - the `use` statements were split, I just joined them.

crates/driver/src/boundary/liquidity/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+pub struct Fetcher {
+    blocks: CurrentBlockStream,
+    inner: LiquidityCollector,
+}
+
+impl Fetcher {
+    /// Creates a new fether for the specified configuration.
```
vkgnosis, line 45:
```suggestion
    /// Creates a new fetcher for the specified configuration.
```

crates/driver/src/boundary/liquidity/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub async fn new(eth: &Ethereum, config: &liquidity::fetcher::Config) -> Result<Self> {
+        let blocks = current_block::Arguments {
+            block_stream_poll_interval_seconds: BLOCK_POLL_INTERVAL,
+            block_stream_retriever_strategy: BlockRetrieverStrategy::EthCall,
+        }
+        .stream(boundary::web3(eth))
+        .await?;
```
vkgnosis, line 52: Isn't the block stream more of a global component that should be passed in?

crates/driver/src/boundary/liquidity/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// Fetches liquidity for the specified auction.
+    pub async fn fetch(&self, _auction: &auction::Auction) -> Result<Vec<liquidity::Liquidity>> {
+        todo!()
+    }
+}
+
+impl std::fmt::Debug for Fetcher {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
vkgnosis, line 89: Why manual impl?

crates/driver/src/boundary/liquidity/mod.rs
nlordell: I considered this, but the only place this is currently used is here. If we need to create a block stream for something else, then I agree that we should pass it in.

crates/driver/src/boundary/liquidity/mod.rs
nlordell: It was because `LiquidityCollector: !Debug`.

crates/driver/src/boundary/liquidity/mod.rs
vkgnosis: This will probably be forgotten if someone else needs a pool cache. I'd rather pass it in but it's no big deal either.

vkgnosis (approved):

PR #1086, opened 2023-01-18
Gzip solver instance before S3 upload

vkgnosis:
Follow up to https://github.com/cowprotocol/services/pull/1027#discussion_r1062177319 .

Gzipping the instances makes them 90% of their original size which is a large win for low effort. They can easily be gunzipped after downloading. There are other choices for even better compression but I didn't want to spend too much time on the choice and we can change it later if needed.

I chose the flate2 crate because it is already in our dependency tree through reqwest's gzip support.

### Test Plan

I manually tested that my local `gunzip` correctly decompresses the output.

### Release notes

Optionally add notes, suggestions and warnings for the releaser. They will be included in the release changelog.


nlordell (approved):
I _think_ you should be able to upload with `Content-Encoding: gzip` header, that way the file looks like a `.json` file on the S3 bucket and `curl`-ing it will "just work" (i.e. S3 will set the `Content-Encoding` header on the response and `curl` will gunzip it auto-magically).

It my suggestion works, then the ergonomics of this would be much nicer.

vkgnosis:
I saw your comment in the linked PR but I couldn't find documentation saying this works. We want S3 to store the gzipped version. It's not about in what form it is sent over the wire to S3. There might be a way to automatically have this header set when retrieving but that's a separate thing.

crates/solver/src/s3_instance_upload.rs
```diff
@@ -40,14 +42,25 @@ impl S3InstanceUploader {
...
+        // fast.
+        let mut encoder = GzEncoder::new(json, Compression::best());
+        let mut encoded: Vec<u8> = Vec::with_capacity(json.len());
+        encoder.read_to_end(&mut encoded).context("gzip encoding")?;
+        Ok((key, encoded))
     }
 
     fn filename(&self, auction: AuctionId) -> String {
-        format!("{}{auction}.json", self.filename_prefix)
+        format!("{}{auction}.json.gzip", self.filename_prefix)
```
MartinquaXD, line 63: The auction ids could collide across staging/prod. Do we have separate S3 buckets or some other strategy to deal with that? I guess this could also be handled with different `filename_prefix`es.

MartinquaXD: Also do you think we should compress all the files we already have as a follow up task to have a consistent file format?

MartinquaXD (approved):

crates/solver/src/s3_instance_upload.rs
vkgnosis: > The auction ids could collide across staging/prod. 

The intention is to have the prefix handle conflicts

> Also do you think we should compress all the files we already have as a follow up task to have a consistent file format?

No. My plan is to enable the new code (it isn't enabled yet) and leave the old bucket as is until it expires and we delete it.

nlordell:
> I couldn't find documentation saying this works.

I think it should be possible with [`content_encoding`](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/client/fluent_builders/struct.PutObject.html#method.content_encoding) and [`content_type`](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/client/fluent_builders/struct.PutObject.html#method.content_type) methods on the `PutObject` builder.

I also found some relevant information [here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html).

Note that this isn't blocking the merge for me, but it should be easy to test that it works.

nlordell:
(sorry for closing the PR, fat fingers).

Issue #1084, opened 2023-01-18
Seasolver token support - multiple token holders


pata-eth:
COW protocol checks seasolver's endpoint `/token_holders` for a list of token addresses with a corresponding holder to determine whether a token is supported. Seasolver refreshes the list once every couple of days, so it's possible that a token may be considered unsupported if a holder in the list does no longer keep a balance of such token. 

It'd be useful if the driver checked for multiple token holders until it finds a valid one instead of determining validity based on a single datapoint. This will dramatically increase the chance of success when determining token validity, and therefore improve trading experience. Also, it's been difficult for Seasolver to learn about when orders are being dropped because their tokens were no longer supported because a holder wasn't found anymore. 

Current setup:

```json
"0x0327112423F3A68efdF1fcF402F6c5CB9f7C33fd": "0xd4dbf96db2fdf8ed40296d8d104b371adf7dee12"
```

Proposed setup:
```json
"0x0327112423F3A68efdF1fcF402F6c5CB9f7C33fd": [
    "0xd4dbf96db2fdf8ed40296d8d104b371adf7dee12",
    "0x592349f7dedb2b75f9d4f194d4b7c16d82e507dc",
    "0x16de73a47c9d663b02e625315d3c03f7aa3b8d7b"
]
```

Thank you, folks.

cc: @sunce86

nlordell:
@pata-eth - what do you think about changing the API to return token holders to token instead of a large JSON map. Something like:

```
GET tokens/0x0327112423F3A68efdF1fcF402F6c5CB9f7C33fd/holders
```

Which would return:
```json
[
    "0xd4dbf96db2fdf8ed40296d8d104b371adf7dee12",
    "0x592349f7dedb2b75f9d4f194d4b7c16d82e507dc",
    "0x16de73a47c9d663b02e625315d3c03f7aa3b8d7b"
]
```

The reason I am proposing this API as it more closely matches the inner interface we use:

https://github.com/cowprotocol/services/blob/cabe79274f7ea61d554791c54ea567e01d9759ba/crates/shared/src/bad_token/token_owner_finder.rs#L40-L46C2

sunce86:
@nlordell Wouldn't that slow down the implementation? We are now fetching holders in background thread and with your proposed solution it would require us to send http request each time.

nlordell:
> We are now fetching holders in background thread and with your proposed solution it would require us to send http request each time.

[...]

PR #1088, opened 2023-01-18
Support multiple owners for token

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1084

### Release notes

Requires new version of endpoint provided by solvers in order to have a smooth transition.


crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
```diff
@@ -34,7 +34,7 @@ impl TokenOwnerSolverApi for SolverConfiguration {
         self.query().await.map(|token_owner_pairs| {
             token_owner_pairs
                 .into_iter()
-                .filter_map(|(token, owner)| owner.map(|owner| (token, vec![owner])))
+                .filter_map(|(token, owners)| owners.map(|owners| (token, owners)))
```
MartinquaXD, line 37: I found it a bit difficult to parse that expression. Not sure if my attempt is any improvement, though. 😄 
```suggestion
                .filter_map(|(token, owners)| Some((token, owners?)))
```

crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
```diff
@@ -16,7 +16,7 @@ pub struct SolverConfiguration {
 impl SolverConfiguration {
     /// Return type is `Token, Option<Owner>` because there are
     /// entries containing `Null` instead of owner address.
-    async fn query(&self) -> Result<HashMap<Token, Option<Owner>>> {
+    async fn query(&self) -> Result<HashMap<Token, Option<Vec<Owner>>>> {
```
MartinquaXD, line 19: This function expects a JSON format. I believe that means the only additional "feature" `Option<Vec<Owner>>` has over `Vec<Owner>` is that this would parse:
```json
{"0xeee...eee": null}
```
If we don't care about that case then we wouldn't even have to convert the response in `get_token_owner_pairs()` if we expect `Vec<Owner>` in the response.

MartinquaXD (approved):
Implementation is fine as it is. Just wanted to mention that we might be making things more complicated than they have to be.

crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
vkgnosis: That was the case before too. The result is already a Map so why would you add a key and then a null value. From an API point of view it doesn't make sense to store Option in the Map here. Dunno if we have properly documented this API anywhere or or what solvers do.

vkgnosis:
> Requires new version of endpoint provided by solvers in order to have a smooth transition.

So what happens if we merge this now? Requests fail until solvers update, right? You could make it backward compatible by attempting to parse the old and the new format. Up to you if that is worth it.

crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
MartinquaXD: Not sure if I'm misunderstanding you here but I totally agree that supporting the `null` case with `Option<T>` is a bit weird and simply expecting `Vec<Owner>` makes things easier for us. Maybe I expressed myself wrong.

nlordell: An empty `Vec` is a reasonable default value for `null` JSON IMO (unless we need to have this distinction somewhere, which I don't think we do).

vkgnosis: I was trying to say:
- You are right that Option<Vec<Owner>> should be Vec<Owner>.
- Before this PR already, Option<Owner> should have been plain Owner.
- There is no reason for this API to ever associate null with a key rather than not having the key in the first place.

crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
sunce86: Deprecated. Thanks for noticing though.

crates/shared/src/bad_token/token_owner_finder/solvers/solver_api.rs
sunce86: Seasolver removed the `null` so now it's safe to remove Option. Done.

sunce86:
> So what happens if we merge this now? Requests fail until solvers update, right? You could make it backward compatible by attempting to parse the old and the new format. Up to you if that is worth it.

Seasolver created a separate version already. My plan is to merge and apply configuration change at the same time (for staging). For prod it's easier because the new configuration will be applied together with the new code release.

nlordell (approved):

vkgnosis (approved):

PR #1089, opened 2023-01-18
Fetch Uniswap V2 Liquidity

nlordell:
This PR introduces boundary logic for fetching Uniswap V2 liquidity, and mapping it to domain liquidity models.

This is mostly just mapping `solver` types to `driver` types.

One notable change is the introduction of a `SettlementHandling::as_any` method, in order to downcast `dyn SettlementHandling` trait objects into concrete types. This is needed to extract pool information that is relevant to the domain liquidity object. I think that this is a bit of a hack, but mostly harmless and should be removed (i.e. purged with 🔥) once we complete the boundary liquidity mappings and finally move code from `solver` to `driver` crate. The plan for removing the use of `Any` is to change `settlement_handler` to use concrete types for each pool type. This is quite hard to do ATM because `settlement_handler` being a trait object is needed in multiple places in the code - but once the `solver` binary can be decommissioned, then this should no longer be an issue to change (as the code that relies on `settlement_handler` being a trait object can be removed).

### Test Plan

Rust compiler, E2E test that verifies liquidity fetching is coming once we integrate liquidity fetching into the driver's solving code.

### DIFF
--- a/crates/driver/src/boundary/liquidity/mod.rs
+++ b/crates/driver/src/boundary/liquidity/mod.rs
@@ -1,18 +1,26 @@
 use {
     crate::{
         boundary,
-        domain::{competition::auction, eth, liquidity},
+        domain::{
+            competition::{auction, order},
+            eth,
+            liquidity,
+        },
         infra::blockchain::Ethereum,
     },
     anyhow::Result,
     futures::future,
     itertools::Itertools,
+    model::TokenPair,
     shared::{
         baseline_solver::BaseTokens,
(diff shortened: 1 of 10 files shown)


vkgnosis (approved):
> I think that this is a bit of a hack, but mostly harmless and should be removed

We shouldn't rely on using Any and casting long term. I am ok with having it short term with a specific plan of getting rid of it again.

nlordell:
> I am ok with having it short term with a specific plan of getting rid of it again.

Added a quick comment in the PR description on my plan for removing it.

Issue #1033, opened 2023-01-05
EthFlow Orders Missing From Order Creation Metrics


nlordell:
It appears that created EthFlow orders are not being tracked with metrics, this is causing our order creation metrics to report incorrect data.

This issue captures the work to:
- Start tracking **created** EthFlow orders (or, more generally speaking on-chain orders) with Prometheus metrics
- Update existing Grafana dashboards to account for on-chain orders.

josojo:
@fedgiac Could you tackle this issue this week?

fedgiac:
We will be changing how eth-flow orders are handled and orders are counted. This task will be handled there.

fedgiac:
As far as I know, metrics are still missing. However, we want to revisit how the ETH flow works, so this would be included in the refactor.

PR #1090, opened 2023-01-18
Add onchain order error metrics

fedgiac:
In partial fulfillment of #1033.

Adds new metrics to track errors encountered when picking up onchain orders.

Some less related changes that I included in this PR:
- `convert_onchain_order_placement` returned a result but there was no code path where it errored, so I unwrapped it.
- I upgraded an error from an `Err` to `unreachable!` since hitting this error means that the code is trying to parse an event that is not an order creation event as an order creation event.

### Test Plan

I tried the modified ignored unit test and it doesn't error.
I ran the autopilot on Görli with `export ETHFLOW_CONTRACT="0x40A50cf069e992AA4536211B23F286eF88752187"` and `export ETHFLOW_INDEXING_START=8000000`. I waited a bit for some orders to be indexed and I saw at `http://localhost:9589/metrics`:

```
# HELP gp_v2_autopilot_onchain_orders_onchain_order_errors Keeps track of errors in picking up onchain orders.
# TYPE gp_v2_autopilot_onchain_orders_onchain_order_errors counter
gp_v2_autopilot_onchain_orders_onchain_order_errors{error_type="quote not found"} 78
```
which makes sense since the quote that was used to create that order does not come from my own autopilot instance.


crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -419,6 +423,32 @@ fn extract_invalidated_order_uids(
...
 
+/// Temporarily stores metrics for onchain order parsing, which will later be written to the global
+/// metrics object.
+struct MetricsOnchainOrderParsing {
+    errors: Arc<RwLock<HashMap<String, u64>>>,
+}
+impl MetricsOnchainOrderParsing {
+    fn new() -> Self {
+        Self {
+            errors: Arc::new(RwLock::new(HashMap::<String, u64>::new())),
```
MartinquaXD, line 434:
```suggestion
            errors: Arc::new(Default::default()),
```

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -419,6 +423,32 @@ fn extract_invalidated_order_uids(
...
+        let MetricsOnchainOrderParsing { errors } = self;
+        if let Ok(read_lock) = errors.read() {
+            for (error_label, increment) in read_lock.iter() {
+                metrics
+                    .onchain_order_errors
+                    .with_label_values(&[error_label])
+                    .inc_by(*increment);
+            }
+        } else {
+            tracing::warn!("Unable to write onchain order error metrics, poisoned RW lock");
```
MartinquaXD, line 447: Usually we just unwrap those errors because they can only happen when a thread panics while holding a lock.
This can not happen during this call to `.read()` or the 2 calls to `.write()`.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -419,6 +423,32 @@ fn extract_invalidated_order_uids(
...
+        }
+    }
+    fn write_to_metrics(self, metrics: &'static Metrics) {
+        let MetricsOnchainOrderParsing { errors } = self;
+        if let Ok(read_lock) = errors.read() {
+            for (error_label, increment) in read_lock.iter() {
+                metrics
+                    .onchain_order_errors
+                    .with_label_values(&[error_label])
+                    .inc_by(*increment);
```
MartinquaXD, line 444: I believe if all you do is increasing a labeled counter there is probably no advantage to having the intermediate `MetricsOnchainOrderParsing` instance because there is probably no harm in having partially updated metrics.
And even with your intermediate struct the update to the metrics will not be atomic anyway.

MartinquaXD (commented):
I think this can be simplified without the intermediate metrics collector struct.
Also `for a in list.iter()` is equivalent to `for a in &list`.

crates/autopilot/src/database/onchain_order_events.rs
fedgiac: Related code was simplified away.

crates/autopilot/src/database/onchain_order_events.rs
fedgiac: I removed the intermediate struct.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -425,61 +429,74 @@ type GeneralOnchainOrderPlacementData = (
...
-            };
-            let event = match data {
-                ContractEvent::OrderPlacement(event) => event,
-                _ => {
-                    return Err(anyhow!(
-                        "parse_general_onchain_order_placement_data should not reach this state"
-                    ))
+    let futures = order_placement_events_and_quotes_zipped.into_iter().map(
+        |(EthContractEvent { data, meta }, event_timestamp, quote_id)| {
+            let metrics_clone = metrics.clone();
```
MartinquaXD, line 442: Last nit to clear this diff up:
`metrics` doesn't have to be cloned into the future which means the whole block does not have to be reformatted due to the increased level of indentation. This should make the diff way smaller.

crates/autopilot/src/database/onchain_order_events.rs
fedgiac: This comment doesn't apply anymore after the simplification but I'm still curious about it: the increment was handled in multiple futures, which in theory could be handled by multiple threads, one of which could panic (for example in `unreacable!`). Is this still unnecessary to handle?

fedgiac:
Worth mentioning that I had to rerun the tests again because the test `maintenance::tests::block_stream_retries_failed_blocks` failed ([logs](https://github.com/cowprotocol/services/actions/runs/3950785530/jobs/6763787990#step:11:417)). I don't think it's related to these changes and I couldn't reproduce the issue locally.

crates/autopilot/src/database/onchain_order_events.rs
MartinquaXD: I think there was a misconception. "While holding the lock" wasn't supposed to mean "while the thread has access to the `RwLock`. It rather means the lock gets poisoned if the thread panics in the critical section:
```rust
let lock_guard = rwlock.read(); // lock was acquired; critical section begins
// if the thread panics here the rwlock gets poisoned
drop(lock_guard); // lock gets released; critical section ends
// if the thread panics here the rwlock is unaffected
```

In the earlier version of the PR you never did anything that could cause the thread to panic while in the critical section. You only incremented a counter. Technically this could panic in a debug build when the counter overflows but that would never happen.

nlordell:
Dang... flaky test... the worst!

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -3,7 +3,7 @@ pub mod event_retriever;
 
 use super::{
     events::{bytes_to_order_uid, meta_to_event_index},
-    Metrics, Postgres,
+    Metrics as DatabaseMetrics, Postgres,
```
nlordell, line 6: Just a style thing (so 100% feel free to ignore), but here you can:
```rust
use crate::database;
```

Then you can reference it as `database::Metrics` instead of `DatabaseMetrics`.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -461,18 +473,20 @@ async fn parse_general_onchain_order_placement_data(
...
                     gas_amount: quote.data.fee_parameters.gas_amount,
                     gas_price: quote.data.fee_parameters.gas_price,
                     sell_token_price: quote.data.fee_parameters.sell_token_price,
                     sell_amount: u256_to_big_decimal(&quote.sell_amount),
                     buy_amount: u256_to_big_decimal(&quote.buy_amount),
-                })
-            } else {
-                None
+                }),
+                Err(err) => {
```
nlordell, line 486: Should we log the error as well?

crates/database/src/onchain_broadcasted_orders.rs
```diff
@@ -15,6 +15,21 @@ pub enum OnchainOrderPlacementError {
...
+    pub fn to_metrics_label(&self) -> &str {
+        match self {
+            Self::QuoteNotFound => "quote not found",
+            Self::InvalidQuote => "invalid quote",
+            Self::PreValidationError => "pre-validation error",
+            Self::DisabledOrderClass => "disabled order class",
+            Self::ValidToTooFarInFuture => "valid to too far in future",
+            Self::InvalidOrderData => "invalid order data",
+            Self::InsufficientFee => "insufficient fee",
+            Self::Other => "unspecified error",
```
nlordell, line 28: These long names might be annoying to type in Prometheus metrics, might makes sense to make them more concise and indentifier-like.

100% OK as is though.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -655,6 +669,28 @@ fn extract_order_data_from_onchain_order_placement_event(
     Ok((order_data, owner, signing_scheme, order_uid))
 }
 
+#[derive(prometheus_metric_storage::MetricStorage, Clone, Debug)]
+#[metric(subsystem = "onchain_orders")]
+struct Metrics {
+    /// Keeps track of errors in picking up onchain orders.
+    /// Note that an order might be created even if an error is encountered.
+    #[metric(labels("error_type"))]
+    onchain_order_errors: prometheus::IntCounterVec,
```
nlordell, line 678: Does it make sense to initialize all counters with 0 on startup so that you don't have missing values?

MartinquaXD (approved):

josojo (approved):
LGTM, agree with some of the nits of the others.

crates/autopilot/src/database/onchain_order_events.rs
fedgiac: I prefer not to because `database` already represents the `database` crate in this file, while the metrics come from `autopilot::database` so there would be some extra work needed to disambiguate the two that imho isn't improving readability.

crates/autopilot/src/database/onchain_order_events.rs
nlordell: I think it is fine to leave as is and revisit if ever this causes issues.

crates/autopilot/src/database/onchain_order_events.rs
vkgnosis: Generally we assume that the whole program exits if a thread panics. It is too much of a hassle to try to write code that keeps working when any thread could panic at any time. This is enforced by a panic handler that exits the process on any panic. You can always unwrap mutex lock calls because those only error when the mutex is poisoned by a thread panicing while holding the lock.

PR #1094, opened 2023-01-18
Upload Auction Instance JSON with Gzip Content Encoding

nlordell:
Implemented suggestion from https://github.com/cowprotocol/services/pull/1086#pullrequestreview-1253174388 (I wasn't able to explain my suggestion very well, so thought it was easier to just write it out).

This PR changes the S3 upload logic to explicitly specify that:
- the file is a JSON document
- the file is Gzip encoded

This does a couple of nice things for us:
- It is stored compressed on S3, so saves space 🎉 (same as before)
- AWS automatically sets the correct `Content-*` headers when retrieving the file, this means that HTTP clients will do the Gzip decoding for us, removing the manual step of `gunzip` and making using these uploaded files slightly more ergonomic (motivation for this change)

### Test Plan

Manually run the ignored test.

Notice how it is stored in compressed format on S3 (4.6 KB):

Screenshot 2023-01-18 at 15 00 03

Downloading the file, the correct `Content-*` headers are set, the total JSON size is 11.8 KB, while the HTTP body is only 4.6 KB, and the gunzipping happens as part of "regular HTTP things" and I end up with a JSON document without manual a `gunzip`:

Screenshot 2023-01-18 at 15 04 25

### DIFF
--- a/crates/solver/src/s3_instance_upload.rs
+++ b/crates/solver/src/s3_instance_upload.rs
@@ -43,32 +43,31 @@ impl S3InstanceUploader {
     /// bucket.
     ///
     /// The final filename is the configured prefix followed by `{auction_id}.json.gzip`.
-    pub async fn upload_instance(&self, auction: AuctionId, value: Vec<u8>) -> Result<()> {
-        let (key, bytes) = self.prepare_instance(auction, &value)?;
-        self.upload(key, bytes).await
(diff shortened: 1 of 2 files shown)


MartinquaXD (approved):
Nice improvement!

vkgnosis:
I understood what you meant but I had trouble finding S3 documentation stating that this works. I tested it manually with this PR and it definitely works. Probably just badly documented on S3's side. Good change.

vkgnosis (approved):

PR #1101, opened 2023-01-19
Make auction update interval smarter

vkgnosis:
Currently we wait a fixed interval of 2 seconds after every auction update loop. This commit change the behavior to wait as long is needed to reach the configured target effective update interval.

If creating the auction is slow then we sleep less or not at all. If creating the auction is fast then we sleep the for longer.

This an improvement because we get more auctions on slow environments and less on fast environments. For example, at the moment creating the auction on staging Goerli is very fast which means there is a new auction every 2 seconds. This creates unnecessary churn for the solvers and for the data that we keep track of per auction like instance uploading.

This PR makes the target interval configurable and sets the default to 10 seconds from the previous 2 seconds. This won't make auction creation slower on prod mainnet because creating the auction already takes this long.

### Test Plan
observe after merging

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -139,6 +139,10 @@ pub struct Arguments {
     /// How many quotes the limit order quoter updates in parallel.
     #[clap(long, env, default_value = "5")]
     pub limit_order_quoter_parallelism: usize,
+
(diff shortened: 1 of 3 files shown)


nlordell (approved):
Nice change!

PR #1111, opened 2023-01-20
Use Quote Deadline Parameter

nlordell:
This PR changes the `/quote` endpoint to use a deadline that is passed in with the request instead of a configured value.

The rationale for this change is that the protocol (i.e. `orderbook` and `autopilot`) should be communicating their desired timeout for the quoting. By passing it in, it allows:
- Quoting timeouts to be use-case specific (i.e. we can set a shorter deadline for `fast` price quotes than what we use for `optimal` quotes)
- Avoids the need to synchronize quote timeout values across remote driver configurations.

A small note on implementations. I introduced a separate `quote::Deadline` type instead of re-using the `auction::Deadline`. The reason for this is that I wanted to ensure that quote deadlines are treated differently from auction ones (i.e. they translate to different `SolverTimeout` values).

Additionally, I moved the `domain::competition::quote` module to `domain::quote`. My rationale was that quoting is orthogonal to solver `competition`, so it shouldn't be in the same domain model. Also, while quoting is implemented in terms of types in `competition`, it doesn't, for example, make use of the "main" abstraction `competition::Competition`. I made it as a separate commit so it is easy to revert if others disagree.

### Test Plan

E2E tests were adjusted and pass 🎉.


vkgnosis (approved):

sunce86 (approved):

crates/driver/src/domain/quote.rs
```diff
@@ -62,15 +60,40 @@ pub struct Order {
     pub amount: order::TargetAmount,
     pub side: order::Side,
     pub gas_price: eth::EffectiveGasPrice,
+    pub deadline: Deadline,
+}
+
+/// The deadline for computing a quote for an order.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Deadline(chrono::DateTime<chrono::Utc>);
```
MartinquaXD, line 68: It's a bit unfortunate that we end up with duplicated type implementations. What do you think about having a single shared `Deadline` type but giving it some `PhantomData` so we can have something like `Deadline<Solving>` and `Deadline<Quoting>`.
This could also be used for an execution deadline.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -211,20 +212,44 @@ impl SolverTimeout {
...
+            .ok_or(auction::DeadlineExceeded)
+    }
+
+    /// The time limit passed to the solver when quoting.
+    ///
+    /// Similar to [`SolverTimeout::for_solving`] but for handling quotes. The
+    /// reason for a different value is that less post-processing work is
+    /// required for quoting, meaning a smaller "time buffer" is needed on the
+    /// auction deadline.
+    pub fn for_quoting(
```
ennmichael, line 228: There is a reason why I initially left the quoting deadline calculation away from the `solution`. The quoting process depends on the solving process. On the contrary, the solving process shouldn't know anything about quoting. Quote-related methods do not belong here.

crates/driver/src/domain/mod.rs
```diff
@@ -1,5 +1,6 @@
 pub mod competition;
 pub mod eth;
 pub mod liquidity;
+pub mod quote;
```
ennmichael, line 4: Nice! I wanted to move this here as well but never quite got around to it.

crates/driver/src/domain/quote.rs
ennmichael: Ehhh not sure. It works but it's not pretty. There's not _that much_ duplicated code.

OTOH since quoting is just a form of solving, I don't think I'd disagree with just using `solution::Deadline` here.

crates/driver/src/infra/api/quote/dto/order.rs
```diff
@@ -41,3 +45,9 @@ enum Kind {
     Sell,
     Buy,
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("received an order with an exceeded deadline")]
```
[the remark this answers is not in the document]
ennmichael: We refer to the "quote request" as the "quote order". It's neither here nor there and I wouldn't mind renaming `quote::Order` to `quote::Request` but I don't think that either name is much better. I personally avoid the word `Request` like the plague since it's so damn pervasive.

crates/driver/src/domain/quote.rs
nlordell: > I don't think I'd disagree with just using solution::Deadline here.

One reason I'm not super fond of this, is that it prevents us from using different "timeout buffers" for both solving and quoting (which should be different, the post-quote work is **much** quicker - not involving any external API roundtrip). If we did this, we would leak the choice of "time buffer" to use outside of the `solution` module which I thought wasn't ideal.

I think we can get rid of the duplicated validity check by moving it to when we render a  `SolverTimeout`, then the `Deadline` is as much boiler-plate as a marker type.

crates/driver/src/domain/competition/solution/mod.rs
nlordell: Very good point.

nlordell: Moved. c9409ae3fb16acf921bfcb507bf0865add9728ae

crates/driver/src/domain/quote.rs
nlordell: Alright, so I ended up settling on something like this: c9409ae3fb16acf921bfcb507bf0865add9728ae

Specifically, instead of `SolveTimeout::new(deadline, now)`, we now have `deadline.timeout(now)`. This change makes is so that the `time_buffer` details are tied to the specific deadlines (we have `auction::Deadline::time_buffer` and `quote::Deadline::time_buffer`) and aren't "leaked" into the `solution` module. Also, to reduce boilerplate, `Deadline`'s aren't checked against `now`, but instead this only happens when rendering a `SolverTimeout`. This is nice because:
- We only check once that the timeout is reasonable when actually "using" a deadline, making it so we don't have any copied logic, and just boilerplate (conversion trait implementations) and deadline-specific logic (different timeouts and error messages).
- Simplifies code (`now` doesn't need to be passed in as many places)

Let me know what you think.

docker-compose.yaml
```diff
@@ -47,7 +47,9 @@ services:
       context: .
       dockerfile: docker/Dockerfile.dev-geth
     restart: on-failure
-    network_mode: host
```
nlordell, line 50: `network_mode: host` is not supported on Apple Silicon 😞.

crates/driver/src/domain/quote.rs
ennmichael: Niiice.

docker-compose.yaml
ennmichael: How are you able to run tests then?

ennmichael: This is the only thing I'm unsure about, this is necessary so that `dev-geth` can expose the ports that `geth` is listening to.

ennmichael (approved):

PR #1113, opened 2023-01-20
Fetch Liquidity when `/quote`-ing

nlordell:
This PR changes `/quote` endpoint to first fetch and pass in liquidity to the solver engine.

Additionally, it introduces a new `quote::Tokens` struct to maintain the invariant that `sellToken != buyToken` for quotes (which doesn't make sense).

### Test Plan

Driver tests continue to pass.

Also, manually request a quote with a driver + baseline solver pair and see the baseline solver find a price!
```
2023-01-20T12:48:10.736Z TRACE driver::infra::solver: got response from solver self.config.endpoint=http://localhost:7872/ res=Ok("{\"prices\":{\"0x6b175474e89094c44da98b954eedeac495271d0f\":\"1000000000000000000\",\"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2\":\"1547346991108415943371\"},\"trades\":[{\"kind\":\"fulfillment\",\"order\":\"0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\",\"executedAmount\":\"1000000000000000000\"}],\"interactions\":[{\"kind\":\"liquidity\",\"internalize\":false,\"id\":\"0\",\"inputToken\":\"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2\",\"outputToken\":\"0x6b175474e89094c44da98b954eedeac495271d0f\",\"inputAmount\":\"1000000000000000000\",\"outputAmount\":\"1547346991108415943371\"}]}")
```

We can't quite E2E this yet, since interaction encoding for quotes is not done yet.

### DIFF
--- a/crates/driver/src/domain/quote.rs
+++ b/crates/driver/src/domain/quote.rs
@@ -2,6 +2,7 @@ use crate::{
     domain::{
         competition::{self, order, solution},
         eth,
+        liquidity,
     },
     infra::{
         solver::{self, Solver},
@@ -25,12 +26,12 @@ impl Quote {
     fn new(order: &Order, solution: competition::Solution) -> Result<Self, Error> {
         let sell_price = solution
             .prices
-            .get(&order.sell_token)
+            .get(&order.tokens.sell)
             .ok_or(Error::QuotingFailed)?
             .to_owned();
         let buy_price = solution
             .prices
-            .get(&order.buy_token)
+            .get(&order.tokens.buy)
             .ok_or(Error::QuotingFailed)?
             .to_owned();
         let amount = match order.side {
@@ -55,39 +56,11 @@ impl Quote {
 /// An order which needs to be quoted.
 #[derive(Debug)]
 pub struct Order {
-    pub sell_token: eth::TokenAddress,
-    pub buy_token: eth::TokenAddress,
+    pub tokens: Tokens,
     pub amount: order::TargetAmount,
     pub side: order::Side,
     pub gas_price: eth::EffectiveGasPrice,
     pub deadline: Deadline,
-}
-
-/// The deadline for computing a quote for an order.
-#[derive(Clone, Copy, Debug, Default)]
-pub struct Deadline(chrono::DateTime<chrono::Utc>);
-
-impl Deadline {
-    /// Computes the timeout for solving an auction.
-    pub fn timeout(self, now: time::Now) -> Result<solution::SolverTimeout, DeadlineExceeded> {
-        solution::SolverTimeout::new(self.into(), Self::time_buffer(), now).ok_or(DeadlineExceeded)
-    }
-
-    pub fn time_buffer() -> chrono::Duration {
-        chrono::Duration::seconds(1)
-    }
-}
-
-impl From<chrono::DateTime<chrono::Utc>> for Deadline {
-    fn from(value: chrono::DateTime<chrono::Utc>) -> Self {
-        Self(value)
-    }
-}
-
-impl From<Deadline> for chrono::DateTime<chrono::Utc> {
-    fn from(value: Deadline) -> Self {
-        value.0
-    }
 }
 
 impl Order {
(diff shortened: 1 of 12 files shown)


vkgnosis (approved):

MartinquaXD (approved):

crates/driver/src/domain/quote.rs
```diff
@@ -148,16 +132,66 @@ impl Order {
...
-                token: self.sell_token,
+                token: self.tokens.sell,
             },
         }
     }
 }
 
+/// The deadline for computing a quote for an order.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Deadline(chrono::DateTime<chrono::Utc>);
```
ennmichael, line 147: I like that you moved this down here :+1:

In general I try to also pay attention to this and make it so that the most "important" types come first, and the less important ones after.

crates/driver/src/infra/liquidity/fetcher.rs
```diff
@@ -24,8 +25,32 @@ impl Fetcher {
     }
 
     /// Fetches all relevant liquidity for the orders.
-    pub async fn fetch(&self, orders: &[order::Order]) -> Result<Vec<liquidity::Liquidity>, Error> {
-        let liquidity = self.inner.fetch(orders).await?;
+    pub async fn for_auction(
```
ennmichael, line 28: This is not a dealbreaker, but I'd design this slightly differently.

One thing that you're doing right is you're defining the infrastructural component `Fetcher` to work with domain objects, which is good. But in this case I think that the methods `for_auction` and `for_quote` are too specific to warrant being defined in the `Fetcher` type. `for_quote` belongs in the `quote` domain module, whereas `for_auction` probably belongs to `competition`.

For example, `Simulator` works with `eth::Tx` because it's a component which allows transactions to be simulated - it doesn't (and shouldn't) know what a settlement is.

The `Mempool` type, OTOH, is defined to work with a `settlement::Simulated` which is also a mistake from a puristic point of view, but it's a necessary evil because of the boundary integration :confused: But ideally `Mempool` should also accept an `eth::Tx` and not know about settlements.

crates/driver/src/infra/liquidity/mod.rs
```diff
@@ -10,7 +10,30 @@
...
+use {crate::domain::eth, std::cmp::Ordering};
+
 pub mod config;
 pub mod fetcher;
 
 pub use self::{config::Config, fetcher::Fetcher};
+
+/// An ordered token pair.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct TokenPair(eth::TokenAddress, eth::TokenAddress);
```
ennmichael, line 22: Following up from the previous comment - I'd personally probably move this into `domain` and then expose the `Fetcher` interface to work with this type.

nlordell: Love it! I changed `boundary::liquidity::Fetcher` to work on this type, which should have been a hint that I should have done the same thing for `infra::liquidity::Fetcher`. This way, I can move the decision of "what token pairs matter" to the domain, where it makes more sense anyway (otherwise, my `infra::liquidity::Fetcher` is deciding what matters and what doesn't for my `domain::Auction`).

crates/driver/src/infra/liquidity/fetcher.rs
nlordell: 982c0efe6a358148cb29cf396b4957090df0a74e

`infra::liquidity::Fetcher` now operates on `domain::liquidity::TokenPair`, and each individual domain module is responsible for determining which token pairs matter to it for liquidity, which is much nicer IMO (so `domain::competition` computes the set of token pairs it cares about for solving, and `domain::quote` computes the set of token pairs it cares about for quoting).

crates/driver/src/infra/liquidity/mod.rs
nlordell: 982c0efe6a358148cb29cf396b4957090df0a74e

PR #1115, opened 2023-01-20
Encode Quote Interaction Calldata

nlordell:
This PR adds boundary logic for encoding `Quote` interactions! This means that the `/quote` endpoint is actually doing work and producing real price estimates with calldata 🚀.

It uses the same strategy for encoding interactions as the simulating price estimator. Namely, it encodes approvals for **all** required allowances (and, unlike for settlements, does not check existing allowances to see which approvals are actually needed). This is a small optimization to avoid a roundtrip to the node, at the cost of gas estimation inaccuracy. I think this is fine since:
- We can offset this gas inaccuracy by configuring flat subsidy values in the `orderbook` and `autopilot`
- If an account storage slot within an Ethereum transaction starts with a non-0 value and ends at a non-0 value, then the gas cost for the `SSTORE` opcode is reduced (which is how allowances typically work for ERC20s), meaning the gas inaccuracy is not so large.

### Test Plan

We can run a manual test!

1. Run a driver and baseline solver (see #1100 for instructions)
2. Request a quote:
   ```bash
   curl -s localhost:11088/baseline/quote -H 'Content-Type: application/json' --data '@-' <<JSON | jq
   {
     "sellToken": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
     "buyToken": "0x6b175474e89094c44da98b954eedeac495271d0f",
     "kind": "sell",
     "effectiveGasPrice": "15000000000",
     "deadline": "2024-01-01T00:00:00.000Z"
   }
   JSON
   ```
   
   ```json
   {
     "amount": "1560394000750060862668",
     "interactions": [
       {
         "value": "0",
         "callData": "0x095ea7b30000000000000000000000007a250d5630b4cf539739df2c5dacb4c659f2488d0000000000000000000000000000000000000000000000000000000000000000"
       },
       {
         "value": "0",
         "callData": "0x095ea7b30000000000000000000000007a250d5630b4cf539739df2c5dacb4c659f2488dffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
       },
       {
         "target": "0x7a250d5630b4cf539739df2c5dacb4c659f2488d",
         "value": "0",
         "callData": "0x8803dbee00000000000000000000000000000000000000000000005496d11c5c45fd7ccc0000000000000000000000000000000000000000000000000e043da61725000000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab41ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f"
       }
     ]
   }
   ```
3. Check out this manually created [Tenderly simulation of for this swap](https://dashboard.tenderly.co/nlordell/scratch/simulator/a2368e76-3749-4e9f-ac92-c3a962787a98).


crates/driver/src/infra/api/quote/mod.rs
```diff
@@ -12,8 +12,8 @@ async fn quote(
     let order = order.0.into_domain(state.now()).unwrap();
     let liquidity = state.liquidity().for_quote(&order).await.unwrap();
     let quote = order
-        .quote(state.solver(), &liquidity, state.now())
+        .quote(state.eth(), state.solver(), &liquidity, state.now())
```
nlordell, line 15: It feels like `quote` has too many parameters...

I think we could group everything into a:
```
struct Quoter {
  eth: Ethereum,
  solver: Solver,
  liquidity: liquidity::Fetcher,
  now: time::Now,
}
```

Which would play a similar role as the `competition::Competition` type, and keep all quoting-specific state in a single value. WDYT? Just not sure what to call it though...

We could call it `Quote` and rename the existing `Quote` to `Swap`?

vkgnosis: I'm fine with the number of parameters. The only goal of this struct would be to collect the parameters of the Order::quote function, right? In that case the most natural name to me is QuoteParameters.

vkgnosis (approved):

crates/driver/src/infra/api/quote/mod.rs
nlordell: > The only goal of this struct would be to collect the parameters of the Order::quote function, right?

I would probably invert the call though:
```
let thing = quote::Thing { ... };
let order = quote::Order { ... };
thing.quote(&order).await;
```

crates/driver/src/boundary/quote.rs
```diff
@@ -0,0 +1,77 @@
...
+    let slippage_calculator = SlippageCalculator::from_bps(DEFAULT_QUOTE_SLIPPAGE_BPS, None);
+    let external_prices = ExternalPrices::new(eth.contracts().weth().address(), Default::default())
+        .expect("empty external prices is valid");
+    let slippage_context = slippage_calculator.context(&external_prices);
+
+    let mut settlement = solver::settlement::Settlement::new(Default::default());
+
+    for interaction in interactions {
+        let allowances = interaction.allowances();
+        for allowance in allowances {
```
MartinquaXD, line 30: Is it possible that we would issue multiple identical approval interactions here?

crates/driver/src/infra/api/quote/mod.rs
MartinquaXD: Naming that `Thing` is tricky. One option could be `QuoteProvider` or something like that but that doesn't tell you much.
Alternatively I thought `Router` might fit, although technically it's not just executing a predetermined trade route but computing it in the first place.

nlordell: I like `Router`.

crates/driver/src/boundary/quote.rs
nlordell: It is, and I think it is fine. At worst, we over-estimate the gas of the quote, but it shouldn't have an actual bearing on the actual traded amounts.

crates/driver/src/infra/api/quote/mod.rs
ennmichael: I don't think I like `Router`. Here's a little tidbit from DDD applied to the OO world:

DDD recognizes domain objects which always model abstractions, but also many DDD authors add an additional abstraction layer called the Application Layer where the services reside. Service types are types which don't really have a single representation as a domain object, but rather they represent processes. So these are things like `Quoter`, `PriceEstimator`, and all the other `-er`s that don't belong to the infrastructure layer.

My personal opinion on this in the context of Rust: I don't think that such service types are idiomatic, I think that simply using a function is better (why `quoter.quote()` when you can just say `quote(...)`?) and if you start having too many parameters (roughly the point where clippy starts warning you), try to figure out better domain objects.

ennmichael: OTOH `Competition` does model a relatively coherent abstraction: the competition process.

ennmichael: But if you insist on having a type to model this process, I think the only good name is `Quoter`.

crates/driver/src/boundary/quote.rs
ennmichael, line 30: I'm really sorry about leaving this comment but the fact that these are two lines really hurts my eyes
```suggestion
        for allowance in interaction.allowances() {
```

crates/driver/src/domain/competition/solution/interaction.rs
```diff
@@ -16,6 +16,32 @@ impl Interaction {
...
+            Interaction::Custom(interaction) => interaction.allowances.clone(),
+            Interaction::Liquidity(interaction) => {
+                let address = match &interaction.liquidity.kind {
+                    liquidity::Kind::UniswapV2(pool) => pool.router.into(),
+                    liquidity::Kind::UniswapV3(_) => todo!(),
+                    liquidity::Kind::BalancerV2Stable(_) => todo!(),
+                    liquidity::Kind::BalancerV2Weighted(_) => todo!(),
+                    liquidity::Kind::Swapr(_) => todo!(),
+                    liquidity::Kind::ZeroEx(_) => todo!(),
+                };
```
ennmichael, line 33: I would probably move this block into the `Liquidity` type (e.g. `Liquidity::allowances(&self)`), not because it's better to have shorter methods but because extensively matching on the liquidity kind belongs there more than it does in the `solution` module.

crates/driver/src/infra/api/quote/mod.rs
ennmichael: Also https://github.com/cowprotocol/services/issues/1122 can help reduce the parameters by 1 since the `Now` type is a bit ugly to begin with.

ennmichael (approved):

crates/driver/src/infra/api/quote/mod.rs
nlordell: Left it out for now.

Issue #1091, opened 2023-01-18
Better `driver` test setup


ennmichael:
Improve `dev-geth` to allow running multiple `geth` instances. The test setup would involve making a `POST` request to `dev-geth` which starts a new `geth` process at a random port, returns that port, the test uses it, and then once the test is done it can make a `DELETE` request with that path passed as a parameter, which shuts down the appropriate `geth` process. And now the tests can run in parallel. 

_Originally posted by @ennmichael in https://github.com/cowprotocol/services/pull/1077#discussion_r1072283489_

PR #1121, opened 2023-01-23
Parallelize the driver tests

ennmichael:
Fixes https://github.com/cowprotocol/services/issues/1091.

Adds the ability for `dev-geth` to run multiple instances of `geth` on different ports at the same time using docker host networking. Fixes two nasty race conditions which lead to flaky tests:

- Not waiting for a block to be mined after publishing each transaction, which could lead to two transactions getting mined in the same block, which changes the gas price in the block where the settlement is published and thereby changes the objective value causing a test to fail.
- There was also a problem where the gas price used by the settlement in the tests was not deterministic because of the way that the Uniswap V2 pool uses `block.timestamp` in the contract. This was fixed by waiting for at least 1 second at a specific point during the test setup.

Fixes the `Dockerfile.dev-geth` which incorrectly used `COPY` instead of `cp` in some places, this was missed before due to the way that `docker-compose` caching works.

Adds a base datadir for the geth so that the seeded admin account and genesis block are always the same.

### Test Plan

Automated tests.


crates/dev-geth/src/main.rs
```diff
@@ -10,96 +10,138 @@
...
-    }
+#[derive(Debug, Default)]
+struct Processes(DashMap<Port, tokio::process::Child>);
 
-    async fn spawn() -> tokio::process::Child {
-        let child = tokio::process::Command::new(dbg!(GETH_PROGRAM))
-            .args(GETH_ARGS)
+impl Processes {
+    async fn start(&self) -> Port {
+        let port = Port::free();
```
vkgnosis, line 76: This is a race condition. `Port::free` only holds onto the port while TcpListener is alive. Afterwards the port can be reused. We should hold a mutex during `Processes::start`. In practice the current code is unlikely to go wrong because linux counts the port number up and this is the only thing running in the container but would still be better completely correct.

vkgnosis: Even without this problem, I feel Mutex<HashMap> would be better than DashMap. We discussed this in another PR before. IMO Mutex is simpler than DashMap. I would only use DashMap when it is important for performance that there is no lock. This is just a nit pick and doesn't block the PR but I wanted to mention it again.

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -20,8 +20,62 @@ pub async fn primary_address(web3: &DynWeb3) -> ethcontract::H160 {
...
+        format!("http://localhost:{}", self.port)
+    }
+}
+
+impl Drop for Geth {
+    fn drop(&mut self) {
+        // Use ureq because it's blocking and doesn't rely on an async runtime.
+        ureq::delete(&format!("http://localhost:{DEV_GETH_PORT}/{}", self.port))
+            .call()
+            .unwrap();
```
vkgnosis, line 41: Instead of using a separate http crate just to make a blocking request, a better solution is something like
```rust
tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap().block_on(async { ... });
```
I haven't tested it, but should work.

nlordell: reqwest has blocking requests which can be enabled with a feature flag. I think we already use `reqwest` elsewhere (although - it may spin up a Tokio runtime which may be what you are trying to avoid).

docker/base-datadir/geth/chaindata/CURRENT
```diff
@@ -0,0 +1 @@
+MANIFEST-000003
```
nlordell, line 1: the `base-datadir` should show up in the `.gitignore` right?

crates/dev-geth/src/main.rs
ennmichael: > This is a race condition.

Point taken - I was aware of this but figured it's not worth worrying about since this is just testing setup and the chance that this causes a problem is really low.

> We should hold a mutex during Processes::start

Holding a mutex during `Process::start` is a bottleneck because now you can't start another geth instance until the previous one has been bound and responsive and this can take a few seconds. So now instead of having the setup for four tests take a few seconds, the setup takes four * a few seconds, so you lose the parallelism.

I even thought that maybe better than trying to bind to `0.0.0.0:0` would be to just generate a random port and hope that it's free and accept the chance that it isn't.

The only "perfect" solution is to start geth with `0.0.0.0:0` and parse the logging output to try to figure out the port, but obviously parsing logging output is extremely sketchy and prone to error later down the line if the format changes.

> linux counts the port number up

I'm not sure, it seems to just be handing out random ports when I run it :shrug:

docker/base-datadir/geth/chaindata/CURRENT
ennmichael: I don't think so? Why do you think this directory should be ignored? It's required to build `Dockerfile.dev-geth`.

crates/driver/src/tests/setup/blockchain/mod.rs
ennmichael: Both of your solutions lead to this:

```
thread 'tests::cases::solve::test' panicked at 'Cannot start a runtime from within a runtime. This happens because a function
 (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', crates
/driver/src/tests/setup/blockchain/mod.rs:43:14
```

[`async` and `Drop` don' mix?](https://www.youtube.com/watch?v=sR09sACUzFU)

docker/base-datadir/geth/chaindata/CURRENT
nlordell: Oh shoot:
> The base datadir used to ensure that the genesis block is always the same.

Should have reviewed before mindless-ly posting.

Small suggestion - you should be able to put these files into a single `base-datadir.tar.xz` file [as Docker `ADD` will automatically extract tar archives](https://docs.docker.com/engine/reference/builder/#add).

> If `<src>` is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory.

This way, the "base-datadir" will show up as a single file in Git.

crates/driver/src/tests/setup/blockchain/mod.rs
vkgnosis: Oh, right this Drop happens in async code. So it shouldn't be blocking inside of drop anyway. In that case we can do this:
```rust
tokio::task::spawn(async { issue_delete_request_through_reqwest().await; });
```
This means t drop will return before the call finishes, which is fine.

crates/dev-geth/src/main.rs
vkgnosis: If starting geth takes a while until it actually binds to the port then I agree that we shouldn't mutex there. In that case let's add a comment to this code mentioning the race condition and that it is unlikely to happen.

Maybe another reasonable solution is that you keep track of port numbers yourself. If you know that the docker container only runs this rust program then you can have a global atomic int next_port  that keeps track of the next port you want to give out.

ennmichael: > Maybe another reasonable solution is that you keep track of port numbers yourself. If you know that the docker container only runs this rust program then you can have a global atomic int next_port that keeps track of the next port you want to give out

That's a good idea, I could allocate a fixed range of ports since this is the only program running in the container.

crates/driver/src/tests/setup/blockchain/mod.rs
ennmichael: > which is fine

This also occurred to me but I was unsure if it works - is the runtime going to wait for this task?

docker/base-datadir/geth/chaindata/CURRENT
ennmichael: Nice!

crates/dev-geth/src/main.rs
ennmichael: https://github.com/cowprotocol/services/pull/1121/commits/ea95df09eed6a3ffc4161d776351b2132bf45874

crates/driver/src/tests/setup/blockchain/mod.rs
vkgnosis: From the top of my head I don't know if `#[tokio::main]` or `#[tokio::test]` wait for spawned tasks. That should be documented in the tokio docs.

ennmichael: https://github.com/cowprotocol/services/pull/1121/commits/d1f6804c11466da7670f2142afceeb004a51c049

docker/base-datadir/geth/chaindata/CURRENT
ennmichael: https://github.com/cowprotocol/services/pull/1121/commits/6a5903b698e0c1cf6c90ceec2aa03a6e50394d54

vkgnosis (approved):

crates/driver/src/tests/cases/solve.rs
```diff
@@ -190,5 +192,5 @@ async fn test() {
     assert_eq!(result.as_object().unwrap().len(), 2);
     assert!(result.get("id").is_some());
     assert!(result.get("score").is_some());
-    assert_eq!(result.get("score").unwrap(), -79291602683462.0);
+    assert_eq!(result.get("score").unwrap(), -94756205354950.0);
```
nlordell, line 195: Interesting that this changed.

nlordell (approved):
LGTM.

nlordell:
This is so cool!

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -20,8 +20,67 @@ pub async fn primary_address(web3: &DynWeb3) -> ethcontract::H160 {
...
+pub async fn wait_for<T>(web3: &DynWeb3, fut: impl Future<Output = T>) -> T {
+    let block = web3.eth().block_number().await.unwrap();
+    let result = fut.await;
+    tokio::time::timeout(std::time::Duration::from_secs(15), async {
+        loop {
+            let next_block = web3.eth().block_number().await.unwrap();
+            if next_block > block {
+                break;
+            }
+            tokio::time::sleep(std::time::Duration::from_millis(300)).await;
```
MartinquaXD, line 80: Having a timeout as small as possible would be great here. Did you test if there is a point of diminishing returns?
i.e. does it usually take at least 300ms to mine a new block?

crates/driver/src/tests/setup/blockchain/uniswap.rs
```diff
@@ -48,88 +48,123 @@ pub async fn setup() -> Uniswap {
...
-    web3.eth()
-        .send_transaction(web3::types::TransactionRequest {
-            from: super::primary_address(&web3).await,
-            to: Some(admin),
-            value: Some(balance / 2),
-            ..Default::default()
-        })
-        .await
-        .unwrap();
+    super::wait_for(
```
MartinquaXD, line 51: Wrapping every `send_tx` in a `super::wair_for` seems a bit tedious and indents every call one level to the right.
What do you think about using an extension trait for those calls to be able to do something like 👇 ?
```rust
web3.eth().send_transaction(Default::default()).wait_for_block(&web3).await;
```

MartinquaXD (approved):
Code looks code. Just a few minor suggestions.

crates/driver/src/tests/setup/blockchain/uniswap.rs
nlordell: I just remembered that `ethcontract` supports transaction confirmations. You could write for example:
```rust
        authenticator
            .initialize_manager(admin)
            .from(admin_account.clone())
            .with_confirmations(0)
            .send(),
```

crates/dev-geth/src/main.rs
```diff
@@ -10,96 +10,138 @@
...
+    dashmap::DashMap,
     std::{process::Stdio, sync::Arc},
-    tokio::sync::Mutex,
 };
 
 const PORT: &str = "8547";
-const GETH_PORT: &str = "8545";
 const GETH_PROGRAM: &str = "geth";
-const GETH_ARGS: [&str; 8] = [
+const GETH_ARGS: [&str; 6] = [
```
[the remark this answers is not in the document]
ennmichael: https://github.com/cowprotocol/services/pull/1121/commits/9b1e9fc3c4447d05931cbf1108267c8b2f74397a

crates/driver/src/tests/setup/blockchain/mod.rs
ennmichael: https://github.com/cowprotocol/services/pull/1121/commits/318497ffed2fd8dadc6415424a1f4864780cb4c3

PR #1124, opened 2023-01-24
Use config file instead of CLI args for `driver`

ennmichael:
Instead of CLI args, load configuration from the TOML config file. This is a prerequisite for deploying this in our infrastructure, which is the next task on my TODO list.

### Test Plan

Automated tests continue to pass.


crates/driver/src/tests/setup/driver.rs
```diff
@@ -162,39 +149,60 @@ pub async fn setup(config: Config<'_>) -> Client {
...
+    contracts: &ContractsConfig,
+) {
+    #[rustfmt::skip]
+    let contracts_config = format!(
+        "[contracts]\n\
+        gp-v2-settlement = \"0x{}\"\n\
+        weth = \"0x{}\"",
+        hex::encode(contracts.gp_v2_settlement.unwrap().as_bytes()),
+        hex::encode(contracts.weth.unwrap().as_bytes()),
+    );
```
vkgnosis, line 171:
```suggestion
    let contracts_config = format!(
        r#"
        [contracts]
        gp-v2-settlement = "0x{}"
        weth = "0x{}"
        "#,
        hex::encode(contracts.gp_v2_settlement.unwrap().as_bytes()),
        hex::encode(contracts.weth.unwrap().as_bytes()),
    );
```
This is a bit nicer. You can see the config file like you would when editing it yourself because there is no need for escaping. Also doesn't need rustfmt-skip.

Could create the full config this way too instead of joining the strings afterwards.

vkgnosis (approved):

crates/driver/src/domain/eth/mod.rs
```diff
@@ -170,6 +170,24 @@ impl From<Address> for H160 {
     }
 }
 
+// TODO This type should be more complex and ensure that the private key is
+// valid, use secp256k1?
+/// A private key that can be used to sign transactions.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct PrivateKey(pub H256);
```
nlordell, line 177: Why not use `ethcontract::PrivateKey`?

crates/driver/example.toml
```diff
@@ -3,13 +3,19 @@ name = "mysolver" # Arbitrary name given to this solver, must be unique
 endpoint = "http://0.0.0.0:33219/solve"
 absolute-slippage = "12" # Denominated in wei, optional
 relative-slippage = "0.1" # Percentage in the [0, 1] range
-address = "0xE33bff104C4E1140A3f86f855AD5bDAb8Dc6CA8C" # The ETH address of this solver
+address = "0x2cf0a2Fef084b6DB3fb3161c8f2b6B61bcE2F73c" # The ETH address of this solver
+private-key = "0xaa1de59084f3b7e501f2d48cafb2ee04cd06a79ea126fd27ddf3d1b8903bb85a" # The private key of the solver
```
nlordell, line 7: From a configuration standpoint, is there a reason we need to specify both address and private key? I would suggest:
- Only specifying the `address` for use with a node that manages your PK (for example, what we would use in CI) or for "dry-run" mode
- Only specify a `private-key` if you want to actually sign transactions.

crates/driver/src/tests/setup/driver.rs
nlordell: Note that these aren't equivalent - the `r#` notation keeps leading white-space in the string (don't know if is an issue with TOML).

nlordell (approved):
LGTM.

crates/driver/src/tests/setup/driver.rs
vkgnosis: That's true. I assumed TOML doesn't care about whitespace.

nlordell: > TOML doesn't care about whitespace.

It might be the case, just wanted to point it out. If it doesn't your suggestion definitely looks a lot nicer!

crates/driver/example.toml
ennmichael: This is correct.

crates/driver/src/tests/setup/driver.rs
ennmichael: https://github.com/cowprotocol/services/pull/1124/commits/7ebee4f0f543572dba7c785e43ef967a01bab393

crates/driver/src/domain/eth/mod.rs
ennmichael: Good point https://github.com/cowprotocol/services/pull/1124/commits/fc33c8cb1bc4b8dc8a5ada1cafdccdce5958caf5

PR #1126, opened 2023-01-25
Make sure geth is killed in driver test

vkgnosis:
This is a follow up to a discussion in a previous PR. I carefully checked the tokio docs and code. Tokio does not wait for spawned tasks to complete when the runtime is dropped (when a tokio test returns). In order to ensure that the geth process really is killed, we have to wait for it.

The most correct way to implement this would be to not use Drop, but make tests manually kill geth. That's error prone or requires more test setup (like a helper function that does it for you) so I went with blocking in Drop.

Also changed the order of process killing and directory deletion because it ensures the geth process won't try to use the directory while it is being deleted.

### Test Plan

ran geth tests

By the way, for people who don't want to use docker-compose, the commands to run the container are:

```sh
cd services
podman build . -f docker/Dockerfile.dev-geth -t dev-geth
podman run --interactive --tty --network=host dev-geth
```


nlordell (approved):
🕵️

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -33,17 +33,30 @@ impl Geth {
     }
 }
 
+// What we really want here is "AsyncDrop", which is an unsolved problem in the
+// async ecosystem. As a workaround we create a new runtime so that we can block
+// on the delete request. Spawning a task for this isn't enough because tokio
+// runtimes when they exit drop background tasks, like when a #[tokio::test]
+// function returns.
```
ennmichael, line 40: I think this comment makes sense in the context of the PR but contains a little too much information on its own.

```suggestion
/// What we really want here is "AsyncDrop", which is an unsolved problem in the
/// async ecosystem. As a workaround we create a new runtime so that we can block
/// on the delete request.
```

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -33,17 +33,30 @@ impl Geth {
...
                 .await
                 .unwrap();
-        });
+        };
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap();
+        // block_on call must happen in a new thread because tokio forbids nesting
+        // runtimes.
```
ennmichael, line 57: Nit, sorry
```suggestion
        // block_on must be called in a new thread because tokio forbids nesting
        // runtimes.
```

ennmichael (commented):
At this point just using `ureq` would be vastly simpler :smile: Creating a runtime, spawning a new thread so you can call `block_on`... it's a lot.

The dependency isn't a big deal since you can make it a `dev-dependency` and it's only used it test code.

vkgnosis:
I don't agree with "vastly simpler". Ureq is simpler in two ways:
1. In this call site, there is just one ureq call.
2. The implementation details of Ureq are simpler than those of tokio + reqwest.

1 can be made as simple by moving the runtime setup into a helper that hides the details.

2 Is true. But it isn't true when considering that we already use tokio + reqwest for http requests everywhere.

 It feels better to me to have these hacky 3 lines of code than to have another http library just for making blocking requests. Doing blocking Drop in async code is itself a hack too.

Another thing to consider is that the most correct way of approaching this is to not rely on drop running in the first place. This is applies to async and sync code. There are cases in which Drop doesn't run. We could scrap this whole thing in favor of making the geth manager auto detect when a node instance is no longer needed. But that's kind of complicated too.

It's not terrible to keep the current `tokio::task::spawn` solution and hoping that it usually runs fast enough before the test exits but this change still feels worthwhile.

crates/driver/src/tests/setup/blockchain/mod.rs
vkgnosis: The reason I included that extra sentence is that I feel someone reading this later is going to be confused why we not just spawn a task. They could find the original commit that made the change but that can be tedious. I'm fine with either wording.

PR #1128, opened 2023-01-25
Driver API error handling

ennmichael:
Properly report errors in the API layer.

I'm just returning 400 errors since I'm not sure it makes sense to differentiate here. E.g. `BAD_REQUEST` doesn't really make sense as a response when liquidity fetching fails, but what is? Feel free to propose improvements.

We should strive to minimize the usage of `Type::Unknown` in the future but I think this is OK for a start.

### Test Plan

Will add automated tests in a follow-up.


crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+        domain::competition::{self, quote, solution},
+        infra::api,
+    },
+    hyper::StatusCode,
+    serde::Serialize,
+};
+
+#[derive(Debug, Clone, Copy, Serialize)]
+#[serde(rename_all = "PascalCase")]
+enum Type {
```
nlordell, line 12: Should this be called `Kind` to be more Rust-y?

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+    TransactionPublishingFailed,
+    InvalidAuctionId,
+    MissingSurplusFee,
+    LiquidityError,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Error {
+    error_type: Type,
```
nlordell, line 28: nit: `error.error_type` feels redundant, I would suggest:
```suggestion
    kind: Type,
```

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            status_code,
+            axum::Json(Error {
+                error_type: value,
+                description,
+            }),
+        )
+    }
+}
+
+impl From<quote::Error> for (StatusCode, axum::Json<Error>) {
```
nlordell, line 68: I like that the conversions are all here in this module 🎉

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Error {
+    error_type: Type,
+    description: &'static str,
+}
+
+impl From<Type> for (StatusCode, axum::Json<Error>) {
+    fn from(value: Type) -> Self {
+        let (description, status_code) = match value {
```
nlordell, line 34: I don't agree with all the HTTP codes here.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            Type::QuotingFailed => ("No valid quote found", StatusCode::BAD_REQUEST),
```
nlordell, line 35: To me, this is more that the server couldn't find a quote, not that the request made by the client was bad, I think a 404 is more appropriate:
```suggestion
            Type::QuotingFailed => ("No valid quote found", StatusCode::NOT_FOUND),
```

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            Type::SolverFailed => (
+                "Solver engine returned an invalid response",
+                StatusCode::BAD_REQUEST,
```
nlordell, line 38: This doesn't seem like a client error (so it should not be a 4xx error code). I think the most appropriate error would be a 502:

> 502 Bad Gateway
>    The server was acting as a gateway or proxy and received an invalid response from the upstream server.

Rationale being the driver is a "gateway" to a solver engine.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            ),
+            Type::SolutionNotFound => ("No solution found for given ID", StatusCode::BAD_REQUEST),
```
nlordell, line 40: 400 bad request kind of makes sense here, as the request being made was incorrect (not for a valid solution), but from just from the error description, it looks more like a 404 not found error.

Don't feel strongly about this one.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            Type::DeadlineExceeded => ("Exceeded solution deadline", StatusCode::BAD_REQUEST),
+            Type::SimulationFailed => ("Solution simulation failed", StatusCode::BAD_REQUEST),
+            Type::Unknown => ("An unknown error occurred", StatusCode::BAD_REQUEST),
```
nlordell, line 43: Unknown errors sound more like 500 errors:
> 500 Internal Server Error
>     A generic error message, given when an unexpected condition was encountered and no more specific message is suitable.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+            ),
+            Type::InvalidAuctionId => (
+                "Invalid ID specified in the auction",
+                StatusCode::BAD_REQUEST,
+            ),
+            Type::MissingSurplusFee => (
+                "Auction contains a limit order with no surplus fee",
+                StatusCode::BAD_REQUEST,
+            ),
+            Type::LiquidityError => ("Failed to fetch onchain liquidity", StatusCode::BAD_REQUEST),
```
nlordell, line 56: I think this should be a server error of some kind (5xx error). I don't know which one though. Either a generic 500 error, or a 503 (temporarily unavailable because it can't fetch liquidity).

I would probably stick with 500 for simplicity, or use a non-standard 5xx error code that we can make up.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+                StatusCode::BAD_REQUEST,
+            ),
+            Type::SolutionNotFound => ("No solution found for given ID", StatusCode::BAD_REQUEST),
+            Type::DeadlineExceeded => ("Exceeded solution deadline", StatusCode::BAD_REQUEST),
+            Type::SimulationFailed => ("Solution simulation failed", StatusCode::BAD_REQUEST),
+            Type::Unknown => ("An unknown error occurred", StatusCode::BAD_REQUEST),
+            Type::TransactionPublishingFailed => (
+                "Failed to publish the settlement transaction",
+                StatusCode::BAD_REQUEST,
+            ),
```
nlordell, line 47: Like for `LiquidityError` error, this doesn't sound like a 4xx client error, but a 5xx server error.

nlordell (commented):
LGTM, just some nits on the chosen error codes.

vkgnosis:
I don't want to bikeshed the http error codes like this. I wrote about this in the PR where I introduced the autopilot<->solver openapi.yml. Http status codes were not meant for applications like this and it is not useful to spend time thinking about how we want to divide errors across status codes. This has no practical benefit to us.

The spec as I wrote it only differentiates between error 400 for client errors (not providing valid json, not upholding invariants of an api) and error 500 for server error (server can't handle the request like if it can't talk to ethereum node). I definitely would not do it more fine grained than this.

Now I kind of want to get rid of this distinction too because I see it still causes discussion. We don't need http status codes to carry extra information. Our api level is defined on the json level, not the http level. The only use of the status code is to indicate "this is the ok response" versus "this is the err response" so it can be deserialized differently. Or, we could even always return 200 and have the caller deserialize into Enum to figure out whether an error happened. The only way a non 200 should show up then, is if some intermediary software like nginx created the error.

crates/driver/src/infra/api/error.rs
```diff
@@ -0,0 +1,104 @@
...
+impl From<Type> for (StatusCode, axum::Json<Error>) {
+    fn from(value: Type) -> Self {
+        let (description, status_code) = match value {
+            Type::QuotingFailed => ("No valid quote found", StatusCode::BAD_REQUEST),
+            Type::SolverFailed => (
+                "Solver engine returned an invalid response",
+                StatusCode::BAD_REQUEST,
+            ),
+            Type::SolutionNotFound => ("No solution found for given ID", StatusCode::BAD_REQUEST),
+            Type::DeadlineExceeded => ("Exceeded solution deadline", StatusCode::BAD_REQUEST),
```
MartinquaXD, line 41: Since we have multiple different deadlines (quoting, solving, execution) having a more generic `Deadline exceeded` error message seems fine to me.

crates/driver/src/infra/api/error.rs
MartinquaXD: I'm not sure the `driver` should even error out due to an error while fetching liquidity. Technically it could be able to find a solution without fetching any additional liquidity (liquidity orders in the auction, perfect CoW, market making).
Also would that error already occur when a single liquidity fetch request fails or only when all of them fail?

MartinquaXD (approved):
I agree with Valentin that bike shedding the status codes probably doesn't add much value to the API. As long as the error messages are helpful I'm happy.

crates/driver/src/infra/api/error.rs
nlordell: That is a good point. I think it would make sense for liquidity fetching to never error and instead just return "empty liquidity" in case of an error and just print a `WARN` message. Even Quasimodo, which relies on liquidity, would be able to find CoWs in the absence of any pools.

nlordell:
It feels like the general sentiment is that we shouldn't try hard to find HTTP status codes. Maybe as a compromise we can differentiate input errors (bad JSON for example), as `400` and server errors (a Solver error for example) as `500` as @vkgnosis suggested.

ennmichael:
I agree with @vkgnosis that even differentiating between 400/500 is not very valuable TBH.

@nlordell The reason why I used the name `error_type` is because that's the format used for errors in the `orderbook`. Do you still think we should change it? I would prefer to call it `kind` as well.

nlordell:
> Do you still think we should change it?

To me, it is a different API, I don't think we need to keep it as `error_type`.

> I agree with @vkgnosis that even differentiating between 400/500 is not very valuable TBH.

In this case, I think we should use 500 for everything, then we could even get rid of specifying the status code in conversion trait (since I would even expect axum to default to 500 status code on `Result::Err`).

crates/driver/src/infra/api/error.rs
ennmichael: https://github.com/cowprotocol/services/issues/1130

crates/driver/src/infra/api/error.rs
ennmichael: https://github.com/cowprotocol/services/pull/1128/commits/abda069a433854826d5d895ec5ddfbc901532527

crates/driver/src/infra/api/error.rs
ennmichael: https://github.com/cowprotocol/services/pull/1128/commits/abda069a433854826d5d895ec5ddfbc901532527

vkgnosis (approved):

vkgnosis:
> Do you still think we should change it?

This is a new API so we can change it how we want to. I'm fine with either name. Would be nice to update the openapi doc with this too.

nlordell (approved):

PR #1129, opened 2023-01-25
Add metric to count onchain orders

fedgiac:
In partial fulfillment of https://github.com/cowprotocol/services/issues/1033.

Adds a metric that counts orders created from onchain transactions that is robust to reorgs if excluding pod restarts.
The metric labels mirror the ones that are already available in the orderbook for new orders.

Note that no metric currently tracks invalidated order (despite it being already initialized to zero).

### Test Plan

I don't know how to properly test this code as I'd need to trigger a reorg. I tested it by running the autopilot, creating some orders, and seeing the metrics increase from zero with the expected labels.

### DIFF
--- a/crates/autopilot/src/database/onchain_order_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events.rs
@@ -1,6 +1,8 @@
 pub mod ethflow_events;
 pub mod event_retriever;
+mod metrics;
 
+use self::metrics::{get_most_recent_block, Metrics, OrdersByBlockNumber};
 use super::{
     events::{bytes_to_order_uid, meta_to_event_index},
     Metrics as DatabaseMetrics, Postgres,
@@ -56,6 +58,10 @@ pub struct OnchainOrderParser<EventData: Send + Sync, EventRow: Send + Sync> {
     settlement_contract: H160,
     liquidity_order_owners: HashSet<H160>,
     metrics: &'static Metrics,
+    /// A helper to keep track of metrics changes on reorg. It collects orders
+    /// that were already considered in the metrics so that they can be deleted
+    /// in case a reorg occurs.
+    revertible_orders: OrdersByBlockNumber,
 }
 
 impl<EventData, EventRow> OnchainOrderParser<EventData, EventRow>
@@ -72,6 +78,8 @@ where
         settlement_contract: H160,
         liquidity_order_owners: HashSet<H160>,
     ) -> Self {
+        let metrics = Metrics::get();
+        metrics.init();
         OnchainOrderParser {
             db,
             web3,
@@ -80,7 +88,8 @@ where
             domain_separator,
             settlement_contract,
             liquidity_order_owners,
-            metrics: Metrics::get(),
+            metrics,
+            revertible_orders: Default::default(),
         }
     }
 }
@@ -155,9 +164,17 @@ impl<T: Sync + Send + Clone, W: Sync + Send + Clone> EventStoring<ContractEvent>
             .collect();
         let invalidation_events = get_invalidation_events(events)?;
         let invalided_order_uids = extract_invalidated_order_uids(invalidation_events)?;
-        let (custom_onchain_data, quotes, broadcasted_order_data, orders) = self
+        let onchain_order_data = self
             .extract_custom_and_general_order_data(order_placement_events)
             .await?;
+        let new_orders_for_metrics =
+            OrdersByBlockNumber::prepare_created_orders(&onchain_order_data);
+
+        let (custom_onchain_data, quotes, broadcasted_order_data, orders) =
+            OnchainOrderData::unzip(onchain_order_data);
+
+        let most_recent_block =
+            get_most_recent_block(&invalided_order_uids, &broadcasted_order_data);
 
         let _timer = DatabaseMetrics::get()
             .database_queries
@@ -213,6 +230,15 @@ impl<T: Sync + Send + Clone, W: Sync + Send + Clone> EventStoring<ContractEvent>
             .await
             .context("insert_orders failed")?;
         transaction.commit().await.context("commit")?;
+
+        self.revertible_orders
+            .remove_orders_in_range(range, self.metrics);
+        self.revertible_orders
+            .add_orders(new_orders_for_metrics, self.metrics);
+        if let Some(most_recent_block) = most_recent_block {
+            self.revertible_orders.purge_old_orders(most_recent_block);
+        }
+
         Ok(())
     }
 
@@ -226,9 +252,17 @@ impl<T: Sync + Send + Clone, W: Sync + Send + Clone> EventStoring<ContractEvent>
             .collect();
         let invalidation_events = get_invalidation_events(events)?;
         let invalided_order_uids = extract_invalidated_order_uids(invalidation_events)?;
-        let (custom_order_data, quotes, broadcasted_order_data, orders) = self
+        let onchain_order_data = self
             .extract_custom_and_general_order_data(order_placement_events)
             .await?;
+        let new_orders_for_metrics =
+            OrdersByBlockNumber::prepare_created_orders(&onchain_order_data);
+
+        let (custom_onchain_data, quotes, broadcasted_order_data, orders) =
+            OnchainOrderData::unzip(onchain_order_data);
+
+        let most_recent_block =
+            get_most_recent_block(&invalided_order_uids, &broadcasted_order_data);
 
         let _timer = DatabaseMetrics::get()
             .database_queries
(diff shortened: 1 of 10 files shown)


vkgnosis:
I don't understand the ethflow code in detail, but I'm getting a bad feeling based on how much state there is for keeping track of metrics. Did you consider alternative approaches? I feel like we're solving the wrong problem.

The issue says "this is causing our order creation metrics to report incorrect data". What does that mean? How does the data become incorrect? I *think* that issue is referring to the "matched user orders" metric going over 100%. That is a problem with that graph. It is a rough approximation. It made some sense when we only had off chain orders with low duration. With long lasting limit orders, this graph can also go over 100% and lost a lot of meaning. I don't want to have complicated metrics code for the sake of this graph. The correct solution there is probably to measure "matched user *market* orders" only, rather than making the graph work with all orders.

Maybe we *should* have separate metrics for onchain and offchain orders. Or if we want to measure the total orders in the system, maybe that is better accomplished with a metric created by a database query. That gives us the true number for certain and is stateless.

We can still have individual metrics for "number of times the API created an order" and "number of times the onchain order finder created an order" if that is useful.

fedgiac:
The reason why there is so much state is that the metric has to be decreased in the case of reorgs. Something that would eliminate the need for state would be:
1. Adding orders multiple times in case a reorg happens (meaning that related metrics will be unreliable);
2. Querying the database every time we create/replace an onchain order to check which orders already exist in the database.

> The issue says "this is causing our order creation metrics to report incorrect data". What does that mean? How does the data become incorrect? I think that issue is referring to the "matched user orders" metric going over 100%.

This is correct.

> The correct solution there is probably to measure "matched user market orders" only, rather than making the graph work with all orders

Eth-flow orders are (for the most part) market orders, so I don't think that only considering market orders would actually change the complexity of this PR (it would just remove at most two parameters out of three from the hash map key).

> We can still have individual metrics for "number of times the API created an order" and "number of times the onchain order finder created an order" if that is useful.

This code is still needed to create the second metric. (Naturally I might be missing something and there could easier ways to do it.) So it's true that some of the complexity is related to merging the order creation metrics together, but I'd say not that much extra.

vkgnosis:
> * Adding orders multiple times in case a reorg happens (meaning that related metrics will be unreliable);

This could be fine. It really depends on what we want to measure. Or you could have two counters, one for added orders, including reorgs and one for deleted orders because of reorgs. Take the difference to get without reorgs. But don't know if that information is easily available.

Another idea is reporting onchain orders only after the reorg limit so that we don't have to deal with reorgs.

> * Querying the database every time we create/replace an onchain order to check which orders already exist in the database.

The metric could be changed to reporting the result of `SELECT COUNT(*) FROM ORDERS` every 10 seconds. This is what I meant with a stateless approach.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -56,6 +58,10 @@ pub struct OnchainOrderParser<EventData: Send + Sync, EventRow: Send + Sync> {
     settlement_contract: H160,
     liquidity_order_owners: HashSet<H160>,
     metrics: &'static Metrics,
+    /// A helper to keep track of metrics changes on reorg. It collects orders
+    /// that were already considered in the metrics so that they can be deleted
+    /// in case a reorg occurs.
+    revertible_orders: OrdersByBlockNumber,
 }
 
 impl<EventData, EventRow> OnchainOrderParser<EventData, EventRow>
```
ennmichael, line 67: The design of `OnchainOrderParser` is weird.

There's an `OnchainOrderParsing` trait which is not implemented for `OnchainOrderParser`, but the `OnchainOrderParser` does wrap a `Box<dyn OnchainOrderParsing>`. The comment for `OnchainOrderParsing` (which is not a doc comment) reads:

```
// The following trait allows to implement custom onchain order parsing for
// differently placed orders. E.g., there will be a implementation for ethflow
// and presign orders. For each of the customs types, the trait allows to
// implement parsing the on-chain data and storing the event data
```

However the trait is in fact only implemented for your `EthFlowOnchainOrderParser` type and that is the single implementation of it. :question: 

It's also odd that this module is a submodule of `database` when it seems to be doing _a lot_ more than database access.

crates/autopilot/src/database/onchain_order_events/metrics.rs
```diff
@@ -0,0 +1,139 @@
...
+        let class = db_order_class_label(class);
+        let op = operation_label(op);
+        self.orders.with_label_values(&[class, op]).add(amount);
+    }
+}
+
+type OrderDetailsForMetrics = (u64, OrderClass, OrderOperation);
+pub type OrderDetailsToOrderUids = HashMap<OrderDetailsForMetrics, HashSet<OrderUid>>;
+#[derive(Debug, Default)]
+pub struct OrdersByBlockNumber(OrderDetailsToOrderUids);
```
ennmichael, line 63: Would love to see a doc comment justifying the name `OrdersByBlockNumber`.

crates/shared/src/current_block.rs
```diff
@@ -298,4 +302,15 @@ mod tests {
         assert_eq!(blocks.last().unwrap().0, 5);
         assert_eq!(blocks.first().unwrap().0, 0);
     }
+
+    #[test]
+    fn range_contains() {
+        let range = RangeInclusive::try_new(42, 1337).unwrap();
```
ennmichael, line 308: Why not use https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html?

crates/shared/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub mod network;
 pub mod oneinch_api;
 pub mod order_quoting;
 pub mod order_validation;
+pub mod orderbook_metrics;
```
ennmichael, line 34: :+1:

crates/autopilot/src/database/onchain_order_events.rs
josojo: >However the trait is in fact only implemented for your EthFlowOnchainOrderParser type and that is the single implementation of it. ❓

We created that general concept on the [OnchainOrder defined via this interface ](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapOnchainOrders.sol) that can be inherited by any contract. Yes, currently there is only one implementation, but we intended to add other ones for each contract.

>It's also odd that this module is a submodule of database when it seems to be doing a lot more than database access.

yeah, you are right about this aspect. We should give it its own module.

crates/autopilot/src/database/onchain_order_events/metrics.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl OrdersByBlockNumber {
```
josojo, line 65: Do you think the following way would be less complex and easier to maintain? 
run the following  SQL query 
`select count(*), orderclass from onchain_orders oo inner join orders o o.uid = oo.uid group by orderclass`
before a reorg and after a reorg and then simply update the counter by the difference. I hope this would be simpler code with less maintenance.

crates/autopilot/src/database/onchain_order_events.rs
fedgiac: The idea is that all ochain orders are created by emitting an EVM event of a standard format ([this](https://github.com/cowprotocol/ethflowcontract/blob/9c74c8ba36ff9ff3e255172b02454f831c066865/src/interfaces/ICoWSwapOnchainOrders.sol#L39-L44)). This is not specific to eth-flow orders, but indeed eth-flow orders are, right now, the only contract that makes use of that.

When reading an order from the blockchain, however, there might be the need for reading data that is specific to a particular implementation (from [this field](https://github.com/cowprotocol/ethflowcontract/blob/9c74c8ba36ff9ff3e255172b02454f831c066865/src/interfaces/ICoWSwapOnchainOrders.sol#L43)). How the data is parsed changes based on what implements onchain orders, and what we expect from this parsing is specified by `OnchainOrderParsing`. This is also why only  `EthFlowOnchainOrderParser` implements it, since only eth-flow orders are creating onchain orders right now.

I agree the naming is confusing. Maybe `OnchainOrderParsing` should be `OnchainOrderDataParsing`?

crates/shared/src/current_block.rs
fedgiac: I don't know, I see the event parsing was handled by @sunce86 so he might have more details. The only practical difference I see is that the custom implementation cannot be created if `start >= end`.

fedgiac:
Considering the feedback I received so far, I'll be trying to get the information I need from the database rather than keeping extra state. I'll create another PR for comparison. You might want to wait before reviewing this PR further.

fedgiac:
We will be changing how eth-flow orders are handled and orders are counted. This task will be handled there.

PR #1131, opened 2023-01-26
Continue autopilot competition implementation

vkgnosis:
I'm making some PRs focused on getting something actually runnable without perfect correctness. This PR implements the main competition logic in autopilot. Next one will like contain the finding of the transaction.

### Test Plan

not yet

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -67,7 +67,7 @@ impl RunLoop {
             {
                 Ok(()) => tracing::info!("settled"),
                 Err(err) => {
-                    tracing::error!(?err, "solver {} failed to execute", err);
+                    tracing::error!(?err, "solver {index} failed to execute");
                 }
             }
         }
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -1,11 +1,15 @@
 use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
 use model::auction::{Auction, AuctionId};
+use rand::seq::SliceRandom;
 use shared::current_block::CurrentBlockStream;
 use std::time::Duration;
+use tracing::Instrument;
 
 use crate::{
-    database::Postgres, driver_api::Driver, driver_model::solve,
+    database::Postgres,
+    driver_api::Driver,
+    driver_model::{execute, solve},
     solvable_orders::SolvableOrdersCache,
 };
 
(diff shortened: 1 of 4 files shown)


crates/autopilot/src/run_loop.rs
```diff
@@ -21,57 +25,105 @@ pub struct RunLoop {
...
+                Ok(result) => Some((index, result)),
+                Err(err) => {
+                    tracing::warn!(?err, "driver {} solve error", err);
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Execute the solver's solution. Returns Ok when the corresponding transaction has been mined.
```
nlordell, line 116: I think we should split `execute` and `wait` steps. Since, in the design we specify that solvers are allowed to reject execution (in case of price movement, meaning they know that their proposed solution wouldn't work). Specifically, I think the loop would look like:

```rust
let execution_promise = loop {
    let solution = match solution.pop() {
        Some(value) => value,
        None => break None,
     };
     
    match self
        .execute(auction, id, &self.drivers[index], &solution)
        .await
    {
        Ok(promise) => break Some(promise),
        Err(err) => {
            tracing::warn!(?err, "solver {} failed to execute", err);
            continue;
        }
    }
};
```

nlordell (approved):
LGTM. One comment about the run loop, but OK to not incorporate in this PR.

crates/autopilot/src/run_loop.rs
```diff
@@ -21,57 +25,105 @@ pub struct RunLoop {
...
+        // TODO: Keep going with other solutions until some deadline.
+        if let Some((index, solution)) = solutions.pop() {
+            tracing::info!("executing with solver {}", index);
+            match self
+                .execute(auction, id, &self.drivers[index], &solution)
+                .await
+            {
+                Ok(()) => tracing::info!("settled"),
+                Err(err) => {
+                    tracing::error!(?err, "solver {} failed to execute", err);
```
nlordell, line 70: I imagine you meant something more like this? Otherwise error is duplicated.
```suggestion
                    tracing::error!(?err, "solver {index} failed to execute");
```

crates/autopilot/src/run_loop.rs
vkgnosis: Yes, will be something like this eventually.

crates/autopilot/src/run_loop.rs
vkgnosis: Good catch

PR #1132, opened 2023-01-26
Document transaction identifier in openapi

vkgnosis:
When the spec was written we hadn't decided yet that we were using an arbitrary tag of a couple of bytes to identify transactions yet.

### DIFF
--- a/crates/driver/openapi.yml
+++ b/crates/driver/openapi.yml
@@ -223,18 +223,18 @@ components:
         auctionId:
           description: Id of the auction that should be executed.
           type: integer
+        transactionIdentifier:
+          description: |
+            Extra calldata that is appended to the settlement contract invocation by the Solver to
+            identify the transaction.
+
+            Hex encoded bytes. No fixed length.
+          type: string
+          example: "0x00112233"
     ExecuteResponse:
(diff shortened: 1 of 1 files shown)


crates/driver/openapi.yml
```diff
@@ -227,12 +227,14 @@ components:
...
-        account:
-          description: The address that executes the transaction.
-          $ref: "#/components/schemas/Address"
-        nonce:
-          description: The nonce of the transaction.
-          type: integer
+        transactionIdentifier:
+          description: |
+            Extra calldata that is appended to the settlement contract invocation by the Solver to
+            identify the transaction. Chosen by the Solver.
```
nlordell, line 233: I think @MartinquaXD had decided to have this decided by the autopilot to avoid any issues with implementation-specific choices of identifiers that may lead to collision.

See https://github.com/cowprotocol/services/pull/1106#issuecomment-1400691371.

vkgnosis: I've moved it into the request (from the response) now. Also made it "arbitrary length" so the api doesn't specify that it has to be 4 bytes.

nlordell (approved):

PR #365, opened 2022-07-14
Reduce critical section in rate limiter code

MartinquaXD:
While investigating a deadlock which started happening recently [[0](https://cowservices.slack.com/archives/C0361CDD1FZ/p1657698857978559?thread_ts=1657656045.424389&cid=C0361CDD1FZ)] I noticed that `match`ing on a temporary value extends that temporaries lifetime to the duration of the `match` statement. It can be extended even further when you return the `match` expression as a temporary value itself.
This is usually not a problem but can lead to surprising deadlocks when the temporary value contains a `MutexGuard` or something similar [[1](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0171d62e85c06c628a0f88ad5b9d5629)] [[2](https://github.com/rust-lang/rust/issues/37612)]. Then the resource gets locked longer than expected.

I originally wrote the rate limiter code with the intention of **NOT** locking `self.strategy` for the entire duration of the log statement so it makes sense to fix that now.

However, I'm not convinced that this bug caused the investigated deadlock. The code always acquired the locks in following order 1. `self.strategy` 2. `stdout` and should have released it in the reverse order. For this lifetime extension to cause the deadlock there would also have to be a critical section which first locks `stdout` then `self.strategy` which I don't think it does anywhere.

### Test Plan
existing tests

PR #1135, opened 2023-01-26
Implement autopilot transaction finding

vkgnosis:
I decided on the approach of using the existing event infrastructure to find recent settlement transactions. Otherwise it would be a lot more work to implement reorg robustness and event fetching.

For this PR I set the tag to the auction id. This can be considered a placeholder, although it works well in practice too as long as different environments on the same network do not use the same id at the same time. I'm going to make an issue for this because it isn't part of this PR.

### Test Plan

sql has a test, rest needs to wait for a proper end to end test

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -75,17 +75,20 @@ pub mod solve {
 }
 
 pub mod execute {
+    use derivative::Derivative;
     use model::{bytes_hex, order::OrderUid, u256_decimal};
     use primitive_types::{H160, U256};
     use serde::{Deserialize, Serialize};
     use serde_with::{serde_as, DisplayFromStr};
     use std::collections::BTreeMap;
 
-    #[derive(Clone, Debug, Default, Deserialize, Serialize)]
+    #[derive(Clone, Derivative, Default, Deserialize, Serialize)]
     #[serde(rename_all = "camelCase")]
+    #[derivative(Debug)]
     pub struct Request {
         pub auction_id: i64,
         #[serde(with = "bytes_hex")]
+        #[derivative(Debug(format_with = "shared::debug_bytes"))]
         pub transaction_identifier: Vec<u8>,
     }
 
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -86,7 +86,7 @@ pub mod execute {
     pub struct Request {
         pub auction_id: i64,
         #[serde(with = "bytes_hex")]
-        pub tag: Vec<u8>,
+        pub transaction_identifier: Vec<u8>,
     }
 
     #[serde_as]
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -85,6 +85,8 @@ pub mod execute {
     #[serde(rename_all = "camelCase")]
     pub struct Request {
         pub auction_id: i64,
+        #[serde(with = "bytes_hex")]
+        pub tag: Vec<u8>,
     }
 
     #[serde_as]
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -164,7 +164,13 @@ impl RunLoop {
         // We do keep track of hashes we have already seen to reduce load from the node.
 
         let mut seen_transactions: HashSet<H256> = Default::default();
-        while self.current_block.borrow().number <= deadline {
+        loop {
+            // This could be a while loop. It isn't, because some care must be taken to not
+            // accidentally keep the borrow alive, which would block senders. Technically this is
+            // fine with while conditions but this is clearer.
+            if self.current_block.borrow().number <= deadline {
+                break;
+            }
             let mut hashes = self
                 .database
                 .recent_settlement_tx_hashes(start..deadline + 1)
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -1,10 +1,12 @@
 use anyhow::{anyhow, Context, Result};
 use chrono::Utc;
 use model::auction::{Auction, AuctionId};
+use primitive_types::H256;
 use rand::seq::SliceRandom;
-use shared::current_block::CurrentBlockStream;
-use std::time::Duration;
+use shared::{current_block::CurrentBlockStream, ethrpc::Web3};
+use std::{collections::HashSet, time::Duration};
 use tracing::Instrument;
+use web3::types::Transaction;
 
 use crate::{
     database::Postgres,
@@ -20,6 +22,7 @@ pub struct RunLoop {
     database: Postgres,
     drivers: Vec<Driver>,
     current_block: CurrentBlockStream,
+    web3: Web3,
 }
 
 impl RunLoop {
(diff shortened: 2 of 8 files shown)


crates/autopilot/src/run_loop.rs
```diff
@@ -121,9 +124,68 @@ impl RunLoop {
...
+    ///
+    /// Returns None if no transaction was found within the deadline.
+    pub async fn wait_for_settlement_transaction(&self, tag: &[u8]) -> Result<Option<Transaction>> {
+        // Start earlier than current block because there might be a delay when receiving the
+        // Solver's /execute response during which it already started broadcasting the tx.
+        let start_offset = 10;
+        let max_wait_time = 20;
+        let current = self.current_block.borrow().number;
+        let start = current.saturating_sub(start_offset);
+        let deadline = current.saturating_add(max_wait_time);
```
vkgnosis, line 149: This needs to be configurable because different networks have different block throughput. Will do in future PR.

crates/autopilot/src/run_loop.rs
```diff
@@ -121,9 +124,68 @@ impl RunLoop {
...
+        // calldata.
+        //
+        // If the block range was large, we would make the query more efficient by moving the
+        // starting block up while taking reorgs into account. With the current range of 30 blocks
+        // this isn't necessary.
+        //
+        // We do keep track of hashes we have already seen to reduce load from the node.
+
+        let mut seen_transactions: HashSet<H256> = Default::default();
+        while self.current_block.borrow().number <= deadline {
```
nlordell, line 163: This drops the `burrow()` right? According to the docs:

> Outstanding borrows hold a read lock. This means that long lived borrows could cause the send half to block. It is recommended to keep the borrow as short lived as possible.

Maybe it might make sense to:
```suggestion
        while {
            self.current_block.borrow().number <= deadline
        } {
```

Or more explicitly:
```suggestion
        while {
            let current = self.current_block.borrow();
            let within_deadline = current.number <= deadline
            drop(current);
            within_deadline
        } {
```

(The reason being I remember something about this with `match` statements and don't know the exact semantics here).

crates/autopilot/src/run_loop.rs
```diff
@@ -121,9 +124,68 @@ impl RunLoop {
...
+                    None => continue,
+                };
+                if tx.input.0.ends_with(tag) {
+                    return Ok(Some(tx));
+                }
+                seen_transactions.insert(hash);
+            }
+            // It would be more correct to wait until just after the last event update run, but
+            // that is hard to synchronize.
+            tokio::time::sleep(Duration::from_secs(5)).await;
```
nlordell, line 187: This timeout might be too long for some layer 2s.

crates/database/src/settlements.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::TransactionHash;
+
+pub async fn recent_settlement_tx_hashes(
+    ex: &mut PgConnection,
+    start_block: i64,
+) -> Result<Vec<TransactionHash>, sqlx::Error> {
+    const QUERY: &str = r#"
+SELECT tx_hash
+FROM settlements
+WHERE block_number >= $1
```
nlordell, line 12: Should we also:
```suggestion
WHERE block_number >= $1 and block_number <= $deadline
```

I don't foresee this actually happening, but technically there could be a race condition where `self.current_block.borrow().number <= deadline` is true, but we are on block `deadline + 1` and a transaction indexed (for example if the future doesn't get scheduled for a while for whatever reason).

I'm not sure if this is actually possible in practice, but it might make sense to only accept transactions in the range.

crates/autopilot/src/run_loop.rs
nlordell: I think that the `tokio::sleep` also needs to be configurable for the same reason (different block rates on different chains).

crates/autopilot/src/run_loop.rs
MartinquaXD: I ran into that [problem](https://github.com/cowprotocol/services/pull/365) before.
You could also turn the block stream into an actual stream with [`into_stream()`](https://github.com/cowprotocol/services/blob/main/crates/shared/src/current_block.rs#L119-#L121). Then you wouldn't have to manage the timeout yourself.

crates/autopilot/src/driver_model.rs
```diff
@@ -85,6 +85,8 @@ pub mod execute {
     #[serde(rename_all = "camelCase")]
     pub struct Request {
         pub auction_id: i64,
+        #[serde(with = "bytes_hex")]
+        pub tag: Vec<u8>,
```
MartinquaXD, line 89: Does this get `Debug` printed as something readable?

crates/autopilot/src/run_loop.rs
```diff
@@ -121,9 +124,68 @@ impl RunLoop {
...
+                    .transaction(web3::types::TransactionId::Hash(hash))
+                    .await
+                    .with_context(|| format!("web3 transaction {hash:?}"))?;
+                let tx: Transaction = match tx {
+                    Some(tx) => tx,
+                    None => continue,
+                };
+                if tx.input.0.ends_with(tag) {
+                    return Ok(Some(tx));
+                }
```
MartinquaXD, line 182: I'm not sure how we have to handle reorgs here. What would the intended outcome be if we quickly find a tx hash where the calldata matches the tag but then it gets reorged away a bit later?
Would we be able to easily disregard such entries in our DB for example for the solver rewards payout? (It's even unclear to me whether they should get rewarded in that case or not).
If that is not the case we might have to keep checking transactions until `current_block >= deadline_block + reorg_buffer` to be sure that our tx will not get reorged away.

crates/autopilot/src/run_loop.rs
vkgnosis: Temporaries in the while loop condition don't get their lifetime extended for the while itself so it isn't a problem here. I think I'll still change it to be more obviously correct.

crates/database/src/settlements.rs
vkgnosis: Is it a problem if we are accidentally more lenient here? I was considering this too but decided against it, because if we get the results of the query, we might as well use them.

crates/autopilot/src/driver_model.rs
vkgnosis: It gets debug printed like all integer vectors: `vec[0, 1, 2, 3]`. I think it's fine because we don't debug this struct and the tag is short. Doesn't feel worth the effort of a manual Debug implementation.

crates/autopilot/src/run_loop.rs
vkgnosis: This component can't deal with reorgs. It takes too long to be sure it can't get re-orged. Note that the un-colocated driver competition loop also does not deal with re-orgs. The rewards payout script has to be sure it only considers transactions that really happened and are behind the re-org time. I believe it does that.

The settlement events in the database do deal with reorgs so we are fine if we record in the database for what auction id we sent /execute to what solvers with what tag.

crates/autopilot/src/driver_model.rs
MartinquaXD: This could be derived with:
```rust
    #[derive(Clone, Debug, Default, Deserialize, Serialize)]
    #[derivative(Debug)]
    pub struct Request {
        pub auction_id: i64,
        #[derivative(Debug(format_with = "shared::debug_bytes"))]
        #[serde(with = "bytes_hex")]
        pub tag: Vec<u8>,
    }
```

ennmichael, line 89: Why are you calling it `tag` when I called it `id`? Note that this "tag" would also be used to identify the transaction for `/execute` as well. I still feel like `id` is a better name, or `solutionId` if you want to be wordy.

crates/autopilot/src/run_loop.rs
```diff
@@ -121,9 +124,68 @@ impl RunLoop {
...
+            .context("wait for settlement transaction")?;
         Ok(())
     }
+
+    /// Tries to find a `settle` contract call with calldata ending in `tag`.
+    ///
+    /// Returns None if no transaction was found within the deadline.
+    pub async fn wait_for_settlement_transaction(&self, tag: &[u8]) -> Result<Option<Transaction>> {
+        // Start earlier than current block because there might be a delay when receiving the
+        // Solver's /execute response during which it already started broadcasting the tx.
```
ennmichael, line 144: Wouldn't the proper solution be to ensure that you start "waiting for" or "expecting" the transaction _before_ calling `/execute`? Instead of going back some random X number of blocks.

ennmichael (approved):

crates/autopilot/src/driver_model.rs
vkgnosis: Mistake in this PR. It is called `transactionIdentifier` in the api docs so I will name it that here too. Do you think it should be named something else in the api?

crates/autopilot/src/run_loop.rs
vkgnosis: I don't think so because of the peer to peer to nature of the blockchain. With re-orgs and different delays of when blocks are observed it can be unpredictable for what block the transaction shows up. Maybe with proof of stake this can't happen or is extremely unlikely but in general I think we should have this buffer.

crates/autopilot/src/run_loop.rs
nlordell: Cool! Just wanted to be sure.

> Temporaries in the while loop condition

I guess this only applies to `while boolean_condition` and not `while let`?

crates/autopilot/src/run_loop.rs
nlordell: I see #1141 which will likely resolve this.

crates/database/src/settlements.rs
nlordell: > Is it a problem if we are accidentally more lenient here?

I think it would only be an issue for debugging. I think the solver payout computation should happen with block chain data (and not count these solutions). However, if we make decisions based on success or failure (i.e. store a "success" flag in the database for example), then I don't think we should be lenient.

nlordell: Actually, given the discussion below, I think we don't want to make decisions about validity because of potential reorgs (which we can't reasonably deal with here).

With that in mind, the only issue I see is logging a "successfully mined" instead of "mined too late" which isn't a big deal.

crates/autopilot/src/run_loop.rs
vkgnosis: I *think* it is about whether you can refer to the temporary later. With `match` and `while let` you are binding something that you can then make use of. For these cases the temporary stays alive because there could be lifetime parameters from the temporary into the binding. With `while` you can't do this so there is no reason for the temporary to stay alive. `while` can be transformed into `loop`, where it also makes sense that the temporary wouldn't stay alive.

MartinquaXD (approved):

PR #1139, opened 2023-01-26
Make native price cache prefetch time configurable

MartinquaXD:
Ideally the `autopilot`'s native price cache has the prices warm in the cache at all times. For that to work the `CachingNativePriceEstimator` supports updating prices in the background before they actually expire.
The component already supported customizing that prefetch time but so far we didn't have a CLI argument for that.
This PR adds that argument: `--native-price-cache-prefetch-time-secs`. Configuring this should help us reduce the number of orders that don't make it into the current auction due to missing native prices.

### Test Plan
Ran `autopilot` locally:
Prints CLI arguments:
```
...
native_price_cache_max_age_secs: 60s
native_price_prefetch_time_secs: 2s
native_price_cache_max_update_size: 100
...
```
Prints a help text:
```
      --native-price-prefetch-time-secs <NATIVE_PRICE_PREFETCH_TIME_SECS>
          How long before expiry the native price cache should try to update the price in the background. This is useful to make sure that prices are usable at all times. This value has to be smaller than `--native-price-cache-max-age-secs`

          [env: NATIVE_PRICE_PREFETCH_TIME_SECS=]
          [default: 2]
```

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -823,7 +823,7 @@ mod tests {
             Duration::from_secs(10),
             Duration::MAX,
             None,
-            None,
+            Default::default(),
             1,
         );
 
--- a/crates/e2e/tests/e2e/services.rs
+++ b/crates/e2e/tests/e2e/services.rs
@@ -146,7 +146,7 @@ impl OrderbookServices {
             Duration::from_secs(10),
             Duration::from_secs(10),
             None,
-            None,
+            Default::default(),
             1,
         ));
         let quoter = Arc::new(OrderQuoter::new(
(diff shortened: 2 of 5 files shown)


sunce86 (approved):
LG.
So, by increasing this parameter we can increase the number of hits? To me, logic is that decreasing this parameter would allow us to process more tokens therefore lead to more hits, unless `update_interval` messes up things?

MartinquaXD:
> To me, logic is that decreasing this parameter would allow us to process more tokens therefore lead to more hits...

It's correct that increasing the prefetch time decreases the maximum throughput of the background task keeping the cache warm. But increasing that value would mean that we can update more tokens right before they expire (instead of right after).
So the complete strategy is to over provision the background worker slightly to give us enough head room to refresh native token prices before they actually expire. Then we should end up with fewer orders discarded due to missing native prices.

ennmichael (approved):

nlordell:
I'm going to merge this to include in next weeks release.

Issue #1114, opened 2023-01-20
SolverRewards: Add `score` to solvers response json


sunce86:
Issue tracker: https://github.com/cowprotocol/services/issues/1108

Solvers need to report one more value in their response json, which is the `score`.

nlordell:
This also means we need to add a "score" to the existing Gnosis solvers right? Is there a plan for that?

sunce86:
> Solvers need to report one more value in their response json, which is the `score`.

Indeed, added https://github.com/cowprotocol/services/issues/1147

sunce86:
Closed with https://github.com/cowprotocol/services/pull/1140

Issue #1108, opened 2023-01-19
Auction-based Solver Rewards


harisang:
There has been discussion for quite some time for restructuring the solver rewards and move towards a model where solvers control how much they want to get reimbursed for the solutions they provide. Concretely:

1. Each solver would report a solution and the `score` it assigns to the solution, where the `score` is just a quantity denominated in ETH. 

2. The protocol would simulate all submitted solutions, and the solutions that successfully simulate get ranked in decreasing order of their score.

3. The winner is the solution that passed simulation and has the highest score, breaking ties randomly.

More discussion and details about the proposed model can be found [here](https://forum.cow.fi/t/cip-draft-auction-model-for-solver-rewards/1405).

**Required changes**
1. - [x] Solvers need to report one more value in their response json, which is the `score`. https://github.com/cowprotocol/services/issues/1114

2. - [x] Implement default `score` calculation.  https://github.com/cowprotocol/services/issues/1147

3. - [x] The score and ranking should be stored in the solver competition db, as it will be needed in order to analyse the `auction-based` ranking during the transition period and for general debugging purposes. https://github.com/cowprotocol/services/issues/1138

5. - [x] Solutions with negative score should be filtered out. This means that an auction where all solutions have a negative score will be rejected.

6. - [x] The driver simulates all submitted solutions, like it does now, but ranks solutions according to the score, which is reported by solvers themselves. However, since we want to have some checks, at least at an initial stage of the model, we would still want to compute the objective function, as it is computed now by the driver, and test whether `score > objective value`. There is an option here to always enforce that the score should be at most equal to the objective function, i.e., we could enforce that `score = min(score, obj value)`, but this has yet to be decided.

7. - [ ] Store all data needed for rewards calculation in a separate tables. 

8. - [x] Since the above model moves significant complexity to the solvers, we would like to do a 2-3 week trial run of the scheme. This would mean that solvers in production would report their solution and a score, and then we will do the "official" ranking and the payouts as we do now, but we would also like to have a shadow version of the competition, where the ranking would be based on score. Ideally this would also mean that if the winner changes when ranking according to score, we would simulate the alternative winner in the same block/index where the winning settlement ended up getting mined. This shadow version data should also be made accessible to all solver teams, so that they can evaluate their performance and get a good idea of what to expect from the model.

nlordell:
> Since the above model moves significant complexity to the solvers, we would like to do a 2-3 week trial run of the scheme

[...]

PR #1140, opened 2023-01-26
SolverRewards: Add score to Settlement

sunce86:
Issue tracker: https://github.com/cowprotocol/services/issues/1108

Starts with implementing https://github.com/cowprotocol/services/issues/1114

### DIFF
--- a/crates/solver/src/settlement.rs
+++ b/crates/solver/src/settlement.rs
@@ -218,7 +218,8 @@ impl Interaction for NoopInteraction {
 #[derive(Debug, Clone, Default)]
 pub struct Settlement {
     pub encoder: SettlementEncoder,
-    pub submitter: SubmissionPreference,
+    pub submitter: SubmissionPreference, // todo - extract submitter and score into a separate struct
+    pub score: Option<BigRational>,
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -266,6 +267,7 @@ impl Settlement {
         Self {
             encoder,
             submitter: self.submitter.clone(),
+            score: self.score.clone(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


nlordell:
I would suggest setting up a "blank" feature branch, and then having PRs merge to it (instead of already having code in this one). This would allow you to "rebase merge" the feature branch and preserve the complete commit history (i.e. each commit is a single PR).

crates/solver/src/settlement.rs
```diff
@@ -218,7 +218,8 @@ impl Interaction for NoopInteraction {
 #[derive(Debug, Clone, Default)]
 pub struct Settlement {
     pub encoder: SettlementEncoder,
-    pub submitter: SubmissionPreference,
+    pub submitter: SubmissionPreference, // todo - extract submitter and score into a separate struct
+    pub score: Option<BigRational>,
```
MartinquaXD, line 222: I'm not sure I fully get the purpose of the `score` yet. Is this intended to be a promise that an objective value greater or equal to the `score` will be achieved with this solution?
Wouldn't this only make sense together with the colocated driver which supports RFQ? (RFQ would be the reason why a solver has to promise some result instead of being able to prove it by responding with the entire call data of the solution upfront).
If that is the case wouldn't we have to have a new `Settlement` type which also supports RFQ where the `score` field could be added?

sunce86: The new field `score` added in this PR corresponds to the `reference_score` mentioned in the [CIP17](https://forum.cow.fi/t/cip-draft-auction-model-for-solver-rewards/1405) and it means how much of the reward solver is willing to give up in order to win the auction and right to submit a solution onchain.

Not sure about the rest of the question but my instinct says this fits well into the colocated driver and eliminating simulations and doing proofs afterwards (meaning it is a step towards it). Right now, all changes are related to changing the logic that happens before submission.

Anyhow, since this is a big change itself, I would prefer to only do the changes needed exclusively for solver rewards change, and do the refactoring in a follow up.

vkgnosis (approved):

nlordell (approved):

PR #1141, opened 2023-01-27
Configure network's time between block

vkgnosis:
In the colocated autopilot we need to know the time between blocks on the network. Instead of making several individual block timeouts configurable I feel it makes more sense to derive them automatically from the block interval.

### Test Plan

CI, double check that staging pods still start


nlordell (approved):

nlordell:
> I feel it makes more sense to derive them automatically from the block interval.

Great idea!

ennmichael (approved):

Issue #1030, opened 2023-01-05
Skip `surplus_fee` computation for orders without balance


MartinquaXD:
Currently we usually filter out [~300 orders](https://g-0263500beb.grafana-workspace.eu-central-1.amazonaws.com/d/3Z-RJHuGz/gpv2?orgId=1&from=1672900647290&to=1672921477548&viewPanel=212) from our auction because their owners don't have the required balances.
Because getting the available balance of an order is way cheaper than getting a price estimate we could adapt the limit order quoting such that we only quote orders that have the required balance.

Right now we decide which orders to update completely with an sql query. If we want to continue doing that it would make sense to add a new column `has_required_balances` that gets continually updated by some background task.
Then [`limit_orders_with_most_outdated_fees()`](https://github.com/cowprotocol/services/blob/main/crates/database/src/orders.rs#L605-L628) would be able to safely ignore those orders until the owner has enough money and therefore reducing the number of orders we keep quoting by a whole lot.

PR #1037, opened 2023-01-06
Bulk update surplus fees

MartinquaXD:
Fixes #1029

After computing a quote for the `surplus_fee` of an order we'll use the same quote for all identical orders (`sell_token`, `buy_token`, `sell_amount`).
This will make sure that users with identical orders will get treated fairly (same `surplus_fee`) and that we issue ~40% fewer price estimation requests for limit order quoting.

### Test Plan
Updated existing unit tests
[Manual test](https://github.com/cowprotocol/services/pull/1037#issuecomment-1376037932) to verify that orders get picked up as quoted and that the performance is alright.

PR #1142, opened 2023-01-27
Don't quote orders with insufficient funds

MartinquaXD:
Fixes: #1030

A lot of our limit orders don't make it into the auction because the owner doesn't have the required `sell_token` balance.
Because determining whether the order's owner has enough balances is a lot cheaper than getting a `surplus_fee` update this PR  adds the flag `has_sufficient_balance` to the `orders` table and only quotes orders where this is true.
A new background task keeps this flag up-to-date on every new block.
I added a DB migration step which sets that flag to true for all existing orders and initializes it to true for new orders. Also in case of an error I tried to err on the safe side to avoid discarding orders unjustly.
This behavior is feature gated by `--skip-quoting-unfunded-orders` so we can quickly disable that in case some problem occurs.
I also updated the existing metrics and added a few new ones to measure this optimization's performance.

We know that around half of the quoted orders (600) are missing the required funds but it's unclear how many of them will still have to be discarded in the `SolvableOrdersCache` due to #1037. However I'm still confident that this is still a useful optimization.

### Test Plan
Updated/extended existing unit tests and added a new one.
Manual tests are looking good.
Prints the CLI argument:
```
...
limit_order_quoter_parallelism: 5
skip_quoting_unfunded_orders: true
```
Prints a help text:
```
      --skip-quoting-unfunded-orders
          Skip updating the `surplus_fee` for orders where the owner currently doesn't have enough balance

          [env: SKIP_QUOTING_UNFUNDED_ORDERS=]
```

### DIFF
--- a/crates/autopilot/src/limit_orders/balance_tracker.rs
+++ b/crates/autopilot/src/limit_orders/balance_tracker.rs
@@ -1,0 +1,135 @@
+use crate::database::Postgres;
+use anyhow::Result;
+use futures::StreamExt;
+use primitive_types::U256;
+use prometheus::IntGaugeVec;
+use shared::{
+    account_balances::{BalanceFetching, Query},
+    current_block::{into_stream, CurrentBlockStream},
+};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
+use tracing::Instrument;
+
+#[derive(prometheus_metric_storage::MetricStorage)]
+pub struct Metrics {
+    /// Results of last run loop of the BalanceTracker
+    #[metric(labels("result"))]
+    limit_orders_balances: IntGaugeVec,
+}
+
+/// Updates `has_sufficient_balance` column of open orders.
+pub struct BalanceTracker {
+    balance_fetcher: Arc<dyn BalanceFetching>,
+    database: Postgres,
+    metrics: &'static Metrics,
+}
+
+fn get_initialized_merics() -> &'static Metrics {
+    let metrics = Metrics::instance(global_metrics::get_metric_storage_registry()).unwrap();
+
+    let gauges = &metrics.limit_orders_balances;
+    gauges.with_label_values(&["error"]).set(0);
+    gauges.with_label_values(&["sufficient"]).set(0);
+    gauges.with_label_values(&["insufficient"]).set(0);
+
+    metrics
+}
+
+impl BalanceTracker {
+    pub fn new(balance_fetcher: Arc<dyn BalanceFetching>, database: Postgres) -> Self {
+        Self {
+            balance_fetcher,
+            database,
+            metrics: get_initialized_merics(),
+        }
+    }
+
+    // Updates the `has_sufficient_balance` flag of open limit orders in the background on every
+    // new block.
+    pub fn spawn(self, block_stream: CurrentBlockStream) {
+        tracing::debug!("spawning background task of BalanceTracker");
+        tokio::spawn(
+            self.background_task(block_stream)
+                .instrument(tracing::info_span!("balance_tracker")),
+        );
+    }
+
+    async fn background_task(self, block_stream: CurrentBlockStream) {
+        let mut block_stream = into_stream(block_stream);
+        while block_stream.next().await.is_some() {
+            if let Err(err) = self.update_payable_orders().await {
+                tracing::error!(?err, "error updating has_sufficient_balance flag, consider disabling --skip-quoting-unfunded-orders");
+            }
+        }
+        tracing::error!("current block stream terminated unexpectedly");
+    }
+
+    async fn update_payable_orders(&self) -> Result<()> {
+        let orders = self
+            .database
+            .open_limit_orders_without_pre_interactions()
+            .await?;
+
+        // Collect orders in a HashSet first to avoid duplicated requests.
+        let queries: HashSet<_> = orders.iter().map(Query::from_order).collect();
+        let queries: Vec<_> = queries.into_iter().collect();
+
+        let start = std::time::Instant::now();
+        let balances = self.balance_fetcher.get_balances(&queries).await;
+        let balances: HashMap<_, _> = queries
+            .iter()
+            .zip(balances.into_iter())
+            .map(|(query, result)| (query, result.unwrap_or_default()))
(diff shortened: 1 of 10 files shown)


vkgnosis:
It is a good observation that checking the balance is cheaper than calculating the fee, so we should do the balance first. It also makes sense that we cache the balance information for some time, just like the surplus fee.

I feel missing from this PR is integrating this properly with the solvable_orders module. When we create the auction, we already fetch all balances there. (I realize now that this actually has become a problem because of how many orders there are. When this code was written we didn't expect there to be so many orders in the auction.) But we also fetch balances for the new component added in this PR. This is a lot of duplicate work.

We should think about having a single component responsible for balance updates like we have for native prices to make sure that work is not duplicated and to make sure that creating an auction doesn't take too long. It might be reasonable to treat market and limit orders differently, like requiring market order balance to be more up to date, but it should be handled in the same component.

We should also think about whether it makes more sense to keep all balances in memory or in the database. Native prices we keep in memory but balances are per user per token so might be too much. Need to do some math to answer this.

vkgnosis:
I only saw https://github.com/cowprotocol/services/issues/1143 after making the comment. I would like to figure what I just wrote about first, before merging this PR.

MartinquaXD:
I think having a similar caching component for balances as we have for native prices makes sense and would be fairly easy because it's obvious when to update the cache (when a new block arrives) and when to invalidate the cache (also when a new block arrives).
While doing manual tests for this PR I saw that fetching balances for our orders isn't a crazy amount of work. I used a snapshot of the prod DB which is already a bit older but I think the findings should still be valuable.
When fetching balances for 609 orders I only had to do 164 unique requests which took ~500ms - 1s to complete. That means keeping a possible balances cache warm (at least for open orders) would be trivial. 
So given the fact that this cache can be filled and updated very quickly I think it's enough to keep it in memory instead of the DB. And with that same argument I think the currently duplicated work is not so big that we need to have this caching component before merging this PR.

Another separate discussion is how to work with limit orders in the future. Right now we do the selection which orders to update completely in the DB which makes certain things annoying. For example I had to add this new flag which seems weird and a future optimization where we skip quoting orders which are way off the current market price would require us to keep the native prices in the DB as well. I think we could avoid a lot of trouble if instead picking, sorting and limiting the number of quotes to do in the DB we would just copy all the order data into our process and select orders there. That would allow us to work way more easily with data we only have cached in memory at the moment.
I didn't do the benchmarks yet but if copying all orders into memory during every surplus fee quoting round is not too much overhead I would be strongly in favor of that.

crates/autopilot/src/limit_orders/balance_tracker.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/// Updates `has_sufficient_balance` column of open orders.
+pub struct BalanceTracker {
+    balance_fetcher: Arc<dyn BalanceFetching>,
+    database: Postgres,
+    metrics: &'static Metrics,
+}
+
+fn get_initialized_merics() -> &'static Metrics {
```
ennmichael, line 30:
```suggestion
fn get_initialized_metrics() -> &'static Metrics {
```

crates/autopilot/src/limit_orders/balance_tracker.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    async fn update_payable_orders(&self) -> Result<()> {
+        let orders = self
+            .database
+            .open_limit_orders_without_pre_interactions()
+            .await?;
+
+        // Collect orders in a HashSet first to avoid duplicated requests.
+        let queries: HashSet<_> = orders.iter().map(Query::from_order).collect();
+        let queries: Vec<_> = queries.into_iter().collect();
```
ennmichael, line 78: [`Itertools::unique`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique)

crates/autopilot/src/limit_orders/balance_tracker.rs
```diff
@@ -0,0 +1,135 @@
...
+        let updates: Vec<_> = orders
+            .iter()
+            .map(|order| {
+                let balance = balances.get(&Query::from_order(order));
+
+                let has_sufficient_balance = if let Some(balance) = balance {
+                    let has_sufficient_balance = match order.data.partially_fillable {
+                        true => balance >= &U256::one(), // any amount would be enough
+                        false => balance >= &order.data.sell_amount,
+                    };
```
ennmichael, line 101: `if`

crates/database/src/orders.rs
```diff
@@ -727,17 +765,20 @@ pub async fn count_limit_orders_with_outdated_fees(
     ex: &mut PgConnection,
     max_fee_timestamp: DateTime<Utc>,
     min_valid_to: i64,
+    quote_unfunded_orders: bool,
```
ennmichael, line 768: The name `quote_unfunded_orders` may be incorrect; there's no quoting going on at this layer.

crates/database/src/orders.rs
```diff
@@ -1809,10 +1850,11 @@ mod tests {
         .await
         .unwrap();
 
-        let orders: Vec<_> = order_parameters_with_most_outdated_fees(&mut db, timestamp, 2, 100)
-            .try_collect()
-            .await
-            .unwrap();
+        let orders: Vec<_> =
+            order_parameters_with_most_outdated_fees(&mut db, timestamp, 2, 100, true)
```
ennmichael, line 1854: This is why specifying parameters as bare `bool`s is not the best idea: it is not clear to the reader what this `true` stands for at a glance.

crates/autopilot/src/limit_orders/balance_tracker.rs
nlordell: Personally, I like the the esthetic of `match` more than `if` for expressions and prefer `if` for control flow. Just personal preference.

nlordell:
Personally, I think we should keep this in memory. I would like to keep the `orders` table as append-only as possible (I know this already isn't the case with EthFlow orders and limit order surplus fees). Maybe this is misguided by me though.

MartinquaXD:
Closed in favor of #1153 which achieves the same thing in-memory.

PR #1144, opened 2023-01-27
Fix rust 1.67.0 warnings

vkgnosis:
This release introduced a new clippy warnings. The most common one is about not inlining variables into format strings.

### Test Plan

CI

### DIFF
--- a/crates/orderbook/src/api/get_trades.rs
+++ b/crates/orderbook/src/api/get_trades.rs
@@ -86,7 +86,7 @@ mod tests {
         };
 
         let owner = H160::from_slice(&hex!("0000000000000000000000000000000000000001"));
-        let owner_path = format!("/v1/trades?owner=0x{:x}", owner);
+        let owner_path = format!("/v1/trades?owner=0x{owner:x}");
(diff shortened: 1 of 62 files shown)


crates/orderbook/src/api/get_trades.rs
```diff
@@ -95,7 +95,7 @@ mod tests {
         assert_eq!(result.order_uid, None);
 
         let uid = OrderUid([1u8; 56]);
-        let order_uid_path = format!("/v1/trades?orderUid={:}", uid);
+        let order_uid_path = format!("/v1/trades?orderUid={uid:}");
```
MartinquaXD, line 98: Does `{:}` even do anything special? Same for the change under this one.

MartinquaXD (approved):

crates/orderbook/src/api/get_trades.rs
vkgnosis: It doesn't and it already didn't before this PR. Will fix it in a follow up because I had auto merge on.

PR #1145, opened 2023-01-27
Fix unused `:` in format strings

vkgnosis:
Martin noticed in https://github.com/cowprotocol/services/pull/1144 that we sometimes include `:` in format strings even though this doesn't have any effect.

I also inlined some more cases that clippy doesn't catch because they appear in macros that aren't std's `format!`.

### Test Plan

CI

### DIFF
--- a/crates/refunder/src/submitter.rs
+++ b/crates/refunder/src/submitter.rs
@@ -55,7 +55,7 @@ impl Submitter {
             .eth()
             .transaction_count(self.account.address(), None)
             .await
-            .map_err(|err| anyhow!("Could not get latest nonce due to err: {:}", err))
+            .map_err(|err| anyhow!("Could not get latest nonce due to err: {err}"))
     }
 
     pub async fn submit(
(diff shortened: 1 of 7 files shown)


crates/refunder/src/submitter.rs
```diff
@@ -55,7 +55,7 @@ impl Submitter {
             .eth()
             .transaction_count(self.account.address(), None)
             .await
-            .map_err(|err| anyhow!("Could not get latest nonce due to err: {:}", err))
+            .map_err(|err| anyhow!("Could not get latest nonce due to err: {err}"))
```
MartinquaXD, line 58: I noticed you changed one log from `{err}` to `{err:?}` to get the error log with more details. Should this also happen here?

MartinquaXD (approved):

crates/refunder/src/submitter.rs
vkgnosis: The one I changed was logging an anyhow error, which should always be done with Debug printing to make it show the causes. This is a web3 error. I think it doesn't matter for these but haven't tested.

ennmichael (approved):

PR #1149, opened 2023-01-28
Record Metrics for S3 Uploads

nlordell:
This PR changes the instance cache to record metrics instead of issuing error alerts when uploading to S3 fails. This is because uploading can intermittently fail, and these instance are used for debugging anyway, so temporary failures to upload is not a big issue. With metrics, we can setup an alert if it starts failing for, say, 10 minutes straight which means that there is probably something more seriously wrong that requires some attention.

### Test Plan

CI


MartinquaXD (approved):
Depending on how important we find these uploads we could also collect failed uploads until we get a successful upload again and then retry the failed ones one after another.

nlordell:
RN they are just for debugging, but this might be a good strategy for the driver (as a solver may want to collect signatures for disputes in the future).

vkgnosis (approved):

PR #1150, opened 2023-01-28
Fix Naive Solver Uniswap Output Rounding

nlordell:
What a doozy!!!

In investigating Naive Solver simulation alerts, I came across [this simulation](https://dashboard.tenderly.co/gp-v2/staging/simulator/cc9f2d65-13db-4360-9078-d576e195f661/debugger?trace=0.7.0.0.1.2) where the Naive Solver was requesting more out of a Uniswap pool than it had reserves.

It turns out that the rounding computation is to blame. Basically, we have this rounding for the Uniswap output amount so that the total traded "shortage" is exactly equal to the exact Uniswap amount we request. This is to prevent off-by-1 errors that can happen because of how the smart contract rounds in favour of the trader (as the original Naive Solver equations were developed without integer math in mind). Therefore, this is expected to use the **synthetic** traded amounts (so, as if the `surplus_fee` was a regular fee) and not the actual executed amount (as that would be too large - the surplus fee is meant to stay in the settlement contract and not trade against the pool).

### Test Plan

Added a unit test that demonstrated the issue. If you `git checkout main -- crates/solver/src/solver/naive_solver/multi_order_solver.rs` and then run the test, you will see that the requested output amount exceeds the pool reserves.

Existing Native Solver tests continue to pass, including existing tests around the rounding computation for market orders.


crates/solver/src/solver/naive_solver/multi_order_solver.rs
```diff
@@ -148,18 +148,36 @@ fn solve_with_uniswap(
...
     // be that we actually require a bit more from the Uniswap pool in order to
     // pay out all proceeds. We move the rounding error to the sell token so that
-    // it either comes out of the fees or existing buffers. Compute that amount:
-    let uniswap_out_with_rounding = big_int_to_u256(&settlement.trade_executions().fold(
+    // it either comes out of the fees or existing buffers. It is also important
+    // to use the **synthetic** order amounts for this, as this output amount
+    // needs to be computed for what is actually intented to be traded against
+    // an AMM and not how the order will be executed (including things like
+    // surplus fees).
+    let uniswap_out_with_rounding = big_int_to_u256(&orders.iter().try_fold(
```
vkgnosis, line 156: I don't really understand the transformation that took place here. I have some intuition why this makes sense and it's plausible, but I'm not very confident how the PR description and change of comment above, lead to the change of code here.

Could you clarify that more?

crates/solver/src/solver/naive_solver/multi_order_solver.rs
```diff
@@ -148,18 +148,36 @@ fn solve_with_uniswap(
...
-            if execution.sell_token == uniswap_out_token {
-                total -= execution.sell_amount.to_big_int();
+        |mut total, order| {
+            if order.sell_token == uniswap_out_token {
+                total -= match order.kind {
+                    OrderKind::Sell => order.sell_amount,
+                    OrderKind::Buy => order
+                        .buy_amount
+                        .checked_mul(uniswap_out)?
+                        .checked_div(uniswap_in)?,
```
MartinquaXD, line 165: Naively I would think that the multiplications and divisions in the different cases need to be inverted to each other, right? (line 164/165 and line 172/173).

In the first branch we compute `sell_amount = buy_amount * sell_price / buy_price` with `uniswap_out` as the `sell_price` and `uniswap_in` as the `buy_price`.
In the second branch we compute `buy_amount = sell_amount * buy_price / sell_price`  with `uniswap_out` as the `buy_price` and `uniswap_in` as the `sell_price`.

Shouldn't it be consistent across the 2 branches which uniswap value we use as the `buy_price` and which as the `sell_price`?

crates/solver/src/solver/naive_solver/multi_order_solver.rs
nlordell: I'll try my best 🙈.

Basically, given `[Orders]` and `Pool`, we compute the *exact* input and output amount that satisfies the original OBA equation (basically, clearing prices are the executed Uniswap pool price).

However, the equations were originally created assuming real executed amount. Since executed amounts can't be fractional **and** the smart contract always rounds up in favour of the trader, we have:
```
(
  ∑ executed_buy_amount_of_uniswap_out_token -
  ∑ executed_sell_amount_of_uniswap_out_token
) ∈ [uniswap_out_amount, uniswap_out_amount + n_orders)
```

Additionally, when we encode the Uniswap interaction, we use `swapTokensForExactTokens`, meaning that we only ever swap exactly `uniswap_out_amount` and slippage is paid for in the input token. This means that, in some cases, we may be missing a few wei of `uniswap_out_token` which can cause settlements to revert in case of no buffers in that token.

To address this, we compute the sum above and set the actual output amount to be that, so we are never short a wei or two.

What is critical here, is that the OBA equation and trade executions are applied to the **synthetic** orders and not the real orders. This is what the change is about.

crates/solver/src/solver/naive_solver/multi_order_solver.rs
nlordell: Unintuitively, prices are `amount ^ -1`:
```rust
    let mut settlement = Settlement::new(maplit::hashmap! {
        uniswap_in_token => uniswap_out,
        uniswap_out_token => uniswap_in,
    });
```

So, if `sell_token == uniswap_out_token`, then
```
buy_amount * uniswap_out / uniswap_in
  = buy_amount * uniswap_in_token_price / uniswap_out_token_price
  = buy_amount * buy_price / sell_price
```

Which I think is correct.

vkgnosis (approved):

crates/solver/src/solver/naive_solver/multi_order_solver.rs
MartinquaXD: But isn't the equation we try to uphold `sell_amount / sell_price == buy_amount / buy_price`? And solved for `sell_amount` it would be `sell_amount = buy_amount * sell_price / buy_amount` which is different from what you end up with.
Not sure where I'm going wrong but I guess if all the existing tests still pass it must be alright. 🤔

nlordell: No, its:
```
sell_amount * sell_price == buy_amount * buy_price
```

Or, this is how I remember it:
```
1 beer * 4.00€ == 8 chocolates * 0.50€
```

MartinquaXD: Yeah, that makes more sense than what I had in mind. 🤦‍♂️

MartinquaXD (approved):

PR #1153, opened 2023-01-29
Skip quoting unfunded limit orders

### LINKED ISSUE #1030: Skip `surplus_fee` computation for orders without balance

MartinquaXD:
Fixes #1030

Alternative approach to #1142. Instead of keeping the logic to select the orders we want to quote the `surplus_fee` for next in the DB we do everything in memory. This allows us to use data we only have in cache way easier. This would enable us to very easily add the mechanism to only quote limit orders that are close to the current market price using the cached native prices.
The downside is that we copy a lot of data from the DB in each quoting cycle but that seems okay.

### Test Plan
added unit test


crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -58,10 +58,13 @@ impl LimitOrderQuoter {
...
     /// Returns whether it is likely that there is no more work.
     async fn update(&self) -> Result<bool> {
-        let order_specs = self
+        let orders = self
             .database
-            .order_specs_with_outdated_fees(self.limit_order_age, self.parallelism)
+            .open_limit_orders(self.limit_order_age)
             .await?;
+        let orders = self.orders_with_sufficient_balance(orders).await;
+        let order_specs: Vec<_> = orders.into_iter().map(order_spec_from).unique().collect();
```
nlordell, line 66: Not a suggestion, but something that I like (subjective and esthetic) in projects that already import `itertools` is to
```suggestion
        let order_specs = orders.into_iter().map(order_spec_from).unique().collect_vec();
```

crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -58,10 +58,13 @@ impl LimitOrderQuoter {
...
+
```
nlordell, line 67: This changes semantics in a pretty big way, we used to get order_parameters_with_most_outdated_fees which was limited to `self.parallelism` orders and ordered by how long the order has been waiting for a price.

Now, we are getting much more orders and trying to update them all in an arbitrary order.

I would propose to return `database::orders::open_limit_orders` ordered ascending by `surplus_fee_timestamp` and to `.take(self.parallelism)`. `Itertools::unique` is stable, so the `order_specs` would retain the sorting from `open_limit_orders`.

nlordell:
Personally, I prefer the in-memory strategy. I don't think that the amount of data being read per block is huge (i.e. Postgres should be able to handle it - it is already sending much more data per auction building step).

crates/autopilot/src/arguments.rs
```diff
@@ -147,6 +148,10 @@ pub struct Arguments {
     /// The time in seconds between new blocks on the network.
     #[clap(long, env, value_parser = shared::arguments::duration_from_seconds)]
     pub network_block_interval: Option<Duration>,
+
+    /// Defines which strategies to apply when updating the `surplus_fee` of limit orders.
+    #[clap(long, env, use_value_delimiter = true)]
+    pub quoting_strategies: Vec<QuotingStrategy>,
```
vkgnosis, line 154: Why Vec instead of a single strategy?

crates/database/src/orders.rs
```diff
@@ -704,6 +666,41 @@ pub async fn update_limit_order_fees(
...
+    pub uid: OrderUid,
+    pub owner: Address,
+    pub sell_token: Address,
+    pub buy_token: Address,
+    pub sell_amount: BigDecimal,
+    pub sell_token_balance: SellTokenSource,
+    pub partially_fillable: bool,
+}
+
+/// Returns all limit orders than are currently waiting to be filled.
```
vkgnosis, line 682:
```suggestion
/// Returns all limit orders that are currently waiting to be filled.
```

crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -58,10 +58,23 @@ impl LimitOrderQuoter {
...
+        let orders = match self.strategies.contains(&QuotingStrategy::SkipUnfunded) {
+            true => orders_with_sufficient_balance(&*self.balance_fetcher, orders).await,
+            false => orders,
+        };
+
+        let order_specs = orders
+            .into_iter()
+            .map(order_spec_from)
+            .unique()
+            .take(self.parallelism)
```
vkgnosis, line 75: Related to discussion below. If we are fetching all orders from the DB anyway, then it doesn't make much sense to only `take self.parallelism` here.

The implementation that I originally wrote was designed to take as few orders from the DB as needed to have small individual update loops. This has a benefit over fetching all then updating all, because a single update of all orders can take a long time and by the end some orders might no longer be needed to be updated at all.

Since then we've had the change to group orders (used to happen in DB, now happens here), and now to also filter out orders without enough balance.

Now it seems more wasteful to always get all db orders, and all their balance to then only take ~5 order specs per update. I don't know what the right value is but we can do something like `5 * self.parallelism` or we can update as many order specs as possible at the current parallelism level, until a time limit of 60 seconds is exceeded.

crates/autopilot/src/arguments.rs
MartinquaXD: I wanted to be able to compose different strategies in the future. For example I have in mind to prefer updating orders which have an achievable price instead of going by the `surplus_fee_tiemstamp`.
Eventually this should get dropped entirely when we tested each strategy and decided on whether to keep it or not.

nlordell: Feature flags for testing how different strategies

crates/autopilot/src/limit_orders/quoter.rs
nlordell: I think that is a good proposal. I think we need to stream the price estimates like we used to do in `solvable_orders` then.

crates/autopilot/src/arguments.rs
MartinquaXD: Here is the issue: https://github.com/cowprotocol/services/issues/1168

crates/autopilot/src/limit_orders/quoter.rs
MartinquaXD: It's not obvious to me how using a timeout would be better than simply picking more orders to update and wait until they are done. This is trivial to implement and we can't "waste" any requests by starting one and dropping it before we get a response.

vkgnosis: The reason I suggest a timeout is

> This has a benefit over fetching all then updating all, because a single update of all orders can take a long time and by the end some orders might no longer be needed to be updated at all.

It is a minor optimization.

nlordell: For now, should we specify `self.update_batch_size`, which gets parallelized by `self.parallelism` as @vkgnosis suggested?

crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -156,16 +169,146 @@ impl LimitOrderQuoter {
...
 #[derive(prometheus_metric_storage::MetricStorage, Clone, Debug)]
 #[metric(subsystem = "limit_order_quoter")]
 struct Metrics {
     /// Categorizes order quote update results.
     #[metric(labels("type"))]
     update_result: prometheus::IntCounterVec,
+
+    /// Tracks how many orders don't get quoted because their
+    /// owners don't have sufficient balance.
+    orders_skipped_for_missing_balance: prometheus::IntGauge,
```
nlordell, line 238: Oh wow! the `#[metrics()]` really isn't needed then 🤔.

crates/database/src/orders.rs
```diff
@@ -704,6 +666,41 @@ pub async fn update_limit_order_fees(
...
+    pub owner: Address,
+    pub sell_token: Address,
+    pub buy_token: Address,
+    pub sell_amount: BigDecimal,
+    pub sell_token_balance: SellTokenSource,
+    pub partially_fillable: bool,
+}
+
+/// Returns all limit orders that are currently waiting to be filled.
+pub fn open_limit_orders(
```
nlordell, line 683: Can we add a comment here indicating that they are sorted by surplus_fee_timestamp (so the most outdated orders come first)?

nlordell (approved):
Changes look good! One small suggestion is a knob for update batch size (but would be OK with merging as-is).

crates/autopilot/src/limit_orders/quoter.rs
MartinquaXD: It's only needed if you want to add some configuration to that metric. For example bucket sizes or labels. Since the defaults are fine we don't need to use the macro here.

vkgnosis (approved):

PR #1154, opened 2023-01-29
Implement CachingBalanceFetcher

### LINKED ISSUE #1143: Cache balances in `Web3BalanceFetcher`
This component gets used in the 3 different places and it would be great if we could share and cache requests across those components. I think to do that correctly we would have to adapt `dyn BalanceFetching` slightly to also take a target block number. (Not sure how much we would like that)
This could allow us to build auctions more quickly and decrease the time we take to validate an order before creating it in the orderbook.

MartinquaXD:
Fixes #1143

With #1142 on the horizon it makes sense to start caching users balances for `sell_tokens` of their orders.
This means the `SolvableOrdersCache` and `LimitOrderQuoter` can easily share the balances which reduces the number of `eth_call`s we have to do.
I also added a very simple background task which refetches all cached balances whenever it sees a new block. This should further reduce the time required to build an auction or quote limit orders because they should completely hit the cache most of the time.

### Limitations
I decided against caching the `can_transfer()` calls since they would not really profit from caching anyway.

Also this cache currently doesn't try to prevent avoiding duplicate requests in case one of the components requires the balances right after the background task invalidated the cache but since fetching all balances is fairly quickly this shouldn't be too bad. To address that we could only invalidate the cached data when the background process has fetched the new balances. This is technically less accurate (using old balances from the previous block for ~0.5s -1s longer) but would mostly resolve the (minor) problem of duplicate requests. Let me know what you think about that trade-off.

### Test Plan
Added unit tests


crates/shared/src/caching_balance_fetcher.rs
```diff
@@ -0,0 +1,146 @@
...
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
+use tracing::Instrument;
+
+#[derive(Default)]
+struct BalanceCache {
+    /// Block number the cached data was collected from.
+    block: u64,
+    data: HashMap<Query, U256>,
```
nlordell, line 20: Can't we use the `RequestSharing` to synchronize updates and requests?

crates/shared/src/caching_balance_fetcher.rs
```diff
@@ -0,0 +1,146 @@
...
+        let updates = missing
+            .iter()
+            .zip(new_balances.iter())
+            .filter_map(|(i, result)| Some((queries[*i], *result.as_ref().ok()?)))
+            .collect();
+        Self::update_balance_cache(&self.balance_cache, initial_block, updates);
+
+        cached.extend(missing.into_iter().zip(new_balances.into_iter()));
+        cached.sort_by_key(|(i, _)| *i);
+        cached.into_iter().map(|(_, balance)| balance).collect()
```
nlordell, line 132: Correct me if I'm wrong, but I think the cache grows in an unbounded way.

crates/shared/src/caching_balance_fetcher.rs
MartinquaXD: Yeah, I guess I could just put that in the `Web3BalanceFetcher` as a follow up PR.

crates/shared/src/caching_balance_fetcher.rs
vkgnosis: It feels wasteful to have auto update on block and no expiry. Over time there will be lots of unused balance updates. If a balance hasn't been requested in X minutes it could drop from the cache.

For the size we don't need a limit but it would be good to have  metric for it.

crates/shared/src/caching_balance_fetcher.rs
```diff
@@ -0,0 +1,265 @@
...
+        Self {
+            inner,
+            balance_cache: Default::default(),
+        }
+    }
+}
+
+struct CacheResponse {
+    // The indices and results of queries that were in the cache.
+    cached: Vec<(usize, Result<U256>)>,
```
vkgnosis, line 39: Seems like this could be without Result because it is never set to Err.

MartinquaXD: This was done to avoid creating yet another vector to merge the cached and requested values together.
But I'm fine with changing that.

crates/shared/src/caching_balance_fetcher.rs
nlordell: We should be able to use the `cached` crate for this (which has a similar interface to `HashMap` and shouldn't have an impact on the complexity here).

crates/shared/src/caching_balance_fetcher.rs
vkgnosis: If there is a good reason for it then I don't mind. Would maybe double check that that is really needed and can't be avoided with iterators or something like that.

crates/shared/src/caching_balance_fetcher.rs
nlordell: Perfect! Great idea.

vkgnosis (approved):

nlordell (approved):

PR #1155, opened 2023-01-29
Don't Overcharge for Onchain ERC-1271 Orders

nlordell:
This PR changes the quote API to set `verification_gas_limit` to 0 for on-chain orders. This is in-line with our design decision from:

https://github.com/cowprotocol/services/blob/13b73e0033071417744c78615d5dff89141e7187/crates/autopilot/src/database/onchain_order_events.rs#L529-L533

### Test Plan

Added a unit test.


MartinquaXD (approved):

vkgnosis (approved):
This might lead to a future bug when we do parse other onchain orders.

josojo:
> This might lead to a future bug when we do parse other onchain orders.

I kind a agree. Maybe its better to fix it in the front-end only? (https://github.com/cowprotocol/cowswap/pull/1937)

nlordell:
Closing as not needed anymore.

PR #1157, opened 2023-01-30
Add driver library crate

vkgnosis:
This is a similar setup as in older crates. By having a library crate, we can make use of the library for e2e tests. I would like to add a colocation e2e test which instantiates a real autopilot, driver, solver.

`main.rs` has effectively been renamed to `run.rs`, but Github doesn't show it as a rename.

Some imports changed because we were accidentally importing names `use`d in `main.rs`.

### Test Plan

code still compiles, no logic change


nlordell (approved):
Fine by me.

ennmichael (approved):
I absolutely despise this but I don't see a better solution. Please document the ugliness, why `lib.rs` exists.

nlordell:
The other option is to have the `e2e` crate `process::Command::new("driver")` to start up the driver. This would give us "true" E2E test without having access to poke at the inside bits like we do for our current E2E tests.

vkgnosis:
Why is it so despicable? It just means you can run  the driver as a library instead of being forced to run it as a binary. This is something that I for example wish we could do with geth too, so that we wouldn't need a docker container for it.

PR #1158, opened 2023-01-30
Improve Metrics fo Auction Building

nlordell:
This PR adds a new gauge for the number of **market** orders that are filtered out because of missing prices. This is because filtering out market orders for this reason is more serious than limit orders. This new metric will allow us to monitor this separately and potentially create alerts to deal with missing market order native token prices.

I also changed the `checkpoint` to log the order's class to make debugging filtered orders in the log easier.

### Test Plan

Rust compiler

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -11,7 +11,7 @@ use model::{
 };
 use number_conversions::u256_to_big_decimal;
 use primitive_types::{H160, H256, U256};
-use prometheus::{IntCounter, IntGaugeVec};
+use prometheus::{IntCounter, IntGauge, IntGaugeVec};
 use shared::{
     account_balances::{BalanceFetching, Query},
     bad_token::BadTokenDetecting,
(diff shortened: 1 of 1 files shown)


sunce86 (approved):

crates/autopilot/src/solvable_orders.rs
```diff
@@ -45,6 +45,10 @@ pub struct Metrics {
     /// Auction filtered orders grouped by class.
     #[metric(labels("reason"))]
     auction_filtered_orders: IntGaugeVec,
+
+    /// Auction filtered market orders due to missing native token price.
+    #[metric()]
```
MartinquaXD, line 50: Does this line serve a purpose?

MartinquaXD (approved):

crates/autopilot/src/solvable_orders.rs
```diff
@@ -485,10 +494,21 @@ fn get_orders_with_native_prices(
                 used_prices.insert(*t1, *p1);
                 true
             }
-            _ => false,
+            _ => {
+                if order.metadata.class == OrderClass::Market {
+                    filtered_market_orders += 1;
+                }
```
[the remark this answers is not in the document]
nlordell: Love it.

crates/autopilot/src/solvable_orders.rs
nlordell: It appears to be needed. We also use it in other places:

https://github.com/cowprotocol/services/blob/5dcfab50fa1a5b8b3189a2d38685fa52181e2d73/crates/shared/src/maintenance.rs#L134-L146

vkgnosis: It's not required.

PR #1159, opened 2023-01-30
Do not update native prices in background in Orderbook

vkgnosis:
The auto updating behavior makes sense for Autopilot, but not in Orderbook. This removes some load from our price estimators, especially because this cache never expires its entries (TODO future PR).

### Test Plan

CI

### DIFF
--- a/crates/shared/src/price_estimation/factory.rs
+++ b/crates/shared/src/price_estimation/factory.rs
@@ -321,6 +321,7 @@ impl<'a> PriceEstimatorFactory<'a> {
         &mut self,
         kinds: &[PriceEstimatorType],
         drivers: &[Driver],
+        spawn_update_task: bool,
     ) -> Result<Arc<CachingNativePriceEstimator>> {
         anyhow::ensure!(
             self.args.native_price_cache_max_age_secs > self.args.native_price_prefetch_time_secs,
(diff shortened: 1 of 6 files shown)


nlordell (approved):

crates/shared/src/price_estimation/factory.rs
```diff
@@ -321,6 +321,7 @@ impl<'a> PriceEstimatorFactory<'a> {
         &mut self,
         kinds: &[PriceEstimatorType],
         drivers: &[Driver],
+        spawn_update_task: bool,
```
nlordell, line 324: In general, I tend to agree with @ennmichael that these bool parameters are hard to understand at the callsite (you just see a true/false without much context).

Not blocking PR merge though.

MartinquaXD:
I think updating native prices still makes sense for the order book to get faster price estimates (which was its initial goal).
One thing that seems problematic at the moment is that the background task doesn't always wait a constant amount of time but instead waits less time the longer the run loop took:
```rust
        tokio::time::sleep(update_interval.saturating_sub(now.elapsed())).await;
```
If that gets replaced by a constant sleep we can easily scale back the number of requests emitted from the background task without having to kill it entirely.

vkgnosis:
I feel it is worth more to reduce load from estimators than speculatively decrease delay but not sure.

MartinquaXD:
> I feel it is worth more to reduce load from estimators than speculatively decrease delay but not sure.

We can already reduce the load by tweaking the parameters of the background task.
I think this PR still has merit because splitting the arguments for the native price estimator from the arguments for the background task is nice but to me it seems we should tweak settings first instead of axing a feature immediately.

ennmichael (approved):

crates/shared/src/price_estimation/factory.rs
nlordell: Sorry, didn't mean to bikeshed. As mentioned, doesn't block the PR, was just indicating personal preference.

PR #1161, opened 2023-01-30
Add solver address to solver competition

sunce86:
This PR adds solver address to each solution saved to solver competition db.

It allows data team to do queries more efficiently and avoid matching the `solver` name with the list of solver addresses.


vkgnosis:
We should think about whether this makes sense. Solvers can change their address. Some solvers might not have a single address that they submit from like if they use Gelato (?). From this POV it is more correct to keep using the solver name, even if that means a slight overhead for the data team. On the other hand, maybe a solver can also change it's name. This change isn't problematic, but I'm not sure if it makes sense.

bh2smith:
If the new CIP is going to retroactively distribute remaining rewards to solvers that submitted non-winning solutions, we would need to know an address for these solutions that never made it on chain to attribute to them. Yes. Solvers can change their address, but it is the solver who solved the batch that should be attributed to the retroactive rewards.

bh2smith (approved):

vkgnosis:
The name and the address are connected. We have a document where we list all solver names, their address, their reward address. The question is what the best "primary" key to identify solvers is. It could be any of these fields.

nlordell (approved):

nlordell:
Personally, I think `address` is a better unique identifier for this (as it is how they are identified on-chain).

Agree this means we can technically drop the `name`, although it is useful.

nlordell:
> It allows data team to do queries more efficiently and avoid matching the solver name with the list of solver addresses.

Hmm, given this is the motivation, I'm not sure this is the correct way to do it. I don't think the `solver_competitions` table was intended to be anything more than debug data.

bh2smith:
> Hmm, given this is the motivation, I'm not sure this is the correct way to do it. I don't think the `solver_competitions` table was intended to be anything more than debug data.

Note that, we do not currently have al alternative solution for looking up past competitions with this retroactive disbursement of remaining rewards balance. So this table will likely be the place we must go to get a glimpse of historical solver performance metrics.

ennmichael (approved):
:shrug: Code is OK, depends on what we decide to do

sunce86:
> If we need something for the core protocol like solve payouts, we should add it to a separate table so that it is clear that it isn't generic debug data.

I addressed this here: https://github.com/cowprotocol/services/pull/1166#issuecomment-1423969119
With that I believe it is ok to merge this PR.

PR #1162, opened 2023-01-30
Prioritize updates of market order native token prices

vkgnosis:
The native price cache cannot keep up with the amount of tokens it has to process in production. This can lead to market orders not settling because the orders are not considered solvable because of missing native prices.

This PR prioritizes the updates of market order tokens over other tokens. This functionality is only used by the autopilot's solvable order module.

### Test Plan

new unit test for the harder to understand part of the change

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -4,7 +4,7 @@ use itertools::{Either, Itertools};
 use primitive_types::H160;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     sync::{Arc, Mutex, MutexGuard, Weak},
     time::{Duration, Instant},
 };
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -17,18 +17,28 @@ struct Metrics {
...
-    price: f64,
-    updated_at: Instant,
-    requested_at: Instant,
+/// Wrapper around `Box<dyn PriceEstimating>` which caches successful price estimates for some time
+/// and supports updating the cache in the background.
+/// The size of the underlying cache is unbounded.
+pub struct CachingNativePriceEstimator {
+    inner: Arc<Inner>,
+    max_age: Duration,
+    metrics: &'static Metrics,
```
vkgnosis, line 26: This has been moved up because I found that easier to understand in the module.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -136,15 +146,27 @@ impl Inner {
...
+            .map(|(token, cached)| (*token, cached.requested_at))
+            .collect();
+        let high_priority = self.high_priority.lock().unwrap().clone();
+        let priority = |token: &H160| high_priority.contains(token) as u8;
+        outdated.sort_unstable_by_key(|entry| {
+            (
+                std::cmp::Reverse(priority(&entry.0)),
+                std::cmp::Reverse(entry.1),
+            )
+        });
```
vkgnosis, line 167: This is a logic change.

sunce86 (approved):

MartinquaXD (approved):

nlordell (approved):
Nice!

PR #1163, opened 2023-01-30
Single order solvers always try settling market orders first

MartinquaXD:
The match rate for market orders currently doesn't look great. This PR makes it so that our single order solvers always try to settle market orders first and only then work on limit orders.

### Test Plan
CI


vkgnosis (approved):

nlordell (approved):
Nice! Great idea.

PR #1164, opened 2023-01-31
Add more metrics to native price cache

vkgnosis:
Allows us to measure the size of the cache and the update throughput.

### Test Plan

CI, see metrics after merging

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -2,7 +2,7 @@ use crate::price_estimation::native::{NativePriceEstimateResult, NativePriceEsti
 use futures::stream::StreamExt;
 use itertools::{Either, Itertools};
 use primitive_types::H160;
-use prometheus::IntCounterVec;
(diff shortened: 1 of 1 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -14,7 +14,11 @@ use tracing::Instrument;
 struct Metrics {
     /// native price cache hits misses
     #[metric(labels("result"))]
-    native_price_cache: IntCounterVec,
+    native_price_cache_access: IntCounterVec,
+    /// number of items in cache
+    native_price_cache_size: IntGauge,
```
MartinquaXD, line 19: Should we maybe also add a gauge that tracks the number of outdated prices during each update cycle?
Then we could see when we hit a point of diminishing returns when we increase the throughput with configuration changes.

MartinquaXD (approved):

crates/shared/src/price_estimation/native_price_cache.rs
vkgnosis: added

PR #1165, opened 2023-01-31
Use a config file instead of CLI args for `solvers`

ennmichael:
Move solvers to use config files to be consistent with the `driver` and also because experience with the driver shows that it's easier and more predictable to define configuration in files rather than swaths of CLI arguments.

The idea is to expect a different config file format based on the `command` being executed.

Also cleaned up the approach to setting the log filter for both `driver` and `solvers`.

### Test Plan

Updated the automated tests.


crates/solvers/src/main.rs
```diff
@@ -15,29 +16,24 @@ use std::net::SocketAddr;
...
-    command: cli::Command,
-    bind: Option<oneshot::Sender<SocketAddr>>,
-) {
-    let cli::Command::Baseline(baseline) = command;
+async fn run(args: impl Iterator<Item = String>, bind: Option<oneshot::Sender<SocketAddr>>) {
+    let args = cli::Args::parse_from(args);
+    boundary::initialize_tracing(&args.log);
+    tracing::info!("running solver engine with {args:#?}");
 
+    // TODO In the future, should use different load methods based on the command being executed
```
nlordell, line 33: To let the compiler tell you when you need to change this:
```suggestion
    // TODO In the future, should use different load methods based on the command being executed
    let cli::Command::Baseline = args.command;
```

nlordell (approved):

MartinquaXD (approved):

vkgnosis (approved):

PR #1166, opened 2023-01-31
[DB] Create Table Settlement Rewards

bh2smith:
This is merely a discussion PR to align expectations between backend and data team. In reference to [this issue](https://github.com/cowprotocol/services/issues/1108#issuecomment-1410110608) we implement a draft migration to get the tables in shape for Solver Auction CIP.

My primary concerns are;

1. Not back filling settlements with id and execution cost is a RISK. Technically these columns should not be nullable and allowing the historical values (which can easily be replayed and filled) to be NULL leave room for future bugs like accidentally leaving a cell empty. 

2. Furthermore, the primary key of this table could be changed from `(block_number, log_index)` to ID and this would allow us to build a proper relational mapping between `settlement_rewards.settlement_id` and `settlements.id`

3. [Idea] We may want to split `execution_cost` into `gas_used | gas_price` for an enriched dataset.

cc @sunce86 and @vkgnosis


vkgnosis (commented):
I don't understand why all the new fields are needed. In the CIP the reward is this:

```
reward = observedQuality - referenceScore
observedQuality = observedSurplus + observedFee
```

I think to get the *observed* values we can't just simulate in advance, we actually need to look at the transaction later, to determine them exactly. The only required to be stored value is `referenceScore`, which is the score of the second best solution. And whatever data is needed to associate a settlement transaction and with the auction/competition.

bh2smith:
> I think to get the _observed_ values we can't just simulate in advance, 

This is merely a bad choice of naming convention. These "observed" values are indeed the values observed during simulation. cc @fhenneke

The only part of this computation that uses actual after-the-fact transaction data is the execution cost, which is already available data in the part of the service linking settlements to actions.

Would someone from the backend who is already aware of all these details like to chime in on this discussion? Seem like a communication gap.

fhenneke:
> I think to get the observed values we can't just simulate in advance, we actually need to look at the transaction later, to determine them exactly.

The observed quality is supposed to be the quality of the solution as seen on chain. But there should only be two cases: in case of a successful settlement, the observed quality is exactly equal to the simulated quality; in case of an unsuccessful settlement, the observed quality is zero.

nlordell:
> The only required to be stored value is referenceScore, which is the score of the second best solution. And whatever data is needed to associate a settlement transaction and with the auction/competition.

The CIP also stipulates a soft constraint on `observedValue >= winningScore`, so we should also store that IMO.

To me, this table should look like:
```sql
CREATE TABLE settlement_rewards (
  settlement_id bytea, -- the ID is calldata bytes, small technical detail
  winning_score numeric(78,0) NOT NULL,
  reference_score numeric(78,0) NOT NULL
);
```

> These "observed" values are indeed the values observed during simulation.

This is not my understanding either - this should be the values from the on-chain transaction and not from the simulation.

nlordell:
>  in case of an unsuccessful settlement, the observed quality is zero.

I also think we need to have a hard deadline for when a settlement is included on-chain. I mentioned this in the CIP but didn't get much discussion around it. Not mining is a failure condition as well IMO.

bh2smith:
As @fhenneke claims - the "observed" valuations should Equal to the simulated values when the transaction is successful. Not sure if this is true (because of slippage).

If we are to be getting the observed values from the transaction data, then we will need to know EXACTLY how this is computed and I do not believe that we can do that with any sort of reasonable precision (note that these observed values are in ETH and will rely on a price feed). 

As discussed yesterday, these values are computed from "externalPrices" which NEVER make it on chain. We could Opt for using "clearingPrices", but these are not guaranteed to be relative to ETH. This means, the ONLY way we could compute the "observed" values is by mixing off-chain data from our orderbook. This would amount to needing to store another table "externalPrices" instead of just `surplus` and `fee` proposed here. I hope this clears things up about how we landed on the fact that the `settlement_rewards` table  should include both `observed_surplus` and `observed_fees`.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE settlements
+  ADD COLUMN id integer default NULL,
+  ADD COLUMN execution_cost integer default NULL;
```
bh2smith, line 3: execution cost could be split into `gas_used` and `gas_price`.

nlordell:
> the observed quality is exactly equal to the simulated quality

This is not the case. Simulation happens in a component that will be run by external solvers in the near future (the driver). The source of truth for observed value needs to be the blockchain.

bh2smith:
> The source of truth for observed value should be the blockchain.

We cannot deterministically or reliably with any precision compute this (in ETH) without having `clearingPrices` relative to ETH.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,11 @@
...
+
+CREATE TABLE settlement_rewards (
+  settlement_id integer,
+  simulated_surplus integer NOT NULL,
+  simulated_fee integer NOT NULL,
```
bh2smith, line 8: Ok - so it seems the problem everyone has with this is that we should include "observed" values and that they belong in a different table. I will make the suggested changes, but then the name of this table will change.

nlordell:
> We cannot deterministically or reliably with any precision compute this (in ETH) without having clearingPrices relative to ETH.

The observed value is based on `external_prices` which is part of the auction data that we do have. This is off-chain and orthogonal to whether or not we use onchain data or simulations. The question is how to compute the surplus in traded tokens? To me, there is no difference in "simulated" and "observed" in this regard - the surplus in traded tokens is entirely in the calldata and I don't understand what additional information a "simulation" will give.

My point is that the calldata needs to be taken from what ends up on chain (observed) and not what was promised to end up on chain (simulated).

bh2smith:
I have applied changes according to the feedback so far. It appears there was some serious misunderstanding about yesterday's conclusions.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE settlements
+  ADD COLUMN id bytea default NULL,
+  ADD COLUMN gas_usd integer default NULL,
+  ADD COLUMN gas_price integer default NULL;
+
+CREATE TABLE settlement_scores (
+  settlement_id bytea NOT NULL,
+  winning_score numeric(78,0) NOT NULL,
+  reference_score numeric(78,0) NOT NULL
```
nlordell, line 9: This isn't part of the CIP, but I do think we also need to store the block deadline somewhere. Not sure if this should be tied to `auction` or `settlement_scores`, but it should go somewhere.

See https://github.com/cowprotocol/services/pull/1166#issuecomment-1411869152

nlordell:
Sorry for closing the issue - misclick.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE settlements
+  ADD COLUMN id bytea default NULL,
+  ADD COLUMN gas_usd integer default NULL,
+  ADD COLUMN gas_price integer default NULL;
```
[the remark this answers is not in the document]
nlordell: @vkgnosis - do you think it makes sense to stick these into `settlement_observations` table instead? This way the `settlements` table is just for indexing events and is append only (I think we took a similar strategy with `auction_transaction` table).

We can read this as part of the transaction receipt as we index settlements. It could even replace the existing `auction_transaction` table.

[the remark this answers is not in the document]
bh2smith: Whoops that was a typo: was supposed to say `gas_used`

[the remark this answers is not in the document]
nlordell: Unresolved the thread as there was an open question directed at @vkgnosis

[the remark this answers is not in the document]
bh2smith: Sorry didn't see that.

fhenneke:
> To me, there is no difference in "simulated" and "observed" in this regard - the surplus in traded tokens is entirely in the calldata and I don't understand what additional information a "simulation" will give.

A simulation does not give additional information on the amounts traded by users compared to the call data itself. Together with corresponding `external_prices` it determines the surplus in ETH.
If (non-discounted) fees can also be derived from the call data, they can be aggregated into fees in ETH, again using `external_prices`.
(Both of those values in ETH are currently being stored in the `solver_competitions` table.)

nlordell:
>  it determines the surplus in ETH

Just clarify here 'it' is calldata or simulation?

database/sql/V048__create_settlement_rewards.sql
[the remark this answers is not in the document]
vkgnosis: I agree this should not go into the settlements table because that table can get overriden by event updates and it is clearer separation to put it somewhere else. Don't know what the best place is, but not that table.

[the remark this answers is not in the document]
bh2smith: The event. Contains the transaction receipt with the details of gas used and gas price. Why would we even consider not putting it here. In any case I am fine either way since it is also an "observation" of the settlement.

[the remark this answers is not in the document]
nlordell: We actually need transaction calldata for extracting the observed fee and surplus values for `settlement_observations`. I would propose not `ALTER`-ing the `settlements` table and instead adding these values to the `settlement_observations` (this should make your life easier @bh2smith as you won't need as many `JOIN`s and we don't need to deal with `NULL` values for old data):

```sql
CREATE TABLE settlement_observations (
  -- the appended settlement ID read from the calldata
  id bytea NOT NULL,
  -- the transaction hash and log index to be able to `JOIN` on the `settlement`
  -- table
  tx_hash bytea NOT NULL,
  log_index bigint NOT NULL,
  -- the gas characteristics observed from the transaction receipt
  cumulative_gas_used numeric(78,0) NOT NULL,
  effective_gas_price numeric(78,0) NOT NULL;
  -- the surplus observed from the transaction calldata, and converted to ETH
  -- with the auction external prices.
  surplus numeric(78,0) NOT NULL,
  fee numeric(78,0) NOT NULL
);
```

And, we should also store auction external prices for traded tokens, as they are important to keep around and will be needed for indexing settlement transactions (for computing surplus and fee in ETH, we can also use the `auctions` table, but that gets cleaned up periodically and I'm reluctant to do so for that reason):

```sql
CREATE TABLE auction_prices (
  auction_id bigint NOT NULL,
  token bytea NOT NULL,
  price numeric(78,0) NOT NULL,
);
```

I'm not sure how to optimize storage for `auction_prices`, as we only need them for traded tokens, but given the co-location model we don't have a reliable way for determining which tokens we need to store ATM. (We can always incentivize solvers to be upfront about this by saying "price is 0 for surplus and fee computation for tokens that you did not say were traded" - so they get fewer rewards in that case).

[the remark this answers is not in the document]
bh2smith: I think this thread has been very helpful and brought forth some great points. Have updated the schema (and now there is no need to alter `settlements` table at all). I hope we are getting closer to a reasonable model for this CIP.

Going to close this issue for now. Please feel free to add further comments on the recently proposed schema.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,29 @@
...
+  gas_used numeric(78,0) NOT NULL,
+  effective_gas_price numeric(78,0) NOT NULL,
+  -- the surplus observed from the transaction call data,
+  -- and converted to ETH with the auction external prices.
+  surplus numeric(78,0) NOT NULL,
+  fee numeric(78,0) NOT NULL
+);
+
+CREATE TABLE auction_prices (
+  auction_id bigint NOT NULL,
```
sunce86, line 26: These are `external_prices` right?

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,29 @@
...
+-- (after the submissions have been ranked)
+CREATE TABLE settlement_scores (
+  settlement_id bytea NOT NULL,
+  winning_score numeric(78,0) NOT NULL,
+  reference_score numeric(78,0) NOT NULL
+);
+
+-- Populated after block finalization via transactionReceipt.
+CREATE TABLE settlement_observations (
+  -- the appended settlement ID read from the call data
```
sunce86, line 11: If this table is populated after the auction is finished (for example lagging 64 blocks behind the latest block) that would be problematic, since we lose the `external_prices` vector once the auction is finished (and we also do not store the external prices anywhere right now I believe).
So, in order to avoid saving the whole `external_prices` vector, we would need to inspect the winning settlement before submission and save the subset of prices to db, and hope that the onchain transaction will be the same (just thinking out loud).

bh2smith: But we can add the prices when we have them and create the link to here later...

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,29 @@
...
+  settlement_id bytea NOT NULL,
+  -- the transaction hash and log index to be able to `JOIN` on the `settlement`
+  -- table, read from the transaction receipt
+  tx_hash bytea NOT NULL,
+  log_index bigint NOT NULL,
+  gas_used numeric(78,0) NOT NULL,
+  effective_gas_price numeric(78,0) NOT NULL,
```
MartinquaXD, line 18: Which `gas_price` is used here?
1. while building the auction
2. while adding the observed settlement
3. price of the block the settlement occurred

Do we maybe need multiple?

bh2smith: The only gas price that matters to me is the gas price that is used to compute the transaction cost. 


Screenshot 2023-02-07 at 10 28 25

database/sql/V048__create_settlement_rewards.sql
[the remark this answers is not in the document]
sunce86: > I'm not sure how to optimize storage for auction_prices, as we only need them for traded tokens

The main concern with adding full list of external prices is the storage right? 
On each auction cut we can store full list of external prices to `auction_prices`. After transaction being mined, when we update the `settlement_observations` we can also update the `auction_prices` entry and eliminate not traded tokens.

database/sql/V048__create_settlement_rewards.sql
sunce86: My current plan is to handle this as explained here: https://github.com/cowprotocol/services/pull/1166#discussion_r1098571919
Closing the thread.

sunce86:
@bh2smith I'm in favour of moving this PR to review mode.

bh2smith:
Is there any chance we could get the base branch updated so that our CI won't fail on an irrelevant error?

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,29 @@
...
+);
+
+-- Populated after block finalization via transactionReceipt.
+CREATE TABLE settlement_observations (
+  -- the appended settlement ID read from the call data
+  settlement_id bytea NOT NULL,
+  -- the transaction hash and log index to be able to `JOIN` on the `settlement`
+  -- table, read from the transaction receipt
+  tx_hash bytea NOT NULL,
+  log_index bigint NOT NULL,
```
vkgnosis, line 16: Let's make this blog_number and log_index, instead of tx_hash. That is slightly more consistent to how we uniquely identify settlements in other places. `(block_number, log_index)` should be marked as the primary key too.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE TABLE settlement_scores (
+  settlement_id bytea NOT NULL,
+  winning_score numeric(78,0) NOT NULL,
+  reference_score numeric(78,0) NOT NULL
+);
+
+-- Populated after block finalization via transactionReceipt.
+CREATE TABLE settlement_observations (
+  -- the appended settlement ID read from the call data
+  settlement_id bytea NOT NULL,
```
vkgnosis, line 12: Is this settlement_id the name we want for this long term? In `crates/driver/openapi.yml` this is called `transaction_identifier`.

bh2smith: Sounds fine with me, as long as it's "id" and not "identifier" - since id means the same thing and is way shorter.

bh2smith: I'm actually more worried that the EVM can not enforce this id to exist and so it might not always be there. Perhaps we should consider allowing it to be nullable. We can identify the transaction and solver by the event --- although I guess then there are other fields that could not be linked (the reference score). 

If we were to follow traditional financial accounting principles we would have to make these two fields nullable and someone would have to manually try to match missing records at the end of each accounting period as best as possible.

vkgnosis: Technically, instead of being null it would be an empty byte array. So either way it doesn't need to be nullable but in some situations there might not be a meaningful id.

vkgnosis: Another discussion is whether this should be UNIQUE. The way we used to talk about the transaction_identifier, is that we needed it as a way to identify settlement transactions when solvers settle them. This means that from the perspective of the backend they only have to be unique temporarily but it is not a problem if we reuse these later.

vkgnosis: > Sounds fine with me with a strong preference for "id" and not "identifier" - since it means the same thing and is way shorter.

id or identifier is fine. But first word word should match the rest of the code.

MartinquaXD: > Another discussion is whether this should be UNIQUE.

If we read that value from the block chain we can't guarantee that it's unique. Some external party might have a bug and reuse an old id or there is an adversary trying to mess with us by submitting settlements with duplicated ids on purpose.

bh2smith: While it it true that is may not be unique -- we are going to have a pretty hard time (non-deterministic) matching things which are not. We may want to come up with some preventative measures or at least an approach for handling these kinds of situations early on.

bh2smith: Note that I am changing both occurrences of `settlement_id` to `transaction_id` now. I hope this naming convention is really what we want.

```
CREATE TABLE settlement_scores (
  transaction_id bytea NOT NULL,
  winning_score numeric(78,0) NOT NULL,
  reference_score numeric(78,0) NOT NULL
);

-- Populated after block finalization via transactionReceipt.
CREATE TABLE settlement_observations (
  -- the appended transaction identifier read from the call data
  transaction_id bytea NOT NULL,
```

bh2smith: Changed name here, but there seems to be a few other things in this thread -- feel free to "resolve" whenever anyone feels its all been sorted.

bh2smith:
For some reason this keeps failing CI on unrelated code. Right now it's only nightly so I will ignore it.

sunce86:
Additionally, for each auction we need to store the list of solvers (adresses) that submitted valid solutions (part of the solver rewards budget is payed to each solver proportionally to the number of auctions each solver submitted a valid solution to).
I don't see we have this relation anywhere (expect `solver_competitions` table which we should not use for this AFAIU).
I think we need another table for this, for example `auction_participants`? 

```
CREATE TABLE auction_participants (
 id bigserial PRIMARY KEY,
 participans bytea[]
);
```

database/sql/V048__create_settlement_rewards.sql
nlordell: > Which gas_price is used here?

The transaction's "effective gas price". This has a specific meaning post-EUP-1559 to be the effective/actual gas price that was changed to the EOA for the transaction.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,31 @@
...
+  block_number bigint NOT NULL,
+  log_index bigint NOT NULL,
+  gas_used numeric(78,0) NOT NULL,
+  effective_gas_price numeric(78,0) NOT NULL,
+  -- the surplus observed from the transaction call data,
+  -- and converted to ETH with the auction external prices.
+  surplus numeric(78,0) NOT NULL,
+  fee numeric(78,0) NOT NULL,
+
+  PRIMARY KEY (block_number, log_index)
```
nlordell, line 24: I think that the primary key should be `settlement_id`. I think we should add an index on `(block_number, log_index)` and that they should be unique, but I don't agree that it the primary key.

vkgnosis: I suggested this as the primary key before. Note that `settlement_id` (now called `transaction_id`) can't be the primary key because we cannot control whether some onchain settlement transaction reuses an id.

bh2smith:
Hey @sunce86 - thanks for bringing this up! Just a small point of concern with the ID column:

> id bigserial PRIMARY KEY,

I believe this would have to correspond to the auction_id that is being used elsewhere in the DB. However I think the type then would have to be bound to whatever type is already there?

vkgnosis:
> I believe this would have to correspond to the auction_id that is being used elsewhere in the DB. However I think the type then would have to be bound to whatever type is already there?

Yes. We already have the auction_id bigserial somewhere. It doesn't make sense to create a new bigserial. You would use a bigint that semantically refers to the auction_id bigserial.

bh2smith:
> Yes. We already have the auction_id bigserial somewhere. It doesn't make sense to create a new bigserial. You would use a bigint that semantically refers to the auction_id bigserial.

so then is this more like a "foreign key"? Just not sure how to propose the SQL be changed.

sunce86:
Let's just make it

> id bigint PRIMARY KEY

I would like to push this PR forward to unblock further progress on this feature. We can make additional tweaks in following PRs (which I expect to exist anyway).

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,31 @@
+-- Populated before settlement occurs on chain
+-- (after the submissions have been ranked)
+CREATE TABLE settlement_scores (
+  transaction_id bytea NOT NULL,
```
vkgnosis, line 4: Let's change this to use `auction_id bigint PRIMARY KEY`. Remove `transaction_id`.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,31 @@
...
+  winning_score numeric(78,0) NOT NULL,
+  reference_score numeric(78,0) NOT NULL
+);
+
+-- Populated after block finalization via transactionReceipt.
+CREATE TABLE settlement_observations (
+  -- the appended transaction identifier read from the call data
+  transaction_id bytea NOT NULL,
```
vkgnosis, line 12:
```suggestion
```

vkgnosis (commented):
I went over this again and realized that we aren't using the transaction id right. Sorry for not catching this earlier. IMO the way we should identify things is as follows:

- Settlement transactions on chain are identified by (block_number, log_index).
- Auction details, possibly before we can associate them with a transaction, by their auction  id.

`transaction_id` is an implementation detail of this mapping. The mapping is currently handled by the `auction_transaction` table and the `auction_transaction` autopilot module. When we change to colocation we will use the a `transaction_id` to tag transactions on chain.

vkgnosis:
> so then is this more like a "foreign key"? Just not sure how to propose the SQL be changed or if its fine as is.

Doesn't need to be tagged as foreign key. Just a comment (or clear from context) that it refers to the same auction id that we always use.

database/sql/V048__create_settlement_rewards.sql
bh2smith: will do.

bh2smith:
Have updated according to the discussion's progression. If we are getting close to finalizing this PR -- I may start preparing solver-rewards to make use of this data.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,36 @@
...
+
+CREATE TABLE auction_prices (
+  auction_id bigint NOT NULL,
+  token bytea NOT NULL,
+  price numeric(78,0) NOT NULL
+);
+
+CREATE TABLE auction_participants (
+ -- This refers to the auction_id (from `auctions`)
+ id bigint PRIMARY KEY,
```
vkgnosis, line 33: In the other tables this is called auction_id so I'd also use the full name here.

vkgnosis (approved):

sunce86:
> > part of the solver rewards budget will be payed to each solver proportionally to the number of auctions each solver submitted a valid solution to
> 
> This is confusing with regards to colocation. The way I understand colocation is that we can't verify whether a solver's solution is "valid" unless we pick them as the winner. For non winners the only information we have is that they claimed to have a solution at score X.

Indeed, however it is how solver rewards are planned to be distributed with this CIP. I assume after the colocation is done, this part of budget will be either killed or distributed to all solvers that submitted a score regardless if their solution is valid or not.

database/sql/V048__create_settlement_rewards.sql
bh2smith:
```suggestion
 -- This links to the `auctions` table
 auction_id bigint PRIMARY KEY,
```

bh2smith:
I am assuming we have (more or less) landed on a schema that we can start to base our work off of. I am not sure if this PR will be merged as is or if there is intended to be a follow up that actually implements the insertions into this table.

Will someone from the backend continue from here?

sunce86:
@bh2smith  Yes, let's merge this and I will implement read/write functions.

nlordell:
> I assume after the colocation is done, this part of budget will be either killed or distributed to all solvers that submitted a score regardless if their solution is valid or not.

The plan is to make it so solvers can submit their calldata in order to receive participation bonus. So this incentivizes them to share their calldata, but aren't required.

database/sql/V048__create_settlement_rewards.sql
nlordell: I think this comment was not addressed before merging.

bh2smith: Yes! Great point.... Should we prepare an issue to track this better?

sunce86: Addressed here: https://github.com/cowprotocol/services/issues/1108#issuecomment-1438371881

PR #1171, opened 2023-01-31
Separate transaction submission from transaction notification nodes

fedgiac:
Initial discussion [here](https://cowservices.slack.com/archives/C0375NV72SC/p1675076444783699).

An external solver wants to include its own node to the list of transaction submission nodes for testing purposes. As far as I understand, the node is not guaranteed to be reliable and its main purpose is seeing when we are submitting a transaction.
The goal of this PR is separating this node from our own submission nodes to decrease the impact that a malfunctioning of the new node has on the services.

This PR splits submission nodes in two categories: `broadcast` and `notification`. The former are normally nodes we control, while the latter are third-party nodes that we only trust partially.
Broadcast nodes behave like submission nodes do in the current code.
Notification node are the same except that they come from a different env variable and the solver service doesn't panic if the node is invalid on startup. After startup they behave exactly like broadcast nodes.

### Test Plan

End to end tests.


nlordell:
> After startup they behave exactly like broadcast nodes.

In theory, shouldn't we ignore `eth_getTransactionReceipt` requests and the like for these nodes if we don't fully trust them? I.e. just send the transaction and forget?

Not for this PR (I think the current scope is fine and a nice improvement), but something to discuss.

crates/solver/src/main.rs
```diff
@@ -330,37 +333,41 @@ async fn main() -> ! {
...
-        .collect::<Vec<_>>();
-    for (node, url) in &submission_nodes_with_url {
-        let node_network_id = node
-            .net()
-            .version()
-            .await
-            .with_context(|| {
-                format!(
-                    "Unable to retrieve network id on startup using the submission node at {url}"
+    let submission_nodes = future::join_all(
```
nlordell, line 336: I think you can use `try_join_all` here to avoid the `into_iter().collect()`.

vkgnosis:
> while the latter are third-party nodes that we only trust partially

I'm missing context for why we have notification nodes at all. I think it was something like a solver team wanting to debug why how transactions get settled and adding their node as another submission node was the path of least resistance to achieve that. Or are these nodes potentially faster at bringing transactions on chain? Understanding what their purpose is, is important for evaluating this change.

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
```diff
@@ -119,6 +130,77 @@ impl TransactionSubmitting for PublicMempoolApi {
...
+    /// eventually be included in a block.
+    Broadcast(Web3),
+    /// A notification node is an endpoint that is not expected to submit transactions to the
+    /// mempool once a transaction has been received. Its purpose is notifying the node owner that a
+    /// transaction has been submitted.
+    /// In general, there are lower expectations on the availability of this node variant.
+    Notification(Web3),
+}
+
+impl Deref for SubmissionNode {
```
nlordell, line 145: Nice, I was reviewing last night and asking myself: “where is the code change that unwraps the inner web3 instance in the submission logic?”. The answer is “Rust magic!”.

nlordell (approved):

fedgiac:
> I'm missing context for why we have notification nodes at all. I think it was something like a solver team wanting to debug why how transactions get settled and adding their node as another submission node was the path of least resistance to achieve that.

I added more context to the PR description. But in short, your summary is correct and there is no direct benefit for us in adding the new node.

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
```diff
@@ -119,6 +130,77 @@ impl TransactionSubmitting for PublicMempoolApi {
...
+
+pub async fn validate_submission_node(node: &Web3, expected_network_id: &String) -> Result<()> {
+    let node_network_id = node
+        .net()
+        .version()
+        .await
+        .context("Unable to retrieve network id on startup")?;
+    if &node_network_id != expected_network_id {
+        bail!("Network id doesn't match expected network id");
+    };
```
MartinquaXD, line 201: nit:
```suggestion
    anywhow::ensure!(&node_network_id == expected_network_id, "Network id doesn't match expected network id");
```

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
```diff
@@ -119,6 +130,77 @@ impl TransactionSubmitting for PublicMempoolApi {
...
+        url: Url,
+        name: impl ToString,
+        expected_network_id: &String,
+    ) -> Self {
+        let node = ethrpc::web3(ethrpc_configs, http_factory, &url, name);
+        if let Err(err) = validate_submission_node(&node, expected_network_id).await {
+            tracing::error!("Error validating submission notification node {url}: {err}");
+        }
+        Self::Notification(node)
+    }
```
MartinquaXD, line 183: These function seem to do the same (validating that a node belongs to the correct network). Why not have a single function `validate_node()` and then wrap the result in a `SubmissionNode` variant?
If you absolutely want to make sure that you can only create a `SubmissionNode` from a validated node you can put it in a new type like
```rust
struct ValidatedNode(Web3);
fn validate_node() -> Result<ValidatedNode>;
SubmissionNode::new(node: ValidatedNode) -> Self;
```

I think this should allow you to instantiate the nodes in `solver/main.rs:335` with a lot less code.

MartinquaXD (commented):
I'm a bit worried that would lead to a lot of solvers giving us debugging nodes which seems weird. However, since this would be resolved when we have co-located drivers I don't think it's a huge problem but the approach still seems weird to me.

crates/solver/src/main.rs
```diff
@@ -389,8 +393,8 @@ async fn main() -> ! {
             }
             TransactionStrategyArg::PublicMempool => {
                 assert!(
-                    !submission_nodes.is_empty(),
-                    "missing transaction submission nodes"
+                    submission_nodes.iter().any(|node| node.can_broadcast()),
+                    "At least one submission node that can broadcast transactions must be available"
```
MartinquaXD, line 397: Why is that the case? Is there any solver that wants us to use a broadcast node on `goerli` of `gc`?

fedgiac: Note that a node that can broadcast is our usual submission node type. This line says that at lease one node must be able to reliably get the transaction mined.
I don't think we have had any requests for GC or Görli.

MartinquaXD: Yes, of course. I mixed up the node types. 🙈

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
fedgiac: True. I changed this [here](https://github.com/cowprotocol/services/pull/1171/commits/3006c858309b6f11af6d56f6eea18fde382a1167), but I'm not sure I like it more because it puts more logic in `main.rs` that might be useful to know when reading `main`.

MartinquaXD: I had it a bit different in mind. More like this:
```rust
            .map(|(index, url)| {
                let name = format!("broadcast {index}");
                (name, url, true)
            })
            .chain(
                args.transaction_notification_nodes
                    .into_iter()
                    .enumerate()
                    .map(|(index, url)| {
                        let name = format!("notify {index}");
                        (name, url, false)
                    })
            )
            .map(|(name, url, panic_on_error)| async move {
                let node = ethrpc::web3(&args.shared.ethrpc, &http_factory, &url, name);
                if let Err(err) = validate_submission_node(&node, &network_id).await {
                    tracing::error!(
                        "Error validating submission notification node {url}: {err}"
                    );
                    assert!(!panic_on_error);
                }
                Ok(SubmissionNode::Notification(node))
            })
```

fedgiac: I agree this is nicer! I implemented this suggestion with some changes [here](https://github.com/cowprotocol/services/pull/1171/commits/7661a3d98b8ee755774a960a0ac8270c6f8f67cb) if you want to take a look.

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
vkgnosis: This is a bit of an anti pattern in Rust imo. I'd rather make this a normal function or make users of SubmissionNode access the web3 field directly.

vkgnosis (approved):

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
fedgiac: The docs do say "Deref should only be implemented for smart pointers to avoid confusion" so it's a bit eccessive.
Now that `SubmissionNode` is a struct, the unwrapping is just `node.web3` so I decided to remove `Deref` from the code.

MartinquaXD (approved):

PR #1172, opened 2023-01-31
Ignore transaction creation reports from notification nodes

fedgiac:
Builds on top of #1171.

Change the behavior of notification nodes so that we don't accept transactions that were created through them.
This is done to avoid that a malicious node stops our search immediately even if the transaction hasn't been submitted to the blockchain.

### Test Plan

E2e tests to show that broadcast nodes still work.
Then I tried to run an e2e test with `SubmissionNode::Notification` instead of `SubmissionNode::Broadcast` and the test failed as expected.


crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
```diff
@@ -56,24 +56,31 @@ impl TransactionSubmitting for PublicMempoolApi {
...
+                    if can_broadcast {
+                        tracing::debug!(%label, "created transaction with hash: {:?}", tx_hash);
+                        return Ok(TransactionHandle {
+                            tx_hash,
+                            handle: tx_hash,
+                        });
+                    } else {
+                        tracing::debug!(%label, "non-broadcasting node reports transaction creation with hash: {:?}", tx_hash);
+                        futures = rest;
+                    }
```
vkgnosis, line 83: It might be even simpler to fire and forget by `tokio::task:spawn`ing the requests to notification nodes. That way we don't need to handle the response at all. But also that might induce more control flow changes so not sure.

vkgnosis (approved):

sunce86 (approved):

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
```diff
@@ -56,24 +56,31 @@ impl TransactionSubmitting for PublicMempoolApi {
...
-                    });
+                    if can_broadcast {
+                        tracing::debug!(%label, "created transaction with hash: {:?}", tx_hash);
+                        return Ok(TransactionHandle {
+                            tx_hash,
+                            handle: tx_hash,
+                        });
+                    } else {
+                        tracing::debug!(%label, "non-broadcasting node reports transaction creation with hash: {:?}", tx_hash);
+                        futures = rest;
```
MartinquaXD, line 82: nit: At this point it seems less error prone to move the `futures = rest;` to after the `match` statement and remove it from the `Err` case (so we can't do it twice). That way it communicates this always has to happen unless we take an early exit.

MartinquaXD (approved):

nlordell (approved):

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
fedgiac: I implemented this [here](https://github.com/cowprotocol/services/pull/1178).

crates/solver/src/settlement_submission/submitter/public_mempool_api.rs
fedgiac: Very true, while implementing it I spotted a bug in these changes that could cause a panic if the last node in the loop was a notification node: we wouldn't exit the loop if the vector was empty, and `future::select_all` panics if the input is an empty vector.

PR #1173, opened 2023-01-31
CIP-17 ranking with score

sunce86:
Issue tracker: https://github.com/cowprotocol/services/issues/1108

Implements points (3),(4),(5),(7).
Edit: not covering (7) completely, will have a follow up PR for it.

Once this code is merged the transition period starts. It means that we are:
1. Doing the real ranking by objective value (as we did before).
2. At the same time, doing the simulations of ranking by score and saving the simulation results to the solver competition database for analysis.

Once the transition period ends, we can activate feature `auction-rewards` to do the real ranking by score, first in shadow solver, and later on others.


crates/solver/src/settlement_ranker.rs
```diff
@@ -191,17 +184,85 @@ impl SettlementRanker {
...
+                        )),
+                    );
+                    return false;
+                }
+                true
+            });
+
+            // Before sorting, make sure to shuffle the settlements. This is to make sure we don't give
+            // preference to any specific solver when there is an objective value tie.
+            rated_settlements.shuffle(&mut rand::thread_rng());
```
MartinquaXD, line 209: I think we can safely move the `suffle` before the `if cfg!(feature = "CIP17") {} else {}` block, no?

crates/solver/src/settlement_ranker.rs
```diff
@@ -191,17 +184,85 @@ impl SettlementRanker {
...
+                },
+            );
+        }
         Ok((rated_settlements, errors))
     }
 }
 
+// TODO: remove this once CIP-17 is implemented
+// Sort settlements by CIP-17 rules and return hashmap of settlement id to ranking
+fn cip17_ranking(settlements: Vec<RatedSettlement>) -> HashMap<usize, usize> {
```
MartinquaXD, line 253: It's enough to work with references so you don't have to clone all the settlements.

```suggestion
fn cip17_ranking(settlements: Vec<&RatedSettlement>) -> HashMap<usize, usize> {
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -191,17 +184,85 @@ impl SettlementRanker {
...
+    let mut settlements = settlements;
+    settlements.retain(|settlement| settlement.score >= BigRational::zero());
+    settlements.sort_by(|a, b| compare_solutions_cip17(a, b, 0));
```
MartinquaXD, line 256: This should probably use `self.decimal_cutoff` since the `CIP17` version also uses that.
As a side note: do we still need to put settlements into "buckets" with `CIP17`?

crates/model/src/solver_competition.rs
```diff
@@ -74,6 +76,9 @@ pub struct CompetitionAuction {
 pub struct SolverSettlement {
     pub solver: String,
     pub objective: Objective,
+    #[serde(with = "ratio_as_decimal")]
+    pub score: BigRational, // CIP17 score
```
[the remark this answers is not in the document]
nlordell: I think this should be `U256` as it represents an Ether value in wei.

At least this is my interpretation of the CIP.

crates/solver/src/settlement_ranker.rs
sunce86: Obsolete, but anyway, nice catch!

crates/model/src/solver_competition.rs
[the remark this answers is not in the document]
sunce86: Went with `f64` since the `score` have the same semantics as rewards, because it represents the amount of reward the solver is willing to give up for the submitted solution. Since rewards are `f64` then `f64` makes most sense probably.

crates/solver/src/settlement_ranker.rs
sunce86: Fixed.

crates/model/src/solver_competition.rs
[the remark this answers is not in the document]
vkgnosis: I'm unrsolving this to double check that Nick is ok with f64. I'm still not sure.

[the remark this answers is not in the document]
vkgnosis: I'm unrsolving this to double check that Nick is ok with f64. I'm still not sure.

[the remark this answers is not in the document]
sunce86: @nlordell pinging in case Nick missed the message.

vkgnosis (approved):
The question about the type of the score is still open but other than that this is all fine.

crates/model/src/solver_competition.rs
```diff
@@ -77,6 +77,9 @@ pub struct CompetitionAuction {
 pub struct SolverSettlement {
     pub solver: String,
     pub objective: Objective,
+    #[serde(flatten)]
+    pub score: Score, // auction based score
```
nlordell, line 81: Why do we distinguish between the different score types in the settlement? Do we need to share this information (i.e. share with other solvers how their competition are computing scores)?

crates/model/src/solver_competition.rs
```diff
@@ -77,6 +77,9 @@ pub struct CompetitionAuction {
...
+    pub ranking: usize, // auction based ranking
```
nlordell, line 82: At first I thought this was redundant. Could you add to the documentation that this is temporarily needed as the scored settlements are ordered by objective value ATM and this represents how they would be ranked after switching to the new scoring scheme?

crates/solver/src/settlement_ranker.rs
```diff
@@ -200,17 +192,90 @@ impl SettlementRanker {
...
-            self.metrics
-                .settlement_simulation(solver.name(), SolverSimulationOutcome::Success);
-        }
 
+        // Before sorting, make sure to shuffle the settlements. This is to make sure we
+        // don't give preference to any specific solver when there is an
+        // objective value tie.
+        rated_settlements.shuffle(&mut rand::thread_rng());
+
+        if cfg!(feature = "auction-rewards") {
```
nlordell, line 201: I'm not a big fan of `cfg!(feature)` here. Mostly because I think this should be a runtime and not a compile time flag.

nlordell (commented):
Code looks good in general, left some comments for mostly smaller things.

crates/shared/src/http_solver/model.rs
```diff
@@ -173,6 +173,19 @@ impl Interaction for InteractionData {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Score {
+    /// The score used for ranking.
+    Score(f64),
```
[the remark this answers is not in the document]
vkgnosis: Note that this is being discussed in a previous thread. I'm against BigDecimal and BigRational because they have infinite size.

crates/model/src/solver_competition.rs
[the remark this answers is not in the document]
nlordell: > same semantics as rewards

To me, they have different semantics. One is a CoW value (whose computation in the CIP is defined for floating point numbers, and not integers). The score is an Ether amount (and I think we should use either `wei: U256` or `units: BigDecimal`.

crates/shared/src/http_solver/model.rs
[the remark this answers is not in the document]
nlordell: Commented in thread ☝️

crates/model/src/solver_competition.rs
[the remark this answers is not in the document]
nlordell: @vkgnosis makes a valid point that `BigDecimal` is unbounded numerical value. I think we should use `U256` for this as a wei amount in Ether.

crates/solver/src/settlement_ranker.rs
sunce86: Switched to compile time check.

vkgnosis: It's still a compile-time check at the moment with `#[cfg(feature = "auction-rewards")]` . Maybe didn't commit?

vkgnosis: Or maybe misunderstanding. Nick is saying (and I agree) that this should *not* be a compile time check. It should be a normal configuration option.

sunce86: I thought Nick confused and wanted to say the opposite (but i forgot to comment/ask here). 
Originally I wanted to use the feature flag but ended up using the flag only at 1 place in the code, so turning on/off with configuration parameter here makes sense. 
Added.

sunce86:
@nlordell I will merge this PR so I can continue with the work on solver rewards. 
Since the merge goes to feature branch we will have 1 more chance to fix something from this PR if needed.

crates/solver/src/settlement_ranker.rs
nlordell: This is what I meant, I think it should be a configuration flag and _not_ a compile time flag.

PR #1174, opened 2023-02-01
Small native price cache refactor

vkgnosis:
While working with this module I noticed a couple of details that I'd like to change:
- Create a new struct for the update task to group its data.
- Move all other fields into `Inner`.
- Don't explicitly store `Metrics`.

### Test Plan

existing tests

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -45,6 +45,14 @@ struct Inner {
     max_age: Duration,
 }
 
+struct UpdateTask {
+    inner: Weak<Inner>,
+    update_interval: Duration,
+    update_size: Option<usize>,
+    prefetch_time: Duration,
+    concurrent_requests: usize,
+}
+
 #[derive(Debug, Clone)]
 struct CachedPrice {
     price: f64,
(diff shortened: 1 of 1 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -175,6 +190,50 @@ impl Inner {
     }
 }
 
+impl UpdateTask {
+    /// Single run of the background updating process.
+    async fn single_update(&self, inner: &Inner) {
+        let metrics = Metrics::get();
```
vkgnosis, line 196: This is the same as the freestanding function deleted below, except that the outer while loop is in its own function and we use parameters from self.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -175,6 +190,50 @@ impl Inner {
...
+        metrics
+            .native_price_cache_size
+            .set(inner.cache.lock().unwrap().len() as i64);
```
MartinquaXD, line 199: One thing I missed earlier is that the `inner.cache` gets locked twice in quick succession with the introduction of the `native_price_cache_size` metric. However, since this only happens in the background task that doesn't get run very frequently making the code more ugly to save one lock is probably not worth it.
An easy "fix" would be to put the metric update into `sorted_tokens_to_update()` which wouldn't fit well with the function, though.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -194,32 +253,29 @@ impl CachingNativePriceEstimator {
...
+        let update_task = UpdateTask {
+            inner: Arc::downgrade(&inner),
+            update_interval,
+            update_size,
+            prefetch_time,
+            concurrent_requests,
         }
+        .run()
+        .instrument(tracing::info_span!("caching_native_price_estimator"));
+        tokio::spawn(update_task);
```
MartinquaXD, line 268: Should this PR be merged instead of https://github.com/cowprotocol/services/pull/1159? Asking because they seem incompatible.
Also in the other PR I found the separation of the `CachingNativePriceEstimator` and it's update task a bit nicer. By introducing a new function that spawns the update task you only had to pass the arguments related to the task if you actually intend to spawn the task. Don't get me wrong I still think it makes sense to have the update task for the `orderbook` but the separation of concerns is an added bonus IMO.

MartinquaXD (approved):

crates/shared/src/price_estimation/native_price_cache.rs
vkgnosis: Yes. I was considering another refactor later where the task holds the lock and passes the reference around to the functions using it. As you said this is just ugly but doesn't have a performance impact in practice so I wasn't sure about it.

nlordell (approved):
LGTM

PR #1177, opened 2023-02-02
Don't discard orders with pre-interactions

MartinquaXD:
In case an order has pre interactions it's possible that those would transfer enough money to the owner before settling the order in case they currently don't have the required funds.

### Test Plan
Extended existing test cases for the SQL query and the filtering logic.

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -677,6 +677,7 @@ pub struct OrderQuotingData {
     pub sell_amount: BigDecimal,
     pub sell_token_balance: SellTokenSource,
     pub partially_fillable: bool,
+    pub pre_interactions: i32,
 }
 
 /// Returns all limit orders that are currently waiting to be filled sorted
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

crates/database/src/orders.rs
```diff
@@ -677,6 +677,7 @@ pub struct OrderQuotingData {
     pub sell_amount: BigDecimal,
     pub sell_token_balance: SellTokenSource,
     pub partially_fillable: bool,
+    pub pre_interactions: i32,
```
vkgnosis, line 680: Is this the right type for the maximum length of a Postgres list?

MartinquaXD: I initially thought `i64` surely would be the correct type but when I did a local test run I encountered this error:
```
2023-02-02T10:18:44.593Z ERROR limit_order_quoter: autopilot::limit_orders::quoter: limit order quoter update error err=error occurred while decoding column "pre_interactions": mismatched types; Rust type `i64` (as SQL type `INT8`) is not compatible with SQL type `INT4`

Caused by:
    mismatched types; Rust type `i64` (as SQL type `INT8`) is not compatible with SQL type `INT4`
```

vkgnosis: TIL

PR #1180, opened 2023-02-06
Refactor uniswapv2-like liquidity

vkgnosis:
This changes the way we create uniswapv2-like liquidity to not use macros and modules. The motivation is that I wanted to add a custom univ2 liquidity argument, like what we have in the new solvers crate, to other crates. This turned out surprisingly tedious so I made this refactor first.

### Test Plan

CI, double check that I didn't make a typo when moving the init digests, do one manual test


crates/shared/src/sources/uniswap_v2.rs
```diff
@@ -1,42 +1,102 @@
...
+    };
+    let router = contracts::IUniswapLikeRouter::at(web3, router);
+    let factory = router.factory().call().await.context("factory")?;
+    let provider = pair_provider::PairProvider {
+        factory,
+        init_code_digest: init_code_digest.0,
+    };
+    let pool_reader = match pool_reader(source, provider, web3) {
+        Some(inner) => inner,
+        None => return Ok(None),
```
MartinquaXD, line 95: Wouldn't this be an error case? It probably doesn't make sense to be able to get the router but not the reader for a given liquidity source.
Also the implementation details should probably be encapsulated in this file so we could make most of the functions and constants private.

MartinquaXD (approved):

crates/shared/src/sources/uniswap_v2.rs
vkgnosis: Yes it would be an error. This stops mattering in the [follow up PR](https://github.com/cowprotocol/services/pull/1181) so I might not fix it here.

crates/shared/src/sources/uniswap_v2.rs
```diff
@@ -1,42 +1,102 @@
...
-            mainnet_pair_provider.pair_address(&mainnet_pair),
-            addr!("3e8468f66d30fc99f745481d4b383f89861702c6")
-        );
+/// If the baseline source is uniswapv2-like, returns the address of the router
+/// contract and the init code digest needed for calculating pair addresses.
+pub fn from_baseline_source(source: BaselineSource, net_version: &str) -> Option<(H160, H256)> {
+    let (contract, init_code_digest) = match source {
+        BaselineSource::BalancerV2 => None,
+        BaselineSource::ZeroEx => None,
+        BaselineSource::UniswapV3 => None,
```
ennmichael, line 38: `|` would make this slightly shorter.

crates/shared/src/sources/uniswap_v2.rs
```diff
@@ -1,42 +1,102 @@
...
+    web3: &Web3,
+) -> Option<Box<dyn PoolReading>> {
+    let default_reader = DefaultPoolReader {
+        pair_provider,
+        web3: web3.clone(),
+    };
+    match source {
+        BaselineSource::BalancerV2 => None,
+        BaselineSource::ZeroEx => None,
+        BaselineSource::UniswapV3 => None,
```
ennmichael, line 69: Here too.

crates/shared/src/sources/uniswap_v2.rs
```diff
@@ -1,42 +1,102 @@
...
+        BaselineSource::UniswapV2 => Some(Box::new(default_reader)),
+        BaselineSource::Honeyswap => Some(Box::new(default_reader)),
+        BaselineSource::SushiSwap => Some(Box::new(default_reader)),
+        BaselineSource::Baoswap => Some(Box::new(default_reader)),
```
ennmichael, line 73: Same here.

crates/shared/src/sources/uniswap_v2/pool_fetching.rs
```diff
@@ -54,6 +52,17 @@ pub trait PoolReading: Sized + Send + Sync {
...
+impl PoolReading for Box<dyn PoolReading> {
+    fn read_state(
+        &self,
+        pair: TokenPair,
+        batch: &mut Web3CallBatch,
+        block: BlockId,
+    ) -> BoxFuture<'_, Result<Option<Pool>>> {
+        (**self).read_state(pair, batch, block)
+    }
+}
```
ennmichael, line 64: Do we really need `Arc<Box>`? This seems weird.

ennmichael (commented):
I'm for anything that gets rid of macros, God bless.

ennmichael (approved):

crates/shared/src/sources/uniswap_v2/pool_fetching.rs
vkgnosis: There is no Arc<Box>

Issue #901, opened 2022-12-02
Allow running `solver` binary without any built-in solvers


MartinquaXD:
If someone is using the `solver` crate (the current driver) to work on a new solver it would be great to allow running the binary only with the new solver enabled to avoid having the logs spammed by solvers you are not interested in.
This is currently not possible because the CLI expects at least one argument to be passed to `--solvers` (`--solvers ""` doesn't work).

Note that this is not a high priority and should no longer be a problem with the new `driver`.

devanoneth:
FYI @MartinquaXD this wasn't automatically closed but can be now.

PR #1181, opened 2023-02-07
Allow specifying custom univ2 like liquidity in command line arguments

vkgnosis:
Allows us to configure custom univ2 instances. This is useful in e2e tests. A custom univ2 instance is the address of a router contract along with the init_hash for calculating pool addresses.

As a follow up it could allow us to get rid of all of the univ2 fork hardcodes that we currently have in the code. We would remove them as baseline sources and move them into the new argument in our cluster configuration.

### Test Plan

manually test and observe after merging

### DIFF
--- a/crates/shared/src/sources.rs
+++ b/crates/shared/src/sources.rs
@@ -7,22 +7,17 @@ pub mod uniswap_v3;
 pub mod uniswap_v3_pair_provider;
 
 use {
-    self::uniswap_v2::{
-        pair_provider::PairProvider,
-        pool_fetching::{Pool, PoolFetching},
-    },
-    crate::{ethrpc::Web3, recent_block_cache::Block},
+    self::uniswap_v2::pool_fetching::{Pool, PoolFetching},
+    crate::recent_block_cache::Block,
     anyhow::{bail, Result},
     model::TokenPair,
-    std::{
-        collections::{HashMap, HashSet},
-        sync::Arc,
-    },
+    std::{collections::HashSet, sync::Arc},
 };
 
 #[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, clap::ValueEnum)]
 #[clap(rename_all = "verbatim")]
 pub enum BaselineSource {
+    None,
     UniswapV2,
     Honeyswap,
     SushiSwap,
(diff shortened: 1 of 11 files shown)


crates/shared/src/sources.rs
```diff
@@ -7,22 +7,17 @@ pub mod uniswap_v3;
...
-        collections::{HashMap, HashSet},
-        sync::Arc,
-    },
+    std::{collections::HashSet, sync::Arc},
 };
 
 #[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, clap::ValueEnum)]
 #[clap(rename_all = "verbatim")]
 pub enum BaselineSource {
+    None,
```
MartinquaXD, line 20: Why would we need a `None` variant here? Shouldn't an optional `BaselineSource` rather be `Option<BaselineSource>`?

vkgnosis: I forgot to explain this in the PR description. The problem is that in our Clap arguments we specify baselines sources as `Option<Vec<BaselineSource>>`. I have not found a way to have this parsed as `Some(vec![])` to indicate that we don't want any hardcoded baseline source.

MartinquaXD: I remember running into the same [issue](https://github.com/cowprotocol/services/issues/901). I did a bit of research just now and all the documentation suggests that it should work out of the box (which it apparently doesn't).
But I was able to make it work by specifying `num_args = 0..` in the clap derive macro:
```rust
    /// Which Liquidity sources to be used by Price Estimator.
    #[clap(
        long,
        env,
        ignore_case = true,
        use_value_delimiter = true,
        num_args = 0..
    )]
    pub baseline_sources: Option<Vec<BaselineSource>>,
```

```
cargo run -p solver -- --baseline-sources
...
base_tokens: []
baseline_sources: Some([])
pool_cache_blocks: 10
...
```

```
cargo run -p solver -- --baseline-sources "UniswapV2"
...
base_tokens: []
baseline_sources: Some([UniswapV2])
pool_cache_blocks: 10
...
```

MartinquaXD: I just remembered that we configure everything with env variables and with those I wasn't able to make it work yet. 🤔

nlordell: >  I have not found a way to have this parsed as Some([]) to indicate that we don't want any hardcoded baseline source.

You can't. It's annoying. I opened an issue/discussion in the past, but never got anywhere.

nlordell (approved):
Welcome change! We can try out Pancake swap in staging with this.

crates/shared/src/sources.rs
MartinquaXD: > You can't. It's annoying.

@nlordell  Did you read my comment? It's certainly possible. Just not with environment variables.

nlordell: I did! Sorry for the short reply, what I meant was to confirming that I came to the same conclusion a while ago (can't specify an empty list with environment variables).

I tried to dig up my clap-rs discussion in their repo but I can't find it anymore.

MartinquaXD (approved):

crates/shared/src/sources.rs
ennmichael: It sounds like a problem that would be solved by not reusing the CLI args DTO for the business logic.

ennmichael (approved):

crates/shared/src/sources.rs
MartinquaXD: > It sounds like a problem that would be solved by not reusing the CLI args DTO for the business logic.

How so? If `clap` doesn't properly support differentiating between passing no argument and passing an empty argument then we still can't "recover" the difference by using more intermediary structs.

vkgnosis: If we had a separate type for the argument parsing then we could convert into another type that doesn't have the `None` variant. `Vec<ArgEnum>` gets turned into `Vec<DomainEnum>` and the `ArgEnum::None`s are dropped. I can see the appeal of it but won't include it in this PR.

PR #1186, opened 2023-02-07
Add pre-interactions support for foreign liquidity orders

sunce86:
SeaSolver wants to integrate their own liquidity orders, but need pre-interactions support because, for these liquidity orders to work, they first need to transfer funds to their trading contract and then execute the foreign liquidity order by using our `foreign_liquidity_orders` field.

Since a pretty low hanging fruit decided to add it regardless if they use it at the end or end up using custom interactions after all.


vkgnosis (approved):

crates/shared/src/http_solver/model.rs
```diff
@@ -258,6 +258,7 @@ pub struct NativeLiquidityOrder {
     pub data: OrderData,
     #[serde(flatten)]
     pub signature: Signature,
+    pub interactions: Option<Interactions>,
```
MartinquaXD, line 261: This might be a bit nicer to work with:
```rust
#[serde(default)]
pub interactions: Interactions,
```

MartinquaXD (approved):

crates/shared/src/http_solver/model.rs
sunce86: Thanks, added.

PR #1187, opened 2023-02-08
Update colocated driver API details

vkgnosis:
While writing the autopilot-driver e2e test I noticed some mistakes in the api. This PR changes the driver's rust API types and populates the openapi spec with the details of the Order struct.

This leaves some fields that were previously populated unset. Autopilot should not provide this information. Driver should fetch it itself, which I have not implemented.

### Test Plan

CI

### DIFF
--- a/crates/driver/src/infra/api/routes/solve/dto/auction.rs
+++ b/crates/driver/src/infra/api/routes/solve/dto/auction.rs
@@ -6,23 +6,25 @@ use {
     itertools::Itertools,
     serde::Deserialize,
     serde_with::serde_as,
-    std::{collections::HashMap, str::FromStr},
+    std::collections::HashMap,
 };
 
 impl Auction {
     pub fn into_domain(self) -> Result<competition::Auction, Error> {
         Ok(competition::Auction {
-            id: Some(FromStr::from_str(&self.id).map_err(|_| Error::InvalidAuctionId)?),
+            id: Some((self.id as u64).into()),
             tokens: self
-                .tokens
+                .prices
                 .into_iter()
-                .map(|(address, token)| competition::auction::Token {
-                    decimals: token.decimals,
-                    symbol: token.symbol,
-                    address: address.into(),
-                    price: token.reference_price.map(Into::into),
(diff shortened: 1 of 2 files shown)


crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -6,23 +6,25 @@ use {
...
-                    decimals: token.decimals,
-                    symbol: token.symbol,
-                    address: address.into(),
-                    price: token.reference_price.map(Into::into),
-                    available_balance: token.available_balance,
-                    trusted: token.trusted,
+                // TODO: Populate currently hardcoded fields.
+                .map(|(key, value)| competition::auction::Token {
+                    decimals: None,
+                    symbol: None,
```
nlordell, line 22: I don't think this is necessarily part of the auction. To me, token metadata is "side-car" data (much like liquidity) which the driver "augments" before sending to the solver engine.

crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -6,23 +6,25 @@ use {
...
+                    address: key.into(),
+                    price: Some(value.into()),
+                    available_balance: 0.into(),
+                    // TODO: Does autopilot communicate this to drivers?
+                    trusted: false,
```
nlordell, line 27: Personally, I think this should be determined by the `autopilot` (as this is a "core protocol" concept).

crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -113,7 +115,8 @@ impl Auction {
                     })
                 })
                 .try_collect()?,
-            gas_price: self.effective_gas_price.into(),
+            // TODO: Populate hardcoded value.
+            gas_price: eth::U256::from(0).into(),
```
nlordell, line 119: Given the new auction model, I agree that this shouldn't be sent by the `autopilot`.

nlordell (approved):
LGTM

PR #1188, opened 2023-02-08
Allow configurable settlement contract and native token address

vkgnosis:
Needed for e2e tests.

### Test Plan

CI, new functionality is not yet used or tested

### DIFF
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -438,9 +438,13 @@ impl Display for Arguments {
         display_option(
             f,
             "settlement_contract_address",
-            &self.settlement_contract_address,
+            &self.settlement_contract_address.map(|a| format!("{a:?}")),
         )?;
-        display_option(f, "native_token_address", &self.native_token_address)?;
+        display_option(
+            f,
+            "native_token_address",
+            &self.native_token_address.map(|a| format!("{a:?}")),
+        )?;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


nlordell (approved):

nlordell:
Do we do a `GPv2Settlement::deployed()` anywhere in code?

vkgnosis:
We're not supposed to. Maybe a year ago I remember going through the code and removing all such instances because they were hard to maintain. I just checked by grepping for `::deployed` and it seems still fine. Only happens in tests.

crates/shared/src/arguments.rs
```diff
@@ -427,6 +435,12 @@ impl Display for Arguments {
             self.balancer_pool_deny_list
         )?;
         display_secret_option(f, "solver_competition_auth", &self.solver_competition_auth)?;
+        display_option(
+            f,
+            "settlement_contract_address",
+            &self.settlement_contract_address,
+        )?;
+        display_option(f, "native_token_address", &self.native_token_address)?;
```
MartinquaXD, line 443: Didn't `H160` need to use `Debug` formatting to get the complete readable address?

nlordell: Nice catch!

vkgnosis: fixed

MartinquaXD (approved):

crates/shared/src/arguments.rs
```diff
@@ -292,6 +292,14 @@ pub struct Arguments {
     /// The number of pools to initially populate the UniswapV3 cache
     #[clap(long, env, default_value = "100")]
     pub max_pools_to_initialize_cache: u64,
+
+    /// Override address of the settlement contract.
+    #[clap(long, env)]
+    pub settlement_contract_address: Option<H160>,
+
+    /// Override address of the settlement contract.
```
sunce86, line 300: Comment copy paste

PR #1191, opened 2023-02-09
Fix duration overflow bug

ennmichael:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f36f362ce1571bcb34ed0debab7afc6c

The `u64::MAX` limit is not the right value to use here, and can cause panics. This showed up in staging.

### Test Plan

Tests still pass. After applying to staging we'll see if it goes away.


josojo (approved):
🎉

vkgnosis (approved):

MartinquaXD:
[`Duration::try_from_secs_f64()`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.try_from_secs_f64) is now stabilized so we can entirely replace the code like the comment suggests.

PR #1193, opened 2023-02-09
Deny unknown fields in colocation APIs

vkgnosis:
This helps catch cases where the types used in serialization don't line up.

Also added some `camelCase` annotations where they were missing and removed one of of the deserialize/serialize derives when the type is only used in one direction.

### Test Plan

CI, ran existing colocation tests

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -8,7 +8,7 @@ pub mod quote {
         serde::{Deserialize, Serialize},
     };
 
-    #[derive(Clone, Debug, Default, Deserialize, Serialize)]
+    #[derive(Clone, Debug, Default, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Request {
         pub sell_token: H160,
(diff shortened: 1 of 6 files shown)


sunce86 (approved):

MartinquaXD:
Not sure I agree with this change. It's definitely nice to catch errors while building the new driver but it would make it harder to remove fields later on when the new driver is already deployed. Same goes for `deny_unknown_fields` on config arguments.

vkgnosis:
When we intentionally remove fields later, we can continue to use `deny_unknown_fields` and temporarily keep the field in the struct.  I think this is better than not catching these errors. Sending fields that don't exist should be treated as a bug.

For example, if we wanted to remove an existing field `foo: Bar`, we could change it to

```rust
#[serde(rename = "foo")]
_: json::Value
```

until all users of the API have been updated with the change.

MartinquaXD (approved):

nlordell (approved):
Makes sense. I like these "stricter" APIs for internal requests.

One thing to note is that `deny_unknown_fields` does not play nice with `flatten`. I don't see an issue here, just something to keep in mind in case someone runs into this in the future.

PR #1197, opened 2023-02-09
Make solvers crate usable as library

vkgnosis:
This allows it to be easily used in e2e tests.

### Test Plan

no logic change

### DIFF
--- a/crates/solvers/src/lib.rs
+++ b/crates/solvers/src/lib.rs
@@ -1,0 +1,12 @@
+// TODO remove this once the crate stabilizes a bit.
+#![allow(dead_code)]
+
+mod api;
+pub mod boundary;
+mod cli;
+mod config;
+mod domain;
+pub mod run;
+#[cfg(test)]
+mod tests;
+mod util;
--- a/crates/solvers/src/run.rs
+++ b/crates/solvers/src/run.rs
@@ -1,0 +1,43 @@
+#[cfg(unix)]
+use tokio::signal::unix::{self, SignalKind};
+use {crate::domain::baseline, clap::Parser, std::net::SocketAddr, tokio::sync::oneshot};
+
+pub async fn run(args: impl Iterator<Item = String>, bind: Option<oneshot::Sender<SocketAddr>>) {
+    let args = crate::cli::Args::parse_from(args);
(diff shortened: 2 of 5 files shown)


crates/solvers/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+// TODO remove this once the crate stabilizes a bit.
+#![allow(dead_code)]
+
+mod api;
+pub mod boundary;
+mod cli;
+mod config;
+mod domain;
+pub mod run;
```
MartinquaXD, line 9: Maybe we should only reexport the `run` function from the `run` module. So we don't have to specify `crate::run::run()`. (seems a bit weird)

MartinquaXD (approved):

crates/solvers/src/run.rs
```diff
@@ -0,0 +1,43 @@
+#[cfg(unix)]
+use tokio::signal::unix::{self, SignalKind};
+use {crate::domain::baseline, clap::Parser, std::net::SocketAddr, tokio::sync::oneshot};
+
+pub async fn run(args: impl Iterator<Item = String>, bind: Option<oneshot::Sender<SocketAddr>>) {
```
nlordell, line 5: bikeshed (so feel free to ignore): I would put this in `lib`. It seems like a "top-level" function (so in the `[lib.path]` just like `main` goes in the `[[bin.path]]`).

nlordell (approved):

crates/solvers/src/run.rs
vkgnosis: I find it a little cleaner in it's own module. One thing I noticed in Driver, is that we were importing things for the main function in lib.rs, and then accidentally `use`ing them from other modules. For example, in lib.rs we would `use ethcontract::Address` because it is needed for `fn run()` and then in some unrelated module `foo/bar/mod.rs` we would `use crate::Address` instead of `use ethcontract::Address`. In lib.rs you can't distinguish between "import for this module" or "import for all sub modules".

nlordell: Its a fair point and I definitely agree that re-using parent module `use` statements is brittle (it can cause unrelated changes in refactors).

sunce86 (approved):

PR #1198, opened 2023-02-09
Make config file content accessible without file on disk

vkgnosis:
Currently the configuration file must be passed as a path to the driver and solvers binary. This is annoying in e2e tests because you need to create the file on disk to pass it in. With this change the config file content can be passed directly inside of an argument. This might also be useful for our kubernetes setup.

### Test Plan

colocation e2e test

### DIFF
--- a/crates/driver/src/infra/cli.rs
+++ b/crates/driver/src/infra/cli.rs
@@ -24,6 +24,11 @@ pub struct Args {
     /// Path to the driver configuration file. This file should be in TOML
     /// format. For an example see
     /// https://github.com/cowprotocol/services/blob/main/crates/driver/example.toml.
-    #[clap(long, env)]
-    pub config: PathBuf,
+    #[clap(long, env, conflicts_with = "config_string")]
+    pub config_path: Option<PathBuf>,
+
+    /// Like `config` but reads the configuration directly from the argument
+    /// instead of a file.
+    #[clap(long, env, conflicts_with = "config_path")]
+    pub config_string: Option<String>,
 }
--- a/crates/driver/src/run.rs
+++ b/crates/driver/src/run.rs
@@ -37,7 +37,12 @@ pub async fn run(
 ) {
     let args = cli::Args::parse_from(args);
     boundary::initialize_tracing(&args.log);
-    let config = config::file::load(&args.config).await;
+    let config = match (&args.config_string, &args.config_path) {
+        (Some(string), None) => config::file::load_string(string),
+        (None, Some(path)) => config::file::load_path(path).await,
+        (None, None) => panic!("specify --config-string or --config-path"),
+        (Some(_), Some(_)) => unreachable!(),
+    };
 
     let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
     let eth = ethereum(&config, &args).await;
--- a/crates/driver/src/infra/config/file/load.rs
+++ b/crates/driver/src/infra/config/file/load.rs
@@ -12,12 +12,21 @@ use {
 /// # Panics
 ///
 /// This method panics if the config is invalid or on I/O errors.
-pub async fn load(path: &Path) -> infra::Config {
+pub async fn load_path(path: &Path) -> infra::Config {
     let data = fs::read_to_string(path)
         .await
         .unwrap_or_else(|e| panic!("I/O error while reading {path:?}: {e:?}"));
-    let config: file::Config = toml::de::from_str(&data)
-        .unwrap_or_else(|e| panic!("TOML syntax error while reading {path:?}: {e:?}"));
+    load_string(&data)
+}
+
+/// Load the driver configuration from a string
+///
+/// # Panics
+///
+/// This method panics if the config is invalid.
+pub fn load_string(string: &str) -> infra::Config {
+    let config: file::Config =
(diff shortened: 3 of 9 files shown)


crates/driver/src/infra/cli.rs
```diff
@@ -25,5 +25,10 @@ pub struct Args {
     /// format. For an example see
     /// https://github.com/cowprotocol/services/blob/main/crates/driver/example.toml.
     #[clap(long, env)]
-    pub config: PathBuf,
+    pub config_path: Option<PathBuf>,
+
+    /// Like `config` but reads the configuration directly from the argument
+    /// instead of a file.
+    #[clap(long, env)]
+    pub config_string: Option<String>,
```
MartinquaXD, line 33: It's bit unfortunate that this would introduce another state that needs to be handled.
On the one hand I think creating a file before running a test doesn't seem super annoying on the other hand I'm a bit concerned about temporary files not getting cleaned up when a test panics.
Do you think it would make sense to use `#[cfg(test)]` here? Then we could always require the config string to be passed in directly for tests and still expect the path to be passed in for running the binary in prod.
Not sure about that, though.

vkgnosis: cfg-test doesn't work, because when we depend on the driver crate from the e2e crate, the driver crate is not compiled with cfg-test. It is a normal dependency. Crates are only compiled with cfg-test when their own tests run. 

The extra state is minor imo. It's two more lines in main and this could be useful for the kubernetes setup too. Creating temporary config files on disk is no big deal either, but in general I don't like it when you are forced to create files. I prefer it if e2e tests can be run in memory as much as possible.

crates/driver/src/run.rs
```diff
@@ -37,7 +37,11 @@ pub async fn run(
 ) {
     let args = cli::Args::parse_from(args);
     boundary::initialize_tracing(&args.log);
-    let config = config::file::load(&args.config).await;
+    let config = match (&args.config_string, &args.config_path) {
+        (Some(string), None) => config::file::load_string(string),
+        (None, Some(path)) => config::file::load_path(path).await,
+        _ => panic!("specify exactly one of config_string and config_file"),
```
nlordell, line 43: nit: I think we should use `clap` to make the options mutually exclusive if we go down this path. It should be possible and give better command line help messages when it happens.

vkgnosis: Yes that would be better. I didn't know it was possible. Will check clap documentation for this.

vkgnosis: I think the clap way of doing this is through `group` https://docs.rs/clap/latest/clap/builder/struct.Command.html#method.group but that isn't supported in the drive api yet.

vkgnosis: Another way of doing it would be to have one argument with a custom parsing function. Then it could be used like `--config=path=/path/to/config` and `--config=contents{foo = bar; ... }`. Not sure if that is worth the extra code.

nlordell: You should be able to use `conflicts_with`:

```rust
#[derive(Debug, Parser)]
struct Foo {
    #[clap(short, conflicts_with = "b")]
    a: bool,
    #[clap(short, conflicts_with = "a")]
    b: bool
}
```

crates/driver/src/infra/cli.rs
nlordell: Personally, I'm not a huge fan of the added flag (my rationale is that we have two flags that do the same thing just for testing). I would have preferred using `tempfile` crate for example for tests (as then the E2E is more "complete" in that it checks configuration deserialization from files that get cleaned up by your OS anyway).

That being said, I don't want to bikeshed too much on this issue.

crates/driver/src/run.rs
nlordell: In theory you can use URIs and use a `data://` URI to encode it directly or `file://` URI to read from disk. Also not sure it is worth the extra code.

vkgnosis: `conflicts` doesn't enforce that at least one is set but I've added it.

crates/driver/src/infra/cli.rs
vkgnosis: Another way of accomplishing this would be to pass a file descriptor instead of a file string. `tempfile` does give us a descriptor and the descriptor is available to the component we create because it is the same process.

nlordell: > you can run into problems with not picking a unique name, another process deleting the file, the file not getting cleaned up.

I feel like, while `NamedTempFile` is not perfect, I don't expect us to run into these issues practically (as the temp files are very short lived). That being said, running into one of these in CI would be very difficult to debug...

nlordell (approved):

nlordell:
In the `driver` e2e tests, we already have a mechanism for creating random temporary files. Maybe we can re-use it for this as well?

I'm approving to not bikeshed more, but I am not super happy about the adding a new CLI flag just for testing.

crates/driver/src/infra/cli.rs
nlordell: We can also use the `NamedTempFile::new_in` in the Cargo `target` dir. There is no reasonable reason why another process would mess with that, and at worst you bloat your `target` dir.

PR #1199, opened 2023-02-10
Move fee scaling to autopilot

### LINKED ISSUE #1038: Move Scoring Fee Amount Computation to `autopilot`.
Currently, the fee amount we use for scoring is computed in two places:
- the `full_fee_amount` that is tied to the order
- which is multipled by some `scoring_fee_factor` in the `solver` binary

Moving forward with solver+driver colocation, the computation of the fee used for scoring should move to the `autopilot` (so that we don't need to ensure all drivers are configured with the fee scoring factors).

MartinquaXD:
Fixes #1038 

To not require all driver to be in sync with their `fee_objective_scaling_factor` we want to scale the fee amount on the `autopilot` side.

Working on this issue really shows why one should avoid using the same type for different purposes even if it's convenient at the time. With the `autopilot` pre-scaling the fee we no longer need have the `full_fee_amount` as part of the `OrderMetadata` that gets put into the auction because it doesn't get read anywhere outside the DB module. But because we implemented inserting orders in the DB in terms of that same (API) `Order` type we'll have to refactor DB insertion again to get rid of the redundant field.

### Notes

I renamed a bunch of fields: mostly `unscaled_subsidized_fee -> earned_fee` and `scaled_unsubsidized_fee -> solver_fee`.

### Test Plan
Updated existing tests to see that objective value computation uses the new field.

### Release notes
Requires an update of our configurations.


crates/solver/src/settlement.rs
```diff
@@ -140,14 +140,20 @@ impl Trade {
...
 
-    pub fn executed_unscaled_subsidized_fee(&self) -> Option<U256> {
-        self.compute_fee_execution(self.order.data.fee_amount)
+    /// Returns the actual fees taken by the protocol.
+    pub fn executed_earned_fee(&self) -> Option<U256> {
+        let surplus_fee = match &self.order.metadata.class {
+            // TODO adjust this when limit orders become partially fillable.
+            OrderClass::Limit(details) => details.surplus_fee.unwrap_or_default(),
+            _ => 0.into(),
+        };
```
MartinquaXD, line 155: This was not respected before. But it doesn't matter much, since this function was only used in debug log statements.

crates/shared/src/db_order_conversions.rs
```diff
@@ -54,6 +54,14 @@ pub fn full_order_into_model_order(order: database::orders::FullOrder) -> Result
         sender: onchain_user,
         placement_error: onchain_placement_error,
     });
+    let full_fee_amount =
+        big_decimal_to_u256(&order.full_fee_amount).context("full_fee_amount is not U256")?;
+    let solver_fee = match &class {
+        OrderClass::Limit(limit) => limit.surplus_fee.context("missing surplus_fee")?,
+        OrderClass::Market => full_fee_amount,
+        OrderClass::Liquidity => full_fee_amount,
+    };
```
MartinquaXD, line 63: Not really sure what to put here (`full_fee_amount` or `fee_amount`).
Some of the earlier code suggested to use `full_fee_amount` and would have used the subsidized amount (i.e.`surplus_fee` or `fee_amount`). I think this was a bug so it's probably good that this value gets initialized only once now.
Also different values were used for liquidity vs regular orders which didn't make sense to me.

vkgnosis:
The issue is tagged cololocation but this PR is changing the existing non colocated code, for example the solver crate. Would it be useful to implement this *only* for colocation? Meaning the new APIs for autopilot <-> driver communication?

MartinquaXD:
> Would it be useful to implement this only for colocation? Meaning the new APIs for autopilot <-> driver communication?

My reasoning was that this change makes sense with our without the new driver. The naming in the current code is very confusing (at least for me) and seems to be inconsistent so why not fix it in the current code even if it's supposed to be replaced soon-ish.
But I guess you are thinking about increased potential for regressions?

vkgnosis:
I was just wondering. Maybe you didn't consider it. The downside of changing the current code is that it is more work and that is has a chance to introduce a bug. I'm fine with changing it in the current code if we feel it's worth it.

vkgnosis (approved):
Looks fine. Won't break shadow competition because the new field is tagged default.

nlordell:
> `unscaled_subsidized_fee -> earned_fee`

Nice. These names are concise and descriptive.

crates/model/src/order.rs
```diff
@@ -253,6 +258,12 @@ pub struct OrderData {
     pub buy_amount: U256,
     pub valid_to: u32,
     pub app_data: AppId,
+    /// Fees that will be taken in terms of `sell_token`.
+    /// This is 0 for liquidity orders as they should only ever be used to
+    /// improve the price of a regular order and should not be settled on
+    /// their own. This is 0 for limit orders as their fee gets taken from
+    /// the surplus. This `OrderMetadata::full_fee_amount` modulo possible
```
nlordell, line 265:
```suggestion
    /// the surplus. This is `OrderMetadata::full_fee_amount` modulo possible
```

crates/shared/src/db_order_conversions.rs
nlordell: Intuitively, I would have expected this to be:
```rust
    let solver_fee = match &class {
        OrderClass::Limit(limit) => full_fee_amount,
        OrderClass::Market => full_fee_amount,
        OrderClass::Liquidity => big_decimal_to_u256(&order.fee_amount).context("...")?,
    };
```

The reason here is:
- Full fee amount represents the fee that we actually think the order will use to settle. This is what we want to use for objective value computation (otherwise a subsidy will make the order count less towards an solution's objective value despite having the same complexity - notably, CoW token holder subsidized orders are less attractive). This is the same for limit orders as `surplus_fee` contains subsidies that should not be deducted from the `solver_fee`.
- Liquidity orders should use `fee_amount` (and I think they should probably not get scaled at all). The rationale here is that their fee doesn't contribute to the objective value unless they bribe the protocol with fees.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -283,6 +291,27 @@ impl SolvableOrdersCache {
...
 
+/// Applies fee objective scaling and discard all orders where an overflow
+/// occurred.
+fn apply_fee_objective_scaling_factor(
+    mut orders: Vec<Order>,
+    scaling_factor: &BigRational,
+) -> Vec<Order> {
+    orders.retain_mut(|order| {
+        let fee = u256_to_big_rational(&order.metadata.solver_fee);
+        let scaled_fee = fee * scaling_factor;
```
nlordell, line 302: I don't think we should scale liquidity order fees. Scaling is for artificially inflating the economic viability of a specific order, which we don't want to do for liquidity orders as it could make them be preferred over other "better" solutions over an AMM for example (by increasing the objective value more than the actual benefit they provide).

MartinquaXD: [7b5b8ef](https://github.com/cowprotocol/services/pull/1199/commits/7b5b8ef19ef99244054b02b3c0104834e9b3ce5f)

crates/autopilot/src/solvable_orders.rs
```diff
@@ -283,6 +291,33 @@ impl SolvableOrdersCache {
...
+            // A liquidity order should only be used if it can increase a settlement's total
+            // surplus more than it costs to execute it.
+            // That's why we don't want to promote it artificially by scaling its fee.
+            return true;
+        }
+        let fee = u256_to_big_rational(&order.metadata.solver_fee);
+        let scaled_fee = fee * scaling_factor;
+        let scaled_fee = match big_rational_to_u256(&scaled_fee) {
+            Ok(fee) => fee,
+            Err(_) => return false,
```
nlordell, line 311: This should never happen in practice (so not requesting change here), but semantically, I would expect this to be a saturating operation and not a checked one (i.e. overflowing means `fee == U256::MAX` and not order doesn't get considered).

That being said, since we don't _expect_ this to happen, I think it makes slightly more sense to filter the order out, so we we a log and increment a metric (as it would indicate something terribly wrong with our current assumptions).

crates/model/src/order.rs
```diff
@@ -585,8 +596,22 @@ pub struct OrderMetadata {
     #[serde(flatten)]
     pub class: OrderClass,
     pub settlement_contract: H160,
+    /// This is 0 for liquidity orders as they should only ever be used to
```
nlordell, line 599:
```suggestion
    /// This is `fee_amount` for liquidity orders as they should only ever be used to
```

crates/solver/src/settlement.rs
nlordell: Nice catch!

nlordell (approved):
LGTM

crates/model/src/order.rs
```diff
@@ -253,6 +258,12 @@ pub struct OrderData {
     pub buy_amount: U256,
     pub valid_to: u32,
     pub app_data: AppId,
+    /// Fees that will be taken in terms of `sell_token`.
+    /// This is 0 for liquidity orders as they should only ever be used to
```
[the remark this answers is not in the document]
MartinquaXD: Copy and paste error? This comment explains the value of `fee_amount` which is obviously identical to `fee_amount`. 😄 
A more helpful comment could be:
```
This is 0 for liquidity orders unless the owner bribes the protocol
```

crates/model/src/order.rs
MartinquaXD: How we populate `full_fee_amount` didn't change. We still default initialize it for liquidity orders. But I guess in light of the discussion about bribing the protocol this value should never be less than `fee_amount`.

MartinquaXD: Since this is a slight change coming from your comment could you please check if you agree with this change: [c4e6970](https://github.com/cowprotocol/services/pull/1199/commits/c4e6970eef0c73e2cd10ba82a5236b8a2ce6144d)

nlordell: Looks good.

crates/model/src/order.rs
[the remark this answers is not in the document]
nlordell: Comments look good.

PR #1200, opened 2023-02-10
Add argument to enable autopilot colocation

vkgnosis:
### Test Plan

part of the colocation e2e test


crates/autopilot/src/driver_api.rs
```diff
@@ -3,17 +3,29 @@ use {
     anyhow::{anyhow, Context, Result},
     reqwest::Client,
     shared::http_client::response_body_with_size_limit,
+    std::time::Duration,
     url::Url,
 };
 
 const RESPONSE_SIZE_LIMIT: usize = 10_000_000;
+const RESPONSE_TIME_LIMIT: Duration = Duration::from_secs(60);
```
MartinquaXD, line 11: I think we should probably have different timeouts for different calls. (`/quote`, `/solve`, `/settle`)

crates/autopilot/src/lib.rs
```diff
@@ -569,22 +569,22 @@ pub async fn main(args: arguments::Arguments) -> ! {
         args.limit_order_price_factor
             .try_into()
             .expect("limit order price factor can't be converted to BigDecimal"),
-        true,
+        !args.enable_colocation,
```
MartinquaXD, line 572: nit: The argument is currently named `store_in_db`. Maybe it makes sense to make it refer to the colocated driver to avoid any confusion.

MartinquaXD (approved):

crates/autopilot/src/driver_api.rs
vkgnosis: In general, calls to the endpoints should probably enforce their own timeouts through tokio::timeout. This constant was meant as a safety fallback for all routes.

crates/autopilot/src/lib.rs
vkgnosis: The argument controls whether the solvable_orders module should store the auction in the db, so from that pov it says what it does. Whether it should store in the DB depends on whether colocation is enabled. But the module itself doesn't need to know that. I can see either perspective being reasonable. Will think about it again but not sure what is better.

sunce86, line 572: Why is auction not stored in case of colocation?

vkgnosis: Because the colocation run loop (run_loop.rs) stores the auction instead.

sunce86: Thanks.

sunce86 (approved):

crates/autopilot/src/arguments.rs
```diff
@@ -167,6 +167,14 @@ pub struct Arguments {
     /// limit orders.
     #[clap(long, env, use_value_delimiter = true)]
     pub quoting_strategies: Vec<QuotingStrategy>,
+
+    /// Enable the colocation run loop.
+    #[clap(long, env)]
+    pub enable_colocation: bool,
+
+    /// Driver base URLs for colocation.
```
ennmichael, line 175:
```suggestion
    /// Driver base URLs.
```

"The colocation" will just be the default state of affairs, so I don't think this is needed. We should more or less just forget the term as soon as we're done with it.

crates/autopilot/src/lib.rs
ennmichael: You're right @vkgnosis

crates/autopilot/src/lib.rs
```diff
@@ -607,11 +607,31 @@ pub async fn main(args: arguments::Arguments) -> ! {
...
+                .drivers
+                .into_iter()
+                .map(driver_api::Driver::new)
+                .collect(),
+            current_block: current_block_stream,
+            web3,
+            network_block_interval: _network_time_between_blocks,
+        };
+        run.run_forever().await;
+        panic!("run loop exited");
```
ennmichael, line 632: Slightly slightly better
```suggestion
        unreachable!("run loop exited");
```

crates/autopilot/src/lib.rs
```diff
@@ -607,11 +607,31 @@ pub async fn main(args: arguments::Arguments) -> ! {
...
+    } else {
+        let result = serve_metrics.await;
+        panic!("serve_metrics exited {result:?}");
```
ennmichael, line 635: Same here.

ennmichael (approved):

nlordell (approved):

PR #1202, opened 2023-02-10
Make WETH address optional, use chain ID parameter

nlordell:
As mentioned by @ennmichael in https://github.com/cowprotocol/infrastructure/pull/90#discussion_r1092903722, making the WETH address mandatory to configure is a bit annoying (especially considering that we already have addresses for this contract per chain ID).

In order to be able to get contract addresses, we need to know the chain ID that the solver is running from. Since we don't make use of Ethereum RPC requests at all (if this changes, then we should adjust accordingly). I decided to make a required CLI flag since all of our current solvers need this:
- For our DEXag solvers (0x, ParaSwap, 1Inch) this is needed as it gets appended to the base URL for requests
- For legacy HTTP solvers, this is needed as it gets included in the metadata
- For baseline and naive solvers, they need this for the default WETH address

### Test Plan

Rust compiler, E2E test covers parsing this.


crates/solvers/src/domain/eth/chain.rs
```diff
@@ -0,0 +1,40 @@
+use ethereum_types::U256;
+
+/// A supported Ethereum Chain ID.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ChainId {
+    Mainnet = 1,
+    Goerli = 5,
+    Gnosis = 100,
```
ennmichael, line 8: I don't think you're using the integer values anywhere, I don't think you need them.

crates/solvers/src/domain/eth/chain.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl ChainId {
+    /// Creates a new chain ID returning an error if it is unsupported.
```
ennmichael, line 12: Personally, I think you don't need to restate what the types are already saying. We know that `new()` is used as a constructor and we can see it returns a `Result`. I wouldn't have this comment, I don't think it's very informative - the doc comment that you put on the `ChainId` type is sufficient.

crates/solvers/src/infra/cli.rs
```diff
@@ -17,6 +18,10 @@ pub struct Args {
     #[arg(long, env, default_value = "127.0.0.1:7872")]
     pub addr: SocketAddr,
 
+    /// The chain ID this solver is for.
+    #[arg(long, env, value_parser = parse_chain_id)]
+    pub chain_id: eth::ChainId,
```
ennmichael, line 23: I think this should also be specified in the config file. Why would you specify the WETH address in config but specify this in the CLI? Where do you draw the line?

I'd say everything except the address and log level should go to the config file.

ennmichael (approved):

crates/solvers/src/infra/cli.rs
nlordell: > Why would you specify the WETH address in config but specify this in the CLI?

To me the chain ID is similar to the Ethereum RPC configuration, in that it is for the `solvers` binary and not for a specific solver subcommand.

That being said, you can imagine a "pure" solver that doesn't care about this and only uses passed in liquidity, so I think you are right that it should go in the configuration file.

crates/solvers/src/domain/eth/chain.rs
nlordell: Sure. I liked it because "it was pretty to me" that those are the actual chain ID values, but you are right that they have no place in the domain logic.

crates/solvers/src/domain/eth/chain.rs
nlordell: 0016437aa21683b664984c912239a576cd9e945a

crates/solvers/src/infra/cli.rs
nlordell: 0016437aa21683b664984c912239a576cd9e945a

vkgnosis:
I tried to update my colocation e2e test to main which includes this change, but I see now that this makes e2e tests impossible. In e2e tests we do not have a known chain id and hardcoded contract addresses but this change enforces that the chain must be one of mainnet, goerli, gnosis. Also I don't see why we would artificially restrict ourselves to just those networks when something like a univ2 solver can work anywhere. This is exactly what I recently fixed in orderbook and autopilot by making univ2 sources non hardcoded and allowing to set custom weth and settlement contract addresses.

I'm not sure how to reconcile this with what you had in mind here @nlordell . I could make a PR that makes the chain-id configurable again but you probably know better how to do it.

PR #1203, opened 2023-02-10
Use `TokenAmount` in `AmmOrderExecution`

### LINKED ISSUE #692: Use `TokenAmount` in `AmmOrderExecution` instead of a tuple.
We should refactor `AmmOrderExecution` to contain `TokenAmount` instead of `(H160, U256)`, since the latter one kills the readability... You need to check what `.input_max.0` and `.input_max.1` are to understand the context.

_Originally posted by @sunce86 in https://github.com/cowprotocol/services/pull/659#discussion_r1009718499_

Razz4780:
Replaced the `(H160, U256)` tuple in `AmmOrderExecution` with `TokenAmount` struct. Also replaced separate `token`, `amount` fields or params with `TokenAmount` in related code.

Fixes #692

# Test Plan

GitHub workflows

### DIFF
--- a/crates/driver/src/boundary/liquidity/uniswap/v2.rs
+++ b/crates/driver/src/boundary/liquidity/uniswap/v2.rs
@@ -10,6 +10,7 @@ use {
     futures::StreamExt,
     shared::{
         current_block::{self, CurrentBlockStream},
+        http_solver::model::TokenAmount,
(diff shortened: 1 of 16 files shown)


ennmichael (approved):

josojo (approved):
Nice

PR #1204, opened 2023-02-11
New tracing logs for debugging Paraswap

josojo:
We are getting a lot of errors from paraswap that contain the error message: `Err(Other("EOF while parsing a value at line 1))`.https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/r/s/kXG8f

In order to understand this error better, I want to see the raw response and introduce the same tracing log as we have it for the price API request.


nlordell (approved):

nlordell:
> In order to understand this error better, I want to see the raw response and introduce the same tracing log as we have it for the price API request.

In the past, this has usually been 429 rate limits (as, if I recall correctly, it returns an HTML body instead of JSON that we expect)

Issue #1196, opened 2023-02-09
Fetch gas price in Driver


vkgnosis:
_Originally posted by @nlordell in https://github.com/cowprotocol/services/pull/1187#discussion_r1101430171_

PR #1206, opened 2023-02-13
Fix driver e2e test

vkgnosis:
Some of my recent changes broke the e2e test:
- Didn't update some field names detected by `deny_unknown_fields` .
- Solution score changes because driver no longer sets gas price. See https://github.com/cowprotocol/services/issues/1196 . Getting the current gas price has to be implemented in the driver.

### Test Plan

manually ran driver e2e tests

### DIFF
--- a/crates/driver/src/tests/cases/settle.rs
+++ b/crates/driver/src/tests/cases/settle.rs
@@ -14,6 +14,7 @@ use {
 #[ignore]
 #[tokio::test]
 async fn test() {
+    crate::boundary::initialize_tracing("driver=trace");
     // Set up the uniswap swap.
     let setup::blockchain::Uniswap {
         web3,
@@ -51,7 +52,6 @@ async fn test() {
         domain_separator,
         owner: admin,
     };
-    let gas_price = web3.eth().gas_price().await.unwrap().to_string();
     let now = infra::time::Now::Fake(chrono::Utc::now());
     let deadline = now.now() + chrono::Duration::days(30);
     let interactions = interactions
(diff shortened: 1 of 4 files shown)


MartinquaXD:
Not sure I understand the fix. Why isn't the fix to query the gas price in the driver instead of adapting the tests to pass without the gas price?

crates/driver/src/tests/cases/settle.rs
```diff
@@ -111,7 +111,7 @@ async fn test() {
                     }
                 ],
                 "liquidity": [],
-                "effectiveGasPrice": gas_price,
+                "effectiveGasPrice": "0",
```
ennmichael, line 114: Why? You're introducing yet another unrealistic condition into the test. In the real world, the gas price is not zero.

crates/driver/src/tests/cases/solve.rs
```diff
@@ -13,9 +13,10 @@ use {
 #[ignore]
 #[tokio::test]
 async fn test() {
+    crate::boundary::initialize_tracing("driver=trace");
     // Set up the uniswap swap.
     let setup::blockchain::Uniswap {
-        web3,
+        web3: _web3,
```
ennmichael, line 19: If you don't need it, you can just get rid of it and add `..` at the end of the pattern.

crates/driver/src/tests/cases/settle.rs
vkgnosis: (also re Martin)
As I linked in the description, the correct fix is absolutely to fetch the gas price in the driver. When we do that then the test will fail and we can put back the realistic value here. I didn't do this here because I am fixing multiple things and I felt fetching the real gas price was going to be more work that should go into a separate PR.

ennmichael (approved):

PR #1207, opened 2023-02-14
Optional Chain ID in Baseline Solver Configuration

nlordell:
@vkgnosis pointed out in #1202 that the chain ID parameter being mandatory is not practical for local deployments or when testing out on new chains.

In fact, the baseline solver only needs to know the WETH address, so this PR changes the configuration semantics to require either:
- specifying a `chain-id` to use the canonical WETH contract for that chain.
- specify a custom `weth` address to use that instead, at which point the `ChainId` is no longer needed

### Test Plan

Adjusted test loading the example TOML configuration.

### DIFF
--- a/crates/solvers/example.baseline.toml
+++ b/crates/solvers/example.baseline.toml
@@ -1,4 +1,5 @@
 chain-id = "1"
-weth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+# Alternatively, you can manually specify a WETH contract address:
+#weth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
 base-tokens = []
 max-hops = 0
--- a/crates/solvers/src/infra/config/baseline/file.rs
+++ b/crates/solvers/src/infra/config/baseline/file.rs
@@ -7,28 +7,28 @@ use {
     tokio::fs,
 };
 
+#[serde_as]
 #[derive(Deserialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 struct Config {
-    #[serde(flatten)]
-    pub contracts: ContractsConfig,
+    /// Optional chain ID. This is used to automatically determine the address
+    /// of the WETH contract.
+    #[serde_as(as = "Option<serialize::ChainId>")]
+    chain_id: Option<eth::ChainId>,
+
+    /// Optional WETH contract address. This can be used to specify a manual
+    /// value **instead** of using the canonical WETH contract for the
+    /// configured chain.
+    weth: Option<H160>,
 
     /// List of base tokens to use when path finding. This defines the tokens
     /// that can appear as intermediate "hops" within a trading route. Note that
     /// WETH is always considered as a base token.
-    pub base_tokens: Vec<eth::H160>,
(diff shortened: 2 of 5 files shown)


crates/solvers/example.baseline.toml
```diff
@@ -1,4 +1,5 @@
 chain-id = "1"
-weth = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+# Alternatively, you can manually specify a WETH contract address:
```
MartinquaXD, line 2: nit: Could you make the comment refer to the `chain-id` parameter? This makes the dependency more explicit and less error prone if somebody where to insert a new argument between `chain-id` and `weth`.

crates/solvers/src/infra/config/baseline/file.rs
```diff
@@ -1,21 +1,17 @@
...
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 struct Config {
-    /// The chain ID the solver is configured for.
-    #[serde_as(as = "serialize::ChainId")]
-    pub chain_id: eth::ChainId,
-
-    /// The address of the WETH contract.
-    pub weth: Option<eth::H160>,
+    #[serde(flatten)]
+    pub contracts: ContractsConfig,
```
vkgnosis, line 14: Now we're mixing deny_unknown_fields and flatten which you mentioned before doesn't work. deny_unknown_fields has already been pretty helpful for me, but this is a small config file so doesn't really matter. More generally, I think not using flatten is nice because it makes the structure of the config file more obvious from the code. With complicated serde instructions it tends to get harder to figure out how the serialized file is going to look. Again, this file is small so not super important, so up to you.

MartinquaXD (approved):

crates/solvers/src/infra/config/baseline/file.rs
nlordell: > Now we're mixing deny_unknown_fields and flatten which you mentioned before doesn't work

It doesn't work well. `deny_unknown_fields` and `flatten` together have a whole bunch of prickly issues:
- If you flatten a struct that also has `deny_unknown_fields`, then this won't work (the inner struct will see a whole bunch of fields from the parent and deny them).
- If you `deny_unknown_fields` on untagged enums, then this also [doesn't work](https://github.com/serde-rs/serde/issues/1600)

However, in this particular case of `deny_unknown_fields`, `flatten` and `rename` on an enum with only single values, it does. I tested it out locally, but you can also see an example of this [in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ced40469267152bbf560a20719930bbb).

vkgnosis (approved):

crates/solvers/src/infra/config/baseline/file.rs
nlordell: So, I changed it to use a manual `match (chain_id, weth)`. The reasoning is:
- As you mentioned complicated `serde` instructions make things harder to figure out, also the whole complex interaction between `deny_unknown_fields`, `flatten` and `rename_all` is not obvious (for example, if you don't use the `rename_all`, then serialization will not work as expected)
- You get nicer panic error messages

PR #1208, opened 2023-02-14
Fix orderbook openapi scheme

shoom3301:
Fixes https://github.com/cowprotocol/cowswap/issues/2045

To avoid code duplication and have one source of truth we decided to generate typescript code based on Swagger scheme:
https://github.com/cowprotocol/cow-sdk/pull/97

This PR levels out contradictions between the openapi scheme and the implementation.


crates/orderbook/openapi.yml
```diff
@@ -435,9 +437,14 @@ components:
            Note that for ethflow orders, the valid_to encoded in the smart
            contract is max(uint)
           type: integer
+        isRefunded:
+          description: |
+           Is ETH refunded
+          type: boolean
```
nlordell, line 443: AFAIK, we purposefully left this out as we do not want to support it moving forward. Maybe @josojo can comment (as I'm not 100% sure it was this field, or another field).

nlordell:
Wow, Thanks!

anxolin (commented):
To prevent discrepancies into the API and swagger, would it be a good idea to auto-generate the swaggers using a Macro?

vkgnosis:
> To prevent discrepancies into the API and swagger, would it be a good idea to auto-generate the swaggers using a Macro?

It would be nice to generate the rust code for using and serving the API from the openapi spec, yes. I'm not aware of a tool that can do this. But it's been a while since we checked that.

crates/orderbook/openapi.yml
shoom3301: Removed it, thanks!

crates/orderbook/openapi.yml
```diff
@@ -763,7 +769,7 @@ components:
         - sellAmount
         - sellAmountBeforeFees
         - buyAmount
-        - transactionHash
+        - txHash
```
nlordell, line 772: Should we mark this as `required`? I was under the impression it isn't always available.

crates/orderbook/openapi.yml
```diff
@@ -917,8 +923,7 @@ components:
           description: Quote a sell order given the final total sell amount including fees
           properties:
             kind:
-              type: string
-              enum: [sell]
+              $ref: "#/components/schemas/OrderKind"
```
nlordell, line 926: This is intentionally `[sell]`, as the use with `oneOf` is roughly equivalent to TS:

```ts
type OrderQuoteSide =
    | { kind: "sell", sellAmountBeforeFee: TokenAmount }
    | { kind: "sell", sellAmountAfterFee: TokenAmount }
    | { kind: "buy", buyAmountAfterFee: TokenAmount };
```

crates/orderbook/openapi.yml
shoom3301: It has `nullable: true` and when we mark it as not required it means that it can be `undefined`

crates/orderbook/openapi.yml
```diff
@@ -931,8 +936,7 @@ components:
           description: Quote a sell order given the sell amount.
           properties:
             kind:
-              type: string
-              enum: [sell]
+              $ref: "#/components/schemas/OrderKind"
```
nlordell, line 939: Same here.

crates/orderbook/openapi.yml
```diff
@@ -943,8 +947,7 @@ components:
           description: Quote a buy order given an exact buy amount.
           properties:
             kind:
-              type: string
-              enum: [buy]
+              $ref: "#/components/schemas/OrderKind"
```
nlordell, line 950: Same here.

crates/orderbook/openapi.yml
shoom3301: Oops, sorry. It makes some problems with typescript code based on the scheme, but it's our problems :) 
Fixed

crates/orderbook/openapi.yml
shoom3301: Fixed

crates/orderbook/openapi.yml
shoom3301: Fixed

crates/orderbook/openapi.yml
```diff
@@ -617,6 +621,10 @@ components:
             For ethflow orders - order that are placed onchain with native eth -, this field
             contains a struct with two variables user_valid_to and is_refunded
           $ref: "#/components/schemas/EthflowData"
+        onchainUser:
+          description: |
+            TODO: smth related to ETH flow
```
nlordell, line 626: This represents the actual trader of an on-chain order. In that case, the `owner` would be the EthFlow contract in the case of EthFlow orders and not the actual trader.

vkgnosis (approved):
I'm surprised the missing `type`s where found by our openapi linter.

crates/orderbook/openapi.yml
vkgnosis: Not sure if that gives Alexandr enough context on what to put there. Maybe

> For ethflow orders the `owner` field is set to the ethflow contract and this field is set to the true owner of the order.

shoom3301: I just copied the Nick's message :)
Thank you!

anxolin:
>> To prevent discrepancies into the API and swagger, would it be a good idea to auto-generate the swaggers using a Macro?
> It would be nice to generate the rust code for using and serving the API from the openapi spec, yes. I'm not aware of a tool that can do this. But it's been a while since we checked that.


In case it helps
`Automated OpenAPI (AKA Swagger) document generation for Rust/Rocket projects.`
Not sure how well it works

https://github.com/GREsau/okapi


Also https://lib.rs/crates/utoipa-gen

`Utoipa crate provides auto generated OpenAPI documentation for Rust REST APIs. It treats code first approach as a first class citizen and simplifies API documentation by providing simple macros for generating the documentation from your code.`

shoom3301:
recheck

shoom3301:

shoom3301:
recheck

nlordell (approved):

PR #1213, opened 2023-02-14
toolchain-config: Using channel with a specified version

josojo:
I have installed already 1.67.1 locally. With the current configuration, toolchain-config would pick 1.67.1 as the latest stable version, instead of 1.67.0 - even after running `rustup default 1.67.0` as the toolchain config file would overwrite this rustup default.

How do you guys do it?

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.67.0"
 components = ["clippy", "rustfmt"]
 profile = "minimal"


MartinquaXD:
Not sure I understand the issue. Seems like the intended behavior to use the very latest stable release. Whatever that is at the time.
Also why do you want to make it `1.67.0` when `1.67.1` already exists?

josojo:
>Also why do you want to make it 1.67.0 when 1.67.1 already exists?
because current ci runs with 1.67.0 and 1.67.1 would reformat the project a little bit ( can not compile current code with 1.67.1)

MartinquaXD:
I think the only difference in formatting is that `1.67.0` expects printed arguments to be placed inside the formatting string. If you are running into that problem I think you should just put the arguments there manually for now even though your local `1.67.1` version doesn't care about it. The CI will pick up the new release soon enough.

If that is not your problem then you might be running a different `fmt` command than CI. CI makes use of nightly formatting features so the correct command to run locally is `cargo +nightly fmt --all`.
Hope this helps.

vkgnosis:
> If that is not your problem then you might be running a different `fmt` command than CI. CI makes use of nightly formatting features so the correct command to run locally is `cargo +nightly fmt --all`.
> Hope this helps.

This is it. Unfortunate side effect of our nightly rustfmt config is that for some parts (mostly imports) the formatting changes a lot.

josojo:
Thanks guys!
I installed a newer formatter by `rustup install +nightly` and now the command: `cargo +nightly fmt --all` no longer reformats all project dependencies. Solved it for me. Thx

PR #1214, opened 2023-02-14
Get gas price in driver

### LINKED ISSUE #1196: Fetch gas price in Driver

vkgnosis:
Fixes #1196 .

The driver was not fetching gas prices but they are required for the solvers.

- Add some logging to the `route` function. This is useful because it doesn't make sense to report on all errors in detail through the autopilot-driver API but we do want to log them for debugging.
- Change the score comparison in the test to be approximate. Required because the computed score is not deterministic (unrelated to my change). The test fails if the score deviates by more than 1% from the expected value.

### Test Plan

run the changed test

### DIFF
--- a/crates/driver/Cargo.toml
+++ b/crates/driver/Cargo.toml
@@ -58,8 +58,10 @@ shared = { path = "../shared" }
 solver = { path = "../solver" }
 tracing = { workspace = true }
 warp = { workspace = true }
+tap = "1.0.1"
 
 [dev-dependencies]
+approx = "0.5"
 maplit = { workspace = true }
 mockall = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
--- a/crates/driver/src/infra/api/routes/solve/mod.rs
+++ b/crates/driver/src/infra/api/routes/solve/mod.rs
@@ -1,8 +1,10 @@
-use crate::infra::api::{Error, State};
-
 mod dto;
 
 pub use dto::AuctionError;
+use {
+    crate::infra::api::{Error, State},
+    tap::TapFallible,
+};
 
 pub(in crate::infra::api) fn solve(router: axum::Router<State>) -> axum::Router<State> {
     router.route("/solve", axum::routing::post(route))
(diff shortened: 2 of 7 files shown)


crates/driver/Cargo.toml
```diff
@@ -60,6 +60,7 @@ tracing = { workspace = true }
 warp = { workspace = true }
 
 [dev-dependencies]
+approx = { workspace = true }
```
ennmichael, line 63: There's no reason to share this dependency. The only reason you really need to fix a dependency version across crates is if you are using it in the API of the crate and need consistency. I would say otherwise let the creates pick their own versions, especially considering that we've already been introducing this autonomy to the `driver` and `solvers` in hopes of one day separating them out into their own repos.

crates/driver/src/infra/api/routes/solve/mod.rs
```diff
@@ -12,8 +12,14 @@ async fn route(
...
+    let auction = auction.0.into_domain(state.eth()).await.map_err(|err| {
+        tracing::warn!(?err, "error creating auction");
+        err
+    })?;
     let competition = state.competition();
-    let (solution_id, score) = competition.solve(&auction).await?;
+    let (solution_id, score) = competition.solve(&auction).await.map_err(|err| {
+        tracing::warn!(?err, "error solving auction");
+        err
+    })?;
```
ennmichael, line 23: Tap tap tap tap.

[Tap.](https://lib.rs/crates/tap)

:no_entry: No tap? :no_entry_sign: 

:tada: Tapstravaganza :man_dancing:

ennmichael: image

image

image

image

image

image

image

ennmichael (commented):
Constructive review.

crates/driver/src/infra/api/routes/solve/mod.rs
vkgnosis: I'm aware of this. There is also `inspect_err` in the std but it's nightly. To be clear, you want to use the dependency?

crates/driver/Cargo.toml
vkgnosis: I forgot we were intentionally not sharing dependencies. There is a reason though, which is lower compile times by not accidentally using multiple versions of the same crate, which includes different feature sets too. Anyway, will change it to normal dep.

crates/driver/src/infra/api/routes/solve/mod.rs
vkgnosis: Changed it to tap.

MartinquaXD (approved):

PR #1217, opened 2023-02-15
Don't fail CI on coverage error

vkgnosis:
Coverage often fails and gives an annoying Github notification that the pipeline fails. We don't care much about coverage so it shouldn't notify us if it fails. We might even want to delete the job as a whole to not waste (Github's) resources on coverage generation that no one ever looks at.

### Test Plan

CI


nlordell:
> We might even want to delete the job as a whole to not waste (Github's) resources on coverage generation that no one ever looks at.

Lets delete the job. I don't think anyone uses it and is just noise at this point.

nlordell (approved):

vkgnosis:
Changed to delete the coverage job.

PR #1219, opened 2023-02-15
Check Chain ID on Startup

nlordell:
A couple of times, we have accidentally connected services meant for Goerli to Mainnet. This leads to a bunch of events getting indexed in the wrong database and is very painful to cleanup.

This PR adds a chain ID sanity check on start up to prevent this from happening.

### Test Plan

```
% cargo run -p autopilot -- --node-url https://goerli.infura.io/v3/$INFURA_PROJECT_ID --chain-id 6
...
2023-02-15T15:40:45.363Z ERROR shared::tracing: thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `6`: connected to node with incorrect chain ID'
```

### Release Notes

We should configure the `CHAIN_ID` environment variable on all our pods.

### DIFF
--- a/crates/autopilot/src/lib.rs
+++ b/crates/autopilot/src/lib.rs
@@ -159,6 +159,13 @@ pub async fn main(args: arguments::Arguments) {
         .await
         .expect("Could not get chainId")
         .as_u64();
+    if let Some(expected_chain_id) = args.shared.chain_id {
+        assert_eq!(
+            chain_id, expected_chain_id,
+            "connected to node with incorrect chain ID",
+        );
+    }
+
     let network = web3
         .net()
         .version()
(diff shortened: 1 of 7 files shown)


vkgnosis (approved):

MartinquaXD (approved):
Check looks good. But I guess it could be moved to the top of the function to limit the chance that someone adds some logic before the check.

nlordell:
> Check looks good. But I guess it could be moved to the top of the function to limit the chance that someone adds some logic before the check.

Great point! I moved it immediately after `Web3` instance creation to hopefully avoid such issues.

nlordell:
Disclaimer: I added the chain ID check to the refunder binary as well (as it also writes to the database). I didn't ask for a re-review since it is more of the same.

PR #1222, opened 2023-02-16
Legacy adapter

MartinquaXD:
This adds the `legacy` command to the `solvers` binary that simply converts the new `/solve` request to something the existing legacy http solvers can work with and converts their response back to something the new driver can work with.

### Test Plan
Did a manual test dragging an auction through `autopilot -> driver -> legacy_solver -> cowdexag` and back. The driver was already able to simulate and score the legacy solution.
TODO e2e test (probably in a separate PR)


crates/solvers/src/domain/legacy.rs
```diff
@@ -0,0 +1,8 @@
+//! Legacy HTTP solver adapter implementation.
+//!
+//! In order to faciliate the transition from the legacy HTTP solver API to the
+//! new HTTP API, we provide a solver "wrapper" that just marshals the API
+//! request types to and from the legacy format.
+
+// TODO I guess we can simply move the actual code here?
+pub use crate::boundary::legacy::Legacy;
```
nlordell, line 8: I would do something more like:
```suggestion
pub struct Legacy(boundary::legacy::Legacy);

impl Solver for Legacy {
    // ...
}
```

crates/solvers/src/infra/config/legacy/file.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// The solver name used in metrics.
+    pub solver_name: String,
+
+    /// The URL of the solver.
+    pub base_url: String,
+
+    /// Controls value of the `max_nr_exec_orders` parameter.
+    /// Uses 100 as the default value.
+    pub max_nr_exec_orders: Option<u32>,
```
nlordell, line 27: This is a legacy Quasimodo thing, I would get rid of it and use some hard coded value in the `boundary/legacy` module instead of making it configuratble.

crates/solvers/src/boundary/legacy.rs
```diff
@@ -0,0 +1,549 @@
...
+                client: reqwest::Client::new(),
+                config: SolverConfig {
+                    // Note that we unconditionally set this to "true". This is
+                    // because the auction that we are solving for already
+                    // contains which tokens can and can't be internalized,
+                    // and we don't need to duplicate this setting here. Ergo,
+                    // in order to disable internalization, the driver would be
+                    // configured to have 0 trusted tokens.
+                    use_internal_buffers: Some(true),
+                    max_nr_exec_orders: config.max_nr_exec_orders,
```
nlordell, line 63: (regarding comment ☝️) Specifically, here I would just:
```suggestion
```

The `SolverConfig: Default` implementation already sets this to 100 AFAIU.

nlordell (commented):
LGTM! Just some minor suggestions on moving the `Solver` trait implementation to the `domain` (instead of keeping it in the boundary). The `domain::legacy::Legacy` type would do the mapping from the `boundary::legacy::Legacy` API to what the domain actually needs.

crates/solvers/src/domain/mod.rs
```diff
@@ -3,6 +3,14 @@
...
 pub mod liquidity;
 pub mod order;
 pub mod solution;
+
+pub trait Solver: Send + Sync {
+    fn solve(
+        &self,
+        auction: auction::Auction,
+    ) -> futures::future::BoxFuture<Vec<solution::Solution>>;
+}
```
nlordell, line 16: Since the `solvers` crate knows all possible `Solver` implementations before hand, then having an `enum` makes you not lose type information:

```
pub enum Solver {
    Baseline(baseline::Baseline),
    Legacy(legacy::Legacy),
}
```

This way, we can `match` and figure out what solver kind is being used instead of needing to change the `Solver` trait to be able to extract that information, or use downcasting.

An argument for a trait would be that we have potentially external implementations of `trait Solver` that cannot be known ahead of time, but this shouldn't be the case here.

I don't feel strongly about this point, so I'm fine as is, but we've had this discussion before on trait objects vs enums, so I think it is worth highlighting here.

MartinquaXD: Hmm, I still don't really see the appeal of using enums instead of `dyn Trait` for these kinds of things. It seems to me if we care about the underlying type implementing the `dyn Trait` (for example to handle some weird/specific behavior) then the trait isn't designed correctly.

nlordell: I don't see anything necessarily wrong with a trait here, just that we don't _need_ a trait, since all `Solver` implementations are known upfront: `baseline::Baseline | legacy::Legacy`.

As I said, don't feel strongly either way, just thought it was worth highlighting that we decided to use a trait here.

crates/solvers/src/domain/legacy.rs
```diff
@@ -0,0 +1,68 @@
...
+        infra::config::legacy::LegacyConfig,
+    },
+    anyhow::Result,
+    futures::future::{BoxFuture, FutureExt},
+    shared::http_solver::{DefaultHttpSolverApi, HttpSolverApi, SolverConfig},
+};
+
+pub struct Legacy {
+    solver: DefaultHttpSolverApi,
+    weth: eth::WethAddress,
```
nlordell, line 20: Sorry for the back-and-forth 🙈.

So `solver: DefaultHttpSolverApi` I would hide in the `boundary` module so that types from the `shared` crate don't leak into the domain.

I really think that the `Legacy` type you had in the `boundary` module was exactly what I would expect to be there - the only change I was suggesting was to change:

```rust
pub use crate::boundary::legacy::Legacy;
```

into:

```rust
pub type Legacy(boundary::legacy::Legacy);

impl Legacy {
    pub fn new(...) -> Self {
        Self(boundary::legacy::Legacy::new(...))
    }
}

impl Solver for Legacy {
    ...
}
```

MartinquaXD: I hope now it's how you meant it. I was not sure about leaving the `solve` method in the boundary directory because the boundary could either be seen as giving all the tools to convert the types back and forth or like it's implementing the old solving strategy.
But by having the "old" solve function in the boundary we have to expose fewer functions/types which seemed like a good thing to me.

crates/driver/src/infra/mempool/mod.rs
```diff
@@ -10,6 +10,10 @@ pub async fn send(
     solver: &Solver,
     settlement: settlement::Simulated,
 ) -> Result<(), Error> {
+    if mempools.is_empty() {
+        // handle this case specificially because `select_ok` panics otherwise.
+        return Err(Error::AllMempoolsFailed);
+    }
```
vkgnosis, line 16: Unrelated to this PR but could remember in an issue. If this fails with no mempools then we should catch that during startup.

vkgnosis (approved):

crates/solvers/src/domain/order.rs
```diff
@@ -73,3 +93,87 @@ impl<'a> NonLiquidity<'a> {
         self.0
     }
 }
+
+/// A COW reward amount, in base units.
+#[derive(Clone, Copy, Debug)]
+pub struct Reward(pub f64);
+
+#[derive(Debug, Clone)]
+pub struct CustomInteraction {
```
ennmichael, line 102: Imagine being new to the code and stumbling upon a `CustomInteraction` or a `JitOrder`. You'd have _no clue_ what these are. I think doc comments on these would be super-valuable.

ennmichael (commented):
I haven't reviewed everything yet.

The important parts of the design were done well IMO, I would change some details, like not depending on infra for config (more important) and using an enum rather than a trait (less important).

crates/solvers/src/domain/mod.rs
nlordell: > It seems like a single solver should return a single solution.

The idea is that solver engines return multiple solution proposals and that the `driver` would simulate each of them and pick the best one for solver competition.

crates/solvers/src/domain/order.rs
```diff
@@ -73,3 +93,87 @@ impl<'a> NonLiquidity<'a> {
...
+    pub fn to_bytes(self) -> [u8; 65] {
+        let mut bytes = [0u8; 65];
+        bytes[..32].copy_from_slice(self.r.as_bytes());
+        bytes[32..64].copy_from_slice(self.s.as_bytes());
+        bytes[64] = self.v;
+        bytes
+    }
+}
+
+pub struct AppData(pub [u8; 32]);
```
ennmichael, line 179: I'd leave a doc comment here as well.

I'm sorry if this is annoying, but it's important.

In general I want almost every entity in `domain` to be documented, unless it's something really obvious. The other modules don't need such detailed docs, but they're absolutely mandatory for `domain`.

On the flipside if we end up having docs online which are detailed enough, we could simply leave links in a lot of places. But I don't think we're quite there yet. You can copy large portions of the doc comments from `driver` or elsewhere in the codebase.

crates/solvers/src/domain/solution.rs
```diff
@@ -11,6 +11,11 @@ pub struct Solution {
     pub interactions: Vec<Interaction>,
 }
 
+pub enum Trade {
```
ennmichael, line 14: A simple doc comment would suffice here IMO.

crates/solvers/src/domain/order.rs
```diff
@@ -73,3 +93,87 @@ impl<'a> NonLiquidity<'a> {
         self.0
     }
 }
+
+/// A COW reward amount, in base units.
```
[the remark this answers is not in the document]
nlordell: Base units is the term for units accounting for decimals, for example "1.177 Ξ". This is in contrast to atoms which would be "wei" for Ether (noting that "wei" is specifically a measurement of Ether and not of general tokens).

This is fairly established financial naming AFAIU.

crates/solvers/src/domain/solution.rs
```diff
@@ -23,14 +28,36 @@ impl ClearingPrices {
...
+pub struct Fulfillment {
     order: order::Order,
+    executed: U256,
 }
 
-impl Trade {
+impl Fulfillment {
+    /// Creates a new order filled to the specified amount. Returns `None` if
+    /// the fill amount is incompatible with the order.
+    pub fn new(order: order::Order, executed: U256) -> Option<Self> {
```
ennmichael, line 39: Since the other constructor is called `fill`, it might make sense for this one to be called `partial`.

crates/solvers/src/domain/solution.rs
```diff
@@ -40,18 +67,53 @@ impl Trade {
...
+        };
+
+        eth::Asset {
+            token,
+            amount: self.executed,
         }
     }
 }
 
+pub struct JitTrade {
```
ennmichael, line 82: :memo:

crates/solvers/src/domain/solution.rs
```diff
@@ -40,18 +67,53 @@ impl Trade {
...
+pub struct LiquidityInteraction {
     pub liquidity: liquidity::Liquidity,
     // TODO: Currently there is not type-level guarantee that `input` and
     // output` are valid for the specified liquidity.
     pub input: eth::Asset,
     pub output: eth::Asset,
+    pub internalize: bool,
+}
+
+/// A custom interaction
```
ennmichael, line 104:
```suggestion
/// A custom interaction.
```

`CustomInteraction` has a doc comment saying "A custom interaction". This doesn't document anything, it only restates the name which we already see in the code. I would say: "an arbitrary interaction returned by the solver, which needs to be executed to fulfill the trade" or something similar to that.

crates/solvers/src/domain/mod.rs
MartinquaXD: I think the argument that we wouldn't have to use `BoxFuture` has merit (although the trade-off will be the boilerplate code on the enum itself).
Returning multiple solutions was introduced by the `baseline` solver implementation and is supposed to allow us to compute settlements for individual orders and merge them afterwards.
@nlordell since you have loads of branches depending on this. Should I change it now or should we do it on-top of your changes to avoid possible merge conflicts?

crates/solvers/src/domain/solution.rs
```diff
@@ -40,18 +67,53 @@ impl Trade {
...
+    pub target: Address,
+    pub value: eth::Ether,
+    pub calldata: Vec<u8>,
+    pub inputs: Vec<eth::Asset>,
+    pub outputs: Vec<eth::Asset>,
+    pub internalize: bool,
+    pub allowances: Vec<Allowance>,
+}
+
+/// Approval required to make some custom interaction possible.
```
ennmichael, line 115: On the contrary, this is an example of a really nice doc comment!

crates/solvers/src/infra/cli.rs
```diff
@@ -31,4 +31,6 @@ pub struct Args {
 pub enum Command {
     /// Baseline solver.
     Baseline,
+    /// Wrapper for solvers implementing the lagacy http interface.
```
ennmichael, line 34:
```suggestion
    /// Wrapper for solvers implementing the legacy HTTP interface.
```

crates/solvers/src/domain/mod.rs
nlordell: > @nlordell since you have loads of branches depending on this. Should I change it now or should we do it on-top of your changes to avoid possible merge conflicts?

I would change it now. This should be fairly easy merge conflicts to resolve.

MartinquaXD:
@ennmichael I addressed your comments. It should be ready for a final review. 🤞

crates/solvers/src/boundary/mod.rs
```diff
@@ -2,6 +2,9 @@
 //! "legacy" logic in `shared` and `model`.
```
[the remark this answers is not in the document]
nlordell: That would be me 🙈.

What do you mean though, isn't this doc comment pure poetry? Rich in metaphor and imagery and written so beautifully that it brings a tear to your eye? It puts even Shakespeare to shame!

PR #1224, opened 2023-02-16
Custom 'solve' URL Path for HTTP Solvers

nlordell:
I did something bad...

This **incredibly hacky** PR changes the Yearn solver price estimator to use a custom path for the solve request. This is because the HTTP solver expects requests to `/quote` instead of `/solve` for price estimation.

Why is it done in such a hacky way? Well the new `driver` binary already supports arbitrary paths for solvers (so in this case we would be able to configure the URL to be `https://the.yearn.solver.eth/path/to/quote`). This means that this technical debt will be removed for free in the very near future without any additional work.

The alternative would be to stop appending `/solve` in the `http_solver.rs` implementation and to add `/solve` to the URL for each external solver in our deployment configuration. I chose against this for a couple of reasons:
1. It is a configuration hassle with a non-0 risk.
2. External solvers have a `/notify` URL that gets added to the base for solver competition. This has some pretty ambiguous semantics when paired with changing the `DefaultHttpSolverApi` to be configured with a "solve" URL instead of a base URL: is `/notify` appended to the solve URL, does it `s/solve/notify/`? etc.

As mentioned ☝️, since this is already solved in the new driver, in a correct and non-hacky way, I decided to use this hacky patch for now.

### Test Plan

Run the orderbook API and see it make requests with different paths:
```
% cargo run -p orderbook -- --price-estimators Yearn,Quasimodo --native-price-estimators Yearn,Quasimodo --yearn-solver-url http://localhost:10001 --yearn-solver-path quote --quasimodo-solver-url http://localhost:10002
...
2023-02-16T16:08:14.652Z DEBUG request{id=0}: shared::price_estimation::competition: new price estimate query=Query { from: None, sell_token: 0xdac17f958d2ee523a2206206994597c13d831ec7, buy_token: 0x6b175474e89094c44da98b954eedeac495271d0f, in_amount: 1000000000, kind: Sell } result=Err(Other(failed to send request

Caused by:
    0: error sending request for url (http://localhost:10001/quote?instance_name=2023-02-16_16%3A08%3A14.644312_UTC_Ethereum___Mainnet_1_0&time_limit=2&max_nr_exec_orders=100&use_internal_buffers=false&objective=surplusfeescosts): error trying to connect: tcp connect error: Connection refused (os error 61)
    1: error trying to connect: tcp connect error: Connection refused (os error 61)
    2: tcp connect error: Connection refused (os error 61)
    3: Connection refused (os error 61))) estimator="Yearn"
2023-02-16T16:08:14.652Z DEBUG request{id=0}: shared::price_estimation::competition: new price estimate query=Query { from: None, sell_token: 0xdac17f958d2ee523a2206206994597c13d831ec7, buy_token: 0x6b175474e89094c44da98b954eedeac495271d0f, in_amount: 1000000000, kind: Sell } result=Err(Other(failed to send request

Caused by:
    0: error sending request for url (http://localhost:10002/solve?instance_name=2023-02-16_16%3A08%3A14.646544_UTC_Ethereum___Mainnet_1_0&time_limit=2&max_nr_exec_orders=100&use_internal_buffers=false&objective=surplusfeescosts): error trying to connect: tcp connect error: Connection refused (os error 61)
    1: error trying to connect: tcp connect error: Connection refused (os error 61)
    2: tcp connect error: Connection refused (os error 61)
    3: Connection refused (os error 61))) estimator="Quasimodo"
...
```

**Note**: The errors are expected because I don't have anything actually bound to those ports, but the paths are different for both price estimators as expected.

### Release Notes

We should set `YEARN_SOLVER_PATH` to `quote` once it they switch their routing.

### DIFF
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -62,6 +62,9 @@ pub struct DefaultHttpSolverApi {
     /// Base solver url.
     pub base: Url,
 
+    /// The custom URL path used for the solve request.
+    pub custom_solve_path: Option<String>,
+
     /// An async HTTP client instance that will be used to interact with the
     /// solver.
     pub client: Client,
@@ -117,7 +120,8 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             .checked_sub(Duration::from_secs(1))
             .context("no time left to send request")?;
 
-        let mut url = self.base.join("solve").context("join base")?;
+        let path = self.custom_solve_path.as_deref().unwrap_or("solve");
+        let mut url = self.base.join(path).context("join base")?;
 
         let maybe_auction_id = model.metadata.as_ref().and_then(|data| data.auction_id);
         let instance_name = self.generate_instance_name(maybe_auction_id.unwrap_or(0));
@@ -275,6 +279,7 @@ HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Encoding: gzip\r\n\
             network_name: Default::default(),
             chain_id: Default::default(),
             base: "http://localhost:1234".parse().unwrap(),
+            custom_solve_path: None,
             client: Default::default(),
             config: Default::default(),
         };
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -63,7 +63,7 @@ pub struct DefaultHttpSolverApi {
     pub base: Url,
 
     /// The custom URL path used for the solve request.
-    pub custom_solve_path: Option<String>,
+    pub solve_path: String,
 
     /// An async HTTP client instance that will be used to interact with the
     /// solver.
@@ -120,8 +120,7 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             .checked_sub(Duration::from_secs(1))
             .context("no time left to send request")?;
 
-        let path = self.custom_solve_path.as_deref().unwrap_or("solve");
-        let mut url = self.base.join(path).context("join base")?;
+        let mut url = self.base.join(&self.solve_path).context("join base")?;
 
         let maybe_auction_id = model.metadata.as_ref().and_then(|data| data.auction_id);
         let instance_name = self.generate_instance_name(maybe_auction_id.unwrap_or(0));
@@ -279,7 +278,7 @@ HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Encoding: gzip\r\n\
             network_name: Default::default(),
             chain_id: Default::default(),
             base: "http://localhost:1234".parse().unwrap(),
-            custom_solve_path: None,
+            solve_path: "solve".to_owned(),
             client: Default::default(),
             config: Default::default(),
         };
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -133,6 +133,10 @@ pub struct Arguments {
     #[clap(long, env)]
     pub yearn_solver_url: Option<Url>,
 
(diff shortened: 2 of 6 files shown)


crates/shared/src/http_solver.rs
```diff
@@ -117,7 +120,8 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             .checked_sub(Duration::from_secs(1))
             .context("no time left to send request")?;
 
-        let mut url = self.base.join("solve").context("join base")?;
+        let path = self.custom_solve_path.as_deref().unwrap_or("solve");
```
vkgnosis, line 123: Maybe to be more explicit make `custom_solve_path` non optional and set it to "solve" everywhere?

vkgnosis (approved):

crates/shared/src/http_solver.rs
nlordell: Good idea. This at least makes it more explicit and opens up the possibility of configuring it everywhere.

MartinquaXD:
> Why is it done in such a hacky way? Well the new driver binary already supports arbitrary paths for solvers (so in this case we would be able to configure the URL to be https://the.yearn.solver.eth/path/to/quote). This means that this technical debt will be removed for free in the very near future without any additional work.

The new driver doesn't support that in its current form, no? At least not without another workaround. Because the driver currently also assumes `/solve` can be used for quoting we'll have the same problem there. The only difference AFAIK would be that we could run 2 separate drivers (1 for quoting and 1 for solving) which also doesn't seem optimal. Maybe we should just make alternative `/quote` URLs part of the driver itself then. 🤷‍♂️

MartinquaXD (approved):

nlordell:
> Because the driver currently also assumes /solve can be used for quoting we'll have the same problem there.

The [driver](https://github.com/cowprotocol/services/blob/main/crates/driver/example.toml) accepts [arbitrary URLs](https://github.com/cowprotocol/services/blob/dbacf0c33aeeed128c5e1749123410357391c8eb/crates/driver/src/infra/solver/mod.rs#L116-L120) for solver engines. It just responds to a `/solve` and `/quote` requests itself.

crates/shared/src/http_solver.rs
nlordell: Implemented this and made the `YEARN_SOLVER_PATH` configurable to facilitate the switch over when it happens.

MartinquaXD:
What I meant is that when you send a `/quote` request to a driver it will forward the request to the same endpoint that was configured for `/solve` [[0](https://github.com/cowprotocol/services/blob/main/crates/driver/src/domain/quote.rs#L73-#L90)].
That's why yearn would have to run 2 different drivers. 1 where `/solve` gets forwarded to `yearn_solver_url/solve` and one where `/solve` gets forwarded to `yearn_solver_url/quote`. The first driver would then be strictly used for auction solving whereas the second driver would be strictly used as a price estimator.
The point I'm making is that it would probably be nicer to have a single driver that knows that `/solve` and `/quote` requests should go to different endpoints.
Initially we assumed there would be no reason for a solver to have this requirement but I guess there is one and with the current driver implementation somebody will always have to do a workaround.

nlordell:
Ah, sorry, I misunderstood you.

> The point I'm making is that it would probably be nicer to have a single driver that knows that /solve and /quote requests should go to different endpoints.

I think this is possible now by having two solvers `yearn-quote` and `yearn-solve` with different configurations. This would allow different URLs for solving and quoting. If this becomes a common ask, I agree that we should have the `driver` be aware of potentially different URLs in solvers for quoting and solving.

PR #1228, opened 2023-02-17
Legacy solver tests

MartinquaXD:
Adds some tests for the legacy solver.
These tests are not great since they don't test the `driver <> solvers` interface but they are better than nothing, I guess.


crates/solvers/src/domain/order.rs
```diff
@@ -98,11 +98,13 @@ impl<'a> NonLiquidity<'a> {
 #[derive(Clone, Copy, Debug)]
 pub struct Reward(pub f64);
 
+/// Interaction that can be executed as part of an order's pre- or
+/// post-interactions.
 #[derive(Debug, Clone)]
-pub struct CustomInteraction {
+pub struct OrderInteraction {
```
nlordell, line 104: naming nit: I would call this `Interaction`. Just because it already is in the order module. Personally, I slightly prefer `order::Interaction` over `order::OrderInteraction`.

crates/solvers/src/tests/legacy/jit_order.rs
```diff
@@ -0,0 +1,122 @@
...
+                        "signingScheme": "eip712",
+                        "signature": "0x\
+                            0101010101010101010101010101010101010101010101010101010101010101\
+                            0202020202020202020202020202020202020202020202020202020202020202\
+                            03",
+                        "interactions": {
+                            "pre": [
+                                {
+                                    "target": "0x2222222222222222222222222222222222222222",
+                                    "value": "200",
```
nlordell, line 50: Is this expected behaviour? I feel like we use decimal notation for big integers everywhere. Maybe this was an accidental oversight in the original HTTP API? (cc @sunce86)

crates/solvers/src/domain/order.rs
```diff
@@ -98,11 +98,13 @@ impl<'a> NonLiquidity<'a> {
...
     pub target: Address,
     pub value: eth::Ether,
-    pub calldata: Vec<u8>,
+    pub call_data: Vec<u8>,
```
nlordell, line 107: such a naming nit I almost didn't make this comment:

I've seen this called "Calldata", "Call Data" or just "Data". Personally, I would go for "Calldata" in a single word as it is more distinctive. The only reason we used `callData` in the smart contracts was that `calldata` was a reserved keyword.

Really don't feel strongly at all though, so feel free to hit "resolve thread" and forget I mentioned this.

crates/solvers/src/tests/legacy/jit_order.rs
```diff
@@ -0,0 +1,122 @@
...
+                        "signature": "0x\
+                            0101010101010101010101010101010101010101010101010101010101010101\
+                            0202020202020202020202020202020202020202020202020202020202020202\
+                            03",
+                        "preInteractions": [
+                            {
+                                "target": "0x2222222222222222222222222222222222222222",
+                                // U256::from_str() expects a hex encoded number whereas
+                                // U256::to_string() produces a base 10 encoded number. That's why
+                                // the original `200` gets converted to `512` here.
```
nlordell, line 109: Is this expected behaviour? I feel like we use decimal notation for big integers everywhere. Maybe this was an accidental oversight in the original HTTP API? (cc @sunce86)

crates/solvers/src/tests/legacy/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Creates a temporary file containing the config of the given solver.
+pub fn create_temp_config_file(solver_addr: &SocketAddr) -> tempfile::TempPath {
+    let base_url = format!("http://{solver_addr}");
+    let config = format!(
+        r"
+solver-name = 'legacy_solver'
+base-url = '{}'
+chain-id = '1'
+    ",
```
nlordell, line 60: formatting nit:
```suggestion
        ",
```

nlordell:
> These tests are not great

I disagree, I think these are great! I think these tests test the `solvers` crate at their boundaries which is IMO, what the `solvers` crate tests should do. It's not a full E2E test (and I would argue that should go somewhere else), but it is definitely a full `solvers` crate E2E test.

nlordell (approved):

crates/solvers/src/tests/legacy/jit_order.rs
MartinquaXD: I put it of as expected because I remembered Niksa running into the same thing while writing some test but actually this is indeed weird.
This behavior got introduced with the eth flow. Fortunately our DB currently doesn't hold any interactions where `value != 0` so I think we can change the format without breaking anything: [be54e0e](https://github.com/cowprotocol/services/pull/1228/commits/be54e0e80f25a316318b390fc71bf984d8bfd652)

crates/solvers/src/domain/order.rs
MartinquaXD: I thought it had to be changed because we need 2 different interaction types (1 for custom interactions and 1 for pre interactions) but since they are in different scopes we can get away without renaming anything.

crates/solvers/src/tests/legacy/jit_order.rs
MartinquaXD: @fedgiac can you confirm that this change is harmless for the ethflow?

nlordell:
CI failure probably related to https://github.com/cowprotocol/services/pull/1232

crates/solvers/src/tests/legacy/jit_order.rs
nlordell: As you mentioned, EthFlow pre-interactions exclusively have 0-value, so it shouldn't affect them. This was added to the `foreign_liquidity_orders` interface by @sunce86 for a Yearn use-case, so I just wanted to make sure that they aren't using values for JIT orders (i.e. "foreign liquidity orders").

vkgnosis (approved):

crates/solvers/src/tests/legacy/jit_order.rs
```diff
@@ -0,0 +1,118 @@
+use {
```
ennmichael, line 1: I really like the doc comments you have for other tests, I think they're a great idea. Can you add one here as well?

ennmichael (commented):
I really like how you did these and the comments you left on the tests are very valuable. Nice!

ennmichael (approved):

PR #1229, opened 2023-02-17
Refactor e2e tests to use crate run functions

vkgnosis:
This PR makes all e2e tests use the library run functions instead of manually creating components. This separates the test from the code more which will help future refactoring and makes the tests more extensive because the components are run just like they are as standalone binaries.

As a side effect many tests become simpler. There is potential for a lot more e2e improvements after this but I didn't want to put more than necessary in this PR.

E2e tests become slower because we can no longer drive individual components at the exact moment it is needed. Instead some checks have to be converted to polling with the `wait_for_condition` function. Running all tests on CI goes from 90s to 120s. This is not a big deal. There are low hanging fruits to speed tests up and eventually the plan is to run e2e tests in parallel.

Most of the changed lines are in the individual e2e tests. It's not worth it to review this super closely because the e2e tests still pass and I spent a lot of time debugging them so I'm pretty sure they didn't become less correct than before (most became more correct).

Other changes in this PR that I did not factor out:
- Improve logging. I was looking at a lot of logs while debugging the tests and in the process moved some to trace when they were large and printing duplicate information. We can always change some of these back later if they turn out to have been useful. Other logs were missing information.
- Make balancer contract address configurable. Needed for e2e test.

### Test Plan

CI

### DIFF
--- a/crates/autopilot/src/limit_orders/quoter.rs
+++ b/crates/autopilot/src/limit_orders/quoter.rs
@@ -47,7 +47,7 @@ impl LimitOrderQuoter {
         loop {
             let sleep = match self.update().await {
                 // Prevent busy looping on the database if there is no work to be done.
-                Ok(true) => Duration::from_secs_f32(10.),
+                Ok(true) => Duration::from_secs_f32(1.),
                 Ok(false) => Duration::from_secs_f32(0.),
                 Err(err) => {
                     tracing::error!(?err, "limit order quoter update error");
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -201,8 +201,8 @@ impl SolvableOrdersCache {
                 Ok(balance) => balance,
                 Err(err) => {
                     tracing::warn!(
-                        owner = %query.owner,
-                        token = %query.token,
+                        owner = ?query.owner,
+                        token = ?query.token,
                         source = ?query.source,
                         error = ?err,
                         "failed to get balance"
--- a/crates/e2e/tests/e2e/eth_flow.rs
+++ b/crates/e2e/tests/e2e/eth_flow.rs
@@ -1,26 +1,28 @@
 use {
     crate::{
         deploy::Contracts,
-        local_node::{AccountAssigner, TestNodeApi},
+        local_node::TestNodeApi,
         onchain_components::{
             deploy_token_with_weth_uniswap_pool,
             to_wei,
             MintableToken,
             WethPoolConfig,
         },
-        services::{
-            create_orderbook_api,
-            setup_naive_solver_uniswapv2_driver,
-            wait_for_solvable_orders,
-            OrderbookServices,
-            API_HOST,
-        },
+        services::{get_auction, solvable_orders, wait_for_condition, API_HOST},
     },
     anyhow::bail,
     autopilot::database::onchain_order_events::ethflow_events::WRAP_ALL_SELECTOR,
     chrono::{DateTime, NaiveDateTime, Utc},
     contracts::{CoWSwapEthFlow, ERC20Mintable, WETH9},
-    ethcontract::{transaction::TransactionResult, Account, Bytes, H160, H256, U256},
+    ethcontract::{
+        transaction::{TransactionBuilder, TransactionResult},
+        Account,
+        Bytes,
+        PrivateKey,
+        H160,
+        H256,
+        U256,
+    },
     hex_literal::hex,
     model::{
         app_id::AppId,
@@ -56,10 +58,9 @@ use {
     shared::{
         current_block::timestamp_of_current_block_in_seconds,
         ethrpc::Web3,
-        http_client::HttpClientFactory,
-        maintenance::Maintaining,
         signature_validator::check_erc1271_result,
     },
+    std::time::Duration,
 };
 const ACCOUNT_ENDPOINT: &str = "/api/v1/account";
 const AUCTION_ENDPOINT: &str = "/api/v1/auction";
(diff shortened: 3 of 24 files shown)


crates/e2e/tests/e2e/eth_flow.rs
```diff
@@ -124,6 +135,12 @@ async fn eth_flow_tx(web3: Web3) {
         receiver,
     };
 
+    let client = reqwest::Client::default();
+
+    crate::services::start_api(&contracts, &[]);
+    crate::services::start_autopilot(&contracts, &[]);
+    crate::services::wait_for_api_to_come_up().await;
```
nlordell, line 142: Since these are commonly used together (AFAICT, always?), should we have a:
```suggestion
    crate::services::start_protocol_and_wait(&contracts, &[]).await;
```

crates/autopilot/src/solvable_orders.rs
```diff
@@ -201,8 +201,8 @@ impl SolvableOrdersCache {
                 Ok(balance) => balance,
                 Err(err) => {
                     tracing::warn!(
-                        owner = %query.owner,
-                        token = %query.token,
+                        owner = ?query.owner,
+                        token = ?query.token,
```
nlordell, line 205: Nice catch.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -330,7 +330,7 @@ impl<'a> Submitter<'a> {
                     find_mined_transaction(&self.contract.raw_instance().web3(), &transactions)
                         .await
                 {
-                    tracing::debug!("found mined transaction {:?}", receipt);
+                    tracing::debug!("found mined transaction {:?}", receipt.transaction_hash);
```
nlordell, line 333: Not just this one in particular, but nice log cleanups.

nlordell (commented):
Very cool!

crates/e2e/tests/e2e/eth_flow.rs
vkgnosis: Might do this in the future but don't want to do it in this PR because I'm not sure enough that this is right in every test yet.

crates/autopilot/src/limit_orders/quoter.rs
```diff
@@ -47,7 +47,7 @@ impl LimitOrderQuoter {
         loop {
             let sleep = match self.update().await {
                 // Prevent busy looping on the database if there is no work to be done.
-                Ok(true) => Duration::from_secs_f32(10.),
+                Ok(true) => Duration::from_secs_f32(1.),
```
vkgnosis, line 50: This makes e2e tests faster and during normal operation 1 second sleep is fine too so it doesn't need to be configurable.

nlordell (approved):
LGTM, lots of cleanup I see.

crates/e2e/tests/e2e/eth_flow.rs
```diff
@@ -138,37 +155,57 @@ async fn eth_flow_tx(web3: Web3) {
...
+
+    const REFUNDER_PK: [u8; 32] =
+        hex!("0000000000000000000000000000000000000000000000000000000000000002");
+    let refunder = Account::Offline(PrivateKey::from_raw(TRADER_PK).unwrap(), None);
+    const TRADER_PK: [u8; 32] =
+        hex!("0000000000000000000000000000000000000000000000000000000000000003");
+    let trader = Account::Offline(PrivateKey::from_raw(TRADER_PK).unwrap(), None);
+    const DUMMY_TRADER_PK: [u8; 32] =
+        hex!("0000000000000000000000000000000000000000000000000000000000000004");
+    let dummy_trader = Account::Offline(PrivateKey::from_raw(TRADER_PK).unwrap(), None);
```
MartinquaXD, line 200: Should all those accounts actually be initialized with the same private key?
Also TIL you can use a `const` variable before the line initializing it.

MartinquaXD (approved):
Nice cleanup! 🧹 
I'm just a bit worried about all the changed amounts. Didn't check whether all of them still make sense TBH. 🙈

crates/e2e/tests/e2e/eth_flow.rs
```diff
@@ -83,14 +84,34 @@ async fn local_node_eth_flow_indexing_after_refund() {
...
-        "warn,orderbook=debug,orderbook=debug,solver=debug,autopilot=debug",
+        "e2e=debug,orderbook=debug,solver=debug,autopilot=debug,\
+         orderbook::api::request_summary=off",
     );
     shared::exit_process_on_panic::set_panic_hook();
+
+    crate::services::clear_database().await;
     let contracts = crate::deploy::deploy(&web3).await.expect("deploy");
 
-    let mut accounts = AccountAssigner::new(&web3).await;
```
nlordell, line 91: The `AccountAssigner` seemed useful for making the test setup more succinct - just curious as to the rationale for removing it.

crates/e2e/tests/e2e/eth_flow.rs
nlordell: Looks like copy-pasta!

vkgnosis: fixed

crates/e2e/tests/e2e/eth_flow.rs
vkgnosis: I want to consistently use private key accounts everywhere. Some tests were using node/local accounts which made them harder to debug. Some components in our code can't use node accounts. The only time I use the node local account is to set up initial on chain state and fund other accounts. I should remove the whole abstraction.

vkgnosis: I'm going to add a commit that deletes it soon. I could see adding the component back but this time using private keys internally rather than node accounts.

vkgnosis:
> I'm just a bit worried about all the changed amounts.

I lowered a bunch of amounts because smaller numbers are easier to work with. For example many tests used uniswap pools with with millions of wei which I changed to 1000. Likewise I lowered some order sell and buy amounts. The tests only care about whether an order gets traded. They don't care about the exact price. So I've removed these kind of checks and replaced them with checking that the balancer of the user has changed.

PR #1230, opened 2023-02-17
Fix metrics not being allowed to initialize twice

vkgnosis:
Factored out of https://github.com/cowprotocol/services/pull/1229 .

Without the `MetricStorage` derive, initializing the metrics fails when this is done in tests that run after eachother in the same `cargo test` run. This happens on CI when running e2e tests. With the derive this works fine.

### Test Plan

CI, after merge check staging Grafana dashboard in case of broken metrics due to accidental rename

### DIFF
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -6,16 +6,7 @@ use {
     anyhow::Result,
     ethcontract::U256,
     model::order::{Order, OrderClass},
-    prometheus::{
-        Gauge,
-        Histogram,
-        HistogramOpts,
-        HistogramVec,
-        IntCounter,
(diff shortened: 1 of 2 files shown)


crates/solver/src/metrics.rs
```diff
@@ -128,165 +119,67 @@ pub trait SolverMetrics: Send + Sync {
 
 // TODO add labeled interaction counter once we support more than one
 // interaction
-pub struct Metrics {
+#[derive(prometheus_metric_storage::MetricStorage)]
+struct Metrics_ {
```
nlordell, line 123: Maybe we can call this `Inner` or `Fields`?

nlordell (approved):

PR #1231, opened 2023-02-18
Boundary Naive Solver Implementation

nlordell:
This PR adds the naive solver to the `solvers` crate. It turned out to be harder than anticipated because the naive solver implementation is tightly coupled with the `Settlement` type.

Since the code is fairly simple, I hope to move the logic into the `domain` as soon as possible (hence the test coverage from #1233), especially considering since most of the bugs stem from the fact that we are mixing `LimitOrder` "synthetic" orders with `Order` "signed" orders (and all the weird cases related to custom prices, and synthetic order amounts).

### Test Plan

Tests are included in #1233.

### DIFF
--- a/crates/solvers/src/boundary/naive.rs
+++ b/crates/solvers/src/boundary/naive.rs
@@ -43,11 +43,12 @@ pub fn solve(
 
     // Note that the `order::Order` -> `boundary::LimitOrder` mapping here is
     // not exact. Among other things, the signature and various signed order
-    // fields are missing from the `order::Order` data that have access to in
-    // the solver engines. This means that the naive solver in the `solver`
-    // crate will encode "incorrect" settlements. This is fine, since we give
-    // it just enough data to compute the swapped orders and the swap amounts.
-    // The `driver` is then responsible for encoding the solution into a valid
+    // fields are missing from the `order::Order` data that the solver engines
+    // have access to. This means that the naive solver in the `solver` crate
+    // will encode "incorrect" settlements. This is fine, since we give it just
+    // enough data to compute the correct swapped orders and the swap amounts
+    // which is what the naive solver in the `solvers` crate cares about. The
+    // `driver` is then responsible for encoding the solution into a valid
     // settlement transaction anyway.
     let boundary_orders = orders
         .iter()
@@ -155,14 +156,14 @@ pub fn solve(
 //
 // Joking aside, the existing naive solver implementation is tightly coupled
 // with the `Settlement` and `SettlementEncoder` types in the `solver` crate.
-// This meeans that there is no convinient way to say: "please compute a
-// solution given this list of orders and constant product pool" without it
-// creating a full settlement for encoding. In order to adapt that API into
-// something that is useful in this boundary module, we create a fake slippage
-// context that applies 0 slippage (so that we can recover the exact executed
-// amounts from the constant product pool) and we create capturing settlement
-// handler implementations that record the swap that gets added to each
-// settlement so that it can be recovered later to build a solution.
+// This means that there is no convenient way to say: "please compute a solution
+// given this list of orders and constant product pool" without it creating a
+// full settlement for encoding. In order to adapt that API into something that
+// is useful in this boundary module, we create a fake slippage context that
+// applies 0 slippage (so that we can recover the exact executed amounts from
+// the constant product pool) and we create capturing settlement handler
+// implementations that record the swap that gets added to each settlement so
+// that it can be recovered later to build a solution.
 
 struct Slippage {
     calculator: SlippageCalculator,
--- a/crates/solvers/src/boundary/naive.rs
+++ b/crates/solvers/src/boundary/naive.rs
@@ -41,6 +41,14 @@ pub fn solve(
         _ => return None,
     };
 
+    // Note that the `order::Order` -> `boundary::LimitOrder` mapping here is
+    // not exact. Among other things, the signature and various signed order
+    // fields are missing from the `order::Order` data that have access to in
+    // the solver engines. This means that the naive solver in the `solver`
+    // crate will encode "incorrect" settlements. This is fine, since we give
+    // it just enough data to compute the swapped orders and the swap amounts.
+    // The `driver` is then responsible for encoding the solution into a valid
+    // settlement transaction anyway.
     let boundary_orders = orders
         .iter()
         .map(|order| LimitOrder {
@@ -107,7 +115,7 @@ pub fn solve(
     let boundary_solution =
         multi_order_solver::solve(&slippage.context(), boundary_orders, &boundary_pool)?;
 
-    let mut swap = pool_handler.swap.lock().unwrap();
+    let swap = pool_handler.swap.lock().unwrap().take();
     Some(solution::Solution {
         prices: solution::ClearingPrices::new(
             boundary_solution
@@ -129,7 +137,6 @@ pub fn solve(
             })
             .collect(),
         interactions: swap
-            .take()
             .into_iter()
             .map(|(input, output)| {
                 solution::Interaction::Liquidity(solution::LiquidityInteraction {
@@ -223,8 +230,7 @@ impl SettlementHandling<ConstantProductOrder> for PoolHandler {
         execution: AmmOrderExecution,
         _: &mut SettlementEncoder,
     ) -> anyhow::Result<()> {
-        let mut swap = self.swap.lock().unwrap();
-        *swap = Some((
+        *self.swap.lock().unwrap() = Some((
             eth::Asset {
                 token: eth::TokenAddress(execution.input_max.token),
                 amount: execution.input_max.amount,
--- a/crates/solvers/src/boundary/naive.rs
+++ b/crates/solvers/src/boundary/naive.rs
@@ -1,0 +1,239 @@
+use {
+    crate::{
+        boundary::liquidity::constant_product::to_boundary_pool,
+        domain::{eth, liquidity, order, solution},
+    },
+    ethereum_types::{H160, U256},
+    itertools::Itertools,
+    model::order::{
+        LimitOrderClass,
+        Order,
+        OrderClass,
+        OrderData,
+        OrderKind,
+        OrderMetadata,
+        OrderUid,
+    },
+    num::{BigRational, One},
+    solver::{
+        liquidity::{
+            slippage::{SlippageCalculator, SlippageContext},
+            AmmOrderExecution,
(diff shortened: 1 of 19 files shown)


crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,235 @@
...
+                    output,
+                    internalize: false,
+                })
+            })
+            .collect(),
+    })
+}
+
+// Beyond this point is... well... nameless and boundless chaos. The
+// unfathomable horrors that follow are not for the faint of heart!
```
MartinquaXD, line 112: Sounds like the implementation would be a mess. Do you think it might finally be time to drop the naive solver? AFAIK we were considering that for quite some time now.

nlordell: I was planning on porting it to the `solvers` crate (i.e. moving it from the `boundary` module into the `domain`). I just wanted to setup tests before doing that to make sure there are no bugs/regressions that get introduced. Its actually a pretty small bit of code, and most of the nasty bugs are related to the fact that the code acts on both `LimitOrder` (i.e. the liquidity abstraction) and `Settlement` (i.e. actual orders). By moving it here, we would have a consistent view of liquidity and orders and, in theory, avoid all of the issues we've been seeing with it lately.

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,239 @@
...
+    let pool_handler = Arc::new(PoolHandler::default());
+    let boundary_pool = ConstantProductOrder::for_pool(
+        to_boundary_pool(liquidity.address, pool)?,
+        pool_handler.clone(),
+    );
+
+    let boundary_solution =
+        multi_order_solver::solve(&slippage.context(), boundary_orders, &boundary_pool)?;
+
+    let mut swap = pool_handler.swap.lock().unwrap();
```
MartinquaXD, line 110: nit: Not that it matters here (since this Mutex only gets accessed from the same thread) but I think it would be nice to avoid holding locks if possible.

```suggestion
    let swap = pool_handler.swap.lock().unwrap().take();
```

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+
+    fn encode(
+        &self,
+        execution: AmmOrderExecution,
+        _: &mut SettlementEncoder,
+    ) -> anyhow::Result<()> {
+        let mut swap = self.swap.lock().unwrap();
```
MartinquaXD, line 226: nit: You could prepare the value before locking and then simply assign. That would mean nobody could cause the lock to be held for too long by introducing new code.
Again it wouldn't make a difference here. I just want to promote safer access patterns with locks.

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,239 @@
...
+                        },
+                        solver_fee: order.fee().amount,
+                        ..Default::default()
+                    },
+                    data: OrderData {
+                        sell_token: order.sell.token.0,
+                        buy_token: order.buy.token.0,
+                        sell_amount: order.sell.amount,
+                        buy_amount: order.buy.amount,
+                        fee_amount: order.fee().amount,
```
MartinquaXD, line 84: `order.fee().amount` is used to initialize the user signed `fee_amount` as well as the `solver_fee` that can be scaled by the `autopilot` to promote CoWs. Since those values don't have to be identical it looks like an error.

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,239 @@
...
+                        kind: match order.side {
+                            order::Side::Buy => OrderKind::Buy,
+                            order::Side::Sell => OrderKind::Sell,
+                        },
+                        partially_fillable: order.partially_fillable,
+                        ..Default::default()
```
MartinquaXD, line 90: Don't we need to initialize all fields for the order to be correctly encoded?

crates/solvers/src/boundary/naive.rs
nlordell: Thanks. I think these are good defensive programming suggestions - will definitely apply them!

crates/solvers/src/boundary/naive.rs
MartinquaXD: Maybe I misunderstood how this works. Since the actual encoding will be done later by the driver we probably only have to initialize the fields that matter for the naive algorithm. And I guess that is probably also why we don't care about the difference between fee_amount and solver_fee?

nlordell: > Since the actual encoding will be done later by the driver we probably only have to initialize the fields that matter for the naive algorithm. And I guess that is probably also why we don't care about the difference between fee_amount and solver_fee?

Exactly. I will add a comment blob on top of the type mapping.

nlordell: 6700e0fb429776121c5399e4388c24e763c29367

crates/solvers/src/boundary/naive.rs
nlordell: Added comment (see thread 👇).

crates/solvers/src/boundary/naive.rs
nlordell: 6700e0fb429776121c5399e4388c24e763c29367

crates/solvers/src/boundary/naive.rs
nlordell: 6700e0fb429776121c5399e4388c24e763c29367

MartinquaXD (approved):

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,245 @@
...
+) -> Option<solution::Solution> {
+    let pool = match &liquidity.state {
+        liquidity::State::ConstantProduct(pool) => pool,
+        _ => return None,
+    };
+
+    // Note that the `order::Order` -> `boundary::LimitOrder` mapping here is
+    // not exact. Among other things, the signature and various signed order
+    // fields are missing from the `order::Order` data that have access to in
+    // the solver engines. This means that the naive solver in the `solver`
```
vkgnosis, line 47:
```suggestion
    // the solver engines. This means that the naive solver in the `solvers`
```

vkgnosis (approved):

crates/solvers/src/boundary/naive.rs
ennmichael, line 47: > fields are missing from the `order::Order` data that have access to in the solver engines

There's definitely a typo in here but I can't tell what it is.

nlordell: I meant it as `solver::naive_solver` in that it would produce invalid `Settlement` values, but that `solvers::naive_solver` doesn't care because it can extract the information it needs anyway.

crates/solvers/src/boundary/naive.rs
```diff
@@ -0,0 +1,245 @@
...
+                })
+            })
+            .collect(),
+    })
+}
+
+// Beyond this point is... well... nameless and boundless chaos. The
+// unfathomable horrors that follow are not for the faint of heart!
+//
+// Joking asside, the existing naive solver implementation is tightly coupled
```
ennmichael, line 156:
```suggestion
// Joking aside, the existing naive solver implementation is tightly coupled
```

Whatever happened about that spell checker?

crates/solvers/src/domain/naive.rs
```diff
@@ -0,0 +1,86 @@
+//! "Naive" solver implementation.
+//!
+//! The naive solver is a solver that collects all orders over a single token
+//! pair, and matches the excess over a Uniswap V2 pool.
```
ennmichael, line 4: I don't understand what you mean by "matches the excess over a Uniswap V2 pool".

crates/solvers/src/domain/naive.rs
```diff
@@ -0,0 +1,86 @@
...
+struct Group<'a> {
+    orders: Vec<&'a order::Order>,
+    liquidity: &'a liquidity::Liquidity,
+    pool: &'a liquidity::constant_product::Pool,
+}
+
+type Groups<'a> = HashMap<liquidity::TokenPair, Group<'a>>;
+
+/// Groups an auction by token pairs, where each group contains all orders over
+/// the token pair as well as the **deepest** constant product pool.
```
ennmichael, line 42: It's not clear to me what "deepest" means in this context. From reading the implementation presumably it means that it has the highest `k`? Do you think there may be a more explicit term for this?

crates/solvers/src/domain/naive.rs
```diff
@@ -0,0 +1,86 @@
...
+            .or_insert_with(|| Group {
+                orders: Vec::new(),
+                liquidity,
+                pool,
+            });
+    }
+
+    for order in &auction.orders {
+        // the naive solver algorithm is sensitive to 0-amount orders (i.e. they
+        // cause panics). Make sure we don't consider them.
```
ennmichael, line 69:
```suggestion
        // The naive solver algorithm is sensitive to 0-amount orders (i.e. they
        // cause panics). Make sure we don't consider them.
```

ennmichael (commented):
(Haven't reviewed everything yet)

crates/solvers/src/domain/liquidity/constant_product.rs
```diff
@@ -1,6 +1,10 @@
 //! Constant product pool.
 
-use {crate::domain::eth, std::cmp::Ordering};
+use {
+    crate::domain::{eth, liquidity},
+    ethereum_types::U256,
```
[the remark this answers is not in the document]
nlordell: I didn't put `U256` in `domain::eth` in the `solvers` crate. My reasoning is that it is a primitive type as much as u128 is. I could move it there if you feel strongly, but it would affect more of this crate (and would likely do it in a separate PR).

crates/solvers/src/domain/naive.rs
nlordell: What ever can't be matched peer-to-peer. Will try and clarify.

nlordell: 12e2d3205a6b8ac29ac655129fd3c4d687e74ef5

crates/solvers/src/domain/naive.rs
nlordell: 12e2d3205a6b8ac29ac655129fd3c4d687e74ef5

PR #1233, opened 2023-02-18
Add Naive Solver Tests

nlordell:
This PR is a follow up to #1231, adding interface tests for the naive solver. It ports most existing Naive solver tests (with the exception of a few non-applicable tests related to things like slippage and buying Eth which are taken care by the driver and not the solver).

The reason for this effort is that I **really** want to move the naive solver logic into the `solvers` crate. The boundary implementation is quite hacky as a result of the `naive` solver being tightly coupled with `Settlement` building which we don't want for the `solvers` crate.

Furthermore, the tests caught a bad unit test in the `naive_solver` module. It was checking the pool execution for a settlement that was impossible (as it would violate the order's limit prices) 🎉.

### Test Plan

🥄


crates/solver/src/solver/naive_solver.rs
```diff
@@ -445,25 +445,19 @@ mod tests {
...
 
-        settle(SlippageContext::default(), orders, liquidity);
-        let TokenAmount {
-            token: out_token,
-            amount: out_amount,
-        } = amm_handler.calls.lock().unwrap()[0].output;
-        assert_eq!(out_token, usdc);
-        assert!(out_amount < U256::from(32275540_u128));
+        let settlements = settle(SlippageContext::default(), orders, liquidity);
+        assert!(settlements.is_empty());
```
nlordell, line 460: This test's assertion was useless. It was capturing the last swap amounts, but there was no settlement so that data invalid anyway.

It turns out that this settling these orders is actually impossible, so assert that instead.

MartinquaXD (approved):

vkgnosis (approved):

PR #1234, opened 2023-02-19
Port Balancer SOR API

nlordell:
Fixes #issue.

### Test Plan

There are currently no tests. The reason for this is that we don't initialize a Balancer SOR API instance anywhere to be able to set up an E2E test. However, I manually added a test to see it work, and it did:

<details><summary><code>manual-test.patch</code></summary>

```diff
diff --git a/crates/solvers/src/infra/dex/balancer/mod.rs b/crates/solvers/src/infra/dex/balancer/mod.rs
index 3c26bb6e..6acda70c 100644
--- a/crates/solvers/src/infra/dex/balancer/mod.rs
+++ b/crates/solvers/src/infra/dex/balancer/mod.rs
@@ -149,3 +149,52 @@ pub enum Error {
     #[error(transparent)]
     Http(#[from] reqwest::Error),
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn manual() {
+        shared::tracing::initialize_reentrant("solvers=trace");
+
+        let sor = Sor {
+            client: reqwest::Client::new(),
+            endpoint: std::env::var("BALANCER_SOR_URL").unwrap().parse().unwrap(),
+            vault: boundary::balancer::Vault::new(eth::ContractAddress(
+                "0xBA12222222228d8Ba445958a75a0704d566BF2C8"
+                    .parse()
+                    .unwrap(),
+            )),
+            settlement: eth::ContractAddress(
+                "0x9008D19f58AAbD9eD0D60971565AA8510560ab41"
+                    .parse()
+                    .unwrap(),
+            ),
+        };
+
+        let swap = sor
+            .swap(
+                &dex::Order {
+                    sell: eth::TokenAddress(
+                        "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    buy: eth::TokenAddress(
+                        "0xba100000625a3754423978a60c9317c58a424e3D"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    side: order::Side::Sell,
+                    amount: dex::Amount::new(1_000_000_000_000_000_000_u128.into()),
+                },
+                dex::Slippage::new(0.01).unwrap(),
+                auction::GasPrice(eth::Ether(15_000_000_000_u128.into())),
+            )
+            .await
+            .unwrap();
+
+        panic!("{swap:#?}");
+    }
+}

```

</details>

### DIFF
--- a/crates/solvers/src/boundary/balancer.rs
+++ b/crates/solvers/src/boundary/balancer.rs
@@ -1,0 +1,93 @@
+//! Balancer boundary module. This contains encoding logic for the Balancer V2
+//! vault contract. The reason this was moved to a boundary module is that it
+//! currently requires `web3` as dependencies (not to mention needing to
+//! instantiate a "dummy" web3 instance) for ABI encoding, which should **not**
+//! be necessary at all. The swap parameter types are here instead of the domain
+//! as they should be generated with the contract bindings (and are more like
+//! DTOs than domain objects anyway).
+
+pub use ethcontract::I256;
+use {
+    crate::domain::{dex, eth},
+    contracts::BalancerV2Vault,
+    ethcontract::Bytes,
+    ethereum_types::{H160, H256, U256},
+};
+
+pub struct Vault(BalancerV2Vault);
+
+#[repr(u8)]
+pub enum SwapKind {
+    GivenIn = 0,
+    GivenOut = 1,
+}
+
+pub struct Swap {
+    pub pool_id: H256,
+    pub asset_in_index: U256,
+    pub asset_out_index: U256,
+    pub amount: U256,
+    pub user_data: Vec<u8>,
+}
+
+pub struct Funds {
+    pub sender: H160,
+    pub from_internal_balance: bool,
+    pub recipient: H160,
+    pub to_internal_balance: bool,
+}
+
+impl Vault {
+    pub fn new(address: eth::ContractAddress) -> Self {
+        Self(shared::dummy_contract!(BalancerV2Vault, address.0))
+    }
+
+    pub fn address(&self) -> eth::ContractAddress {
+        eth::ContractAddress(self.0.address())
+    }
+
+    pub fn batch_swap(
+        &self,
+        kind: SwapKind,
+        swaps: Vec<Swap>,
+        assets: Vec<H160>,
+        funds: Funds,
+        limits: Vec<I256>,
+        deadline: U256,
+    ) -> dex::Call {
+        dex::Call {
+            to: self.address(),
+            calldata: self
+                .0
+                .methods()
+                .batch_swap(
+                    kind as _,
+                    swaps
+                        .into_iter()
+                        .map(|swap| {
+                            (
+                                Bytes(swap.pool_id.0),
+                                swap.asset_in_index,
+                                swap.asset_out_index,
+                                swap.amount,
+                                Bytes(swap.user_data),
+                            )
+                        })
+                        .collect(),
+                    assets,
+                    (
+                        funds.sender,
+                        funds.from_internal_balance,
+                        funds.recipient,
+                        funds.to_internal_balance,
+                    ),
+                    limits,
+                    deadline,
+                )
+                .tx
+                .data
+                .expect("calldata")
+                .0,
+        }
+    }
+}
--- a/crates/solvers/src/domain/dex.rs
+++ b/crates/solvers/src/domain/dex.rs
@@ -55,7 +55,7 @@ pub struct Slippage(f64);
 
 impl Slippage {
     /// Creates a new slippage value. The value represents the slippage
-    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerace is
+    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerance is
     /// accepted, i.e. a swap can return 5% less than promised). Returns `None`
     /// if an invalid value outside of the range [0.0, 1.0] is specified.
     pub fn new(value: f64) -> Option<Self> {
--- a/crates/solvers/src/domain/dex.rs
+++ b/crates/solvers/src/domain/dex.rs
@@ -129,7 +129,7 @@ pub struct Allowance {
     pub amount: Amount,
 }
 
-/// A token amount
+/// A token amount.
 #[derive(Debug)]
 pub struct Amount(U256);
 
--- a/crates/solvers/src/domain/dex.rs
+++ b/crates/solvers/src/domain/dex.rs
@@ -1,0 +1,143 @@
+//! Various solver implementations rely on quoting APIs from DEXs and DEX
+//! aggregators. This domain module models the various types around quoting
+//! single orders with DEXs and turning swaps into single order solutions.
+
+use {
+    crate::{
+        domain::{eth, order, solution},
+        util::fmt,
+    },
+    ethereum_types::U256,
+    std::fmt::{Debug, Formatter},
+};
+
+/// An order for quoting with an external DEX or DEX aggregator. This is a
+/// simplified representation of a CoW Protocol order.
+#[derive(Debug)]
+pub struct Order {
+    pub sell: eth::TokenAddress,
+    pub buy: eth::TokenAddress,
+    pub side: order::Side,
+    pub amount: Amount,
+}
+
+impl Order {
+    /// Returns the order swapped amount as an asset. The token associated with
+    /// the asset is dependent on the side of the DEX order.
+    pub fn amount(&self) -> eth::Asset {
+        eth::Asset {
+            token: match self.side {
+                order::Side::Buy => self.buy,
+                order::Side::Sell => self.sell,
+            },
+            amount: self.amount.0,
+        }
+    }
+}
+
+impl From<order::Order> for Order {
+    fn from(value: order::Order) -> Self {
+        Self {
+            sell: value.sell.token,
+            buy: value.buy.token,
+            side: value.side,
+            amount: Amount(match value.side {
+                order::Side::Buy => value.buy.amount,
+                order::Side::Sell => value.buy.amount,
+            }),
+        }
+    }
+}
+
+/// A slippage tolerance (as a factor) that can be applied to token amounts.
+#[derive(Clone, Copy, Debug)]
+pub struct Slippage(f64);
+
+impl Slippage {
+    /// Creates a new slippage value. The value represents the slippage
+    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerace is
+    /// accepted, i.e. a swap can return 5% less than promised). Returns `None`
+    /// if an invalid value outside of the range [0.0, 1.0] is specified.
+    pub fn new(value: f64) -> Option<Self> {
+        if !(0.0..=1.0).contains(&value) {
+            return None;
+        }
+        Some(Self(value))
+    }
+
+    /// Adds slippage to the specified token amount. This can be used to account
+    /// for negative slippage in a sell amount.
+    pub fn add(&self, amount: U256) -> U256 {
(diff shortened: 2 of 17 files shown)


crates/solvers/src/infra/dex/balancer/dto.rs
```diff
@@ -0,0 +1,155 @@
+use {
```
nlordell, line 1: This is just copied from the `balancer_sor_api.rs` file. One notable exception 👇.

crates/solvers/src/infra/dex/balancer/dto.rs
```diff
@@ -0,0 +1,155 @@
...
+            buy_token: order.buy.0,
+            order_kind: match order.side {
+                order::Side::Buy => OrderKind::Buy,
+                order::Side::Sell => OrderKind::Sell,
+            },
+            amount: order.amount().amount,
+            gas_price: gas_price.0 .0,
+        }
+    }
+}
```
nlordell, line 54: New code that previously was not in `balancer_sor_api.rs`.

crates/solvers/src/domain/dex.rs
```diff
@@ -0,0 +1,143 @@
...
+impl Slippage {
+    /// Creates a new slippage value. The value represents the slippage
+    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerace is
+    /// accepted, i.e. a swap can return 5% less than promised). Returns `None`
+    /// if an invalid value outside of the range [0.0, 1.0] is specified.
+    pub fn new(value: f64) -> Option<Self> {
+        if !(0.0..=1.0).contains(&value) {
+            return None;
+        }
+        Some(Self(value))
```
MartinquaXD, line 65: No strong preference here, just wanted to spread awareness for `Option::then_some()`.
```suggestion
        (0.0..=1.0).contains(&value).then_some(Self(value))
```

crates/solvers/src/infra/dex/balancer/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    ) -> Result<dex::Swap, Error> {
+        let query = dto::Query::from_domain(order, gas_price);
+        let quote = {
+            // Set up a tracing span to make debugging of API requests easier.
+            // Historically, debugging API requests to external DEXs was a bit
+            // of a headache.
+            static ID: AtomicU64 = AtomicU64::new(0);
+            let id = ID.fetch_add(1, atomic::Ordering::Relaxed);
+            let span = tracing::trace_span!("swap", id = %id);
+            let _enter = span.enter();
```
MartinquaXD, line 47: This only works for synchronous code:
> Warning: In asynchronous code that uses async/await syntax, Span::enter may produce incorrect traces if the returned drop guard is held across an await point. See [the method documentation](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code) for details.

For async functions this would do what you expect:
```rust
async {}.instrument(span);
```

This would also be compatible with my next suggestion: It might be nice to have the tracing span for the entire `swap()` function. That way we could also associate any new logs that are related to (de)serializing request/response (should we ever add them).

crates/solvers/src/infra/dex/balancer/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        boundary,
+        domain::{auction, dex, eth, order},
+    },
+    ethereum_types::U256,
+    std::sync::atomic::{self, AtomicU64},
+};
+
+mod dto;
+
+pub struct Sor {
```
MartinquaXD, line 14: Already noted in a follow up PR (reviewed the stack backwards) but I think we should either spell out the acronym here or put a doc comment on this `struct` instead of the module because you would actually see that with IDE inline hints when using the type. (I believe the doc comments would not be displayed)

crates/solvers/src/domain/dex.rs
nlordell: I always forget it exists. Awesome!

crates/solvers/src/infra/dex/balancer/mod.rs
nlordell: Nice catch!

nlordell: > It might be nice to have the tracing span for the entire `swap()` function.

I intentionally only set up a span around the `POST /quote` API request as a way to link the request and response objects.

My general strategy for these single order solvers would be to have a span with `{order = "...uid..."}` for each attempted to solve order, making it easier to debug exactly what happened to a specific order. Also worth noting that, in theory for a partially fillable order, you might have multiple concurrent Balancer SOR requests for a single order (in order to find a partial fill amount).

crates/solvers/src/infra/dex/balancer/mod.rs
nlordell: Agreed. 0bcffecd411b53910a72df8d7b67955c567747ff

crates/solvers/src/infra/dex/balancer/mod.rs
nlordell: Fixed the instrumentation issue in 0bcffecd411b53910a72df8d7b67955c567747ff. Created issue to remember to do a general "instrument `solvers` crate" pass in #1239

MartinquaXD (approved):

vkgnosis (approved):

crates/solvers/src/domain/dex.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+/// A slippage tolerance (as a factor) that can be applied to token amounts.
+#[derive(Clone, Copy, Debug)]
+pub struct Slippage(f64);
+
+impl Slippage {
+    /// Creates a new slippage value. The value represents the slippage
+    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerace is
```
ennmichael, line 58:
```suggestion
    /// tolerance as a factor (so 0.05 means that up to 5% slippage tolerance is
```

crates/solvers/src/domain/dex.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Debug)]
+pub struct Allowance {
+    /// The spender address that requires an allowance in order to execute a
+    /// swap.
+    pub spender: eth::ContractAddress,
+    /// The amount, in tokens, of the required allowance.
+    pub amount: Amount,
+}
+
+/// A token amount
```
ennmichael, line 132:
```suggestion
/// A token amount.
```

crates/solvers/src/boundary/balancer.rs
```diff
@@ -0,0 +1,93 @@
+//! Balancer boundary module. This contains encoding logic for the Balancer V2
+//! vault contract. The reason this was moved to a boundary module is that it
+//! currently requires `web3` as dependencies (not to mention needing to
+//! instantiate a "dummy" web3 instance) for ABI encoding, which should **not**
+//! be necessary at all. The swap parameter types are here instead of the domain
+//! as they should be generated with the contract bindings (and are more like
+//! DTOs than domain objects anyway).
```
ennmichael, line 7: Putting the types here rather than the domain was indeed the right decision. However, if the domain ever needed to access them, that would probably be a good indicator that this needs to be refactored and they should be moved into the domain. But currently AFAIU they're only accessed by `infra` so this is totally fine. :+1:

crates/solvers/src/infra/dex/balancer/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        let (input, output) = match order.side {
+            order::Side::Buy => (quote.return_amount, quote.swap_amount),
+            order::Side::Sell => (quote.swap_amount, quote.return_amount),
+        };
+
+        let (max_input, min_output) = match order.side {
+            order::Side::Buy => (slippage.add(input), output),
+            order::Side::Sell => (input, slippage.sub(output)),
+        };
```
ennmichael, line 82: Let me try to walk you through my thought process while reading this code, because I might be misunderstanding something.

It seems like this particular match statement should be part of the domain logic, because the way that you apply the slippage to the input/output amounts is a business rule and it's the same for all DEXs, it's not specific to just Balancer. (Is that correct?) The first thing that comes to mind is `order.apply_slippage(slippage)` but these `input` and `output` variables show that this is not the right idea. Maybe it's a step in the right direction? Maybe there should be a `dex::Quote` object with `apply_slippage`? WDYT?

ennmichael (commented):
(Haven't reviewed everything yet)

ennmichael (commented):
Please don't merge this yet - I'd like to do another pass through this.

crates/solvers/src/infra/dex/balancer/mod.rs
nlordell: This is a good point, and I _kind_ of tried to do it. The idea is that in the `domain` you have:
- `dex::slippage::Slippage` - the general slippage parameters
- `dex::Order` - simplified orders for DEX and DEX aggregators
- `dex::slippage::Tolerance` - the actual slippage tolerance needed for an order.

The computation of all of these things are done in the `domain` - because, as you pointed out, it is business logic.

The issue is, as you mentioned that applying a slippage tolerance to a Balancer swap. Here, we add slippage to one of the swap amounts **just for computing the swap calldata** - but the actual slippage that should be used, and how it gets determined for a particular order is kept the domain. The reason I thought that this fits here in infra code, is that it is adapter logic to make the the Balancer SOR API quack like a DEX/DEXag API quacks. For example 1Inch, 0x, ParaSwap all have a "slippage" parameter that gets added to the API query and used for computing the calldata - we emulate this here for Balancer SOR as the calldata gets computed locally instead of by the API. Does this make sense?

> Maybe there should be a dex::Quote object with apply_slippage? WDYT?

I actually *almost* did this, and introduced a:

```
struct dex::Execution {
    input: eth::Asset,
    output: eth::Asset,
}
```

kind of how I broke up the `dex::Swap` and added a sub-struct `dex::Call` for use with the `boundary::balancer::Vault` abstraction. By adding a `dex::Execution`, I could have moved this match statement into a `slippage::Tolerance::apply(dex::Execution, order::Side)`. I chose against it because encoding slippage into calldata _felt_ infrastructure-y enough to keep it as is (especially considering that all other DEX/DEXag abstractions would have it this way as well). In fact, if the Balancer contracts were written a different way, it could have just been an additional parameter in the calldata for a `slippage_tolerance_bps` such that the limit amounts get computed on-chain.

ennmichael (commented):
`boundary` isolates `contracts` from `infra`, which doesn't make sense. `infra` already uses `contracts::WETH9` directly. I understand you also wanted to isolate the "dummy web3", but why? There's no problem if `infra` knows about `web3`. Furthermore, the proper solution might be to push constructing the "dummy web3" instance into `contracts::dummy_web3()` and then just call that.

The corollary of the above is that `boundary::balancer` probably should not exist and these types should be defined in `infra`.

Other than that I think that your decision of what to put in `infra` and what to leave in the `domain` is great and the resulting architecture is good. :+1:

crates/solvers/src/infra/dex/balancer/mod.rs
ennmichael: Looking at your follow-up PR and the explanation you gave here, everything seems correct to me.

ennmichael (approved):

nlordell:
> I understand you also wanted to isolate the "dummy web3", but why?

I think needing a `web3` instance to encode calldata bothers me a disproportionate amount and wanted to put it there more as a "this is techdebt" and less of a "this needs a boundary to abstract some legacy code". Your arguments have convinced me, I'll nuke `boundary::balancer` and move that module in to `infra::balancer::vault`.

PR #1236, opened 2023-02-19
Add Balancer SOR Solver E2E Tests

nlordell:
This PR adds E2E tests to the ported Balancer SOR solver.

Nothing too crazy happening here. I also moved the `tempfile` logic out of `create_temp_config_file` into the `tests::SolverEngine` creation in order to make it easier to reuse for future solvers.

Added a `dex::setup` API mock for mocking external DEX HTTP APIs. I made it accept bothe `GET` and `POST` methods right off the bat as they will be needed for other DEX aggregator API solvers.

### Test Plan

🥄

### DIFF
--- a/crates/solvers/src/tests/dex.rs
+++ b/crates/solvers/src/tests/dex.rs
@@ -1,0 +1,113 @@
+use std::{
+    net::SocketAddr,
+    sync::{Arc, Mutex},
+};
+
+#[derive(Debug, Clone)]
+pub enum Expectation {
+    Get {
+        path: String,
+        res: serde_json::Value,
+    },
+    Post {
+        path: String,
+        req: serde_json::Value,
+        res: serde_json::Value,
+    },
+}
+
+/// Set up an mock external DEX or DEX aggregator API.
+pub async fn setup(expectations: Vec<Expectation>) -> SocketAddr {
+    let state = Arc::new(Mutex::new(expectations));
+    let app = axum::Router::new()
+        .route(
+            "/*path",
+            axum::routing::get(
+                |axum::extract::State(state),
+                 axum::extract::Path(path),
+                 axum::extract::RawQuery(query)| async move {
+                    axum::response::Json(get(state, Some(path), query))
+                },
+            )
+            .post(
+                |axum::extract::State(state),
+                 axum::extract::Path(path),
+                 axum::extract::RawQuery(query),
+                 axum::extract::Json(req)| async move {
+                    axum::response::Json(post(state, Some(path), query, req))
+                },
+            ),
+        )
+        // Annoying, but `axum` doesn't seem to match `/` with the above route,
+        // so explicitely mount `/`.
+        .route(
+            "/",
+            axum::routing::get(
+                |axum::extract::State(state), axum::extract::RawQuery(query)| async move {
+                    axum::response::Json(get(state, None, query))
+                },
+            )
+            .post(
+                |axum::extract::State(state),
+                 axum::extract::RawQuery(query),
+                 axum::extract::Json(req)| async move {
+                    axum::response::Json(post(state, None, query, req))
+                },
+            ),
+        )
+        .with_state(State(state));
+    let server = axum::Server::bind(&"0.0.0.0:0".parse().unwrap()).serve(app.into_make_service());
+    let addr = server.local_addr();
+    tokio::spawn(async move { server.await.unwrap() });
+    tokio::time::sleep(tokio::time::Duration::from_millis(1_000)).await;
+    addr
+}
+
+#[derive(Debug, Clone)]
+struct State(Arc<Mutex<Vec<Expectation>>>);
+
+fn get(state: State, path: Option<String>, query: Option<String>) -> serde_json::Value {
+    let mut state = state.0.lock().unwrap();
+    assert!(
+        !state.is_empty(),
+        "got another GET request, but didn't expect any more"
+    );
+
+    let full_path = full_path(path, query);
+    let (expected_path, res) = match state.pop().unwrap() {
+        Expectation::Get { path, res } => (path, res),
+        other => panic!("expected GET request but got {other:?}"),
+    };
+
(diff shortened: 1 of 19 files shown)


crates/solvers/src/tests/legacy/mod.rs
```diff
@@ -48,18 +50,13 @@ pub async fn setup(expectations: Vec<Expectation>) -> SocketAddr {
...
-pub fn create_temp_config_file(solver_addr: &SocketAddr) -> tempfile::TempPath {
-    let base_url = format!("http://{solver_addr}/solve");
-    let config = format!(
+/// Creates a legacy solver configuration for the specified host.
+pub fn config(solver_addr: &SocketAddr) -> tests::Config {
+    tests::Config::String(format!(
         r"
 solver-name = 'legacy_solver'
-solve-endpoint = '{}'
+solve-endpoint = 'http://{solver_addr}/solve'
```
MartinquaXD, line 58: The balancer config calls this just `endpoint`. I don't really care about the specific name but I think we should make it consistent across config files if applicable.

crates/solvers/src/tests/legacy/mod.rs
```diff
@@ -48,18 +50,13 @@ pub async fn setup(expectations: Vec<Expectation>) -> SocketAddr {
...
 chain-id = '1'
-",
-        base_url
-    );
-    let mut file = tempfile::NamedTempFile::new().unwrap();
-    file.write_all(config.as_bytes()).unwrap();
-    file.into_temp_path()
+        ",
```
MartinquaXD, line 60: You nitted on this in my other PR. I don't care myself. Just wanted to point it out since you preferred it that way. Same thing applies in `crates/solvers/src/tests/balancer/mod.rs` as well.
```suggestion
",
```

crates/solvers/src/tests/dex.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Debug, Clone)]
+struct State(Arc<Mutex<Vec<Expectation>>>);
+
+fn get(state: State, path: Option<String>, query: Option<String>) -> serde_json::Value {
+    let mut state = state.0.lock().unwrap();
+    assert!(
+        !state.is_empty(),
+        "got another GET request, but didn't expect any more"
+    );
```
MartinquaXD, line 74: nit to promote safer mutex access patterns (same for `post()`):
```suggestion
    let expectation = state.0.lock().unwrap().pop();
    assert!(
        expectation.is_some(),
        "got another GET request, but didn't expect any more"
    );
```

crates/solvers/src/tests/dex.rs
```diff
@@ -0,0 +1,113 @@
...
+                 axum::extract::Json(req)| async move {
+                    axum::response::Json(post(state, None, query, req))
+                },
+            ),
+        )
+        .with_state(State(state));
+    let server = axum::Server::bind(&"0.0.0.0:0".parse().unwrap()).serve(app.into_make_service());
+    let addr = server.local_addr();
+    tokio::spawn(async move { server.await.unwrap() });
+    tokio::time::sleep(tokio::time::Duration::from_millis(1_000)).await;
```
MartinquaXD, line 62: I assume you copied this `sleep` from the mocking server of the `legacy` tests. I only had it in for debugging something and forgot to drop it. I think it's safe to not wait here.

crates/solvers/src/tests/dex.rs
```diff
@@ -0,0 +1,113 @@
...
+    },
+    Post {
+        path: String,
+        req: serde_json::Value,
+        res: serde_json::Value,
+    },
+}
+
+/// Set up an mock external DEX or DEX aggregator API.
+pub async fn setup(expectations: Vec<Expectation>) -> SocketAddr {
```
MartinquaXD, line 20: Do you think we should rather only have 1 mock server implementation and also use this one for the `legacy` tests?

crates/solvers/src/tests/dex.rs
nlordell: Thanks! Good point on the Mutex usage.

crates/solvers/src/tests/dex.rs
nlordell: Yes, I thought about this but didn't want change too much here. Will do it in a follow up.

crates/solvers/src/tests/legacy/mod.rs
nlordell: The formatting nit was to vertically align the opening and closing `"` characters in the string (which is how it should be here).

crates/solvers/src/tests/legacy/mod.rs
nlordell: I saw you explicitly renamed `endpoint` to `solve-endpoint` (which makes sense for the legacy solver, since the endpoint is for `/solve`-ing). I'm not sure I like the name `solve-endpoint` here, as the Balancer SOR API is more of a DEXag-style API for quoting.

MartinquaXD: I agree that `solve-endpoint` wouldn't fit well for the balancer solver but I think `endpoint` would be good enough for the `legacy` solver. There is only 1 endpoint that makes sense in that context so it's probably not needed to be overly specific. No strong opinion, though.

crates/solvers/src/tests/dex.rs
nlordell: Fixed in 12d4c68aef74da554a89366f6286cad379e7d870

crates/solvers/src/tests/dex.rs
nlordell: Fixed in 12d4c68aef74da554a89366f6286cad379e7d870

MartinquaXD (approved):

crates/solvers/src/tests/dex.rs
nlordell: https://github.com/cowprotocol/services/pull/1244

vkgnosis (approved):

PR #1243, opened 2023-02-20
Refactor token list module

vkgnosis:
Factored out of https://github.com/cowprotocol/services/pull/1229 .

For an e2e test we need a token whose buffers are allowed to be traded. The existing test accomplishes this by manually creating the token list component. With the change to using only the main/run functions of the crates this is no longer possible.

In this PR we change the driver arguments to allow passing in a hardcoded token list and not using a remote URL. We also simplify the token list component by removing extra token information like the symbol, decimals etc that are not relevant for the rust code.

### Test Plan

CI

### DIFF
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -117,10 +117,14 @@ pub struct Arguments {
     )]
     pub solver_time_limit: Duration,
 
-    /// The list of tokens our settlement contract is willing to buy when
-    /// settling trades without external liquidity
-    #[clap(long, env, default_value = "https://files.cow.fi/token_list.json")]
-    pub market_makable_token_list: String,
+    /// The URL of a list of tokens our settlement contract is willing to buy
+    /// when settling trades without external liquidity
+    #[clap(long, env)]
+    pub market_makable_token_list: Option<Url>,
+
+    /// Like `market_makable_token_list` but hardcoded list of tokens.
+    #[clap(long, env, use_value_delimiter = true)]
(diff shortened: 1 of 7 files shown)


crates/solver/src/arguments.rs
```diff
@@ -119,8 +119,12 @@ pub struct Arguments {
 
     /// The list of tokens our settlement contract is willing to buy when
     /// settling trades without external liquidity
-    #[clap(long, env, default_value = "https://files.cow.fi/token_list.json")]
-    pub market_makable_token_list: String,
+    #[clap(long, env)]
+    pub market_makable_token_list: Option<String>,
```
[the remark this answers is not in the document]
vkgnosis: It's already configured.

crates/solver/src/arguments.rs
```diff
@@ -339,10 +343,18 @@ impl std::fmt::Display for Arguments {
...
+            "market_makable_token_list",
+            &self.market_makable_token_list,
+        )?;
+        display_option(
+            f,
+            "market_makable_tokens",
+            &self
+                .market_makable_tokens
+                .as_ref()
+                .map(|list| format!("{list:?}")),
```
MartinquaXD, line 357: Don't we only have tokens here?:
```suggestion
                .map(|token| format!("{token:?}")),
```

MartinquaXD (approved):

crates/solver/src/arguments.rs
MartinquaXD: Ahh, makes sense. 👌

PR #1244, opened 2023-02-20
General HTTP Mocking Helper

nlordell:
This PR is a follow up to https://github.com/cowprotocol/services/pull/1236#discussion_r1111323638.

Specifically, it gets rid of the `legacy::setup` mock HTTP legacy solver function in favour of a general HTTP API mocking helper.

What is nice about this change is that we are additionally checking that the legacy HTTP solver query parameters are being correctly set 🎉.

In terms of a technical decisions, this PR introduces a dependency on the `glob` crate for matching HTTP path patterns that are non-deterministic (such as the date timestamp in the that the legacy solver expects in the URL).

### Test Plan

Tests continue to pass.


crates/solvers/src/tests/mock/http.rs
```diff
@@ -1,22 +1,63 @@
...
+                .debug_tuple("Exact")
+                .field(&format_args!("{exact}"))
+                .finish(),
+            Path::Glob(glob) => f
+                .debug_tuple("Glob")
+                .field(&format_args!("{}", glob.as_str()))
+                .finish(),
+        }
+    }
+}
```
nlordell, line 45: Not strictly needed, but makes the test assertions easier to parse.

crates/solvers/src/tests/mock/http.rs
```diff
@@ -39,7 +80,7 @@ pub async fn setup(expectations: Vec<Expectation>) -> SocketAddr {
             ),
         )
         // Annoying, but `axum` doesn't seem to match `/` with the above route,
-        // so explicitely mount `/`.
+        // so explicitly mount `/`.
```
nlordell, line 83: Look ma! Spell checker!

crates/solvers/src/tests/legacy/jit_order.rs
```diff
@@ -2,13 +2,20 @@
...
 #[tokio::test]
 async fn test() {
-    let legacy_solver = tests::legacy::setup(vec![legacy::Expectation {
+    let legacy_solver = mock::http::setup(vec![mock::http::Expectation::Post {
+        path: mock::http::Path::glob(
+            "solve\
+             [?]instance_name=*_Mainnet_1_0\
+               &time_limit=*\
+               &max_nr_exec_orders=100\
+               &use_internal_buffers=true"
```
MartinquaXD, line 17: I think it would be cool if we could reduce the test code to the things that matter. As long as we have 1 test that covers the query parameters I think adding this glob path to all the tests makes them unnecessarily noisy.

MartinquaXD (approved):
Great that we can also test query parameters now. 👍

crates/solvers/src/tests/legacy/jit_order.rs
nlordell: Good point. I changed the tests to so that only the `market_orders` test verifies the query string (one for `quote` and one for `solve`).

da21fd61d6d8001ec8c06b093e7ae10b0b0c6af7

crates/solvers/src/tests/mock/http.rs
```diff
@@ -1,22 +1,66 @@
...
+        Self::Exact(s.to_string())
+    }
+
+    pub fn glob(s: impl AsRef<str>) -> Self {
+        Self::Glob(glob::Pattern::new(s.as_ref()).unwrap())
+    }
+}
+
+impl PartialEq<Path> for String {
+    fn eq(&self, path: &Path) -> bool {
```
vkgnosis, line 25: Not blocking merge of PR but I think we should not use PartialEq for this. Using equality in the code below made sense when the path was a plain string, now that we're doing complex matching logic this is going to be confusing later. Would change it to `matches(&str) -> bool` and call that below.

vkgnosis (approved):

crates/solvers/src/tests/mock/http.rs
nlordell: > Would change it to matches(&str) -> bool and call that below.

I had this, but then we can't use `assert_eq!` and need to manually build the panic message. Since it was for testing code, I thought it was fine to "abuse" the equality operator and `assert_eq!`.

In general, I agree with you though.

vkgnosis: You could have `assert_matches` which can build the panic message in one place. Anyway, no change required for merging.

PR #1249, opened 2023-02-21
Fix transaction-submission-nodes driver argument

vkgnosis:
My e2e refactor added a default to the argument. This is problematic because in shadow mode we don't want any such node and it is hard to set the argument to no value with clap.

### Test Plan

see that CI succeeds


fedgiac (approved):
Thanks. Not sure if still there, but there was a variable `NODE_HOST` that could be used instead of `http://127.0.0.1:8545`.

PR #1248, opened 2023-02-21
Port 0x Api

nlordell:
Similar to #1234, this PR ports the 0x API to the `solvers` crate.

Nothing too special, although we make use of the `reqwest::RequestBuilder::query(q: impl serde::Serialize)` to avoid manual serialization code 🎉.

Additionally, I added some (hopefully) more complete documentation to the DTOs with links to the online documentation. I also removed fields that we don't need/use to keep the DTOs as slim as possible.

### Test Plan

No E2E test quite yet. However, a manual test can be run:

<details><summary><code>manual-test.patch</code></summary>

```diff
diff --git a/crates/solvers/src/infra/dex/zeroex/mod.rs b/crates/solvers/src/infra/dex/zeroex/mod.rs
index 287385da..d2ff7ce3 100644
--- a/crates/solvers/src/infra/dex/zeroex/mod.rs
+++ b/crates/solvers/src/infra/dex/zeroex/mod.rs
@@ -190,3 +190,55 @@ impl From<dto::Error> for Error {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn manual() {
+        shared::tracing::initialize_reentrant("solvers=trace");
+
+        let dex = ZeroEx {
+            client: reqwest::Client::new(),
+            endpoint: DEFAULT_URL.parse().unwrap(),
+            settlement: eth::ContractAddress(
+                "0x9008D19f58AAbD9eD0D60971565AA8510560ab41"
+                    .parse()
+                    .unwrap(),
+            ),
+            defaults: dto::Query {
+                excluded_sources: vec![
+                    "Uniswap".to_owned(),
+                    "SushiSwap".to_owned(),
+                    "Curve".to_string(),
+                ],
+                ..Default::default()
+            },
+        };
+
+        let swap = dex
+            .swap(
+                &dex::Order {
+                    sell: eth::TokenAddress(
+                        "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    buy: eth::TokenAddress(
+                        "0xE41d2489571d322189246DaFA5ebDe1F4699F498"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    side: order::Side::Sell,
+                    amount: dex::Amount::new(1_000_000_000_000_000_000_u128.into()),
+                },
+                &dex::Slippage::one_percent(),
[...]

PR #1250, opened 2023-02-21
Port 1Inch API

nlordell:
Similar to #1248 but for 1Inch.

I also cleaned up some 0x API issues from the aforementioned PR when I was refactoring out the `serialize::CommaSeparated` to be shared with both. Specifically:
- We don't need to configure the `settlement` contract address
- We weren't actually checking response errors (noticed when copying error types over)

### Test Plan

Like the other PR, can be tested with a manual test:

<details><summary><code>manual-test.patch</code></summary>

```diff
diff --git a/crates/solvers/src/infra/dex/oneinch/mod.rs b/crates/solvers/src/infra/dex/oneinch/mod.rs
index bcad12d0..b6b92989 100644
--- a/crates/solvers/src/infra/dex/oneinch/mod.rs
+++ b/crates/solvers/src/infra/dex/oneinch/mod.rs
@@ -188,3 +188,53 @@ impl From<dto::Error> for Error {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::domain::order};
+
+    #[tokio::test]
+    async fn manual() {
+        shared::tracing::initialize_reentrant("solvers=trace");
+
+        let dex = OneInch::new(Config {
+            endpoint: None,
+            settlement: eth::ContractAddress(
+                "0x9008D19f58AAbD9eD0D60971565AA8510560ab41"
+                    .parse()
+                    .unwrap(),
+            ),
+            liquidity: Liquidity::Exclude(vec!["UNISWAP_V3".to_owned(), "PMM4".to_owned()]),
+            referrer: None,
+            main_route_parts: None,
+            connector_tokens: None,
+            complexity_level: None,
+        })
+        .await
+        .unwrap();
+
+        let swap = dex
+            .swap(
+                &dex::Order {
+                    sell: eth::TokenAddress(
+                        "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    buy: eth::TokenAddress(
+                        "0x111111111117dC0aa78b770fA6A738034120C302"
+                            .parse()
+                            .unwrap(),
+                    ),
+                    side: order::Side::Sell,
+                    amount: dex::Amount::new(1_000_000_000_000_000_000_u128.into()),
+                },
+                &dex::Slippage::one_percent(),
+                auction::GasPrice(eth::Ether(15_000_000_000_u128.into())),
+            )
+            .await
+            .unwrap();
+
+        panic!("{swap:#?}");
+    }
+}

```

</details>.


vkgnosis (approved):
> Like the other PR, can be tested with a manual test

Why not include these in the module? They seem useful. We'll always want to have an ignored test that talks to the actual external API when we make changes.

crates/solvers/src/util/num.rs
```diff
@@ -0,0 +1,63 @@
...
+use {
+    bigdecimal::BigDecimal,
+    num::{BigInt, Signed},
+};
+
+/// Round a [`bigdecimal::BigDecimal`] to the specified precision.
+///
+/// This is the same as [`bigdecimal::BigDecimal::round`], but it does not panic
+/// for high precision values.
+pub fn round(x: &BigDecimal, round_digits: i64) -> BigDecimal {
```
MartinquaXD, line 12: Instead of reimplementing the original `round` couldn't you just detect the case where it would panic and call `with_prec(round_digits + 1)` to avoid the panic. I don't know if this works but I think it's probably less error prone to build the same functionality by composing existing functions.

nlordell: I think this is a good suggestion!

The reason I originally dismissed this was that `with_prec` and `round`-ing have fundamentally different semantics:

```rust
let b = "1234.56789".parse::<BigDecimal>().unwrap();
println!("with_prec: {}", b.with_prec(2));
println!("round:     {}", b.round(2));
```
```
with_prec: 1200
round:     1234.57
```

that being said, we **should** have the invariant that slippage is less than 1 since the limits has the invariant that `slippage.relative ∈ [0, 1]`. This means that we should be able to do it the way you suggested.

A slightly more "fool-proof" approach would be to use `.with_prec(MAX_PRECISION_BEFORE_ROUND_PANICS)`. Computing `MAX_PRECISION_BEFORE_ROUND_PANICS` is actually kind of straight forward - in the `round` implementation, it basically does a `BigInt.try_into::<i128>().unwrap()` which causes the panic for very large precision `BigDecimal` values. This means that `i128::MAX.to_string().len() - 1` is the maximum allowed precision for use with `round`. In fact, you can test this theory and see that it seems to be correct:

```rust
let max = i128::MAX.to_string().len() - 1;
println!("MAX = {max}");
for i in 1.. {
    println!("{i:>2}: {}", b.with_prec(i).round(4));
}
```

Yields:
```
MAX = 38
 1: 40
...
38: 42.1158
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /Users/nlordell/.cargo/registry/src/github.com-1ecc6299db9ec823/bigdecimal-0.3.0/src/lib.rs:594:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So, a pretty pragmatic work around for this would be `slippage.with_prec(38).round(i)`. WDYT?

MartinquaXD: LGTM

nlordell: 26ceb91aa1830092822e6b2573075b6a973aa6d5

MartinquaXD (approved):

PR #1251, opened 2023-02-22
Add uniswap v3 liquidity to the driver

MartinquaXD:
Adds everything needed for the driver to supply solvers with uni v3 liquidity.
* reading config arguments from disk
* driver internal uni v3 representation
* univ3 serialization logic to put it into the auction
* slight corrections to existing uni v3 types
* slight refactor of uni v3 interaction encoding
* 2 tweaks to fulfill things Quasimodo expects from an auction

### Test Plan
Did a manual e2e test that sent an auction to quasimodo that only included univ3 liquidity which produced a simulatable settlement.
Added a `solvers` e2e test that verifies the (de)serialization of concentrated liquidity pools.


crates/solvers/src/domain/liquidity/concentrated.rs
```diff
@@ -37,3 +37,13 @@ pub struct Amount(pub U256);
...
+
+/// The amount of liquidity added (or, if negative, removed) when the tick is
+/// crossed going left to right.
+#[derive(Debug, Copy, Clone)]
+pub struct LiquidityNet(pub i128);
+
+/// Amount of fees accrued when using using this pool.
+/// Can be only one of these values: 5, 30 or 100 bps
+#[derive(Clone, Debug)]
+pub struct Fee(pub eth::Rational);
```
vkgnosis, line 49: Should this be an enum if what the comment says is true?

vkgnosis (approved):

crates/solvers/src/domain/liquidity/concentrated.rs
MartinquaXD: The comment is actually not entirely true. (Didn't update after finding out). These are the current fees but more could theoretically be added by a DAO decision. Will update the comment.

But even if it was true I'm a bit unsure about the exact types of the uniswap v3 pool internals at every stage of the pipeline.
I guess in the driver it makes sense to use the most accurate representation possible because we don't use it in any computations but for example in the `solvers` it might make sense to store values that can only be `u128` as a `U256` if that makes the type more ergonomic.

vkgnosis: Depends on the specific case but in general I prefer more accurate types. Doing `.into()` on a u128 when a U256 is needed for computation isn't a big deal but trying to later go back to u128 for something that was mistakenly declared as U256 is hard.

In this case because the valid values for the fee can change any time, it makes sense to keep it as Rational.

MartinquaXD:
Converting back to draft as there are were some issues with (de)serialization while working on a test for the `legacy` wrapper. Will make this e2e test part of this PR to show which exact format will be supported.

vkgnosis (changes requested):
unapproving for later review

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -250,8 +267,7 @@ struct ConcentratedLiquidityPool {
     #[serde_as(as = "serialize::U256")]
     liquidity: eth::U256,
     tick: i32,
-    #[serde_as(as = "HashMap<serde_with::DisplayFromStr, serialize::U256>")]
-    liquidity_net: HashMap<i32, eth::U256>,
+    liquidity_net: BTreeMap<i32, i128>,
```
nlordell, line 270: Shouldn't this be:
```suggestion
    #[serde_as(as = "HashMap<serde_with::DisplayFromStr, serde_with::DisplayFromStr>")]
    liquidity_net: BTreeMap<i32, i128>,
```

crates/number-conversions/src/lib.rs
```diff
@@ -56,6 +66,12 @@ pub fn big_decimal_to_u256(big_decimal: &BigDecimal) -> Option<U256> {
     big_int_to_u256(&big_int).ok()
 }
 
+pub fn rational_to_big_decimal<T: Clone + Integer + ToPrimitive + ToBigInt>(
+    value: &Ratio<T>,
+) -> Option<BigDecimal> {
+    BigDecimal::from_f64(value.to_f64()?)
```
nlordell, line 72: I _think_ this might lead to precision issues for Balancer (as the denominator is 1e18).

Maybe we should do:

```suggestion
    BigDecimal::new(value.numer(), 0) / BigDecimal::new(value.denom(), 0)
```

vkgnosis: How *does* BigDecimal represent something like `1/3` by the way?

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -16,6 +18,17 @@ impl Auction {
...
+        let liquidity_tokens = liquidity
+            .iter()
+            .flat_map(|l| match &l.kind {
+                liquidity::Kind::UniswapV2(pool) => {
+                    pool.reserves.iter().map(|r| r.token).collect_vec()
+                }
+                liquidity::Kind::UniswapV3(pool) => vec![pool.tokens.get().0, pool.tokens.get().1],
+                _ => vec![],
+            })
+            .map(|t| (t.0.into(), Token::default()));
```
MartinquaXD, line 30: Quasimodo's `/solve` requests will crash if it doesn't find all the liquidity tokens in the list with all the token meta data.. This showed one thing that should probably be redesigned:
Currently we initialize the list of token meta data when we don't yet know which liquidity we'll include in the settlement which means we could miss some.

crates/solvers/src/boundary/legacy.rs
```diff
@@ -114,14 +114,14 @@ fn to_boundary_auction(
...
-                        decimals: info.decimals,
                         alias: info.symbol.clone(),
                         external_price: info
                             .reference_price
                             .map(|price| price.0 .0.to_f64_lossy() / 1e18),
                         internal_buffer: Some(info.available_balance),
                         accepted_for_internalization: info.trusted,
-                        ..Default::default()
+                        decimals: if address.0 == weth.0 { Some(18) } else { None },
+                        normalize_priority: Some(u64::from(address.0 == weth.0)),
```
MartinquaXD: Quasimodo's `/solve` request will crash if it doesn't know the decimals of the reference token. I crudely made the native token the reference token and hardcoded its decimals to 18 until we properly fetch all the token meta data.

crates/number-conversions/src/lib.rs
MartinquaXD: Fixed in [4706a97](https://github.com/cowprotocol/services/pull/1251/commits/4706a97f56b970a144062acd6bda0486cf96a950).

> How does BigDecimal represent something like 1/3 by the way?

It would print repeating numbers with a lot of precision:
```
0.3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333
```
So far I only filled in the blanks of the API that was already defined. I'm not sure if this is a conscious trade-off or if it was an oversight when defining the interface. Do you think this will be fine or should we move to `Ratio<T>` for everything that is not an integer?

crates/solvers/src/domain/liquidity/concentrated.rs
nlordell: I think fees are [`uint24`s on-chain](https://github.com/Uniswap/v3-core/blob/05c10bf6d547d6121622ac51c457f93775e1df09/contracts/UniswapV3Pool.sol#L48), so a `num::Ratio<u32>` is probably the most accurate representation here.

> These are the current fees but more could theoretically be added by a DAO decision.

Exactly, they already added a 4th fee AFAIU.

crates/number-conversions/src/lib.rs
nlordell: Concentrated liquidity fields are always in BPS, so that means a `num::Ratio<u32>` where `assert_eq!(*num.denom(), 10000)`. Because of this, a `BigDecimal` is a perfectly appropriate representation for concentrated liquidity fees (`f64` not, e.g. `0.3_f64 != 0.3`).

The reason for choice of `BigDecimal` in the JSON was purely esthetic: reading `"0.005"` is more natural and more easily associated with the "50 BPS" fee than `[1, 200]` (noting that `num::Ratio` serializes as a tuple and always reduces its numerator and denominator).

In general I agree that `BigDecimal` is not OK for fractional fees, but here we really do have "decimal fees".

vkgnosis (approved):

crates/driver/src/boundary/liquidity/uniswap/v3.rs
```diff
@@ -0,0 +1,132 @@
...
+) -> Result<Box<dyn LiquidityCollecting>> {
+    let web3 = boundary::web3(eth);
+    let router = UniswapV3SwapRouter::at(&web3, config.router.0);
+
+    let pool_fetcher = Arc::new(
+        UniswapV3PoolFetcher::new(
+            eth.chain_id().0.as_u64(),
+            web3.clone(),
+            // TODO: pass client into this function?
+            Default::default(),
```
nlordell, line 116: Not sure I understand this TODO.

crates/driver/src/boundary/liquidity/uniswap/v3.rs
```diff
@@ -0,0 +1,132 @@
...
+        id,
+        gas: eth::Gas(pool.pool.gas_stats.mean_gas),
+        kind: liquidity::Kind::UniswapV3(Pool {
+            router: handler.inner.router.address().into(),
+            address: pool.pool.address.into(),
+            tokens: liquidity::TokenPair::new(
+                pool.pool.tokens[0].id.into(),
+                pool.pool.tokens[1].id.into(),
+            )
+            .unwrap(),
```
nlordell, line 50: I wouldn't `unwrap` here but instead return a `None` or `Error`. Specifically, there is no type-level guarantee that `tokens[0].id != tokens[1].id`. The Subgraph could have a bug where the pool token addresses don't get indexed correctly for example and cause us to panic instead of disregard the liquidity.

I think this is different than if you are unwrapping a `boundary::TokenPair` -> `eth::TokenPair` as they both guarantee the same invariants.

crates/driver/src/boundary/liquidity/uniswap/v3.rs
```diff
@@ -0,0 +1,132 @@
...
+            tick: Tick(pool.pool.state.tick.to_i32().unwrap()),
+            liquidity_net: pool
+                .pool
+                .state
+                .liquidity_net
+                .iter()
+                .map(|(key, value)| {
+                    (
+                        Tick(key.to_i32().unwrap()),
+                        LiquidityNet(value.to_i128().unwrap()),
```
nlordell, line 62: idem.

crates/driver/src/boundary/liquidity/mod.rs
```diff
@@ -76,9 +88,9 @@ impl Fetcher {
         );
 
         Ok(Self {
-            blocks,
+            blocks: block_stream,
             inner: LiquidityCollector {
-                liquidity_sources,
+                liquidity_sources: uni_v2.into_iter().chain(uni_v3.into_iter()).collect(),
```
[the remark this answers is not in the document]
MartinquaXD: Yep, also concluded the same thing when starting with balancer liquidity today. Will fix it in the balancer PR.

crates/driver/src/boundary/liquidity/uniswap/v3.rs
MartinquaXD: It's unclear to me if we need to pass a special `reqwest::Client` into the function or if it is okay to just create it in place. This would affect things like timeout limits for example.

crates/driver/src/boundary/liquidity/uniswap/v3.rs
```diff
@@ -0,0 +1,132 @@
...
+    interaction
+        .encode()
+        .into_iter()
+        .map(|(target, value, call_data)| eth::Interaction {
+            target: eth::Address(target),
+            value: eth::Ether(value),
+            call_data: call_data.0,
+        })
+        .next()
+        .expect("returns exactly 1 interaction")
```
nlordell, line 100: If we expect exactly one interaction, we can use [`Itertools::exactly_one`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one). As it currently is, I don't think it would care if there is two (so the expectation here is more "returns at least 1 interaction").

Normally, I wouldn't suggest changing this, but since we already pull in `itertools`, I thought it was worth being extra explicit.

crates/driver/src/domain/liquidity/uniswap/v3.rs
```diff
@@ -1,5 +1,72 @@
...
+        receiver: &eth::Address,
+    ) -> Option<eth::Interaction> {
+        let tokens_match = (input.0.token == self.tokens.0 && output.0.token == self.tokens.1)
+            || (input.0.token == self.tokens.1 && output.0.token == self.tokens.0);
+
+        tokens_match.then_some(boundary::liquidity::uniswap::v3::to_interaction(
+            self, input, output, receiver,
+        ))
+    }
+}
```
nlordell, line 72: Wow, docs are great! I learned a few things about Uni V3 here, namely:
- Liquidity is a `u128`
- What negative net liquidity means.

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -80,13 +87,32 @@ impl Auction {
...
+                            fee: rational_to_big_decimal(&pool.fee.0),
+                        })
+                    }
                     liquidity::Kind::BalancerV2Stable(_) => todo!(),
                     liquidity::Kind::BalancerV2Weighted(_) => todo!(),
                     liquidity::Kind::Swapr(_) => todo!(),
                     liquidity::Kind::ZeroEx(_) => todo!(),
                 })
                 .collect(),
+            tokens,
```
nlordell, line 115: Nit: I don't think we should rely on the order of field initialization. I think the code is correct here, but I vaguely remember this being a `clippy` lint...

I may be wrong though (maybe @vkgnosis knows).

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -80,13 +87,32 @@ impl Auction {
                             fee: bigdecimal::BigDecimal::new(3.into(), 3),
                         })
                     }
-                    liquidity::Kind::UniswapV3(_) => todo!(),
+                    liquidity::Kind::UniswapV3(pool) => {
+                        tokens.insert(pool.tokens.get().0.into(), Default::default());
+                        tokens.insert(pool.tokens.get().1.into(), Default::default());
```
nlordell, line 92: Doesn't this overwrite existing and correct token metadata with empty token metadata?

crates/solver/src/liquidity/uniswap_v3.rs
```diff
@@ -36,16 +36,34 @@ pub struct UniswapV3Liquidity {
...
     gpv2_settlement: GPv2Settlement,
     // Mapping of how much allowance the router has per token to spend on behalf of the settlement
     // contract
     allowances: Mutex<Allowances>,
 }
 
 pub struct UniswapV3SettlementHandler {
-    inner: Arc<Inner>,
-    fee: Option<u32>,
+    pub inner: Arc<Inner>,
```
nlordell, line 47: Does this need to be public?

crates/driver/src/infra/solver/dto/auction.rs
MartinquaXD: The order of initialization is defined to behave as the code expects it here. Changing this would be a breaking change requiring a new edition also the borrow checker would complain here.

crates/driver/src/infra/solver/dto/auction.rs
nlordell: Do you mind creating an issue for this?

crates/driver/src/boundary/liquidity/uniswap/v3.rs
nlordell: Ah OK, makes sense. Can you add a bit more context to the comment (I think what you wrote here in the PR is enough).

crates/driver/src/infra/solver/dto/auction.rs
nlordell: Is that expected? This was a bit of a surprise to me.

nlordell:
Looks really good to me. The only approval blocking comment was related to overwriting token metadata (or not understanding why we should be doing it).

crates/driver/src/infra/solver/dto/auction.rs
MartinquaXD: No, it's not intended. 👍 was supposed to mean you found a legitimate issue. Will address it shortly.

crates/driver/src/boundary/liquidity/mod.rs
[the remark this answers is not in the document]
MartinquaXD: `concat` didn't want to work so I settled on `[].into_iter().flatten().collect()` which "scales" better so to speak with more liquidity fetchers: [0b36110](https://github.com/cowprotocol/services/pull/1251/commits/0b361102176763a2ddd95305d8596e6f06779674)

crates/driver/src/boundary/liquidity/uniswap/v3.rs
MartinquaXD: added: [c4d033a](https://github.com/cowprotocol/services/pull/1251/commits/c4d033aa665a6a73539aa352488ec03d11720bb5)

crates/driver/src/boundary/liquidity/uniswap/v3.rs
MartinquaXD: added: [9bc2433](https://github.com/cowprotocol/services/pull/1251/commits/9bc2433e41bfa34cc029b8f31877af720c6fb959)

crates/solvers/src/tests/legacy/market_order.rs
```diff
@@ -74,12 +74,12 @@ async fn quote() {
...
                     "external_price": 1.0,
                     "internal_buffer": "1412206645170290748",
-                    "normalize_priority": null
+                    "normalize_priority": 1
                 },
                 "0xdef1ca1fb7fbcdc777520aa7f396b4e015f497ab": {
                     "accepted_for_internalization": true,
                     "alias": "COW",
-                    "decimals": 18,
+                    "decimals": null,
```
nlordell, line 82: Wasn't super clear to me why the `decimals` needed to change in the test.

crates/driver/src/infra/solver/dto/auction.rs
MartinquaXD: added: [b2c52d0](https://github.com/cowprotocol/services/pull/1251/commits/b2c52d07731863536b0df6a649c4fc5cbfbfa0c1)

crates/solver/src/liquidity/uniswap_v3.rs
MartinquaXD: Unfortunately yes.

nlordell (approved):
LGTM

crates/solvers/src/tests/legacy/market_order.rs
MartinquaXD: This was not strictly required but I thought since I added some [logic](https://github.com/cowprotocol/services/pull/1251#discussion_r1115001844) related to that I might as well add a little test for it.

crates/driver/src/boundary/liquidity/uniswap/v3.rs
MartinquaXD: Made the whole thing fallible: [7803165](https://github.com/cowprotocol/services/pull/1251/commits/78031650880bde643422cd2eea419a388f4e40f8), [ad04836](https://github.com/cowprotocol/services/pull/1251/commits/ad0483642e992c7b938c46dfd3bdf5f1e5709a3e)

crates/solvers/src/tests/legacy/market_order.rs
nlordell: Makes sense!

PR #1252, opened 2023-02-22
Improve API bad token error message

vkgnosis:
Currently when a request fails because a token is unsupported the API returns a generic "token ... is unsupported" message. This leads to questions from external teams asking us why a token is not supported. This PR bubbles up the original bad token reason and improves the reasoning string itself.

### Test Plan

CI and manual local test

```
curl https://api.cow.fi/mainnet/api/v1/token/0x8e595470Ed749b85C6F7669de83EAe304C2ec68F/native_price
{"errorType":"UnsupportedToken","description":"Token address 0x8e595470ed749b85c6f7669de83eae304c2ec68f"}
```

```
curl http://localhost:8080/api/v1/token/0x8e595470Ed749b85C6F7669de83EAe304C2ec68F/native_price
{"errorType":"UnsupportedToken","description":"Token 0x8e595470ed749b85c6f7669de83eae304c2ec68f is unsupported: Failed to decode the token's balanceOf response because it did not return 32 bytes. A common cause of this is a bug in the Vyper smart contract compiler. See https://github.com/cowprotocol/services/pull/781 for more information."}
```

### DIFF
--- a/crates/shared/src/bad_token/trace_call.rs
+++ b/crates/shared/src/bad_token/trace_call.rs
@@ -150,7 +150,13 @@ impl TraceCallDetector {
             }
         };
 
-        let message = "Failed to decode the token's balanceOf response because it did not return 32 bytes. A common cause of this is a bug in the Vyper smart contract compiler. See https://github.com/cowprotocol/services/pull/781 for more information.";
+        let message = "\
+            Failed to decode the token's balanceOf response because it did not \
+            return 32 bytes. A common cause of this is a bug in the Vyper \
+            smart contract compiler. See \
+            https://github.com/cowprotocol/services/pull/781 for more \
+            information.\
+        ";
         let bad = TokenQuality::Bad {
             reason: message.to_string(),
         };


crates/shared/src/bad_token/trace_call.rs
```diff
@@ -113,48 +123,56 @@ impl TraceCallDetector {
...
             Err(reason) => {
                 return Ok(TokenQuality::bad(format!(
-                    "can't transfer out of settlement contract: {reason}"
+                    "Transfer token out of settlement contract to arbitrary recipient \
+                     {arbitrary:?} failed: {reason}",
                 )))
             }
         };
 
+        let message = "Failed to decode the token's balanceOf response because it did not return 32 bytes. A common cause of this is a bug in the Vyper smart contract compiler. See https://github.com/cowprotocol/services/pull/781 for more information.";
```
MartinquaXD, line 153: Could this line be wrapped?

crates/shared/src/bad_token/trace_call.rs
```diff
@@ -166,35 +184,44 @@ impl TraceCallDetector {
...
+                    "Transferring {amount} into settlement contract would overflow its balance.",
                 ))
             }
         };
         if balance_after_in != computed_balance_after_in {
             return Ok(TokenQuality::bad(
-                "balance after in transfer does not match",
+                "Transferring {amount} into settlement contract was expected to result in a \
+                 balance of {computer_balance_after_in} but actually resulted in \
+                 {balance_after_in}.",
```
MartinquaXD, line 195: This error case is also pretty common. I think we could also give a best effort hint what might be the exact problem (like for the vyper bug). In this case we could mention that the token could take a fee on transfer. This doesn't explain every such error but probably the majority of them.

MartinquaXD (approved):
Really nice error messages. 👍

crates/shared/src/bad_token/trace_call.rs
vkgnosis: It seems like cargo fmt sometimes autowraps long strings and sometimes doesn't. I'm hesitant to manually wrap this because it means every time we change the text we need to manually reflow a couple of lines. I'd rather have one long line. But no strong feeling.

This is how I would manually format it:
```rust
        let message = "\
            Failed to decode the token's balanceOf response because it did not \
            return 32 bytes. A common cause of this is a bug in the Vyper \
            smart contract compiler. See \
            https://github.com/cowprotocol/services/pull/781 for more \
            information.\
        ";
```

crates/shared/src/bad_token/trace_call.rs
vkgnosis: added

crates/shared/src/bad_token/trace_call.rs
MartinquaXD: I think reflowing the text should not be a problem (I don't expect it to happen often). But I also don't have a strong opinion on this. Feel free to ignore.

nlordell: Not wrapping causes `cargo fmt` to sometimes ignore formatting for long lines. I do think we should should manually wrap until `cargo fmt` can take over.

nlordell (approved):

crates/shared/src/bad_token/trace_call.rs
vkgnosis: Stable rustfmt has never wrapped lines for you (right?). Nightly sometimes does. With this message, regardless of whether I manually wrap it or don't wrap it, nightly rustfmt does not handle it.

fleupold:
This is awesome!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F26xBFT1F9BgskEvTO%2Fgiphy-downsized-medium.gif&token=love" data-gitmeme-token="love"></a>

PR #1253, opened 2023-02-23
Generic DEX Solver

nlordell:
This PR refactors the `Balancer` solver into a generic `Dex` solver.

This allows us to reuse the same implementation as all other DEX/DEX aggregator solvers. The actually solving logic remains the same, I just moved it into a separate function to make instrumenting it easier and also added some new logs.

### Test Plan

Balancer SOR tests continue to pass.


crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        };
+
+        let Some(solution) = swap.into_solution(order) else {
+            tracing::debug!("no solution for swap");
+            return None;
+        };
+
+        tracing::debug!("solved");
+        Some(solution)
```
nlordell, line 75: We talked a bit [in Slack](https://cowservices.slack.com/archives/C049P351GKF/p1677062705741459) about trying to avoid logs in the domain logic.

Notably, logs for the different `Dex::swap` results could in theory be in `infra`. However, I was a bit torn about putting it there as I explicitly want to log every outcome of solving an order (for observability and debugging purposes so we can trace an order from E2E in the logs). With that in mind, it felt a bit weird to log some of the outcomes in `infra` (and rely on logging implementation details of that module), and the outcome of `Swap::into_solution` here (which is domain-logic and can't be pushed into `infra`). Hence, I decided to put the logging of the outcomes of `solve_order` into the domain.

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    async fn solve_order(
+        &self,
+        order: order::Order,
+        prices: &slippage::Prices,
+        gas: auction::GasPrice,
+    ) -> Option<solution::Solution> {
+        let query = dex::Order::new(&order);
+        let slippage = self.slippage.relative(&query.amount(), prices);
```
MartinquaXD, line 51: Should we also check that the absolute slippage gets respected here or does that already happen at a higher level?

MartinquaXD (approved):

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// The DEX API client.
+    pub dex: infra::dex::Dex,
+
+    /// The slippage configuration to use for the solver.
+    pub slippage: slippage::Limits,
+}
+
+impl Dex {
+    pub async fn solve(&self, auction: auction::Auction) -> Vec<solution::Solution> {
+        // TODO: order prioritization, skip liquidity orders, concurrency.
```
vkgnosis, line 27: One important missing thing is respecting the auction's deadline.

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,78 @@
...
+    ) -> Option<solution::Solution> {
+        let query = dex::Order::new(&order);
+        let slippage = self.slippage.relative(&query.amount(), prices);
+
+        let swap = match self.dex.swap(&query, &slippage, gas).await {
+            Ok(swap) => swap,
+            Err(infra::dex::Error::NotFound) => {
+                tracing::debug!("cound not find swap for order");
+                return None;
+            }
```
vkgnosis, line 59: How about just `tracing::debug(err, "skipping order")`? The error already knows what kind of error it is. We don't need to match on all avariants to log a custom a message.

crates/solvers/src/domain/solver/dex.rs
nlordell: That is the plan.

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,78 @@
...
+        prices: &slippage::Prices,
+        gas: auction::GasPrice,
+    ) -> Option<solution::Solution> {
+        let query = dex::Order::new(&order);
+        let slippage = self.slippage.relative(&query.amount(), prices);
+
+        let swap = match self.dex.swap(&query, &slippage, gas).await {
+            Ok(swap) => swap,
+            Err(infra::dex::Error::NotFound) => {
+                tracing::debug!("cound not find swap for order");
```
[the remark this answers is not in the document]
nlordell: I did... but it works with tree-sitter to only check comments. Now I got to figure out how to make it check string literals as well.

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,78 @@
...
+        solutions
+    }
+
+    async fn solve_order(
+        &self,
+        order: order::Order,
+        prices: &slippage::Prices,
+        gas: auction::GasPrice,
+    ) -> Option<solution::Solution> {
+        let query = dex::Order::new(&order);
```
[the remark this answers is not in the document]
nlordell: Correct, `order` is correct here.

crates/solvers/src/domain/solver/dex.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub async fn solve(&self, auction: auction::Auction) -> Vec<solution::Solution> {
+        // TODO: order prioritization, skip liquidity orders, concurrency.
+        let prices = slippage::Prices::for_auction(&auction);
+
+        let mut solutions = Vec::new();
+        for order in auction.orders {
+            let span = tracing::info_span!("solve", order = %order.uid);
+            if let Some(solution) = self
+                .solve_order(order, &prices, auction.gas_price)
+                .instrument(span)
```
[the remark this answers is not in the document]
nlordell: The span was intentional here. I wanted to be able to debug logs related to solving a specific order. We have needed this in the past, and it was annoying to try and match DEX ag requests by token addresses and amounts to a specific order to figure out exactly which DEX ag did what.

PR #1256, opened 2023-02-23
SolverRewards 1: Move out ExternalPrices

sunce86:
Moves `ExternalPrices` struct into `shared` crate.

This is needed to reuse `ExternalPrices` in another shared component (`DecodedSettlement` introduced in follow up PR) and in autopilot.


crates/shared/src/external_prices.rs
```diff
@@ -122,20 +122,6 @@ fn to_native_xrate(price: U256) -> BigRational {
...
-/// Macro for instantiating an `ExternalPrices` instance for testing.
-macro_rules! externalprices {
-    (native_token: $nt:expr $(, $($t:tt)*)?) => {
-        $crate::settlement::external_prices::ExternalPrices::new(
-            $nt,
-            ::maplit::hashmap!($($($t)*)*),
-        )
-        .unwrap()
-    };
-}
```
sunce86, line 135: MooVed to macros.rs

vkgnosis (approved):

crates/shared/src/external_prices.rs
nlordell: > MooVed

🤣

nlordell (approved):

PR #1257, opened 2023-02-23
Cache cow reference price for rewards calculation

### LINKED ISSUE #1238: Special Treatment for CoW Reference Prices
Since CoW reference prices are important for solver rewards computation, we should have some special treatment for them. Specifically, when building the solvable orders, we should use the last known CoW reference price instead of "pausing" the protocol.

Razz4780:
Fixes #1238.

This change introduces a fallback cache for the CoW token price inside one of autopilot's components. The CoW price is used to calculate the risk adjusted solver rewards ([CIP-14](https://forum.cow.fi/t/cip-14-risk-adjusted-solver-rewards/1132)) when building an auction. Therefore, if the autopilot fails to estimate this price, it cannot update the current auction and the protocol stops. This cache will mitigate transient errors (e.g. network failures when fetching CoW price from other APIs) at the cost of using a stale CoW price.

### Test Plan

A unit test for the `BestEffortCowPriceEstimator` struct. Existing tests.


crates/autopilot/src/risk_adjusted_rewards.rs
```diff
@@ -181,6 +203,91 @@ fn uncapped_reward_eth_atoms(
...
+}
+
+/// A caching wrapper over [NativePriceEstimating] used to estimate and cache
+/// CoW token price. Implemented to enable the [Calculator] produce rewards even
+/// when a fresh estimate is not available.
+#[derive(Clone)]
+struct BestEffortCowPriceEstimator {
+    inner: Arc<dyn NativePriceEstimating>,
+    cow_token: H160,
+    fallback_cache: Arc<Mutex<FallbackCache>>,
```
vkgnosis, line 237: This doesn't need Arc. Also the mutex is only needed for `value: Option<(f64, Instant)>`.

(less sure about the following) I would consider inlining `max_age` and `value` into this struct. You can still have the `get` and `set` methods for the unit test.

Razz4780: Indeed, thanks. I flattened the estimator and removed `FallbackCache`

crates/autopilot/src/risk_adjusted_rewards.rs
```diff
@@ -181,9 +203,85 @@ fn uncapped_reward_eth_atoms(
...
+        match fresh {
+            Ok(price) => {
+                self.set_cached(price, current_time);
+                Ok(price)
+            }
+            Err(error) => {
+                let price = self.get_cached(current_time).ok_or(error);
+
+                if price.is_ok() {
+                    tracing::warn!("Using an old CoW token price from a fallback cache");
```
nlordell, line 248: Can we include the original error in the log message? Might be useful for debugging.

MartinquaXD (commented):
This implements a some very specific logic with a very broad tool (trait).
I think I would be lazier about it and try to handle that logic directly in the `Calculator` (the only place where this logic is needed).

crates/autopilot/src/risk_adjusted_rewards.rs
Razz4780: Sure :+1:

Razz4780:
> This implements a some very specific logic with a very broad tool (trait). I think I would be lazier about it and try to handle that logic directly in the `Calculator` (the only place where this logic is needed).

I'm not sure what you mean, I'm not implementing any trait. Just encapsulating the existing estimation flow and the cache inside a struct :eyes:

vkgnosis:
We did some pair programming on this. We felt it was best to implement it privately inside of risk_adjusted_rewards so there is no external trait or anything. It being a separate struct in this module is fine imo.

vkgnosis (approved):

PR #1258, opened 2023-02-23
SolverRewards 2: DecodedSettlement implementation

sunce86:
Adds `DecodedSettlement` structure used for decoding function input of GPv2Settlement::settle function.

Added to autopilot since it's used there but feels like it could live in `shared`, especially if we merge it to already existing `EncodedSettlement` (see comment below)

### Test Plan

Unit tests for surplus and fees calculation.


crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+        U256,            // executedAmount
+        Bytes<Vec<u8>>,  // signature
+    )>,
+    [Vec<(Address, U256, Bytes<Vec<u8>>)>; 3],
+);
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub struct DecodedSettlement {
+    // TODO check if `EncodedSettlement` can be reused
```
sunce86, line 39: I think I could reuse `EncodedSettlement` for this, but did not want to move out EncodedSettlement out of `solver` crate and make even bigger refactor change. But if you agree, I can try to merge these two structs at a later stage.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+
+impl DecodedSettlement {
+    pub fn new(contract: &GPv2Settlement, input: &[u8]) -> Result<Self> {
+        let function = contract
+            .raw_instance()
+            .abi()
+            .function("settle")
+            .context("settle function not found")?;
+        let decoded_input = decode_function_input(function, input)?;
+        <DecodedSettlementTokenized>::from_token(Token::Tuple(decoded_input))
```
sunce86, line 146: @nlordell My gut feelings says this decoding function could be created by our build script and placed inside GPv2Settlement.rs file (but did not have time to investigate in depth). Then, I would avoid having to declare `DecodedSettlementTokenized`

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+                    );
+                    acc
+                }
+            }
+        })
+    }
+}
+
+fn surplus(
+    trade: &DecodedTrade,
```
sunce86, line 195: Surplus calculation is massively copy-pasted from `Settlement` struct. Code duplication could be removed if we merge `DecodedSettlement` and `EncodedSettlement` into one.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+                    .clone()
+                    .into_iter()
+                    .map(Into::into)
+                    .collect(),
+                interactions[2]
+                    .clone()
+                    .into_iter()
+                    .map(Into::into)
+                    .collect(),
+            ],
```
vkgnosis, line 129:
```suggestion
            interactions: interactions.map(|inner| inner.into_iter().map(Into::into).collect()),
```

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+    pub fee_objective_scaling_factor: f64,
+}
+
+impl DecodedSettlement {
+    pub fn new(contract: &GPv2Settlement, input: &[u8]) -> Result<Self> {
+        let function = contract
+            .raw_instance()
+            .abi()
+            .function("settle")
+            .context("settle function not found")?;
```
vkgnosis, line 144:
```suggestion
    pub fn new(input: &[u8]) -> Result<Self> {
        let function = GPv2Settlement::raw_contract()
            .abi
            .function("settle")
            .unwrap();
```

No need to pass in a specific instance of the contract. Unwrap because we know at compile time that this function exists.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
+//! This module contains the logic for decoding the function input for
+//! GPv2Settlement::settle function.
+
+use {
+    anyhow::{anyhow, Context, Result},
+    bigdecimal::{Signed, Zero},
+    contracts::GPv2Settlement,
+    ethcontract::{common::FunctionExt, tokens::Tokenize, Address, Bytes, U256},
+    model::order::{Order, OrderKind},
```
vkgnosis, line 9: Since this is new code I'd like to consider not using the `model::Order` struct. That type contains a lot of information, most of which is not relevant for this component. We should treat it as the order type for the API. For a component like this we can define a new Order struct with the minimal information it needs to work. It can implement `From<model::Order>`. This will make this component easier to maintain.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,511 @@
...
+                let signature = Bytes(
+                    order
+                        .signature
+                        .encode_for_settlement(order.metadata.owner)
+                        .to_vec(),
+                );
+                signature == trade.signature
+            }) {
+                Some(order) => {
+                    acc + fee(trade, external_prices, order, configuration).unwrap_or_default()
```
MartinquaXD, line 165: If the other case results in a warning about "possible incomplete fee calculation" we should probably also get a message if `fee()` returns `None`.

crates/autopilot/src/decoded_settlement.rs
sunce86: Ah yes.

crates/autopilot/src/decoded_settlement.rs
sunce86: Yes, added.

crates/autopilot/src/decoded_settlement.rs
sunce86: Yes, makes sense.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,487 @@
...
+}
+
+fn fee(
+    trade: &DecodedTrade,
+    external_prices: &ExternalPrices,
+    order: &Order,
+    configuration: &FeeConfiguration,
+) -> Option<U256> {
+    let scaled_fee_amount =
+        order.full_fee_amount * U256::from_f64_lossy(configuration.fee_objective_scaling_factor);
```
MartinquaXD, line 246: Turning the `fee_objective_scaling_factor` directly into a `U256` will lose precision as we will probably use very small values (maybe `[1.0, 2.0]`). I think the correct thing would be to turn `full_fee_amount` and `fee_objective_scaling_factor` into `BigRational`, compute the total fee and only then turn the result back into a `U256`.
And in that case you could store `fee_objective_scaling_factor` directly as a `BigRational` in the `FeeConfiguration` to only convert it once.
To have total consistency we could store the `fee_objective_scaling_factor` together with the solutions when we determined the winner.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,487 @@
...
+                    signature: trade.10,
+                })
+                .collect(),
+            interactions: interactions.map(|inner| inner.into_iter().map(Into::into).collect()),
+        }
+    }
+}
+
+pub struct FeeConfiguration {
+    pub fee_objective_scaling_factor: f64,
```
MartinquaXD, line 105: When does this value get populated? I assume it just takes whatever value the `autopilot` was started with. That would mean that we can encounter inconsistencies when:
1. some solver wins right to execute
2. solver tries to bring settlement on-chain
3. we restart the `autopilot` with a different fee scaling value
4. we notice the settlement got mined and we index it with the new fee scaling value opposed to the old one

This is error case would happen almost (or actually) never since we rarely change the fee scaling factor, but I still wanted to bring it up to document it.

crates/autopilot/src/decoded_settlement.rs
sunce86: Ah nice catch, did not test this properly. Thanks.

crates/autopilot/src/decoded_settlement.rs
sunce86: Interesting case. But it feels like it is not a big deal since it just affects the EXACT TIMESTAMP from which we start to calculate the `fee` using the updated parameter.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    /// Returns the total surplus denominated in the native asset for the
+    /// solution.
+    pub fn total_surplus(&self, external_prices: &ExternalPrices) -> U256 {
+        self.trades.iter().fold(0.into(), |acc, trade| {
+            acc + match surplus(trade, &self.tokens, &self.clearing_prices, external_prices) {
+                Some(surplus) => surplus,
+                None => {
+                    tracing::warn!("possible incomplete surplus calculation");
```
vkgnosis, line 152: I'm unsure about this warn. If this is something that shouldn't happen for valid settlements, then this function should return Result. If it is expected to happen then there shouldn't be a warning. There are two similar cases in `total_fees`.

vkgnosis (approved):
I didn't look closely at the exact calculation of the surplus fee etc. Rest looks good.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: I think since this code is important to compute correct rewards we should probably not take any unnecessary shortcuts and avoid loss of precision wherever possible. The code using all `BigRational` is not even more verbose/uglier IMO:
```rust
    let full_fee_amount = u256_to_big_rational(&order.full_fee_amount);
    let scaling_factor = BigRational::from_f64(configuration.fee_objective_scaling_factor)?;
    let scaled_fee_amount = full_fee_amount * scaling_factor;

    let fee = match order.kind {
        model::order::OrderKind::Buy => {
            scaled_fee_amount * u256_to_big_rational(&trade.executed_amount)
                / u256_to_big_rational(&trade.buy_amount)
        }
        model::order::OrderKind::Sell => {
            scaled_fee_amount * u256_to_big_rational(&trade.executed_amount)
                / u256_to_big_rational(&trade.sell_amount)
        }
    };

    let fee = external_prices.try_get_native_amount(order.sell_token, fee)?;
    big_rational_to_u256(&fee).ok()
```

bh2smith (approved):
Seems alright to me.

crates/autopilot/src/decoded_settlement.rs
sunce86: Updated. Set fee_objective_scaling_factor to be BigRational

MartinquaXD (approved):

PR #1260, opened 2023-02-24
[UniV2Like] Add Pancakce Swap to Baseline Liquidity

bh2smith:
Based on a [Request in slack](https://cowservices.slack.com/archives/C0375NV72SC/p1677154605070629) and also (hopefully) to serve as documentation for future liquidity integrations. 

This PR adds Pancake swap Router and Factory contracts.

Also, I am not sure if this documentation [here](https://github.com/cowprotocol/services/tree/main/crates/contracts) is relevant when creating sim links to existing contracts.

## Test Plan

Unfortunately it has been quite a long time since I have touched this and I am not sure how to test this anymore (ideally we see that a pancake pool is being included in an auction).

### DIFF
--- a/crates/contracts/artifacts/PancakeFactory.json
+++ b/crates/contracts/artifacts/PancakeFactory.json
@@ -1,0 +1,1 @@
+IUniswapLikeFactory.json
--- a/crates/contracts/artifacts/PancakeRouter.json
+++ b/crates/contracts/artifacts/PancakeRouter.json
@@ -1,0 +1,1 @@
+IUniswapLikeRouter.json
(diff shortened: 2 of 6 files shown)


vkgnosis:
I recently introduced command line arguments for adding custom univ2 like liquidity without changing code in the repo. We should use that for this. (We should also convert the existing hardcoded sources to the argument but that isn't part of this PR.)

The argument would be `--custom_univ2_baseline_sources=0xeff92a263d31888d860bd50809a8d171709b7b1c|57224589c67f3f30a6b0d7a1b54cf3153ab84563bc609ef41dfb34f8b2974d2d`, which needs to be added to the cluster configuration.

I can do this for you, if you'd like. Let me know here.

bh2smith:
Oh this is interesting thanks @vkgnosis -- I would actually like to see you do it if we could zoom it!?

MartinquaXD:
> My only concern then is that projects are not able to add themselves.

Right now external projects are also not able to add themselves. I mean they could do what you did and add another liquidity source but ultimately this would also have to be enabled in our infrastructure.

bh2smith:
Ok then... so this PR would be the approach for an external project to take?

Issue #1009, opened 2022-12-28
`driver` solution verification


fcked:
The driver should never publish an illegal solution. Therefore, during the solving process process, the following points need to be verified:

- The settlement transaction passes simulation, both with and without respecting the `internalize` flags of the solution.
- The sum of tokens into and out of our settlement contract should be non-negative, i.e. the solver doesn't take any tokens out of the contract.
- Internalized interactions use trusted tokens. This requires checking the internalized interactions in the solution against the trusted tokens in the auction to make sure there's no foul play.

The first prevents the driver from burning gas and paying what would have been their rewards to the protocol. The other two are illegal and if the solvers were to publish such settlements, their rewards would be slashed as punishment, so the driver prevents that from happening.

Idea for implementation: `solution::Verified` as a newtype for `Solution`. `execute` expects a `solution::Verified`.

nlordell:
For additional extra context, the checks besides the "the settlement transaction passes simulation" are for checking off-chain rules that aren't verified on-chain.

nlordell:
Also important to note that the "the settlement transaction passes simulation" check should be done on **both** the "full" settlement (i.e. the settlement including internalized interactions) and the "actual" settlement (i.e. the settlement that only includes non-internalized interactions).

nlordell:
> Why isn't it enough to just check the "actual" settlement?

So, if you 100% trust your settlement, it is enough. However, not being able to provide valid call data for an internalized interaction is a slash-able offense. I think it makes sense for the reference `driver` implementation to check this.

nlordell:
I think this should be:

> Therefore, during ~~both~~ the solving process ~~and the quoting process~~, the following points need to be verified:

Specifically, simulating quotes is not really possible/well defined as there are no orders. In theory you could do some more complex trade simulation (like we do in the `shared::price_estimation::trade_finder` module), but I think this **should** be done in the core protocol services (i.e. `autopilot` and `orderbook`) as quotes are untrusted and need to be verified there anyway.

The reason we want to do these checks specifically for the solving process is to prevent solver engine footguns that lead to slashing.

nlordell:
This should be:

> The first prevents the driver from burning gas **and paying what would have been their rewards to the protocol**.

This is also illegal under the new solver rewards scheme. Drivers would submit the score to the `autopilot` and, if they win and choose to execute the settlement, would need to mine the transaction within a block limit. Failure to do so would make you pay what you would have gotten as rewards to the protocol.

PR #1261, opened 2023-02-24
Internalization verification in the `driver`

sistemd:
Fixes https://github.com/cowprotocol/services/issues/1009.

Verify that internalized interactions only use trusted tokens.

Update the code so that the autopilot specifies which tokens are trusted. Fix the E2E test.

### Test Plan

Automated tests.

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -72,13 +72,9 @@ pub mod solve {
     #[derive(Clone, Debug, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Token {
-        pub decimals: Option<u8>,
-        pub symbol: Option<String>,
         pub address: H160,
         #[serde_as(as = "Option<DisplayFromStr>")]
         pub price: Option<U256>,
-        #[serde_as(as = "DisplayFromStr")]
-        pub available_balance: U256,
         pub trusted: bool,
     }
 
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -153,11 +153,8 @@ impl RunLoop {
                 .prices
                 .iter()
                 .map(|(address, price)| solve::Token {
-                    decimals: Default::default(),
-                    symbol: Default::default(),
                     address: address.to_owned(),
                     price: Some(price.to_owned()),
-                    available_balance: Default::default(),
                     trusted: self.market_makable_token_list.contains(address),
                 })
                 .chain(
@@ -165,11 +162,8 @@ impl RunLoop {
                         .all()
                         .into_iter()
                         .map(|address| solve::Token {
-                            decimals: Default::default(),
-                            symbol: Default::default(),
                             address,
                             price: None,
-                            available_balance: Default::default(),
                             trusted: true,
                         }),
                 )
--- a/crates/driver/src/infra/api/routes/solve/dto/auction.rs
+++ b/crates/driver/src/infra/api/routes/solve/dto/auction.rs
@@ -17,11 +17,11 @@ impl Auction {
                 .tokens
                 .into_iter()
                 .map(|token| competition::auction::Token {
-                    decimals: token.decimals,
-                    symbol: token.symbol,
+                    decimals: None,
+                    symbol: None,
                     address: token.address.into(),
                     price: token.price.map(Into::into),
-                    available_balance: token.available_balance,
+                    available_balance: Default::default(),
                     trusted: token.trusted,
                 })
                 .collect(),
@@ -143,13 +143,9 @@ pub struct Auction {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 struct Token {
-    pub decimals: Option<u8>,
-    pub symbol: Option<String>,
     pub address: eth::H160,
     #[serde_as(as = "Option<serialize::U256>")]
     pub price: Option<eth::U256>,
-    #[serde_as(as = "serialize::U256")]
-    pub available_balance: eth::U256,
     pub trusted: bool,
 }
 
--- a/crates/driver/src/tests/cases/settle.rs
+++ b/crates/driver/src/tests/cases/settle.rs
@@ -166,13 +166,11 @@ async fn test() {
                     {
                         "address": hex_address(sell_token),
                         "price": "1",
-                        "availableBalance": "0",
                         "trusted": false,
                     },
                     {
                         "address": hex_address(buy_token),
                         "price": "2",
-                        "availableBalance": "0",
                         "trusted": false,
                     }
                 ],
--- a/crates/driver/src/tests/cases/solve.rs
+++ b/crates/driver/src/tests/cases/solve.rs
@@ -166,13 +166,11 @@ async fn test() {
                     {
                         "address": hex_address(sell_token),
                         "price": "1",
-                        "availableBalance": "0",
                         "trusted": false,
                     },
                     {
                         "address": hex_address(buy_token),
                         "price": "2",
-                        "availableBalance": "0",
                         "trusted": false,
                     }
                 ],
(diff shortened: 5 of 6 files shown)


crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -180,48 +180,57 @@ impl Solution {
...
-        Ok(())
+        if self
+            .interactions
+            .iter()
+            .filter(|interaction| interaction.internalize())
+            .all(|interaction| {
+                interaction
+                    .inputs()
+                    .iter()
+                    .chain(interaction.outputs().iter())
```
MartinquaXD, line 226: I think we only require that the `input` token is trusted. That is the token that would end up in the settlement contract instead of some liquidity pool so we want to be sure that it's a good token.
It's okay for the `output` token to be not trusted because that means we would get rid of a potential shit coin.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -180,48 +180,57 @@ impl Solution {
...
+
+        // Check that the settlement doesn't steal tokens from the contract.
         if sum_inputs >= sum_outputs {
             Ok(())
         } else {
             Err(VerificationError::AssetFlow)
         }
     }
 
+    /// Check that internalized interactions only use trusted tokens.
```
MartinquaXD, line 213: The idea of internalization is that the settlement contract poses as a liquidity pool to save on fees. This is only supposed to happen for realistic trades which means the solver has to bring proof that this price is achievable on-chain. Proof would be if we can successfully simulate the to-be-internalized transaction to validate that the trade would be legit.

MartinquaXD: AFAIK currently only a single simulation is happening per solution. Not sure if it's the internalized or uninternalized version.

MartinquaXD: The reasoning was already outlined [here](https://github.com/cowprotocol/services/issues/1009#issuecomment-1387621803). Internalization is only legal if the interaction could also be executed on-chain. If you internalize an interaction illegally you would get slashed.
To keep solvers from submitting illegally internalized solutions we should also simulate the uninternalized version so we know that every interaction can be executed on-chain and is therefore allowed to be internalized (if all the `input` tokens are trusted).

MartinquaXD: This would be a sane default. I think if there is a reason solvers why solvers would want to submit a illegal solution knowingly (maybe it's actually more profitable that way for them 🤷‍♂️) it would be more appropriate to make them opt-out of that check with a run-time flag.

crates/autopilot/src/driver_model.rs
```diff
@@ -56,20 +56,32 @@ pub mod solve {
...
 
+    #[serde_as]
+    #[derive(Clone, Debug, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Token {
+        pub decimals: Option<u8>,
+        pub symbol: Option<String>,
+        pub address: H160,
+        #[serde_as(as = "Option<DisplayFromStr>")]
+        pub price: Option<U256>,
```
MartinquaXD, line 79: I think `price`, `address` and `trusted` are the only pieces of information the autopilot has to send. All the other data can get fetched by the `driver` if it's actually needed for its solver.
Also it seems strange to make all prices optional in order to fit in the `trusted` flag somehow. Maybe it makes sense to have 1 entry `prices` as we had before and a list of `trusted_tokens`.

MartinquaXD: Yeah, that's technically correct. Not sure why a `Map<address, price>` felt like stronger guarantee to me. 😅

crates/autopilot/src/driver_model.rs
```diff
@@ -56,20 +56,32 @@ pub mod solve {
...
+        #[serde_as(as = "DisplayFromStr")]
+        pub available_balance: U256,
```
nlordell, line 81: To me, this is something that is augmented by the `driver` and not essential to the protocol. I would be careful in adding more work to the "auction cutting" for unnecessary things.

nlordell: In other words:

`autopilot::Token`
```
    pub struct Token {
        pub address: H160,
        #[serde_as(as = "Option<DisplayFromStr>")]
        pub price: Option<U256>,
        pub trusted: bool,
    }
```

`driver::Token` after "augmenting":
```
    pub struct Token {
        pub decimals: Option<u8>,
        pub symbol: Option<String>,
        pub available_balance: U256,
        ...autopilot::Token
    }
```

crates/autopilot/src/run_loop.rs
```diff
@@ -146,7 +148,19 @@ impl RunLoop {
                     }
                 })
                 .collect(),
-            prices: auction.prices.clone(),
+            tokens: auction
+                .prices
+                .iter()
+                // TODO Fill in the missing fields (Default::default())
```
nlordell, line 154: As per my comment ☝️, I think that the token fields that are set to `Default::default` here should not be part of the `auction` JSON.

crates/autopilot/src/run_loop.rs
```diff
@@ -146,7 +148,19 @@ impl RunLoop {
...
+                .map(|(address, price)| solve::Token {
+                    decimals: Default::default(),
+                    symbol: Default::default(),
+                    address: address.to_owned(),
+                    price: Some(price.to_owned()),
+                    available_balance: Default::default(),
+                    trusted: self.market_makable_token_list.contains(address),
+                } )
```
nlordell, line 162: We need to chain:

```
self
    .market_makable_token_list
    .iter()
    .filter(is_not_already_included)
    .map(|address| Token {
        address,
        price: None,
        trusted: true,
    })
```

crates/autopilot/src/driver_model.rs
nlordell: > I think `price` [...] are the only pieces of information the autopilot has to send

Prices are only required for orders that are traded. This `vec` can include tokens that are trusted but not traded (and hence have no price).

crates/autopilot/src/driver_model.rs
nlordell: Also, I think this can (and likely should) be done in a follow up to not further bloat this PR.

crates/autopilot/src/run_loop.rs
MartinquaXD: I guess you could use `unique_by`.

crates/autopilot/src/driver_model.rs
MartinquaXD: Not sure why this should happen in a different PR as this would actually reduce the diff of this PR so we wouldn't have to verify that the changes have been reverted correctly.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -180,48 +180,56 @@ impl Solution {
     /// Check that the sum of tokens entering the settlement is not less than
     /// the sum of tokens exiting the settlement.
     fn verify_asset_flow(&self) -> Result<(), VerificationError> {
```
nlordell, line 182: This still seems wrong to me. Asset flow verification should be done on individual token amounts and not on estimated reference amounts (which is not always available).

nlordell (approved):

crates/autopilot/src/driver_model.rs
```diff
@@ -72,13 +72,10 @@ pub mod solve {
     #[derive(Clone, Debug, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Token {
-        pub decimals: Option<u8>,
-        pub symbol: Option<String>,
         pub address: H160,
         #[serde_as(as = "Option<DisplayFromStr>")]
         pub price: Option<U256>,
         #[serde_as(as = "DisplayFromStr")]
```
MartinquaXD, line 78: `#[serde_as(as = "DisplayFromStr")]` also needs to get removed to not affect the serialization of `trusted`, I think.

MartinquaXD (approved):
Thanks for already removing the fields. 👍 
One line is still left to remove but I'm already approving to not require another back and forth when it gets removed.

PR #1263, opened 2023-02-24
SolverRewards 3: Updated db schema and read/write functions

sunce86:
Autopilot and orderbook callsite functions will be introduced in a follow up PR.

### DIFF
--- a/crates/database/src/auction_participants.rs
+++ b/crates/database/src/auction_participants.rs
@@ -1,35 +1,39 @@
 use {
     crate::{auction::AuctionId, Address},
-    sqlx::PgConnection,
+    sqlx::{Connection, PgConnection},
 };
 
-/// Participants of a solver competition for a given auction.
+/// Participant of a solver competition for a given auction.
 #[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
-pub struct Participants {
+pub struct Participant {
     pub auction_id: AuctionId,
-    pub participants: Vec<Address>,
+    pub participant: Address,
 }
 
-pub async fn insert(ex: &mut PgConnection, data: Participants) -> Result<(), sqlx::Error> {
+pub async fn insert(
+    ex: &mut PgConnection,
+    participants: Vec<Participant>,
+) -> Result<(), sqlx::Error> {
+    let mut transaction = ex.begin().await?;
     const QUERY: &str =
-        r#"INSERT INTO auction_participants (auction_id, participants) VALUES ($1, $2);"#;
-    sqlx::query(QUERY)
-        .bind(data.auction_id)
-        .bind(data.participants)
-        .execute(ex)
-        .await?;
+        r#"INSERT INTO auction_participants (auction_id, participant) VALUES ($1, $2);"#;
+    for participant in participants {
+        sqlx::query(QUERY)
+            .bind(participant.auction_id)
+            .bind(participant.participant)
+            .execute(&mut *transaction)
+            .await?;
+    }
+    transaction.commit().await?;
     Ok(())
 }
 
 pub async fn fetch(
     ex: &mut PgConnection,
     auction_id: AuctionId,
-) -> Result<Option<Participants>, sqlx::Error> {
+) -> Result<Vec<Participant>, sqlx::Error> {
     const QUERY: &str = r#"SELECT * FROM auction_participants WHERE auction_id = $1"#;
-    sqlx::query_as(QUERY)
-        .bind(auction_id)
-        .fetch_optional(ex)
-        .await
+    sqlx::query_as(QUERY).bind(auction_id).fetch_all(ex).await
 }
 
 #[cfg(test)]
@@ -43,13 +47,18 @@ mod tests {
         let mut db = db.begin().await.unwrap();
         crate::clear_DANGER_(&mut db).await.unwrap();
 
-        let data = Participants {
-            auction_id: 1,
-            participants: (0u8..3).map(|i| ByteArray([i; 20])).collect::<Vec<_>>(),
-        };
-        insert(&mut db, data.clone()).await.unwrap();
-
-        let result = fetch(&mut db, 1).await.unwrap().unwrap();
-        assert_eq!(result, data);
+        let input = vec![
+            Participant {
+                auction_id: 1,
+                participant: ByteArray([2; 20]),
+            },
+            Participant {
+                auction_id: 1,
+                participant: ByteArray([3; 20]),
+            },
+        ];
+        insert(&mut db, input.clone()).await.unwrap();
+        let output = fetch(&mut db, 1).await.unwrap();
+        assert_eq!(input, output);
     }
 }
--- a/crates/database/src/auction_participants.rs
+++ b/crates/database/src/auction_participants.rs
@@ -1,0 +1,62 @@
+use {
+    crate::{auction::AuctionId, Address},
+    sqlx::PgConnection,
+};
+
+/// Participants of a solver competition for a given auction.
+#[derive(Debug, PartialEq, sqlx::FromRow)]
+pub struct Participants {
+    pub auction_id: AuctionId,
+    pub participants: Vec<Address>,
+}
+
+pub async fn insert(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+    participants: Vec<Address>,
+) -> Result<(), sqlx::Error> {
+    const QUERY: &str =
+        r#"INSERT INTO auction_participants (auction_id, participants) VALUES ($1, $2);"#;
+    sqlx::query(QUERY)
+        .bind(auction_id)
+        .bind(participants)
+        .execute(ex)
+        .await?;
+    Ok(())
+}
+
+pub async fn fetch(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+) -> Result<Option<Participants>, sqlx::Error> {
+    const QUERY: &str = r#"SELECT * FROM auction_participants WHERE auction_id = $1"#;
+    sqlx::query_as(QUERY)
+        .bind(auction_id)
+        .fetch_optional(ex)
+        .await
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::byte_array::ByteArray, sqlx::Connection};
+
+    #[tokio::test]
+    #[ignore]
+    async fn postgres_roundtrip() {
+        let mut db = PgConnection::connect("postgresql://").await.unwrap();
+        let mut db = db.begin().await.unwrap();
+        crate::clear_DANGER_(&mut db).await.unwrap();
+
+        let participants = (0u8..3).map(|i| ByteArray([i; 20])).collect::<Vec<_>>();
+        insert(&mut db, 1, participants.clone()).await.unwrap();
+
+        let result = fetch(&mut db, 1).await.unwrap();
+        assert_eq!(
+            result,
+            Some(Participants {
+                auction_id: 1,
+                participants
+            })
+        );
+    }
+}
--- a/crates/database/src/auction_participants.rs
+++ b/crates/database/src/auction_participants.rs
@@ -4,22 +4,18 @@ use {
 };
 
 /// Participants of a solver competition for a given auction.
-#[derive(Debug, PartialEq, sqlx::FromRow)]
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
 pub struct Participants {
     pub auction_id: AuctionId,
     pub participants: Vec<Address>,
 }
 
-pub async fn insert(
-    ex: &mut PgConnection,
-    auction_id: AuctionId,
-    participants: Vec<Address>,
-) -> Result<(), sqlx::Error> {
+pub async fn insert(ex: &mut PgConnection, data: Participants) -> Result<(), sqlx::Error> {
     const QUERY: &str =
         r#"INSERT INTO auction_participants (auction_id, participants) VALUES ($1, $2);"#;
     sqlx::query(QUERY)
-        .bind(auction_id)
-        .bind(participants)
+        .bind(data.auction_id)
+        .bind(data.participants)
         .execute(ex)
         .await?;
     Ok(())
@@ -47,16 +43,13 @@ mod tests {
         let mut db = db.begin().await.unwrap();
         crate::clear_DANGER_(&mut db).await.unwrap();
 
-        let participants = (0u8..3).map(|i| ByteArray([i; 20])).collect::<Vec<_>>();
-        insert(&mut db, 1, participants.clone()).await.unwrap();
+        let data = Participants {
+            auction_id: 1,
+            participants: (0u8..3).map(|i| ByteArray([i; 20])).collect::<Vec<_>>(),
+        };
+        insert(&mut db, data.clone()).await.unwrap();
 
-        let result = fetch(&mut db, 1).await.unwrap();
-        assert_eq!(
-            result,
-            Some(Participants {
-                auction_id: 1,
-                participants
-            })
-        );
+        let result = fetch(&mut db, 1).await.unwrap().unwrap();
+        assert_eq!(result, data);
     }
 }
--- a/crates/database/src/auction_prices.rs
+++ b/crates/database/src/auction_prices.rs
@@ -1,7 +1,7 @@
 use {
     crate::{auction::AuctionId, Address},
     bigdecimal::BigDecimal,
-    sqlx::PgConnection,
+    sqlx::{Connection, PgConnection},
 };
 
 /// External token price for a given auction.
@@ -13,16 +13,18 @@ pub struct AuctionPrice {
 }
 
 pub async fn insert(ex: &mut PgConnection, prices: Vec<AuctionPrice>) -> Result<(), sqlx::Error> {
+    let mut transaction = ex.begin().await?;
+    const QUERY: &str =
+        r#"INSERT INTO auction_prices (auction_id, token, price) VALUES ($1, $2, $3);"#;
     for price in prices {
-        const QUERY: &str =
-            r#"INSERT INTO auction_prices (auction_id, token, price) VALUES ($1, $2, $3);"#;
         sqlx::query(QUERY)
             .bind(price.auction_id)
             .bind(price.token)
             .bind(price.price)
-            .execute(&mut *ex)
+            .execute(&mut *transaction)
             .await?;
     }
+    transaction.commit().await?;
     Ok(())
 }
 
--- a/crates/database/src/auction_prices.rs
+++ b/crates/database/src/auction_prices.rs
@@ -1,0 +1,92 @@
+use {
+    crate::{auction::AuctionId, Address},
+    bigdecimal::BigDecimal,
+    sqlx::PgConnection,
+};
+
+/// External prices for a given auction.
+#[derive(Debug, PartialEq, sqlx::FromRow)]
+pub struct Prices {
+    pub tokens: Vec<Address>,
+    pub prices: Vec<BigDecimal>,
+}
+
+pub async fn upsert(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+    tokens: Vec<Address>,
+    prices: Vec<BigDecimal>,
+) -> Result<(), sqlx::Error> {
+    const QUERY: &str = r#"INSERT INTO auction_prices (auction_id, tokens, prices) VALUES ($1, $2, $3)
+        ON CONFLICT (auction_id) DO UPDATE
+        SET tokens = $2, prices = $3
+        ;"#;
+    sqlx::query(QUERY)
+        .bind(auction_id)
+        .bind(tokens)
+        .bind(prices)
+        .execute(ex)
+        .await?;
+    Ok(())
+}
+
+pub async fn fetch(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+) -> Result<Option<Prices>, sqlx::Error> {
+    const QUERY: &str = r#"SELECT tokens, prices FROM auction_prices WHERE
+auction_id = $1;"#;
+    let prices = sqlx::query_as(QUERY)
+        .bind(auction_id)
+        .fetch_optional(ex)
+        .await?;
+    Ok(prices)
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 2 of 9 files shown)


database/sql/V048__create_settlement_rewards.sql
```diff
@@ -12,20 +14,20 @@ CREATE TABLE settlement_observations (
...
-  gas_used numeric(78,0) NOT NULL,
-  effective_gas_price numeric(78,0) NOT NULL,
+  gas_used numeric(78,0),
+  effective_gas_price numeric(78,0),
   -- the surplus observed from the transaction call data,
   -- and converted to ETH with the auction external prices.
-  surplus numeric(78,0) NOT NULL,
-  fee numeric(78,0) NOT NULL,
+  surplus numeric(78,0),
+  fee numeric(78,0),
```
sunce86, line 23: Removed `NOT NULL` because the table is initially populated (block_number and log_index) when GPv2Settlement event is emitted, while the rest is populated in an autopilot background task similar to `AuctionTransactionUpdater`.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -12,20 +14,20 @@ CREATE TABLE settlement_observations (
...
 );
 
 CREATE TABLE auction_prices (
-  auction_id bigint NOT NULL,
-  token bytea NOT NULL,
-  price numeric(78,0) NOT NULL
+  auction_id bigint NOT NULL PRIMARY KEY,
+  tokens bytea[] NOT NULL,
+  prices numeric(78,0)[] NOT NULL
 );
```
sunce86, line 31: Might not be the perfect design but seemed convenient because we always fetch all prices for an auction.

crates/database/src/events.rs
```diff
@@ -64,7 +65,8 @@ pub async fn delete(
     ex.execute(sqlx::query(QUERY_TRADE).bind(delete_from_block_number))
         .await?;
 
-    const QUERY_SETTLEMENTS: &str = "DELETE FROM settlements WHERE block_number >= $1;";
+    const QUERY_SETTLEMENTS: &str = "WITH S AS (DELETE FROM settlements WHERE block_number >= $1) \
+                                     DELETE FROM settlement_observations WHERE block_number >= $1;";
```
MartinquaXD, line 69: Here you could do 2 queries without needing to refactor anything as you already have a `PgTransaction` available.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -3,7 +3,9 @@
 CREATE TABLE settlement_scores (
   auction_id bigint PRIMARY KEY,
   winning_score numeric(78,0) NOT NULL,
-  reference_score numeric(78,0) NOT NULL
+  reference_score numeric(78,0) NOT NULL,
```
MartinquaXD, line 6: What happens if only a single solver offers a valid solution?
Should we maybe keep `reference_score` nullable for that case instead of using a default value?
Also a comment on what the `reference_score` is would be nice.

database/sql/V048__create_settlement_rewards.sql
nlordell: Hmm, I think we should have a table with a single row per price. Then, we would insert multiple rows per auction. I am not a fan of using arrays like this in this table.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -12,20 +14,20 @@ CREATE TABLE settlement_observations (
...
 CREATE TABLE auction_prices (
-  auction_id bigint NOT NULL,
-  token bytea NOT NULL,
-  price numeric(78,0) NOT NULL
+  auction_id bigint NOT NULL PRIMARY KEY,
+  tokens bytea[] NOT NULL,
+  prices numeric(78,0)[] NOT NULL
 );
 
 CREATE TABLE auction_participants (
```
nlordell, line 33: I missed the PR that added this, but I also think that we should have 1 row per participant and not use arrays.

At least this is my simple understanding of how this "should" be (and allow you to query things like `SELECT COUNT(*) from auction_participants where participant = '\x...'` for querying the number of times an address participated in an auction).

crates/database/src/auction_prices.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct Prices {
+    pub tokens: Vec<Address>,
+    pub prices: Vec<BigDecimal>,
+}
+
+pub async fn upsert(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+    tokens: Vec<Address>,
+    prices: Vec<BigDecimal>,
```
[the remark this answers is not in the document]
sunce86: Sure, but then I need to change the `fetch` functions also. But I think that's ok also.

crates/database/src/events.rs
sunce86: With this I was aiming to get the atomic execution because my old code assumed that `settlements` and `settlement_observations` tables are in sync always. But I think it is ok to decouple them.

crates/database/src/events.rs
```diff
@@ -139,15 +141,18 @@ async fn insert_settlement(
     event: &Settlement,
 ) -> Result<(), sqlx::Error> {
     const QUERY: &str = "\
-        INSERT INTO settlements (tx_hash, block_number, log_index, solver) VALUES ($1, $2, $3, $4) \
-                         ON CONFLICT DO NOTHING;";
+    WITH S AS (INSERT INTO settlements (tx_hash, block_number, log_index, solver) VALUES ($1, $2, \
+                         $3, $4) ON CONFLICT DO NOTHING)
```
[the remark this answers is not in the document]
sunce86: Yes, I did not find other way. But this can be decoupled also, because they can be updated separately.

crates/database/src/auction_participants.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Debug, PartialEq, sqlx::FromRow)]
+pub struct Participants {
+    pub auction_id: AuctionId,
+    pub participants: Vec<Address>,
+}
+
+pub async fn insert(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+    participants: Vec<Address>,
```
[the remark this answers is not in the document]
sunce86: Fixed.

crates/database/src/auction_transaction.rs
```diff
@@ -70,6 +70,26 @@ LIMIT 1
         .await
 }
 
+#[derive(Debug, sqlx::FromRow)]
+pub struct Auction {
+    pub auction_id: AuctionId,
+}
```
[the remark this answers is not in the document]
sunce86: Cool, fixed.

crates/database/src/settlement_scores.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub reference_score: BigDecimal,
+    pub block_deadline: i64,
+}
+
+pub async fn insert(
+    ex: &mut PgConnection,
+    auction_id: AuctionId,
+    winning_score: BigDecimal,
+    reference_score: BigDecimal,
+    block_deadline: i64,
```
bh2smith, line 16: Don't we want to just pass the score struct into this method and unpack it inside the function with dot notation?

crates/database/src/settlements.rs
```diff
@@ -18,6 +18,25 @@ WHERE
         .await
 }
 
+pub async fn get_hash_by_event(
```
bh2smith, line 21: I find the naming a bit awkward here because there is no "event" being passed. If it were actually an Event we and we unpacked the (block_number, log_index) inside the method I wouldn't have an issue. 

Also, I think this should return and Optional Hash (because for some block_numbers there may not be a SettlementEvent). It is pretty clear this function is meant to be for a SettlementEvent (since the raw sql is querying the settlement table).

database/sql/V048__create_settlement_rewards.sql
bh2smith: It was my impression that this table will not be populated when no corresponding transaction exists. This is why we had set these fields to NOT NULL. If this table is going to be populated with vacant data. I may have to change the data export queries to match this.

database/sql/V048__create_settlement_rewards.sql
bh2smith: I would be very much in favour of not using arrays. They are hard to parse and also make transposition a two-step process!

bh2smith (approved):
I like what I see here. There are a few places where some things could be changed. Assuming that most of the key concerns are addressed, I think its close enough to approve.

In particular there are a few places where the structure attributes are passed into methods rather than the structure itself. This leaves the function signature unbound to the object that it is clearly bound to operate on (based on the internal implementation). Should definitely be fixed.

crates/database/src/auction_prices.rs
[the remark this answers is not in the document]
MartinquaXD: I initially thought so, too. But it's indeed true that liquidity orders and limit orders can result in multiple prices for the same tokens. So having 2 lists here should be correct.

database/sql/V048__create_settlement_rewards.sql
sunce86: It will be populated with the vacant data only initially when our system detects mined transaction.
Then, in the background task it will be populated with full data following the FINALIZED_BLOCK rule, meaning it will be lagging for 64 blocks. 
Since you are doing your queries on a historic data I think there is no need to change them.

database/sql/V048__create_settlement_rewards.sql
sunce86: Reverted.

crates/database/src/auction_prices.rs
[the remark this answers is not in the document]
sunce86: Refactored. Can you check again pls.

database/sql/V048__create_settlement_rewards.sql
sunce86: Added comment for reference_score.

database/sql/V048__create_settlement_rewards.sql
sunce86: Fixed.

crates/database/src/events.rs
```diff
@@ -47,6 +47,7 @@ pub async fn last_block(ex: &mut PgConnection) -> Result<i64, sqlx::Error> {
     const QUERY: &str = "\
             SELECT GREATEST( (SELECT COALESCE(MAX(block_number), 0) FROM trades), (SELECT \
                          COALESCE(MAX(block_number), 0) FROM settlements), (SELECT \
+                         COALESCE(MAX(block_number), 0) FROM settlement_observations),(SELECT \
```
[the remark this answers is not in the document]
sunce86: I added it for completeness, it does not hurt. However, the function itself is not reliable since the tables can be left in an inconsistent state (append and delete functions are not atomic). At least this is my understanding.

crates/database/src/settlement_observations.rs
```diff
@@ -0,0 +1,164 @@
+use {bigdecimal::BigDecimal, sqlx::PgConnection};
+
+#[derive(Debug, PartialEq, sqlx::FromRow)]
+pub struct SettlementEvent {
+    pub block_number: i64,
+    pub log_index: i64,
+}
```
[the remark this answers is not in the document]
sunce86: Yes, reused `EventIndex` now.

crates/database/src/settlement_scores.rs
sunce86: Done.

crates/database/src/auction_prices.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// External token price for a given auction.
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+pub struct AuctionPrice {
+    pub auction_id: AuctionId,
+    pub token: Address,
+    pub price: BigDecimal,
+}
+
+pub async fn insert(ex: &mut PgConnection, prices: Vec<AuctionPrice>) -> Result<(), sqlx::Error> {
```
MartinquaXD, line 15: I think it would probably be best to use a DB transaction here. Otherwise half of the inserts might fail and we assume everything is in order when looking at the DB.

crates/database/src/settlements.rs
sunce86: As far as I see, `settlements` table is always fully populated, so no need to query optional by hash.
There is also no `by_block_number` query here.
I fixed the event comment.

crates/database/src/auction_prices.rs
MartinquaXD: I think you only have to a transaction:
```rust 
pub async fn insert(ex: &mut PgConnection, prices: Vec<AuctionPrice>) -> Result<(), sqlx::Error> {
    let mut ex = ex.begin().await?;
    // existing insert logic
    ex.commit().await
}
```

MartinquaXD: The same pattern can be applied across multiple functions by passing in a `&mut PgTransaction<'_>` to each one. This would be nicer in case we want to insert multiple things in different tables at once. (Which might be needed for the new logic)

crates/database/src/auction_participants.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Participants of a solver competition for a given auction.
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+pub struct Participants {
+    pub auction_id: AuctionId,
+    pub participants: Vec<Address>,
+}
+
+pub async fn insert(ex: &mut PgConnection, data: Participants) -> Result<(), sqlx::Error> {
+    const QUERY: &str =
+        r#"INSERT INTO auction_participants (auction_id, participants) VALUES ($1, $2);"#;
```
nlordell, line 15: I don't think participants should be an array here.

crates/database/src/auction_prices.rs
nlordell: I also think we should follow the structure of the `database` crate and pass in a `PgTransaction` here. Kind of what we do for other queries:

https://github.com/cowprotocol/services/blob/43a7f9cb9271c47d7c1cce6b2cda72286e106351/crates/database/src/events.rs#L78-L95

sunce86: Fixed.

crates/database/src/settlement_observations.rs
```diff
@@ -0,0 +1,146 @@
...
+
+pub async fn update(
+    ex: &mut PgConnection,
+    block_number: i64,
+    log_index: i64,
+    gas_used: BigDecimal,
+    effective_gas_price: BigDecimal,
+    surplus: BigDecimal,
+    fee: BigDecimal,
```
nlordell, line 36: Might be nice to group this into an `SettlementObservation` struct.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -12,12 +15,12 @@ CREATE TABLE settlement_observations (
   -- table, read from the transaction receipt
   block_number bigint NOT NULL,
   log_index bigint NOT NULL,
-  gas_used numeric(78,0) NOT NULL,
-  effective_gas_price numeric(78,0) NOT NULL,
+  gas_used numeric(78,0),
+  effective_gas_price numeric(78,0),
```
nlordell, line 19: In my mind, these need to be nullable.

Specifically, the `driver`/`solver` needs to ask to populate this **before** mining the transaction, when `gas_used` and `effective_gas_price` is unknown.

sunce86: Agree they should be nullable, but I don't understand why it needs to be populated before mining the transaction? I insert it for the first time when the `settlement event` is raised - transaction mined.

MartinquaXD (approved):
My suggestions/questions have been addressed. We might have to revisit the DB transaction discussion when we see how those functions actually get used, though. For right now it seems fine to me.

crates/database/src/auction_participants.rs
sunce86: Fixed.

crates/database/src/settlement_observations.rs
sunce86: Fixed.

crates/database/src/auction_participants.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Participant of a solver competition for a given auction.
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+pub struct Participant {
+    pub auction_id: AuctionId,
+    pub participant: Address,
+}
+
+pub async fn insert(
+    ex: &mut PgConnection,
+    participants: Vec<Participant>,
```
vkgnosis, line 15:
```suggestion
    participants: &[Participant],
```

crates/database/src/auction_participants.rs
```diff
@@ -0,0 +1,64 @@
...
+) -> Result<(), sqlx::Error> {
+    let mut transaction = ex.begin().await?;
```
vkgnosis, line 17: You shouldn't start a transaction inside of these functions usually. Instead change the type of `ex` to `&mut PgTransaction` and create the transaction in the caller. See `crates/database/src/lib.rs` for more information.

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -30,7 +33,7 @@ CREATE TABLE auction_prices (
 
 CREATE TABLE auction_participants (
  -- This links to the `auctions` table
- auction_id bigint PRIMARY KEY,
- -- All solvers who submitted a valid solution to the auction.
- participants bytea[]
+ auction_id bigint NOT NULL,
+ -- Solver who submitted a valid solution to the auction.
+ participant bytea NOT NULL
 );
```
vkgnosis, line 39: This table has no  primary key but you are running queries on it of the form `SELECT * FROM auction_participants WHERE auction_id = $1`, which indicates you want an index on at least `auction_id`. Since it doesn't make sense for this table to contain the same `(auction_id, participant)` pair twice, we should probably make `(auction_id, participant) ` the primary key.

sunce86:
I think we sorted out all open questions, merging.

crates/database/src/auction_prices.rs
```diff
@@ -0,0 +1,81 @@
...
+            .bind(price.price)
+            .execute(&mut *transaction)
+            .await?;
+    }
+    transaction.commit().await?;
+    Ok(())
+}
+
+pub async fn delete(ex: &mut PgConnection, auction_id: AuctionId) -> Result<(), sqlx::Error> {
+    const QUERY: &str = "DELETE FROM auction_prices WHERE auction_id = $1";
```
vkgnosis, line 32: This table is missing a primary key on `(auction_id, token)` because it does not make sense to have duplicates and it is important for the efficiency of the query.

crates/database/src/events.rs
vkgnosis, line 50: It doesn't make sense to me that we're handling the settlement_observations table in this function. This function shouldn't touch the table.Settlement events go into the `settlements` table. Why do we need to also put them in another table? This is related to another comment where you said you had to change some fields to be non null. I don't think this is the case. You should query the settlements table joined with the observations table in `get_settlement_event_without_observation` instead.

`auction_transaction` is designed to not have coumns be null because rows only get inserted when we have the data. This PR should work the same way.

crates/database/src/settlement_observations.rs
```diff
@@ -0,0 +1,117 @@
...
+
+pub async fn get_settlement_event_without_observation(
+    ex: &mut PgConnection,
+    max_block_number: i64,
+) -> Result<Option<EventIndex>, sqlx::Error> {
+    const QUERY: &str = r#"
+SELECT block_number, log_index
+FROM settlement_observations
+WHERE gas_used IS NULL AND block_number <= $1
+LIMIT 1
```
vkgnosis, line 11: I don't know how this function is going to be used but it seems like you need an index on `(block_number, gas_used)` in order to make this query efficient.

vkgnosis (changes requested):

database/sql/V048__create_settlement_rewards.sql
vkgnosis: This isn't the right approach. See my review below.

crates/database/src/auction_participants.rs
sunce86: Fixed in https://github.com/cowprotocol/services/pull/1270

crates/database/src/auction_participants.rs
sunce86: Fixed in https://github.com/cowprotocol/services/pull/1270

database/sql/V048__create_settlement_rewards.sql
sunce86: Fixed in https://github.com/cowprotocol/services/pull/1270

bh2smith: How are we supposed to determine the winning participant from this now? Used to the the last entry in the list and we still need to know this info. Specifically for the scenario when the auction is not settled and we want to know who to penalize.

crates/database/src/events.rs
bh2smith: I also mentioned that this shouldn't be here but for slightly different reasons.

crates/database/src/settlements.rs
bh2smith: Cool! If your passing a struct that is known to correspond to an entry in the table and not a random pair of integers then the return value does not need to be optional.

database/sql/V048__create_settlement_rewards.sql
vkgnosis: In another comment somewhere here we changed the type from an sql array to store one participant per row. This change removed all ordering of participants per auction. This is unrelated to my comment here. Postgres doesn't guarantee any order if there is no ordering constraint in the query.

crates/database/src/auction_prices.rs
sunce86: Addressed in https://github.com/cowprotocol/services/pull/1271

crates/database/src/events.rs
sunce86: Addressed in https://github.com/cowprotocol/services/pull/1271

crates/database/src/settlement_observations.rs
sunce86: Function not used anymore.

PR #1270, opened 2023-02-24
SolverRewards 4: Populate scores and participants

sunce86:
Populates database tables `settlement_scores` and `auction_participants`.

### DIFF
--- a/crates/e2e/tests/e2e/services.rs
+++ b/crates/e2e/tests/e2e/services.rs
@@ -329,6 +329,8 @@ pub async fn setup_naive_solver_uniswapv2_driver(
         web3.clone(),
         network_id.clone(),
         Duration::from_secs(30),
+        Duration::from_secs(13),
+        Duration::from_secs(60),
         block_stream,
         SolutionSubmitter {
             web3: web3.clone(),
--- a/crates/model/src/solver_competition.rs
+++ b/crates/model/src/solver_competition.rs
@@ -23,6 +23,17 @@ pub struct Request {
     pub transaction: Transaction,
     pub competition: SolverCompetitionDB,
     pub executions: Vec<(OrderUid, Execution)>,
+    pub scores: Scores,
+    pub participants: Vec<H160>, // solver addresses
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct Scores {
+    #[serde(with = "u256_decimal")]
+    pub winning_score: U256,
+    #[serde(with = "u256_decimal")]
+    pub reference_score: U256,
+    pub block_deadline: u64,
 }
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
@@ -76,6 +87,7 @@ pub struct CompetitionAuction {
 #[serde(rename_all = "camelCase")]
 pub struct SolverSettlement {
     pub solver: String,
+    pub solver_address: H160,
     pub objective: Objective,
     #[serde(flatten)]
     pub score: Score, // auction based score
@@ -179,6 +191,7 @@ mod tests {
             "solutions": [
                 {
                     "solver": "2",
+                    "solverAddress": "0x0000000000000000000000000000000000000000",
                     "objective": {
                         "total": 3.0f64,
                         "surplus": 4.0f64,
@@ -227,6 +240,7 @@ mod tests {
                 },
                 solutions: vec![SolverSettlement {
                     solver: "2".to_string(),
+                    solver_address: H160([0x00; 20]),
                     objective: Objective {
                         total: 3.,
                         surplus: 4.,
--- a/crates/model/src/solver_competition.rs
+++ b/crates/model/src/solver_competition.rs
@@ -29,6 +29,7 @@ pub struct Request {
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct Scores {
+    pub winner: H160,
     #[serde(with = "u256_decimal")]
     pub winning_score: U256,
     #[serde(with = "u256_decimal")]
--- a/crates/orderbook/src/database/solver_competition.rs
+++ b/crates/orderbook/src/database/solver_competition.rs
@@ -55,6 +55,7 @@ impl SolverCompetitionStoring for Postgres {
             &mut ex,
             Score {
                 auction_id: request.auction,
+                winner: ByteArray(request.scores.winner.0),
                 winning_score: u256_to_big_decimal(&request.scores.winning_score),
                 reference_score: u256_to_big_decimal(&request.scores.reference_score),
                 block_deadline: request
@@ -167,6 +168,7 @@ mod tests {
             },
             executions: Default::default(),
             scores: Scores {
+                winner: H160([1; 20]),
                 winning_score: 100.into(),
                 reference_score: 99.into(),
                 block_deadline: 10,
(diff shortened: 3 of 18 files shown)


crates/solver/src/driver.rs
```diff
@@ -435,13 +452,33 @@ impl Driver {
                     .map_err(|err| anyhow!("{err}"))
                     .context("convert nonce")?,
             };
+            let scores = model::solver_competition::Scores {
+                winning_score: scores.last().copied().expect("no score"), // guaranteed to exist
+                // reference score is the second highest score, or 0 if there is only one score (see
+                // CIP20)
+                reference_score: scores.get(scores.len() - 2).copied().unwrap_or(0.into()),
```
MartinquaXD, line 459: You could not `.collect()` the `scores` iterator and instead `.pop()` it twice to get the winner and runner-up. Just bringing it up to avoid indexing which can always a bit error prone.

crates/orderbook/src/database/solver_competition.rs
```diff
@@ -47,6 +51,36 @@ impl SolverCompetitionStoring for Postgres {
             .context("order_rewards::save")?;
         }
 
+        database::settlement_scores::insert(
+            &mut ex,
```
MartinquaXD, line 55: Okay at this point it seems clear that the DB insertions actually should take a `PgTransaction` instead of starting their own. 😅

sunce86: Fixed.

crates/solver/src/arguments.rs
```diff
@@ -328,8 +328,28 @@ pub struct Arguments {
...
+    #[clap(
+        long,
+        env,
+        default_value = "13",
+        value_parser = shared::arguments::duration_from_seconds,
+    )]
+    pub block_time: Duration, // TODO release note: set for each network in configuration
+
+    /// Additional time to wait for a transaction to appear onchain before
+    /// considering the solution invalid and setting the reward to 0.
```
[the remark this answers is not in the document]
sunce86: We need it for cases when our `current_block()` calls are not reliable, meaning the call maybe hits a delayed node. This would cause `block_deadline` to be sooner than expected, and transactions being mined at the end of real submission deadline period (but inside) to miss the reward.
Since I needed some additional value, I thought I better put it under configuration.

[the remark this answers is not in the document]
sunce86: Once we move this to autopilot it will make more sense I guess.

crates/orderbook/src/database/solver_competition.rs
MartinquaXD: I think inserting `scores` should also be made part of a DB transaction now. This was not really clear during the previous review.

crates/e2e/tests/e2e/services.rs
```diff
@@ -329,6 +329,8 @@ pub async fn setup_naive_solver_uniswapv2_driver(
         web3.clone(),
         network_id.clone(),
         Duration::from_secs(30),
+        Duration::from_secs(13),
+        Duration::from_secs(60),
```
bh2smith, line 333: I see a lot of these but not really sure what they are all about.

crates/model/src/solver_competition.rs
```diff
@@ -23,6 +23,17 @@ pub struct Request {
...
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct Scores {
+    #[serde(with = "u256_decimal")]
+    pub winning_score: U256,
+    #[serde(with = "u256_decimal")]
+    pub reference_score: U256,
+    pub block_deadline: u64,
 }
```
bh2smith, line 37: If the auction participants are no longer an array, we may have to record the winning solver here.

crates/solver/src/driver.rs
sunce86: True, ended up with nicer code.

bh2smith (approved):
Seems good assuming the comments are addressed. Note the outstanding issue with not knowing the winning solver now that participants is not an array. 

Furthermore, I noticed some discussion about zeroes vs nulls. I wanted to mention that we (on the data side) believe we are prepared for either. If you want to slim down the DB by vacuously assuming lack of existence corresponds to zero, it's fine by me. As long as the necessary information is recoverable in some way I would prefer a database with stricter constraints like NOT NULL anywhere possible.

crates/model/src/solver_competition.rs
sunce86: Putting the winner here is cleaner and more explicit.

sunce86: Added the winner in `scores` table.

sunce86:
@bh2smith Let's address null/not null questions in the next PR where I populate external prices and observations.

crates/solver/src/driver.rs
```diff
@@ -435,13 +452,34 @@ impl Driver {
...
+                        + self.additional_mining_deadline;
+                    let number_blocks = deadline.as_secs() / self.block_time.as_secs();
+                    block_during_simulation + number_blocks
+                },
+            };
+            let participants = solver_competition
+                .solutions
+                .iter()
+                .map(|solution| solution.solver_address)
+                .collect();
```
MartinquaXD, line 473: This slipped through the cracks earlier but currently Gnosis solvers can return multiple solutions which means they would show up multiple times in the participant list (which is probably not intended). An easy fix would be to make `participants` a `HashSet` in the `model::solver_competition::Request`.

bh2smith: Would it then be reasonable to collect as a set and then cast back to vector (or whatever it is supposed to be).

MartinquaXD: nit: would be simpler to just make the `participants` a `HashSet` in the solver competition data upload. Then `.collect()` would already collect into the right thing. But feel free to merge already due to urgency.

MartinquaXD (approved):

crates/solver/src/arguments.rs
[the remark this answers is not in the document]
vkgnosis: Autopilot already has a configuration value for this because it was needed for colocation.

crates/solver/src/driver.rs
```diff
@@ -435,13 +452,36 @@ impl Driver {
                     .map_err(|err| anyhow!("{err}"))
                     .context("convert nonce")?,
             };
+            let scores = model::solver_competition::Scores {
+                winner: winning_solver.account().address(),
```
vkgnosis, line 456: Double checking: Is it intended that the score you use here might come from a different settlement than winning_solver, winning_settlement` above? Because we don't guarantee that the scores and rated_settlements are sorted the same way.

vkgnosis (approved):

crates/solver/src/arguments.rs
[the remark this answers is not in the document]
sunce86: @vkgnosis moved to `shared` and reused.

crates/solver/src/driver.rs
nlordell: Came to ask the same thing, isn't the `winning_solver` not guaranteed to correspond with the winning score?

sunce86: Yes, it's a bug introduced when I added the winner. Thanks, fixed.

nlordell (approved):

PR #1276, opened 2023-02-27
Second try at asset flow verification

sistemd:
Further work on https://github.com/cowprotocol/services/issues/1009.

Correctly (hopefully) implements asset flow verification for market orders. I will implement this for liquidity and limit orders in following PRs. I won't merge this PR until the entire PR chain is ready.

After that, all that will be left for verification is always doing uninternalized simulations.

### Test Plan

Automated tests.

### DIFF
--- a/crates/driver/src/domain/competition/order/mod.rs
+++ b/crates/driver/src/domain/competition/order/mod.rs
@@ -143,9 +143,11 @@ impl PartialEq<[u8; 56]> for Uid {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Side {
-    /// Buy an exact amount.
+    /// Buy an exact amount. The sell amount can vary due to e.g. partial fills
+    /// or slippage.
     Buy,
-    /// Sell an exact amount.
+    /// Sell an exact amount. The buy amount can vary due to e.g. partial fills
+    /// or slippage.
     Sell,
 }
 
@@ -184,8 +186,11 @@ pub enum Kind {
     /// Order intended to be immediately executed. This is the "regular" type of
     /// order.
     Market,
-    /// Order intended to be executed possibly far into the future, when the
-    /// price is such that the order can be executed.
+    /// Order intended to be fulfilled possibly far into the future, when the
+    /// price is such that the order can be executed. Because the fulfillment
+    /// can happen any time into the future, it's impossible to calculate
+    /// the order fees ahead of time, so the fees are taken from the order
+    /// surplus instead.
     Limit {
         /// The fee to be taken from the order surplus.
         surplus_fee: SellAmount,
--- a/crates/driver/src/domain/competition/solution/trade.rs
+++ b/crates/driver/src/domain/competition/solution/trade.rs
@@ -29,6 +29,10 @@ pub struct Fulfillment {
 #[derive(Debug)]
 pub struct Jit {
     pub order: order::Jit,
+    /// The amount executed by this JIT trade. See
+    /// [`competition::order::Jit::partially_fillable`]. If the order is not
+    /// partially fillable, the executed amount must equal the amount from the
+    /// order.
     pub executed: competition::order::TargetAmount,
 }
 
@@ -115,8 +119,111 @@ impl Trade {
                     },
                 }
             }
-            order::Kind::Liquidity => todo!(),
-            order::Kind::Limit { .. } => todo!(),
+            order::Kind::Liquidity => {
+                // Liquidity orders (including JIT) compute the executed amounts by linearly
+                // scaling the buy/sell amounts in the order.
+                match side {
+                    order::Side::Buy => Execution {
+                        buy: eth::Asset {
+                            amount: executed.into(),
+                            token: buy.token,
+                        },
+                        sell: eth::Asset {
+                            amount: sell
+                                .amount
+                                .checked_mul(executed.into())
+                                .ok_or(Error::Overflow)?
+                                .checked_div(buy.amount)
+                                .ok_or(Error::Overflow)?,
+                            token: sell.token,
+                        },
+                    },
+                    order::Side::Sell => Execution {
+                        sell: eth::Asset {
+                            amount: executed.into(),
+                            token: sell.token,
+                        },
+                        buy: eth::Asset {
+                            amount: buy
+                                .amount
+                                .checked_mul(executed.into())
+                                .ok_or(Error::Overflow)?
+                                .checked_div(sell.amount)
+                                .ok_or(Error::Overflow)?,
+                            token: buy.token,
+                        },
+                    },
+                }
+            }
+            order::Kind::Limit { surplus_fee } => {
+                // Warning: calculating executed amounts for limit orders is complex and
+                // confusing. To understand why the calculations work, it is important to note
+                // that the solver doesn't receive limit orders with the same amounts that were
+                // specified by the users when placing the orders. Instead, the sell amount for
+                // each limit order is reduced by the surplus fee, which is the fee taken by
+                // the network to settle the order. These are referred to as "synthetic" limit
+                // orders. The surplus fees are calculated when cutting the auction.
+                //
+                // See also [`order::Kind::Limit`].
+                //
+                // Similar to market orders, the executed amounts for limit orders are
+                // calculated using the clearing prices.
+                let sell_price = clearing_prices
+                    .0
+                    .get(&sell.token)
+                    .ok_or(Error::ClearingPriceMissing)?
+                    .to_owned();
+                let buy_price = clearing_prices
+                    .0
+                    .get(&buy.token)
+                    .ok_or(Error::ClearingPriceMissing)?
+                    .to_owned();
+                match side {
+                    order::Side::Buy => Execution {
+                        buy: eth::Asset {
+                            amount: executed.into(),
+                            token: buy.token,
+                        },
+                        sell: eth::Asset {
+                            amount: executed
+                                .0
+                                .checked_mul(buy_price)
+                                .ok_or(Error::Overflow)?
+                                .checked_div(sell_price)
+                                .ok_or(Error::Overflow)?
+                                // Because of how "synthetic" limit orders are constructed as
+                                // explained above, we need to simply increase the executed sell
+                                // price by the surplus fee. We know that the user placed an order
+                                // big enough to cover the surplus fee.
+                                .checked_add(surplus_fee.into())
+                                .ok_or(Error::Overflow)?,
+                            token: sell.token,
+                        },
+                    },
+                    order::Side::Sell => Execution {
+                        sell: eth::Asset {
+                            amount: executed.into(),
+                            token: sell.token,
+                        },
+                        buy: eth::Asset {
+                            amount: executed
+                                .0
+                                // Because of how "synthetic" limit orders are constructed as
+                                // explained above, the solver received the sell amount
+                                // reduced by the surplus fee. That's why we're have to reduce the
+                                // executed amount by the surplus fee when calculating the
+                                // executed buy amount.
+                                .checked_sub(surplus_fee.into())
+                                .ok_or(Error::Overflow)?
+                                .checked_mul(sell_price)
+                                .ok_or(Error::Overflow)?
+                                .checked_ceil_div(&buy_price)
+                                .ok_or(Error::Overflow)?,
+                            token: buy.token,
+                        },
+                    },
+                }
+            }
         })
     }
 }
--- a/crates/dev-geth/src/main.rs
+++ b/crates/dev-geth/src/main.rs
@@ -94,6 +94,7 @@ struct Processes {
 impl Processes {
     async fn start(&self) -> Port {
         let port = self.next_port();
+        println!("starting {port}");
         let datadir = format!("/{port}");
         let status = tokio::process::Command::new("cp")
             .arg("-r")
@@ -120,10 +121,12 @@ impl Processes {
             .unwrap();
         Self::wait_for_geth(&port).await;
         self.children.insert(port, child);
+        println!("started {port}");
         port
     }
 
     async fn stop(&self, port: Port) {
+        println!("stopping {port}");
         let (_, mut process) = self
             .children
             .remove(&port)
@@ -132,6 +135,7 @@ impl Processes {
         tokio::fs::remove_dir_all(format!("/{}", port.0))
             .await
             .expect("failed to delete the datadir");
+        println!("stopped {port}");
     }
 
     fn next_port(&self) -> Port {
@@ -142,6 +146,7 @@ impl Processes {
     }
 
     async fn wait_for_geth(port: &Port) {
+        println!("waiting for {port} to come online");
         let web3 = web3::Web3::new(
             web3::transports::Http::new(&format!("http://localhost:{}", port.0))
                 .expect("valid URL"),
@@ -161,6 +166,7 @@ impl Processes {
         })
         .await
         .expect("timed out while waiting for geth to become reachable");
+        println!("{port} has come online");
     }
 }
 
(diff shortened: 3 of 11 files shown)


crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -180,7 +182,49 @@ impl Solution {
...
+        // TODO Write a comment here
+        for trade in self.trades.iter() {
+            if let Trade::Fulfillment(trade) = trade {
+                // TODO Question: the executed amount is specific to the
+                // side, how do I get e.g. the sold amount for a
+                // partially fulfilled buy order? Is it
+                // sold = (
+                //   fulfillment.executed.amount / order.buy.amount
+                // ) * order.sell_amount
+                // ?
```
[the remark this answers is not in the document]
MartinquaXD: I think the `executed.amount` is only supposed to be the additional amount executed during this specific trade (does not take current fill state into account) so we shouldn't have to relate it to the total buy or sell amount. I think you only have to multiple `executed.amount` by `buy_price` or `sell_price` depending on the `side`.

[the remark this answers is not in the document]
nlordell: > The idea is that the executed sell amount is proportional to the executed buy amount for a buy order, but I don't know if this is the right way to go.

It is based on the clearing price. So something like this:

```rust
let (executed_sell, executed_buy) = match order.side {
   order::Side::Buy => (
       (order.buy.amount * solution.prices[order.buy.token]).ceil_div(solution.prices[order.sell.token]),
       order.buy.amount,
   ),
   order::Side::Sell => (
       order.sell.token,
       (order.sell.amount * solution.prices[order.sell.token]).ceil_div(solution.prices[order.buy.token]),
   ),
};
```

Note that this code **does not** handle limit orders (as there is the surplus fee to consider), but it should be the gist of how to compute executed amounts.

Follow the [code here](https://github.com/cowprotocol/services/blob/06270c3c10882404b9e369c484187d1f234dbc10/crates/solver/src/settlement/settlement_encoder.rs#L247) to see some of the nuances for computing executed amounts.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -185,7 +179,45 @@ impl Solution {
     /// Check that the sum of tokens entering the settlement is not less than
     /// the sum of tokens exiting the settlement.
     fn verify_asset_flow(&self) -> Result<(), VerificationError> {
-        Ok(())
+        let mut flow: HashMap<eth::TokenAddress, num::BigInt> = Default::default();
+
+        // Interaction inputs represent flow into the contract, i.e. positive flow.
+        for input in self
```
MartinquaXD, line 185: Isn't it actually the other way around? `input` in this context means tokens that get "used" for a given transaction meaning they leave the contract and the `output` flow back into the settlement contract.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -13,14 +20,72 @@ pub enum Trade {
...
+    }
+}
+
+/// The amounts executed by a fulfillment.
+#[derive(Debug, Clone, Copy)]
+pub struct Execution {
+    /// The amount entering the settlement contract.
+    pub input: eth::Asset,
+    /// The amount exiting the settlement contract.
+    pub output: eth::Asset,
```
MartinquaXD, line 75: The meaning of these is actually opposite to the meaning of `input`/`output` in the context of custom interactions where `inputs` leave the contract and `outputs` enter the contract.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -13,14 +20,72 @@ pub enum Trade {
...
+                    .ok_or(Error::Overflow)?
+                    .checked_ceil_div(&output_price)
+                    .ok_or(Error::Overflow)?,
+                order::Kind::Limit { .. } => todo!(),
+                order::Kind::Liquidity => todo!(),
+            },
+            token: match self.order.side {
+                order::Side::Buy => self.order.sell.token,
+                order::Side::Sell => self.order.buy.token,
+            },
```
MartinquaXD, line 63: The sell token is always the token that flow into the contract and the buy token always flows out. The order `side` is only needed to know which amount needs to be computed with the price.
A `buy` order has a fixed buy amount so we have to use the clearing price to compute the sell amount. And a `sell` order has a fixed sell amount and we need to compute the buy amount from it.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -185,7 +179,45 @@ impl Solution {
...
+        {
+            *flow.entry(output.token).or_default() -= util::conv::u256::to_big_int(output.amount);
+        }
+
+        // For trades, the input and output amounts are determined by the type of trade.
+        for trade in self.trades.iter().filter_map(|trade| match trade {
+            Trade::Fulfillment(trade) => Some(trade),
+            // JIT trades are "fake" trades inserted by the solver to improve the solution score,
+            // and hence should not be considered for asset flow verification.
+            Trade::Jit(..) => None,
```
MartinquaXD, line 207: I'm actually no longer sure `JIT` orders need a special treatment with this kind of asset flow test. cc @nlordell?

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -13,14 +20,72 @@ pub enum Trade {
...
+    /// executed amount must equal the amount from the order.
     pub executed: competition::order::TargetAmount,
 }
 
+impl Fulfillment {
+    /// Calculate the settlement contract input and output amounts executed by
+    /// this trade.
+    pub fn execution(&self, clearing_prices: &ClearingPrices) -> Result<Execution, Error> {
+        let input = self.executed.to_asset(&self.order);
+        let (input_price, output_price) = {
```
MartinquaXD, line 33: I think in this context it makes more sense to think in terms of `sell_amount` and `buy_amount`. Where `sell_amount` always flows into the contract and `buy_amount` always flows out of the contract. Using these names makes the computation easier to follow IMO.

MartinquaXD:
Btw the test could even one step further. Instead of simply summing up positive and negative flow and compare it at the end you could add and subtract the amounts in the correct order and also assert that at no time an over- or underflow would happen.
Imagine a solution contains only these interactions:
T1: 100a -> 100c (trade sells 100a for 100c)
I1: 100a -> 100b (interaction A exchanges 100a for 100b)
I2: 100b -> 100c
Then executing `I1, I2` would result in an executable sequence whereas `I2, I1` would not work because we don't have 100b in the contract when `I2` is supposed to be executed (underflow while adjusting current balances).
For this enhanced test the settlement balances would first be initializes with the sum of all `sell_amounts` then it would go through all interactions in order and subtract `inputs` and add `outputs` accordingly and finally subtract all the `buy_amounts`. If during any of these subtractions or additions an under- or overflow happens this solution can not be executed.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -13,14 +20,72 @@ pub enum Trade {
...
+                .to_owned();
+            let buy_price = clearing_prices
+                .0
+                .get(&self.order.buy.token)
+                .ok_or(Error::ClearingPriceMissing)?
+                .to_owned();
+            match self.order.side {
+                order::Side::Buy => (buy_price, sell_price),
+                order::Side::Sell => (sell_price, buy_price),
+            }
```
nlordell, line 47: I think this match isn't needed. `sell` tokens always enter the contract, `buy` tokens always leave the contract (regardless of the `order.side`). The difference between the `order.side` is only in which of the token amounts are fixed.

crates/driver/src/domain/competition/solution/mod.rs
nlordell: I also think that we defined `input` as inputs to the interaction (so not inputs into the settlement contract contract).

crates/driver/src/domain/competition/solution/mod.rs
nlordell: I think they definitely need to be considered. They are an important part of token conservation.

Take, for example, an order that gets settled directly with a JIT order. It would have a +ve flux of sell token going into the settlement contract, and a -ve flux of buy token out of the settlement contract. This would lead to breaking token conservation (which shouldn't be the case, and wouldn't be if you also consider the JIT order's amounts).

MartinquaXD: Yes, I know. That's why I wrote "I'm actually no longer sure...".

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -25,61 +25,119 @@ pub struct Fulfillment {
...
+                executed: trade.executed,
             },
-            token: match self.order.side {
-                order::Side::Buy => self.order.sell.token,
-                order::Side::Sell => self.order.buy.token,
+            Trade::Jit(trade) => ExecutionParams {
+                side: trade.order.side,
+                // For the purposes of calculating the executed amounts, a JIT order behaves the
+                // same as a regular market order.
+                kind: order::Kind::Market,
```
MartinquaXD, line 58: Jit orders are supposed to be liquidity orders. These orders don't adhere to the uniform clearing prices and always settle at their limit price.

For fill-or-kill JIT orders you can simply take the `sell` and `buy` amount as is. And for partially fillables you have to compute the filled percentage with the `executed_amount`, `side`, `sell`, `buy` and scale the `sell` and `buy` values accordingly.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -25,61 +25,119 @@ pub struct Fulfillment {
...
+            order::Side::Buy => Execution {
+                buy: eth::Asset {
+                    amount: executed.into(),
+                    token: buy.token,
+                },
+                sell: eth::Asset {
+                    amount: match kind {
+                        // TODO Is this really something that should return an error?
+                        // Just use BigInt for the calculation instead and the FINAL result needs
+                        // to be able to fit into a U256
```
MartinquaXD, line 88: Erroring on intermittent overflows mimics the behavior of the smart contract. **Technically** we don't have to do that here but I think it's still useful as the solver would propose a solution that would not work in the settlement contract. We would still detect overflows of the actual settlement during the simulation but I don't see why we shouldn't do it here as well.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -18,7 +18,13 @@ pub struct Order {
...
     pub receiver: Option<eth::Address>,
     pub valid_to: util::Timestamp,
+    /// The amount that this order ideally wants to sell. When the order is
+    /// fulfilled, the actual sold amount isn't always exactly equal to this
+    /// amount, e.g. because the order was partial or due to slippage.
     pub sell: eth::Asset,
+    /// The amount that this order ideally wants to buy. When the order is
+    /// fulfilled, the actual bought amount isn't always exactly equal to this
+    /// amount, e.g. because the order was partial or due to slippage.
     pub buy: eth::Asset,
```
MartinquaXD, line 28: "Ideally wants to sell/buy" seems like an odd choice of words. Ideally you would for example get way more than the specified `buy` amount in case this order is a sell order. Maybe it makes sense to say:
```rust
/// The maximum amount this order is allowed to sell.
/// The final executed amount depends on partial fills and the order side.
pub sell: eth::Asset,
/// The minimum amount this order must buy.
/// The final executed amount depends on partial fills and the order side.
pub sell: eth::Asset,
```
It's tricky to come up with a concise but accurate description because a few different factors are at play here.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -18,13 +18,11 @@ pub struct Order {
     /// The user specified a custom address to receive the output of this order.
     pub receiver: Option<eth::Address>,
     pub valid_to: util::Timestamp,
-    /// The amount that this order ideally wants to sell. When the order is
-    /// fulfilled, the actual sold amount isn't always exactly equal to this
-    /// amount, e.g. because the order was partial or due to slippage.
+    /// The maximum amount this order is allowed to sell when completely filled.
+    /// The actual executed amount depends on partial fills and the order sid
```
MartinquaXD, line 22:
```suggestion
    /// The actual executed amount depends on partial fills and the order side.
```

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -217,13 +215,11 @@ pub enum BuyTokenBalance {
 /// [`Order`].
 #[derive(Debug)]
 pub struct Jit {
-    /// The amount that this order ideally wants to sell. When the order is
-    /// fulfilled, the actual sold amount isn't always exactly equal to this
-    /// amount, e.g. because the order was partial or due to slippage.
+    /// The maximum amount this order is allowed to sell when completely filled.
+    /// The actual executed amount depends on partial fills and the order sid
```
MartinquaXD, line 219:
```suggestion
    /// The actual executed amount depends on partial fills and the order side.
```

MartinquaXD (approved):
The logic looks good to me. 👍

vkgnosis (approved):
I'm not confident on all the details of the asset flow verification but code looks good.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -232,7 +215,11 @@ pub enum BuyTokenBalance {
 /// [`Order`].
 #[derive(Debug)]
 pub struct Jit {
+    /// The maximum amount this order is allowed to sell when completely filled.
+    /// The actual executed amount depends on partial fills and the order side.
     pub sell: eth::Asset,
+    /// The minimum amount this order must buy when completely filled.
+    /// The actual executed amount depends on partial fills and the order side.
     pub buy: eth::Asset,
```
nlordell, line 223: JIT orders are always executed at exactly their limit price. So you can actually consider JIT orders amounts as flux.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -211,6 +240,26 @@ impl Solution {
...
+/// ```
+/// amount_x * price_x = amount_y * price_y
+/// ```
+#[derive(Debug)]
+pub struct ClearingPrices(pub HashMap<eth::TokenAddress, eth::U256>);
+
+impl FromIterator<(eth::TokenAddress, eth::U256)> for ClearingPrices {
+    fn from_iter<T>(iter: T) -> Self
+    where
+        T: IntoIterator<Item = (eth::TokenAddress, eth::U256)>,
```
nlordell, line 257: Generic code... 😮

crates/driver/src/domain/competition/order/mod.rs
MartinquaXD: Oh, true. I overlooked that these comments could be slightly different for JIT orders as they are always liquidity orders.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -24,3 +31,119 @@ pub struct Jit {
...
+                executed: trade.executed,
+            },
+        };
+
+        // For operations which require division, the rounding always happens in favor
+        // of the user.
+        // Errors are returned on 256-bit overflow in certain cases, even though
+        // technically they could be avoided by doing BigInt conversions. The
+        // reason for this behavior is to mimic the onchain settlement contract,
+        // which reverts on overflow.
```
nlordell, line 71: 👍

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -24,3 +31,119 @@ pub struct Jit {
...
+/// The amounts executed by a trade.
+#[derive(Debug, Clone, Copy)]
+pub struct Execution {
+    /// The total amount being sold.
+    pub sell: eth::Asset,
+    /// The total amount being bought.
+    pub buy: eth::Asset,
+}
+
+/// The amounts executed by a trade.
```
nlordell, line 133: This doc-comment seems off to me.

nlordell (approved):
Looks great!

PR #1277, opened 2023-02-27
Balancer response status code

MartinquaXD:
Balancer sometime sends empty responses leading to these logs:
```
2023-02-27T07:44:20.562Z  WARN auction{id=5699489 run=7891}: solver::solver::single_order_solver: Solver BalancerSOR error: EOF while parsing a value at line 1 column 0
```

In the past this usually meant that we got a 429 (rate limiting) error. This PR starts logging the status code and preemptively adds a nicer error message for that case.

### Test Plan
CI

### DIFF
--- a/crates/shared/src/balancer_sor_api.rs
+++ b/crates/shared/src/balancer_sor_api.rs
@@ -46,10 +46,12 @@ impl BalancerSorApi for DefaultBalancerSorApi {
             .post(self.url.clone())
             .json(&query)
             .send()
-            .await?
+            .await?;
+        let status = response.status();
+        let response = response
             .text()
             .await?;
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Do we need to also make the Balancer SOR API "rate limiter" aware like we do for other DEX aggregators? (So we can do things like exponential backoff)?

vkgnosis (approved):

MartinquaXD:
> Do we need to also make the Balancer SOR API "rate limiter" aware like we do for other DEX aggregators? (So we can do things like exponential backoff)?

Yes, I think so. Will have to think about how to integrate that nicely.

PR #1279, opened 2023-02-27
Cip20 Implementation

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1108

Merging this PR would start the transition period as defined in [CIP20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f)

Contains individual PRs:
[SolverRewards: Add score to Settlement](https://github.com/cowprotocol/services/pull/1279/commits/3cc9c8f58f5c84fea305ca60f7e7b6939cc3159e)
[Add score to RatedSettlement](https://github.com/cowprotocol/services/pull/1279/commits/7aecfe71ff1d0f961d177660fe0a07c30e232380)
[[DB] Create Table Settlement Rewards](https://github.com/cowprotocol/services/pull/1279/commits/6baa6b45b06dc0b473098750abb1a9cbc7604f60)
[CIP-20 ranking with score](https://github.com/cowprotocol/services/pull/1279/commits/33782d65381d5ff740f078ed5a58d9dc00a00428)
[SolverRewards 1: Move out ExternalPrices](https://github.com/cowprotocol/services/pull/1279/commits/f8ae9b94fde00ccd9dbeac9e8971d1c6aa2108c1)
[SolverRewards 2: DecodedSettlement implementation](https://github.com/cowprotocol/services/pull/1279/commits/6dfcef3e7dd36f97b866a41bad80adbb0106eaf2)
[SolverRewards 3: Updated db schema and read/write functions](https://github.com/cowprotocol/services/pull/1279/commits/32c6dfbf3790030d9e2e0c44122736ebfd43760e)
[SolverRewards 4: Populate scores and participants](https://github.com/cowprotocol/services/pull/1279/commits/b76196366d2985e1a90c721c545f7a7620bb2326)
[SolverRewards 5: Populate observations and auction prices](https://github.com/cowprotocol/services/pull/1279/commits/4754f78641b5ec8921440870de8420f5a772fda5)
[SolverRewards 7: Fix sql comments and formating](https://github.com/cowprotocol/services/pull/1279/commits/aaf5bb2633b203d6b0d45f37cbe13c5209e6fc4c)
[SolverRewards 6: Populate external prices from driver](https://github.com/cowprotocol/services/pull/1279/commits/9e1dd7be1469e510c8cb52eea25d2d05ead17518)
[SolverRewards 8: Fix executed amount for fee calculation](https://github.com/cowprotocol/services/pull/1279/commits/8ebcafb6a850b5161c5a4ef780978f30a9283396)

### Test Plan

Running in local, observing results.


nlordell:
Just to clarify, this is just requesting to merge the `cip17-solver-rewards-2` into `main` right? So in theory all of the individual changes were already reviewed as part of previous PRs?

Would it be possible to either:
- Clean up Git history so that this can be a direct rebase-merge on top of main?
- If that is too much work (which I suspect it might be), can we add the list of the individual PRs to the commit message for x-referencing?

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,496 @@
...
+        U256,            // feeAmount
+        U256,            // flags
+        U256,            // executedAmount
+        Bytes<Vec<u8>>,  // signature
+    )>,
+    [Vec<(Address, U256, Bytes<Vec<u8>>)>; 3],
+);
+
+#[derive(Debug)]
+#[allow(dead_code)]
```
nlordell, line 37: Might be a left-over from the PR stack.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,496 @@
...
+}
+
+fn fee(
+    trade: &DecodedTrade,
+    external_prices: &ExternalPrices,
+    order: &Order,
+    configuration: &FeeConfiguration,
+) -> Option<U256> {
+    let full_fee_amount = u256_to_big_rational(&order.full_fee_amount);
+    let scaled_fee_amount = full_fee_amount * configuration.fee_objective_scaling_factor.clone();
```
nlordell, line 249: I think we already have a separate field in `Order` that contains the scaled unsubsidized fee amount that should be used for observed value computation. We should use that.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,496 @@
...
...00000000000000005e3734ff2b3127e01070eb225afe910525959ad0000000000000000000000000000000000000000000000000a4f4fa622eb598000000000000000000000000000000000000000000000000000000000000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000044a9059cbb00000000000000000000000005e3734ff2b3127e01070eb225afe910525959ad00000000000000000000000000000000000000000000000000000001cf862866000000000000000000000000000000000000000000000000000000000000000000000000000000001d94bedcb3641ba060091ed090d28bbdccdb7f1d00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000006420cf38cc000000000000000000000000000000000000000000000000405ff0dca143cb520000000000000000000000000000000000000000000001428c970000000000008000000000000000000000002dd35b4da6534230ff53048f7477f17f7f4e7a70000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
nlordell, line 441: nit: It would be nice to break this into multiple lines, `rustfmt` chokes on these long lines.

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -0,0 +1,342 @@
...
+            .with_context(|| format!("get tx {hash:?}"))?
+            .with_context(|| format!("no tx {hash:?}"))?;
+        let tx_from = transaction
+            .from
+            .with_context(|| format!("no from {hash:?}"))?;
+        let tx_nonce: i64 = transaction
+            .nonce
+            .try_into()
+            .map_err(|err| anyhow!("{}", err))
+            .with_context(|| format!("convert nonce {hash:?}"))?;
```
nlordell, line 116: Are these not available in the transaction receipt?

crates/autopilot/src/solvable_orders.rs
```diff
@@ -269,7 +269,16 @@ impl SolvableOrdersCache {
         counter.record(&auction.orders);
 
         if self.store_in_db {
-            let _id = self.database.replace_current_auction(&auction).await?;
+            let auction_id = self.database.replace_current_auction(&auction).await?;
+            self.database
+                .insert_auction_prices(auction_id, &auction.prices)
```
nlordell, line 274: If we do it this way, then we need to clean up the `auction_prices` table for auctions without settlements.

crates/model/src/solver_competition.rs
```diff
@@ -23,6 +23,18 @@ pub struct Request {
     pub transaction: Transaction,
     pub competition: SolverCompetitionDB,
     pub executions: Vec<(OrderUid, Execution)>,
+    pub scores: Scores,
+    pub participants: Vec<H160>, // solver addresses
```
nlordell, line 27: Instead of first adding the `auction_prices` when creating the auction in the autopilot, and then replacing them when we see a settlement, can't we just add the `auction_prices` here? This way we only ever add what is needed and don't need to have additional cleanup code for auction prices (as well as add new logic for cleaning up prices for auctions that didn't settle).

As it is, I think the `auction_prices` would cause a huge explosion of data in the database (as we never clean up `auction_prices` for auctions without settlements, which is most of them).

crates/shared/src/http_solver/model.rs
```diff
@@ -182,6 +182,20 @@ impl Interaction for InteractionData {
...
+pub enum Score {
+    /// The score used for ranking.
+    #[serde(with = "u256_decimal")]
+    Score(U256),
+    /// This option is used to indicate that the solver did not provide a score.
+    /// Instead, the score should be computed by the protocol.
+    /// To have more flexibility, the protocol score can be tweaked by the
+    /// solver by providing a multiplication factor. Expected value: [0,
+    /// inf], 0 being 100% discount, 1 being the same as protocol score
+    MulFactor(f64),
```
nlordell, line 196: nit: I would have called this `scoreFactor` - `mulFactor` doesn't have much information.

crates/solver/src/driver.rs
```diff
@@ -420,13 +445,36 @@ impl Driver {
...
+                .collect();
             tracing::debug!(?transaction, "winning solution transaction");
 
             let solver_competition = model::solver_competition::Request {
                 auction: auction_id,
                 transaction,
                 competition: solver_competition,
                 executions,
+                scores,
+                participants,
```
nlordell, line 477: Here specifically, we can add the relevant `auction_prices` to store. This is compatible with solver+driver colocation as this logic is moved to the `autopilot` anyway.

crates/model/src/solver_competition.rs
sunce86: > can't we just add the auction_prices here?

Yes, but didn't we want to move this responsibility to autopilot? 

I addressed the question with one possible solution here:
https://github.com/cowprotocol/services/pull/1271#issue-1599452805

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,44 @@
+-- Populated before settlement occurs on chain
```
nlordell, line 1: Ubernit: It would be nice to add a description header to this migration script (since it does so much).

database/sql/V048__create_settlement_rewards.sql
```diff
@@ -0,0 +1,44 @@
...
+  price numeric(78,0) NOT NULL,
+
+  PRIMARY KEY (auction_id, token)
+);
+
+CREATE TABLE auction_participants (
+ -- This links to the `auctions` table
+ auction_id bigint NOT NULL,
+ -- Solver who submitted a valid solution to the auction.
+ participant bytea NOT NULL,
```
nlordell, line 41: Could you make the formatting more consistent in the SQL file?

crates/autopilot/src/decoded_settlement.rs
nlordell: cc @MartinquaXD as this is something you worked on if I remember correctly.

crates/model/src/solver_competition.rs
nlordell: > Yes, but didn't we want to move this responsibility to autopilot?

Yes, but it will do it as part of the solver competition loop. Right now I have a very big concern that we never clean up this table and it will grow **very** quickly.

Since we already write a whole bunch of required data in the `solver::driver::run` when uploading the settlement competition, I really think we should move the external prices there as well.

Note that _eventually_ this will move to the `autopilot` here:

https://github.com/cowprotocol/services/blob/6d513aee1326b45365ebd1af7d290f1044423b2d/crates/autopilot/src/run_loop.rs#L71

For now though, I do think we are causing unneeded complication with the current setup.

nlordell:
There are two things I found that I think may be issues:
1. We are using `executed_amount` from the decoded `Trade` from the settlement calldata. This is not correct for [fill-or-kill orders for which this value is ignored](https://github.com/cowprotocol/contracts/blob/7efd5b43fbf46fe40bac9e6de1b6bab05d27c20b/src/contracts/GPv2Settlement.sol#L333-L334) (and should be encoded as 0 for our fill-or-kill orders to save on gas).
2. The `auction_prices` table will grow very quickly if we are storing 100's of prices every 10s per auction run without solutions.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: Yes, this was part of #1199. The field that solvers should use for their objective value computation / fee calculation is `solver_fee`.

crates/autopilot/src/on_settlement_event_updater.rs
bh2smith: is the transaction not the same as transaction receipt?

crates/autopilot/src/solvable_orders.rs
bh2smith: I think the discussion about that was had and conclusion was that it wasn't substantially "large" enough to bother.

database/sql/V048__create_settlement_rewards.sql
bh2smith: Hint: If you are using JetBrains (CMD + Option + L) should unify the SQL format.

crates/autopilot/src/decoded_settlement.rs
nlordell: Ah OK - nevermind then.

crates/model/src/solver_competition.rs
vkgnosis: Alternatively we could add a cleanup task for auctions were no settlement attempt was made. Not sure what's better.

crates/autopilot/src/on_settlement_event_updater.rs
nlordell: No, they are not the same. In this code we are fetching both, and just wanted to double check it is really needed.

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -0,0 +1,342 @@
+//! This module is responsible for updating the database, for each settlement
```
bh2smith, line 1: I really with the file rename were properly rendered as such. Would make it much easier to see the difference here. I would be in favour of not renaming the file until later so see this diff better.

crates/autopilot/src/solvable_orders.rs
nlordell: I disagree, right now, that is around 200 prices every 10s. We've had 845 in the past 24 hours, so assuming the numbers stay roughly the same, that is `24 * 60 * 60 / 10 - 845` auctions with 200 prices, or around 1559000 rows per day.

I think this is too much.

crates/model/src/solver_competition.rs
nlordell: We already write a bunch of required solver rewards data in the `solver` binary run loop (including things needed for this CIP like participants and scores). To me it makes sense to write it there (since we are already writing things) instead of splitting it into writing from two places. Also, in the `solver` run loop, we already have all the required information to write exactly the `auction_price`s we need.

crates/autopilot/src/solvable_orders.rs
bh2smith: That is DEFINITELY too much. Apologies for the confusion.

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: They should exist but I wouldn't rely on it since the fields are `Option<T>`

crates/autopilot/src/decoded_settlement.rs
sunce86: Ah yes, fixed here: https://github.com/cowprotocol/services/pull/1282

crates/autopilot/src/solvable_orders.rs
sunce86: Fixed here: https://github.com/cowprotocol/services/pull/1282

crates/model/src/solver_competition.rs
sunce86: fixed here https://github.com/cowprotocol/services/pull/1282

crates/shared/src/http_solver/model.rs
sunce86: Fixed here https://github.com/cowprotocol/services/pull/1282

crates/solver/src/driver.rs
sunce86: Fixed here https://github.com/cowprotocol/services/pull/1282

database/sql/V048__create_settlement_rewards.sql
sunce86: https://github.com/cowprotocol/services/pull/1283

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -0,0 +1,496 @@
...
+    let fee = match order.kind {
+        model::order::OrderKind::Buy => {
+            scaled_fee_amount * u256_to_big_rational(&trade.executed_amount)
+                / u256_to_big_rational(&trade.buy_amount)
+        }
+        model::order::OrderKind::Sell => {
+            scaled_fee_amount * u256_to_big_rational(&trade.executed_amount)
+                / u256_to_big_rational(&trade.sell_amount)
+        }
+    };
```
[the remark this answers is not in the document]
sunce86: Fixed. Needs additional work later https://github.com/cowprotocol/services/issues/1286

sunce86:
> Just to clarify, this is just requesting to merge the cip17-solver-rewards-2 into main right? So in theory all of the individual changes were already reviewed as part of previous PRs?

Yes.

> Would it be possible to either:
> Clean up Git history so that this can be a direct rebase-merge on top of main?
> If that is too much work (which I suspect it might be), can we add the list of the individual PRs to the commit message for x-referencing?

Let's do the second thing :) this time since I have a new git setup and don't want to mess something.

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: In this case I believe it's fine to look at whole file, because I changed everything except `run_forever` function.

crates/autopilot/src/decoded_settlement.rs
sunce86: Fixed.

bh2smith (approved):

crates/solver/src/driver.rs
```diff
@@ -377,6 +390,19 @@ impl Driver {
...
+        // CIP20 TODO - add to if statement below, once the transition period is over.
+        let mut scores = rated_settlements
+            .iter()
+            .map(|(solver, rated_settlement, _)| {
+                (solver.account().address(), rated_settlement.score.score())
+            })
+            .sorted_unstable_by(|(_, left_score), (_, right_score)| {
+                Ord::cmp(&right_score, &left_score) // descending
+            })
+            .peekable();
```
MartinquaXD, line 405: nit: This iterator does not have to be peekable. You could do this to get both the solver and the score from it:
```rust
let (solver, score) = scores.next().expect("has a winner");
```

crates/solver/src/driver.rs
```diff
@@ -420,13 +446,54 @@ impl Driver {
...
+                    let number_blocks = deadline.as_secs() / self.block_time.as_secs();
+                    block_during_simulation + number_blocks
+                },
+            };
+            let participants = solver_competition
+                .solutions
+                .iter()
+                .map(|solution| solution.solver_address)
+                .collect::<HashSet<_>>(); // to avoid duplicates
+                                          // external prices for all tokens contained in the trades of a settlement
```
MartinquaXD, line 468: `participants` is now a `HashMap` so we no longer have to be descriptive here.
```suggestion
                .collect();
            // external prices for all tokens contained in the trades of a settlement
```

crates/solver/src/settlement.rs
```diff
@@ -456,6 +460,13 @@ impl Settlement {
         Ok(Self {
             encoder: merged,
             submitter: self.submitter,
+            score: match (self.score, other.score) {
+                (Some(left), Some(right)) => match (left, right) {
+                    (Score::Score(left), Score::Score(right)) => Some(Score::Score(left + right)),
+                    _ => None,
+                },
```
MartinquaXD, line 467: Is equivalent and a little easier to read IMO:
```suggestion
                (Some(Score::Score(left)), Some(Score::Score(right))) => {
                    Some(Score::Score(left + right))
                }
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -196,17 +192,65 @@ impl SettlementRanker {
...
-                .settlement_simulation(solver.name(), SolverSimulationOutcome::Success);
-        }
 
+        // Before sorting, make sure to shuffle the settlements. This is to make sure we
+        // don't give preference to any specific solver when there is an
+        // objective value tie.
+        rated_settlements.shuffle(&mut rand::thread_rng());
+
+        if self.enable_auction_rewards {
+            rated_settlements.sort_by(|a, b| a.1.score.score().cmp(&b.1.score.score()));
```
MartinquaXD, line 202:
```suggestion
            rated_settlements.sort_by_key(|s| s.1.score.score());
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -196,17 +192,65 @@ impl SettlementRanker {
...
+            );
+        } else {
+            // TODO: remove this block of code once `auction-rewards` is implemented
+            rated_settlements.sort_by(|a, b| compare_solutions(&a.1, &b.1, self.decimal_cutoff));
+
+            let auction_based_ranking = auction_based_ranking(
+                rated_settlements
+                    .iter()
+                    .map(|(_, settlement, _)| settlement)
+                    .collect::<Vec<_>>(),
```
MartinquaXD, line 220:
```suggestion
                    .collect(),
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -196,17 +192,65 @@ impl SettlementRanker {
...
         Ok((rated_settlements, errors))
     }
 }
 
+// TODO: remove this once `auction-rewards` is implemented
+// Sort settlements by `auction-rewards` rules and return hashmap of settlement
+// id to ranking
+fn auction_based_ranking(settlements: Vec<&RatedSettlement>) -> HashMap<usize, usize> {
+    let mut settlements = settlements;
+    settlements.sort_by(|a, b| a.score.score().cmp(&b.score.score()));
```
MartinquaXD, line 244:
```suggestion
    settlements.sort_by_key(|s| s.score.score());
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -196,17 +192,65 @@ impl SettlementRanker {
...
+        }
         Ok((rated_settlements, errors))
     }
 }
 
+// TODO: remove this once `auction-rewards` is implemented
+// Sort settlements by `auction-rewards` rules and return hashmap of settlement
+// id to ranking
+fn auction_based_ranking(settlements: Vec<&RatedSettlement>) -> HashMap<usize, usize> {
+    let mut settlements = settlements;
```
MartinquaXD, line 243:
```suggestion
fn auction_based_ranking(mut settlements: Vec<&RatedSettlement>) -> HashMap<usize, usize> {
```

MartinquaXD (approved):
Looks really good overall! 👍 
Just a few suggestions which hopefully make some things a bit more readable. Feel free to ignore anything that you don't like or agree with.

vkgnosis (approved):

nlordell (commented):
🎉

nlordell (approved):

crates/solver/src/driver.rs
sunce86: Added.

crates/solver/src/driver.rs
```diff
@@ -377,6 +390,19 @@ impl Driver {
...
+        // sort them.
+        // CIP20 TODO - add to if statement below, once the transition period is over.
+        let mut scores = rated_settlements
+            .iter()
+            .map(|(solver, rated_settlement, _)| {
+                (solver.account().address(), rated_settlement.score.score())
+            })
+            .sorted_unstable_by(|(_, left_score), (_, right_score)| {
+                Ord::cmp(&right_score, &left_score) // descending
+            })
```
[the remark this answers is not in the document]
sunce86: Added.

crates/solver/src/driver.rs
sunce86: Removed.

crates/solver/src/driver.rs
```diff
@@ -420,13 +446,54 @@ impl Driver {
...
+                    [
+                        (
+                            sell_token,
+                            auction_prices.get(&sell_token).cloned().unwrap_or_default(),
+                        ),
+                        (
+                            buy_token,
+                            auction_prices.get(&buy_token).cloned().unwrap_or_default(),
+                        ),
+                    ]
```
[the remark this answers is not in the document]
sunce86: Nice.

crates/solver/src/settlement.rs
sunce86: Even nicer.

crates/solver/src/settlement_ranker.rs
sunce86: Added.

crates/solver/src/settlement_ranker.rs
sunce86: Removed.

crates/solver/src/settlement_ranker.rs
sunce86: Added.

crates/solver/src/settlement_ranker.rs
sunce86: Will keep this one since it's more clear that we are doing thing on a copy. But this function will be removed anyway.

PR #1282, opened 2023-02-27
SolverRewards 6: Populate external prices from driver

sunce86:
Fixes https://github.com/cowprotocol/services/pull/1279#discussion_r1118812860

Decision was made to simplify the insertion of the auction external prices. Now we do it from the driver instead of autopilot. 
Additionally, changes score multiplication factor to be an absolute value - score discount.

### DIFF
--- a/crates/autopilot/src/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/on_settlement_event_updater.rs
@@ -168,7 +168,7 @@ impl OnSettlementEventUpdater {
             fee,
         };
 
-        tracing::debug!("updating settlement details for tx {hash:?}: {update:?}");
+        tracing::debug!(?hash, ?update, "updating settlement details for tx");
 
         self.db
             .update_settlement_details(update.clone())
--- a/crates/autopilot/src/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/on_settlement_event_updater.rs
@@ -26,12 +26,9 @@
 // a) the mined transaction call data
 // b) the auction external prices fetched from orderbook
 // c) the orders fetched from orderbook
-// Auction external prices for all tokens for all solvable orders are stored in
-// the database (auction_prices table) in autopilot before competition.
 // After a transaction is mined we calculate the surplus and fees for each
 // transaction and insert them into the database (settlement_observations
-// table). Now we know which tokens were used in the transaction so we
-// update the auction_prices table with the actual prices that were used.
+// table).
 
 use {
     crate::{
@@ -48,7 +45,7 @@ use {
         event_handling::MAX_REORG_BLOCK_COUNT,
         external_prices::ExternalPrices,
     },
-    std::{collections::HashSet, time::Duration},
+    std::time::Duration,
     web3::types::TransactionId,
 };
 
@@ -159,52 +156,6 @@ impl OnSettlementEventUpdater {
         let surplus = settlement.total_surplus(&external_prices);
         let fee = settlement.total_fees(&external_prices, &orders, &configuration);
 
-        // reduce external prices in `auction_prices` table to only include used tokens
-        // this is done to reduce the amount of data we store in the database
-        let order_tokens = orders
-            .into_iter()
-            .flat_map(|order| vec![order.buy_token, order.sell_token])
-            .collect::<HashSet<_>>();
-        let prices = settlement
-            .trades
-            .iter()
-            .flat_map(|trade| {
-                let buy_token = settlement
-                    .tokens
-                    .get(trade.buy_token_index.as_u64() as usize)?;
-                let sell_token = settlement
-                    .tokens
-                    .get(trade.sell_token_index.as_u64() as usize)?;
-                let buy_token_price = auction_external_prices.get(buy_token);
-                let sell_token_price = auction_external_prices.get(sell_token);
-                match (buy_token_price, sell_token_price) {
-                    (Some(buy_token_price), Some(sell_token_price)) => Some(vec![
-                        (*buy_token, *buy_token_price),
-                        (*sell_token, *sell_token_price),
-                    ]),
-                    _ => {
-                        // for db orders there should be external price for tokens
-                        if order_tokens.contains(buy_token) {
-                            tracing::error!(
-                                "missing buy_token {} auction {}",
-                                auction_id,
-                                sell_token
-                            );
-                        }
-                        if order_tokens.contains(sell_token) {
-                            tracing::error!(
-                                "missing sell_token {} auction {}",
-                                auction_id,
-                                sell_token
-                            );
-                        }
-                        None
-                    }
-                }
-            })
-            .flatten()
-            .collect::<std::collections::BTreeMap<_, _>>();
-
         let update = SettlementUpdate {
             block_number: event.block_number,
             log_index: event.log_index,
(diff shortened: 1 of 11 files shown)


crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -215,9 +166,10 @@ impl OnSettlementEventUpdater {
             effective_gas_price,
             surplus,
             fee,
-            prices,
         };
 
+        tracing::debug!("updating settlement details for tx {hash:?}: {update:?}");
```
MartinquaXD, line 171: You usually get nicer (and easier grepable) log messages by following this pattern:
```suggestion
        tracing::debug!(?hash, ?update, "updating settlement details for tx");
```

crates/model/src/solver_competition.rs
```diff
@@ -24,7 +24,8 @@ pub struct Request {
     pub competition: SolverCompetitionDB,
     pub executions: Vec<(OrderUid, Execution)>,
     pub scores: Scores,
-    pub participants: Vec<H160>, // solver addresses
+    pub participants: Vec<H160>,      // solver addresses
```
MartinquaXD, line 27: This trivially ensures that we'll always have unique participants in the DB.
```suggestion
    pub participants: HashSet<H160>,      // solver addresses
```

crates/solver/src/driver.rs
```diff
@@ -466,6 +467,20 @@ impl Driver {
...
+            // external prices for all tokens contained in the trades of a settlement
+            let prices = winning_settlement
+                .settlement
+                .trades()
+                .filter_map(|trade| {
+                    let sell_token = trade.order.data.sell_token;
+                    let buy_token = trade.order.data.buy_token;
+                    Some(vec![
+                        (sell_token, auction_prices.get(&sell_token).cloned()?),
+                        (buy_token, auction_prices.get(&buy_token).cloned()?),
```
MartinquaXD, line 479: I think these tokens need to be fallible independently of each other. Meaning that only because the `sell_token_price` can not be found we should **not** automatically discard the `buy_token_price` as well. (and the other way around)

crates/shared/src/http_solver/model.rs
```diff
@@ -193,7 +193,7 @@ pub enum Score {
     /// To have more flexibility, the protocol score can be tweaked by the
     /// solver by providing a multiplication factor. Expected value: [0,
     /// inf], 0 being 100% discount, 1 being the same as protocol score
-    MulFactor(f64),
+    ScoreFactor(f64),
```
nlordell, line 196: Sorry for the double nit 🙈 - Here I would actually do:
```suggestion
    #[serde(rename = "scoreFactor")]
    Factor(f64),
```

My rationale is that `Score::Factor` in Rust is less repetitive than `Score::ScoreFactor`, but that in the JSON it should be called `scoreFactor`).

This is very much a "uber-nit", as this only appears in the HTTP solver DTO (i.e. `model.rs`), so feel free to ignore.

crates/solver/src/driver.rs
nlordell: In practice, a missing price only happens for a JIT order, where the token prices don't actually matter.

I actually think the more correct here would be to:
```suggestion
                        (sell_token, auction_prices.get(&sell_token).copied().unwrap_or_default()),
                        (buy_token, auction_prices.get(&buy_token).copied().unwrap_or_default()),
```

This way:
1. We guarantee that there is a price for every token in the auction (and we don't need special cases in other places)
2. Tokens that are included as a result of external solver JIT orders have "no value" WRT the auction, which conceptually makes sense to me.

nlordell (approved):
Changes look good. Just a couple small nits, and a comment on the dealing with missing sell and buy token prices from JIT orders.

crates/shared/src/http_solver/model.rs
fhenneke, line 196: Is there an advantage of using a scoreFactor (as in this PR) over a scoreDiscount as (proposed in the text of the CIP)?
If I understand the code correctly you scale the full objective value and not just the quality. If this is true than a scoreFactor and a scoreDiscount are not equivalent.
An advantage of the latter is that you can be (to some extend) oblivious to transaction costs: If cost estimates by us were accurate and a submitting solver does not know about costs of their solution, then an additive discount can be used to include revert risk into the score. With a multiplicative scaling of the objective you cannot easily do this. You would have to scale the quality only.

Thus I would propose to either use a `scoreDiscount` as in the proposal or to only scale the quality (surplus + fee) instead of the whole objective (surplus + fee - cost).

crates/solver/src/driver.rs
nlordell: > I think these tokens need to be fallible independently of each other.

Since in practice this only affects JIT orders, I _think_ even if we discard both tokens, it won't actually have an impact (I do agree with your point, however, just wanted to point out that there isn't a big practical difference).

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: Fixed.

crates/model/src/solver_competition.rs
sunce86: Fixed.

crates/solver/src/driver.rs
sunce86: My initial reasoning is that these prices will be used exclusively for rewards calculation. If one of the prices for a trade is missing, we cant calculate the rewards and there is no point putting the other one anyway.
But I agree we might use this table for other things in the future. Went with @nlordell suggestion, because it is easier to debug why some value is zero than why is it missing (in this case).

MartinquaXD (approved):

crates/shared/src/http_solver/model.rs
sunce86: I thought it might be easier for solvers to start with the factor, because it is scaled together with the objective_value. But I do agree we should implement as defined in the CIP (if I had any suggestions I should have brought it during the CIP discussion).
Fixed.

sunce86: Fixed. Just with discount name

crates/solver/src/driver.rs
```diff
@@ -463,8 +464,25 @@ impl Driver {
...
-                .into_iter()
+                .collect::<HashSet<_>>(); // to avoid duplicates
+                                          // external prices for all tokens contained in the trades of a settlement
+            let prices = winning_settlement
+                .settlement
+                .trades()
+                .flat_map(|trade| {
+                    let sell_token = trade.order.data.sell_token;
+                    let buy_token = trade.order.data.buy_token;
+                    vec![
```
vkgnosis, line 475:
```suggestion
                    [
```

crates/solver/src/driver.rs
```diff
@@ -463,8 +464,25 @@ impl Driver {
...
+                        (
+                            sell_token,
+                            auction_prices.get(&sell_token).cloned().unwrap_or_default(),
```
vkgnosis, line 478: Is unwrap_or_default correct? I feel like either it is expected that we have prices for all tokens, in which case this should be plain unwrap, or we shouldn't insert 0 prices for tokens we don't have a price for.

sunce86: We discussed it here https://github.com/cowprotocol/services/pull/1282#discussion_r1119313176

vkgnosis (approved):

fhenneke (approved):

PR #1284, opened 2023-02-27
SolverRewards 8: Fix executed amount for fee calculation

sunce86:
Fixes https://github.com/cowprotocol/services/pull/1279#discussion_r1118787688

Core change: 

```
let executed_amount = match order.data.kind {
   OrderKind::Buy => order.metadata.executed_buy_amount,
   OrderKind::Sell => order.metadata.executed_sell_amount_before_fees,
},
```

### DIFF
--- a/crates/autopilot/src/decoded_settlement.rs
+++ b/crates/autopilot/src/decoded_settlement.rs
@@ -8,7 +8,7 @@ use {
     ethcontract::{common::FunctionExt, tokens::Tokenize, Address, Bytes, H160, U256},
     model::order::OrderKind,
     num::BigRational,
-    number_conversions::{big_rational_to_u256, u256_to_big_rational},
+    number_conversions::{big_rational_to_u256, big_uint_to_u256, u256_to_big_rational},
     shared::{conversions::U256Ext, external_prices::ExternalPrices},
     web3::ethabi::{Function, Token},
 };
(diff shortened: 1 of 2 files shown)


crates/autopilot/src/decoded_settlement.rs
```diff
@@ -112,23 +112,30 @@ pub struct Order {
     pub buy_token: H160,
     pub sell_amount: U256,
     pub buy_amount: U256,
+    pub executed_amount: U256,
```
nlordell, line 115: We also have an `executed_fee_amount` part of the order which would avoid the fee computation 👇.

nlordell:
One advantage of using the `Trade` for computing the executed amounts, is that it is independent of the "loading order from database" code.

In theory, you could use the flag "is partially fillable" bit from the `flags` to check whether or not it is the trade's executed amount should be used or the full buy/sell amount.

crates/autopilot/src/decoded_settlement.rs
sunce86: I don't get the same results as in solver competition endpoint when I use `executed_fee_amount`. I will double check.

sunce86: So AFAIU, for objective value computation we use solver fees `Trade::executed_solver_fee()` while the `executed_fee_amount` you mentioned is based on the regular fee `Trade::executed_fee()` which are different so I can't use that.

nlordell (approved):

crates/autopilot/src/decoded_settlement.rs
nlordell: Makes sense.

sunce86:
> In theory, you could use the flag "is partially fillable" bit from the flags to check whether or not it is the trade's executed amount should be used or the full buy/sell amount.

Ah I get it. Ok, I will create additional issue for this and do it afterwards, since I want to merge the solver rewards PR asap.

PR #1287, opened 2023-03-01
Always log 1Inch parsing errors

MartinquaXD:
Right now 1Inch responses only get logged if tracing is enabled. Since this is currently not enabled we don't get to see what caused these [parsing errors](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2023-02-28T21:31:14.884Z',to:'2023-03-01T09:00:00.000Z'))&_a=(columns:!(log),filters:!(),grid:(columns:('@timestamp':(width:164))),index:'6eecf430-220f-11ed-bbdc-2f476b18353d',interval:auto,query:(language:kuery,query:'log:%20%22solver::solver::single_order_solver:%20Solver%201Inch%22%20or%20log:%20%22response%20from%201inch%20api%22'),sort:!(!('@timestamp',desc)))).
We could simply enable trace logs again but that would be quite noisy and AFAIK we currently have to avoid putting to much pressure on the elastic search instance to not crash the logging system.
Instead I decided to always log the response if it can not be parsed.

### Test Plan
only the compiler

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -554,11 +554,16 @@ where
     D: DeserializeOwned,
 {
     tracing::trace!("Query 1inch API for url {}", url);
-    let response = client.get(url).send().await?.text().await;
-    tracing::trace!("Response from 1inch API: {:?}", response);
-    match serde_json::from_str::<RestResponse<D>>(&response?)? {
-        RestResponse::Ok(result) => Ok(result),
-        RestResponse::Err(err) => Err(err.into()),
+    let response = client.get(url).send().await?.text().await?;
+    match serde_json::from_str::<RestResponse<D>>(&response)? {
+        RestResponse::Ok(result) => {
+            tracing::trace!("Response from 1inch API: {:?}", response);
+            Ok(result)
+        }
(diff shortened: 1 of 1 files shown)


nlordell (approved):

nlordell:
> We could simply enable trace logs again but that would be quite noisy and AFAIK we currently have to avoid putting to much pressure on the elastic search instance to not crash the logging system.

You can enable `shared::oneinch_api=trace` to only enable trace logs for the `oneinch_api` module. This should be not noisy at all and allow you to debug without merging this.

I am also fine with this PR though, it makes debugging 1Inch API issues easier for on-call.

vkgnosis (approved):

MartinquaXD:
> You can enable shared::oneinch_api=trace to only enable trace logs for the oneinch_api module. This should be not noisy at all and allow you to debug without merging this.

True. I wanted to do it this way because this will always do the helpful thing without people having to tweak any runtime arguments (or even having to know about it). Also sometimes you can encounter an error which might not reappear after you enabled the trace logs.

PR #1293, opened 2023-03-01
Fix race condition in transaction submission

### LINKED ISSUE #671: Race condition in solution submission leading to missing transaction hash
I noticed this when investigating the one missing settlement transaction for the solver pay out this week.
[logs](https://logs.gnosis.io/goto/4be4fcd0-5531-11ed-b3dc-b32b223fac1c) (prod mainnet, auction{id=2599877 run=51228}, around 2022-10-23T11:16:25.723Z).

I believe what happened here is that in `Submitter::submitter` we see the nonce increase first before we add the potential tx hash to `transactions`. This happens because in `PublicMempoolApi::submit_transaction` the call to `send_raw_transaction` did not return yet but the node actually already broadcasted the transaction.

We should add the transaction hash to the transactions vector even before making the call so that this cannot happen.

vkgnosis:
fixes #671 

See the description of the linked issue. We have  a race condition where a successfully submitted transaction counts as failed. In order to fix this, this PR refactors the `settlement_submission` module to get rid of `TransactionHandle` in favor of passing complete raw transaction bytes around. This allows us to know the tx hash in advance before submitting, which is required to solve the race condition. The handle struct seems to serve no purpose (anymore?) because the handle is never used.

The shadow solver still works because it uses the dry run strategy which does not require an offline account.

Alternatively instead of requiring knowing the tx hash in advance, we could wait for the nonce to increase and inspect the calldata of the new transaction to see if it matches the calldata of the transaction we wanted to settle. That would be a bigger change and is related to how we identify transactions in colocation.

### Test Plan

CI e2e test,

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -29,7 +29,11 @@ use {
     },
     anyhow::{anyhow, ensure, Context, Result},
     contracts::GPv2Settlement,
-    ethcontract::{contract::MethodBuilder, transaction::TransactionBuilder, Account},
+    ethcontract::{
+        contract::MethodBuilder,
+        transaction::{Transaction, TransactionBuilder},
+        Account,
+    },
     futures::FutureExt,
     gas_estimation::{GasPrice1559, GasPriceEstimating},
     primitive_types::{H256, U256},
@@ -99,26 +103,17 @@ pub enum DisabledReason {
     MevExtractable,
 }
 
-#[derive(Debug, Clone, Copy)]
-pub struct TransactionHandle {
-    pub handle: H256,
-    pub tx_hash: H256,
-}
+#[derive(Clone)]
+pub struct RawTransaction(Vec<u8>);
 
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait TransactionSubmitting: Send + Sync {
     /// Submits transaction to the specific network (public mempool, eden,
     /// flashbots...). Returns transaction handle
-    async fn submit_transaction(
-        &self,
-        tx: TransactionBuilder<Web3Transport>,
-    ) -> Result<TransactionHandle>;
+    async fn submit_transaction(&self, tx: RawTransaction) -> Result<()>;
     /// Cancels already submitted transaction using the noop transaction
-    async fn cancel_transaction(
-        &self,
-        tx: TransactionBuilder<Web3Transport>,
-    ) -> Result<TransactionHandle>;
+    async fn cancel_transaction(&self, tx: RawTransaction) -> Result<()>;
     /// Checks if transaction submitting is enabled at the moment
     fn submission_status(&self, settlement: &Settlement, network_id: &str) -> SubmissionLoopStatus;
     /// Returns type of the submitter.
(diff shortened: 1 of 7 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -493,10 +487,14 @@ impl<'a> Submitter<'a> {
...
-            match self.submit_api.submit_transaction(method.tx).await {
+            let (hash, raw) = match method.tx.build().now_or_never().unwrap().unwrap() {
+                Transaction::Request(_) => unreachable!(),
+                Transaction::Raw { bytes, hash } => (hash, RawTransaction(bytes.0)),
+            };
+            transactions.push((hash, gas_price));
+            match self.submit_api.submit_transaction(raw).await {
                 Ok(handle) => {
                     tracing::debug!(?handle, "submitted transaction",);
-                    transactions.push((handle, gas_price));
```
vkgnosis, line 499: Moving this push up fixes the race condition.

vkgnosis:

vkgnosis:
recheck

crates/solver/src/settlement_submission/submitter.rs
MartinquaXD: I think if the tx submission actually fails we will not try to re-submit it with this change. Would it also fix the race condition if this tx gets removed from `transactions` in case we detect an error?

vkgnosis: If we remove in the other-error case, then the following can happen:
- We add hash to transactions.
- Submission errors.
- We remove hash from transactions.
- Submission didn't actually error. We just got a random http error.
- Transactions gets mined but we treat the submission as failed because the hash isn't in transactions.

vkgnosis: The moment we initiate a tx submit, we need to keep the hash in transactions. That's the only reliable way to not miss a transaction. I don't even trust nodes to not accidentally add the new tx to their mempool even if they reply with "already known".

MartinquaXD: > Transactions gets mined but we treat the submission as failed because the hash isn't in transactions.

Hmm, that's true. Maybe we should just don't care that we already tried to submit the tx. Realistically we will only submit a tx every 2s (I think) so we wouldn't overwhelm a node by dropping this optimization but we might make the mining logic more reliable.

vkgnosis: Yeah I think what we should do is always try to submit and if we get "tx already known error" because gas price hasn't increased enough ignore it. I would like to do that in another PR though.

MartinquaXD (approved):

nlordell (approved):
LGTM. @sunce86 do you mind taking a look as well?

> The handle struct seems to serve no purpose (anymore?) because the handle is never used.

This used to be necessary as private networks (such as the old Flashbots protect endpoint) would distinguish between "transaction hash" and some handle that was given to you when submitting the transaction and needed for cancellation.

vkgnosis:
Bumping this for @sunce86 . Not urgent, just reminding.

MartinquaXD:
> @sunce86 - this might be related to the race condition that @MartinquaXD investigated last week.

I think this PR is unrelated to that. If I'm not mistaken the only race condition that was involved with the thing I debugged Friday was that different load balanced nodes might have a different views of the block chain.

nlordell:
Is this still needed? Should we adapt the PR in order to merge it?

MartinquaXD:
I wouldn't touch the submission logic any more unless we still see problems after Felix's latest fix.

nlordell:
OK - Closing to clean up some open PRs.

PR #1294, opened 2023-03-01
SolverRewards 9: Update settlement details if exist

sunce86:
All settlement events are indexed by both staging and production databases.
However, auctions only exist in one of the environments.

This PR inserts settlement observations only if the auction exist.
Thanks @MartinquaXD for figuring out the issue. I did have the same issue when ran CIP20 changes locally, but thought it is because of running the local db. :/

### DIFF
--- a/crates/autopilot/src/database/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/database/on_settlement_event_updater.rs
@@ -5,11 +5,11 @@ use {
     number_conversions::u256_to_big_decimal,
 };
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
 pub struct SettlementUpdate {
     pub block_number: i64,
     pub log_index: i64,
-    pub auction_id: i64,
+    pub auction_id: Option<i64>,
     pub tx_from: H160,
     pub tx_nonce: i64,
     pub gas_used: U256,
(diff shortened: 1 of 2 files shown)


crates/autopilot/src/database/on_settlement_event_updater.rs
```diff
@@ -5,11 +5,11 @@ use {
     number_conversions::u256_to_big_decimal,
 };
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
 pub struct SettlementUpdate {
     pub block_number: i64,
     pub log_index: i64,
-    pub auction_id: i64,
+    pub auction_id: Option<i64>,
```
vkgnosis, line 12: Let's change this so that the required data is grouped together. We always need block_number, log_index, tx_from, tx_nonce.

The rest is needed all or nothing, so it should be something like:
```rust
struct AuctionData {
    pub auction_id: i64,
    pub gas_used: U256,
    pub effective_gas_price: U256,
    pub surplus: U256,
    pub fee: U256,
```

and then in `SettlementUpdate` `pub auction_data: Option<AuctionData>` .

sunce86:

crates/autopilot/src/database/on_settlement_event_updater.rs
sunce86: Done.

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -111,64 +114,75 @@ impl OnSettlementEventUpdater {
...
-            .with_context(|| format!("no auction id for tx {hash:?}"))?;
-        let auction_external_prices =
-            self.db
+        let auction_id = self.db.get_auction_id(tx_from, tx_nonce).await?;
+        let mut update = SettlementUpdate {
+            block_number: event.block_number,
+            log_index: event.log_index,
+            tx_from,
+            tx_nonce,
+            ..Default::default()
```
vkgnosis, line 124: Nit: This is slightly error prone when we add a new field because we can forget to add it here and it would be set through default. Since you're just setting one field to None with this, I would do that explicitly.

vkgnosis (approved):

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: Definitely. Fixed.

MartinquaXD (approved):

Issue #1288, opened 2023-03-01
Feature Gate Partially Fillable Limit Orders


nlordell:
Both the `autopilot` and the `orderbook` API services need to have feature flags for enabling and disabling partially fillable limit orders. It should not be conflated with the existing `--enable-limit-orders` flag, i.e. we should support configurations:
- All limit orders disabled
- Only fill-or-kill limit orders enabled
- All limit orders (so fill-or-kill and partially-fillable) enabled

I'm not sure whether or not we should support the configuration "only partially-fillable limit orders enabled". I would say "yes", since it does not overlap with existing limit order processes (like surplus fee computation), but I don't have a strong opinion either way.

PR #1295, opened 2023-03-01
Add command line arguments to control limit orders

vkgnosis:
Fixes https://github.com/cowprotocol/services/issues/1288 

This PR adds/changes flags for controlling limit orders. We can independently control fill or kill and partially fillable. fill or kill is enabled by default.

### Test Plan

CI

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -148,9 +148,11 @@ pub struct Arguments {
     #[clap(long, env, default_value = "0")]
     pub limit_order_price_factor: f64,
 
-    /// Enable background quoting for limit orders.
-    #[clap(long, env)]
-    pub enable_limit_orders: bool,
+    #[clap(long, env, default_value = "true")]
+    pub process_fill_or_kill_limit_orders: bool,
+
+    #[clap(long, env, default_value = "false")]
+    pub process_partially_fillable_limit_orders: bool,
 
     /// How many quotes the limit order quoter updates in parallel.
     #[clap(long, env, default_value = "5")]
@@ -254,7 +256,16 @@ impl std::fmt::Display for Arguments {
             "limit_order_price_factor: {:?}",
             self.limit_order_price_factor
         )?;
-        writeln!(f, "enable_limit_orders: {:?}", self.enable_limit_orders)?;
+        writeln!(
+            f,
+            "process_fill_or_kill_limit_orders: {:?}",
+            self.process_fill_or_kill_limit_orders
+        )?;
+        writeln!(
+            f,
+            "process_partially_fillable_limit_orders: {:?}",
+            self.process_partially_fillable_limit_orders
+        )?;
         writeln!(
             f,
             "limit_order_quoter_parallelism: {:?}",
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -148,10 +148,10 @@ pub struct Arguments {
     #[clap(long, env, default_value = "0")]
     pub limit_order_price_factor: f64,
 
-    #[clap(long, env, default_value = "true")]
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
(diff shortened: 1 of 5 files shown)


crates/autopilot/src/arguments.rs
```diff
@@ -148,9 +148,11 @@ pub struct Arguments {
     #[clap(long, env, default_value = "0")]
     pub limit_order_price_factor: f64,
 
-    /// Enable background quoting for limit orders.
-    #[clap(long, env)]
-    pub enable_limit_orders: bool,
+    #[clap(long, env, default_value = "true")]
```
MartinquaXD, line 151: Setting `default_value = true` for bool flags means we can only disable it with an environment variable.

crates/orderbook/src/arguments.rs
```diff
@@ -144,10 +144,11 @@ pub struct Arguments {
     #[clap(long, env, default_value = "24")]
     pub solvable_orders_max_update_age_blocks: u64,
 
-    /// Enable limit orders. Once the full limit order flow is implemented, this
-    /// can be removed.
+    #[clap(long, env, default_value = "true")]
```
MartinquaXD, line 147: This should also not get a `default_value`.

crates/autopilot/src/lib.rs
```diff
@@ -637,7 +637,7 @@ pub async fn main(args: arguments::Arguments) {
             .instrument(tracing::info_span!("on_settlement_event_updater")),
     );
 
-    if args.enable_limit_orders {
+    if args.process_fill_or_kill_limit_orders {
```
MartinquaXD, line 640: The `LimitOrderQuoter` currently quotes all limit orders (fill-or-kill and partial fill). This can be fixed very easily and ideally should already happen in this PR, I think.

crates/autopilot/src/arguments.rs
vkgnosis: Does that mean it *has* to become "disable_limit_orders" if I want the default to be to have them enabled?

crates/autopilot/src/lib.rs
vkgnosis: Do partially fillable limit orders need to be quoted? As in, are they never quoted or only quoted if the argument enables them?

I feel changing the quoter isn't part of this task based on the roadmap and issue but I'm not sure. I don't mind doing it here but I'm not sure what behavior is correct.

vkgnosis: That change makes sense to me. I will make another PR for it.

MartinquaXD: But you are right this particular aspect was actually not part of the issue you are tackling right now. I just noticed that because attaching limit order quoting to `process_fill_or_kill_limit_orders` would suggest to me as a reader that partially fillable limit orders don't get quoted which is currently not the case.
Feel free to skip it as it is indeed part of another issue. Just thought it's easy enough to fix and is tangentially related to the introduction of these flags.

crates/autopilot/src/arguments.rs
MartinquaXD: If we want to support disabling and enabling this feature with CLI arguments I would say yes. But it's also not possible to combine CLI arguments and environment variables when you run the binary locally (which is probably the only case where this matters). I personally don't feel very strongly about it and I can see the reason why you would want to make it the default behavior.
Maybe other people also have an opinion on this.

MartinquaXD (approved):
Approving as I don't feel strongly about the points I brought up.

crates/autopilot/src/arguments.rs
vkgnosis: I figured out how to allow default = true while also allowing the argument to be disabled. See most recent commit.

MartinquaXD: Neat! 👍 
This changes the default behavior from passing `--flag` (which only allows to set the flag to true) to `--flag [true|false]` which gives us the total control.

PR #1298, opened 2023-03-01
Fix orders by owner pagination

MartinquaXD:
Elena [reported](https://cowservices.slack.com/archives/C044B50A02F/p1677682870300849) that she got a "Too many limit orders" while the UI showed only 47 orders (when 50 were allowed).
I checked the code that generated the "too many limit orders" and that seems okay. So the other option is that whatever query populates the count of open orders in the UI returns the wrong value.

The previous code seems to assume that `limit` indicates the page size and `offset` indicates how many pages to skip. Since `offset` actually only skips rows and not pages the correct value `limit` in the sub-queries should be `limit + offset`.

However, this currently does not explain what Elena saw yet because even with the incorrect code we should have been able to find all the open orders (`limit = 100, offset = 0`).

Anyway I fixed the computation.

### Test Plan
CI


MartinquaXD:

MartinquaXD:
recheck

vkgnosis (approved):
This seems right but could you double check with git blame and the PR that introduced the wrong pagination to make sure there isn't a reason for it.

nlordell (approved):

MartinquaXD:
> This seems right but could you double check with git blame and the PR that introduced the wrong pagination to make sure there isn't a reason for it.

This specific detail was not discussed during the PR and it appears to be a simple oversight.

PR #1300, opened 2023-03-02
Improve bool argument handling

vkgnosis:
By default a clap bool argument is disabled and enabled when the flag is specified like `--foo`. This makes it impossible to define a flag that is default enabled while also being able to be manually turned off with something like `--foo=false`. In this  PR I change all bool arguments to require passing in a value. This is  more consistent with how other arguments are parsed and more flexibly because it allows default enabled. This changes that that a plain `--foo` without value is not longer accepted, which is fine.

### Test Plan

CI

I have checked that the cluster is fine because it uses environment variables, which always needed a manual `=true`.

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -58,7 +58,7 @@ pub struct Arguments {
     pub db_url: Url,
 
     /// Skip syncing past events (useful for local deployments)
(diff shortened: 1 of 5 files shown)


MartinquaXD (approved):

nlordell:
Very nice!

PR #1301, opened 2023-03-02
SolverRewards: Add temporary logs

sunce86:
Adding temporary logs for fee calculation.

For tx hash 0xa0adf944224c3cda52dfb98e072dd98cb70f3eecb1fc2c8447644c2d70055ecf we get different fees when we compare:
1. what Solver competition endpoint returns
2. what OnSettlementEventUpdater calculated
3. what manual calculation returns using bare db values

I noticed that `full_fee_amount` is different for order when used `https://barn.api.cow.fi/docs/#/default/get_api_v1_orders__UID_` vs manually querying the staging db (super weird).


vkgnosis:
Could you make these trace logs instead and then change the cluster configuration to enable trace logging for this module? That gives us more flexibility so we can change it without code change later.

vkgnosis (approved):

PR #1302, opened 2023-03-02
Asset flow verification liquidity orders

sistemd:
Follow up on https://github.com/cowprotocol/services/pull/1276. Calculate executed amounts for liquidity orders. Test liquidity orders and fix a bug related to not encoding the `partially_fillable` flag when calculating the signature.

Shoutout @nlordell for helping with debugging failing transactions.

### Test Plan
Automated tests.

### DIFF
--- a/crates/driver/src/domain/competition/solution/trade.rs
+++ b/crates/driver/src/domain/competition/solution/trade.rs
@@ -131,7 +131,7 @@ impl Trade {
                         sell: eth::Asset {
                             amount: sell
                                 .amount
-                                .checked_mul(executed.0)
+                                .checked_mul(executed.into())
                                 .ok_or(Error::Overflow)?
                                 .checked_div(buy.amount)
                                 .ok_or(Error::Overflow)?,
@@ -146,7 +146,7 @@ impl Trade {
                         buy: eth::Asset {
                             amount: buy
                                 .amount
-                                .checked_mul(executed.0)
+                                .checked_mul(executed.into())
                                 .ok_or(Error::Overflow)?
                                 .checked_div(sell.amount)
                                 .ok_or(Error::Overflow)?,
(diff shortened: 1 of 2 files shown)


crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -115,7 +119,42 @@ impl Trade {
...
+                match side {
+                    order::Side::Buy => Execution {
+                        buy: eth::Asset {
+                            amount: executed.into(),
+                            token: buy.token,
+                        },
+                        sell: eth::Asset {
+                            amount: executed
+                                .0
+                                .checked_mul(sell.amount)
```
MartinquaXD, line 134: super nit: For some reason I had a super hard time parsing this. Could you consider switching these lines (and the equivalent lines in the other case)?
Then it would be `sell_amount * executed / buy_amount`. Which I would intuitively read like "scaling the `sell_amount` by some factor" where the factor is the relative fill amount of the order.
I might be the only one gets thrown off by this, though.

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
```diff
@@ -0,0 +1,478 @@
...
+                            "trusted": false,
+                        }
+                    ],
+                    "orders": orders,
+                    "deadline": deadline,
+                }),
+            )
+            .await;
+
+        //debug(&web3).await;
```
MartinquaXD, line 459: stray debug log.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -115,7 +119,42 @@ impl Trade {
                     },
                 }
             }
-            order::Kind::Liquidity => todo!(),
+            order::Kind::Liquidity => {
+                // Liquidity orders (including JIT) compute the executed amounts by linearly
+                // scaling the buy/sell amounts in the order.
```
MartinquaXD, line 124: Theoretically the solver could return an `executed_amount` smaller than the total amount for a fill-or-kill order. Not sure if this needs to be caught in the asset flow test or if it's enough to find the error with the failing simulation.
I'm not requesting a change for this PR. Just wanted to bring it up in case we already want to detect that case before simulating.

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
```diff
@@ -0,0 +1,478 @@
...
+        })
+        .await;
+
+        // Call /solve.
+        let mut orders = vec![json!({
+            "uid": boundary.uid(),
+            "sellToken": hex_address(sell_token),
+            "buyToken": hex_address(buy_token),
+            "sellAmount": sell_amount .to_string(),
+            "buyAmount": buy_amount .to_string(),
```
MartinquaXD, line 394:
```suggestion
            "sellAmount": sell_amount.to_string(),
            "buyAmount": buy_amount.to_string(),
```

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
```diff
@@ -0,0 +1,478 @@
...
+    valid: bool,
+    liquidity_orders: fn(sell_token: eth::H160, buy_token: eth::H160) -> Vec<LiquidityOrder>,
+}
+
+#[tokio::test]
+#[ignore]
+async fn valid_liquidity_asset_flow() {
+    crate::boundary::initialize_tracing("debug,hyper=warn");
+
+    let test_cases = [
```
MartinquaXD, line 36: Would be nice to add a small comment explaining the test setup here. Otherwise you have to read the whole test first to make sense of these cases. AFAICS you set up a single user order where all but 50 units of `sell_token` get used by uniswap which returns all but 50 units of the `buy_token`. Meaning that the liquidity orders may use `50 sell_token` and have to produce `50 buy_token` in total. Is that correct?

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
```diff
@@ -0,0 +1,478 @@
...
+                        },
+                        side: competition::order::Side::Sell,
+                        executed: 10.into(),
+                        partially_fillable: true,
+                    },
+                ]
+            },
+        },
+        TestCase {
+            // Single order which attempts to buy too much.
```
MartinquaXD, line 150: I'm not sure why this case should fail. It would sell `5` to get `100` in return which would be a great deal leaving us with `45 sell_token` and `50 buy_token` more than expected.

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
```diff
@@ -0,0 +1,478 @@
...
+                        token: buy_token.into(),
+                    },
+                    side: competition::order::Side::Buy,
+                    executed: 100.into(),
+                    partially_fillable: true,
+                }]
+            },
+        },
+        TestCase {
+            // Two orders which attempt to buy too much.
```
MartinquaXD, line 169: Also surprised that this is supposed to fail for the same reason.

MartinquaXD (commented):
Really nice PR and great test setup! 👍 
Just confused by 2 of the test cases.

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
MartinquaXD: Could it be that the asset flow test requires / assumes that tokens always have the same price? With that assumption it would make sense that this case fails as it would produce `buy_tokens` out of thin air when no equivalent amount of `sell_tokens` was offered in return. However if a solver actually finds market maker that is willing to sell tokens for way below market value this test should probably not fail. Not sure if I'm overlooking something or the underlying test needs special treatment for liquidity orders (and maybe even for limit orders as they are also not guaranteed to trade at the same price).

nlordell: I think the issue here is that you would buy too much. Specifically, only 50 of the buy tokens are "avaiable" for the liquidity order (going in the same direction as the user order), but we are transferring 100 of buy token out for this liquidity order, so I would expect a flux of -50 buy_token (hence not respecting token conservation).

nlordell (commented):
Nice tests! We check token conservation for:
- Market orders
- Liquidity orders (including partial fills)
- Interactions

Wunderbar.

I agree with @MartinquaXD that it would be good to add a general "this is what the test is doing" description. I was aware because of our sync session yesterday, but I probably won't remember the details a month from now.

crates/driver/src/tests/cases/verify_asset_flow/liquidity.rs
MartinquaXD: Ahh, right. I got the "flux" backwards. 🤦‍♂️

MartinquaXD (approved):
Approving as Nick resolved my confusion and my other comments are not blocking.

nlordell (approved):

Issue #1286, opened 2023-02-28
Executed_amount from Trade


sunce86:
https://github.com/cowprotocol/services/pull/1284#issuecomment-1447820866

Get order partially fillable flag from trade, and based on that determine if fee needs adjustment or not. Remove `order` dependency.

nlordell:
This can be done on top of the current CIP implementation IMO. I don't think it is a hard requirement ATM.

PR #1303, opened 2023-03-02
Store executed solver fee

sunce86:
At the moment of execution we need to save the `executed_solver_fee` in `order_execution` table, so we can use it for CIP20 fee calculation in the postprocessing step.

Context: https://cowservices.slack.com/archives/C0361CDD1FZ/p1677771728803909

### DIFF
--- a/crates/autopilot/src/decoded_settlement.rs
+++ b/crates/autopilot/src/decoded_settlement.rs
@@ -102,7 +102,7 @@ impl From<DecodedSettlementTokenized> for DecodedSettlement {
 
 #[derive(Debug)]
 pub struct Order {
-    pub executed_solver_fee: U256,
+    pub executed_solver_fee: Option<U256>,
     pub kind: OrderKind,
     pub sell_token: H160,
     pub buy_token: H160,
@@ -238,7 +238,7 @@ fn surplus(
 }
 
 fn fee(external_prices: &ExternalPrices, order: &Order) -> Option<U256> {
-    let solver_fee = u256_to_big_rational(&order.executed_solver_fee);
+    let solver_fee = u256_to_big_rational(&order.executed_solver_fee?);
     tracing::trace!(?solver_fee, ?order.executed_solver_fee, "executed_solver_fee");
 
     let fee = match order.kind {
@@ -523,7 +523,7 @@ mod tests {
 
         let orders = vec![
             Order {
-                executed_solver_fee: 48263037u128.into(),
+                executed_solver_fee: Some(48263037u128.into()),
                 kind: OrderKind::Sell,
                 buy_amount: 11446254517730382294118u128.into(),
                 sell_amount: 14955083027u128.into(),
@@ -533,7 +533,7 @@ mod tests {
                 signature: hex::decode("155ff208365bbf30585f5b18fc92d766e46121a1963f903bb6f3f77e5d0eaefb27abc4831ce1f837fcb70e11d4e4d97474c677469240849d69e17f7173aead841b").unwrap(),
             },
             Order {
-                executed_solver_fee: 127253135942751092736u128.into(),
+                executed_solver_fee: Some(127253135942751092736u128.into()),
                 kind: OrderKind::Sell,
                 buy_amount: 1236593080.into(),
                 sell_amount: 5701912712048588025933u128.into(),
--- a/crates/autopilot/src/decoded_settlement.rs
+++ b/crates/autopilot/src/decoded_settlement.rs
@@ -106,7 +106,7 @@ pub struct FeeConfiguration {
 
 #[derive(Debug)]
 pub struct Order {
-    pub executed_full_fee_amount: U256,
+    pub executed_solver_fee: U256,
     pub kind: OrderKind,
     pub sell_token: H160,
     pub buy_token: H160,
@@ -121,7 +121,7 @@ impl TryFrom<model::order::Order> for Order {
 
     fn try_from(order: model::order::Order) -> std::result::Result<Self, Self::Error> {
         Ok(Self {
-            executed_full_fee_amount: order.metadata.executed_full_fee_amount,
+            executed_solver_fee: order.metadata.executed_solver_fee,
             kind: order.data.kind,
             sell_token: order.data.sell_token,
             buy_token: order.data.buy_token,
@@ -251,9 +251,9 @@ fn fee(
     order: &Order,
     configuration: &FeeConfiguration,
 ) -> Option<U256> {
-    let full_fee_amount = u256_to_big_rational(&order.executed_full_fee_amount);
-    tracing::trace!(?full_fee_amount, ?order.executed_full_fee_amount, "executed_full_fee_amount");
-    let scaled_fee_amount = full_fee_amount * configuration.fee_objective_scaling_factor.clone();
+    let solver_fee = u256_to_big_rational(&order.executed_solver_fee);
+    tracing::trace!(?solver_fee, ?order.executed_solver_fee, "executed_solver_fee");
+    let scaled_fee_amount = solver_fee * configuration.fee_objective_scaling_factor.clone();
     tracing::trace!(?scaled_fee_amount, ?configuration.fee_objective_scaling_factor, "scaled_fee_amount");
 
     let fee = match order.kind {
@@ -539,7 +539,7 @@ mod tests {
 
         let orders = vec![
             Order {
-                executed_full_fee_amount: 48263037u128.into(),
+                executed_solver_fee: 48263037u128.into(),
                 kind: OrderKind::Sell,
                 buy_amount: 11446254517730382294118u128.into(),
                 sell_amount: 14955083027u128.into(),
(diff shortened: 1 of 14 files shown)


MartinquaXD:
The description mentions that the solver_fee is supposed to be stored (which is probably correct) but the PR currently stores the full_fee_amount.

sunce86:
Yes, this is a core change:

https://github.com/cowprotocol/services/pull/1303/files#diff-a3614105dff26c4a599695c8552d1caf61cee014372000d618ab1ccb4a01bd40R424

But I initially started implementation with `full_fee_amount` which I should rename indeed.

crates/database/src/orders.rs
```diff
@@ -474,7 +475,8 @@ array(Select (p.target, p.value, p.data) from interactions p where p.order_uid =
     where eth_o.uid = o.uid limit 1) as ethflow_data,
 (SELECT onchain_o.sender from onchain_placed_orders onchain_o where onchain_o.uid = o.uid limit 1) as onchain_user,
 (SELECT onchain_o.placement_error from onchain_placed_orders onchain_o where onchain_o.uid = o.uid limit 1) as onchain_placement_error,
-(SELECT surplus_fee FROM order_execution oe WHERE oe.order_uid = o.uid ORDER BY oe.auction_id DESC LIMIT 1) as executed_surplus_fee
+(SELECT surplus_fee FROM order_execution oe WHERE oe.order_uid = o.uid ORDER BY oe.auction_id DESC LIMIT 1) as executed_surplus_fee,
+(SELECT solver_fee FROM order_execution oe WHERE oe.order_uid = o.uid ORDER BY oe.auction_id DESC LIMIT 1) as executed_solver_fee
```
sunce86, line 479: This line won't fail for old entries once the column is added right?

database/sql/V049__add_full_fee_amount.sql
```diff
@@ -0,0 +1,6 @@
+-- Limit orders need to also store the solver_fee at the time of execution.
+-- This is needed because solver_fee is not guaranteed to be constant over time for limit orders.
+
+ALTER TABLE order_execution
+ADD COLUMN solver_fee numeric(78, 0)
+;
```
sunce86, line 6: Can I add this at the end of V48 file? Not sure if it is possible since the migration is already applied in staging for V48.

MartinquaXD: No, adding this to an already applied migration script wouldn't work.

crates/model/src/order.rs
```diff
@@ -614,6 +614,8 @@ pub struct OrderMetadata {
     /// Does not take partial fill into account.
     #[serde(default, with = "u256_decimal")]
     pub solver_fee: U256,
+    #[serde(default, with = "u256_decimal")]
+    pub executed_solver_fee: U256,
```
MartinquaXD, line 618: This is a bit unfortunate. I'm not sure this needs to be added to the `OrderMetadata` as it seems like a value that is only used by us internally. It's currently required because `on_settlement_event_updater.rs` uses the `model::Order` [[0](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/database/orders.rs#L173)] instead of the `database::orders::FullOrder` or some new type that contains exclusively the data needed for the surplus fee computation.
Improving the separation of concerns by not relying on `database::orders::FullOrder` would not leak this implementation detail into the `OrderMetadata`.

crates/solver/src/driver.rs
```diff
@@ -421,11 +421,13 @@ impl Driver {
                         OrderClass::Limit(LimitOrderClass { surplus_fee, .. }) => surplus_fee,
                         _ => None,
                     };
+                    let solver_fee = trade.solver_fee;
```
MartinquaXD, line 424: super nit: Adding this local variable doesn't seem to add anything.

crates/database/src/orders.rs
MartinquaXD: I think this should be fine.

crates/model/src/order.rs
vkgnosis: This is a general problem with us re-using the same type in for the api and internally. You want to use `orders_in_tx` and for that you need to add this field. Doing it differently feels like to big of a change for this PR so I would keep it as is right now, but add a comment to this new field that it is only for internal use.

In the future we definitely should separate the API order type from internal order types. Then we won't have problems like this.

database/sql/V049__add_full_fee_amount.sql
vkgnosis: The moment a migration has been applied to a database, it would be an error to change it. This migration already ran on staging so we can't change it anymore.

crates/model/src/order.rs
MartinquaXD: Makes sense. But then we should at least not expose the field publicly by adding `#[serde(skip_serializing)]` to it.

crates/database/src/orders.rs
vkgnosis: It should be fine because the field is nullable. If it wasn't, one of the tests should fail.

crates/model/src/order.rs
vkgnosis, line 618: This feels like it should be optional. Not all orders have it, and I'd rather have an explicit None than implicit 0.

sunce86: I will leave it as is (but under Option<>) for now. But will probably directly use `FullOrder` when I do refactor described here https://github.com/cowprotocol/services/issues/1286

sunce86: Resolved.

crates/model/src/order.rs
```diff
@@ -614,6 +614,9 @@ pub struct OrderMetadata {
     /// Does not take partial fill into account.
     #[serde(default, with = "u256_decimal")]
     pub solver_fee: U256,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<DecimalU256>")]
+    pub executed_solver_fee: Option<U256>,
```
vkgnosis, line 619: 1. As Martin suggested this doesn't need be serialized because it's not supposed to show up in the api. If we want that, then it should be `#[serde(skip)]` .
2. If we want to have it serialized and use skip-if-none then you need to change it to `#[serde(default, skip_serializing_if = "Option::is_none")]`. There is an added `default` at the start. The current code fails to deserialize if the value is None. Unfortunately that doesn't seem to be caught by any test.

sunce86: I manually tested with `deserialization_and_back` test and it deserializes to `None` if missing (if I understood you well).
Anyway, I added `#[serde(skip)]`.

MartinquaXD (approved):

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -100,13 +100,9 @@ impl From<DecodedSettlementTokenized> for DecodedSettlement {
     }
 }
 
-pub struct FeeConfiguration {
-    pub fee_objective_scaling_factor: BigRational,
-}
-
```
nlordell, line 106: The fact that we don't need to leak fee scaling here is a good sign!

crates/model/src/order.rs
```diff
@@ -614,6 +614,8 @@ pub struct OrderMetadata {
     /// Does not take partial fill into account.
     #[serde(default, with = "u256_decimal")]
     pub solver_fee: U256,
+    #[serde(skip)]
+    pub executed_solver_fee: Option<U256>,
```
nlordell, line 618: This feels like a code-smell that we need to read the `executed_solver_fee` into `OrderMetadata`. I don't think either the FE nor the `solver` need this field, so I don't think we should load it here.

Can we instead change the `autopilot` code to only read the order data it needs for storing the settlement observation?

nlordell (commented):
Code looks good, I don't think we should add a new field to `OrderMetadata` though, it feels like a pretty big codesmell seeing as the `autopilot` can populate this value itself by reading from the DB, and it shouldn't be used anywhere else.

crates/model/src/order.rs
MartinquaXD: We already discussed this a bit [here](https://github.com/cowprotocol/services/pull/1303#discussion_r1124102493).

nlordell: My point is that (AFAICT) the solver observation code doesn't need to load full `Order`s from the database, it can have a custom order type and avoid this.

crates/model/src/order.rs
nlordell: > Doing it differently feels like to big of a change for this PR

I don't think so. We just need to have custom order loading code just for `on_settlement_event_updater` module (it doesn't affect any other existing interface).

crates/model/src/order.rs
MartinquaXD: Totally. I think we all agree on that.

sunce86: I removed the `model::Order` dependency and now `autopilot` reads directly from `FullOrder`.

sunce86:
> I don't think we should add a new field to OrderMetadata though, it feels like a pretty big codesmell seeing as the autopilot can populate this value itself by reading from the DB, and it shouldn't be used anywhere else.

I addressed this. Will merge PR today with 1 approval so it can go into weekly release. @nlordell and @vkgnosis any additional comments will be addressed separately.

Issue #1289, opened 2023-03-01
Allow `partiallyFillable` Flag to Be Set to `true` for Limit Orders


nlordell:
This issue captures the work in the `orderbook` API to accept limit orders with the `partiallyFillable` flag set to `true`. Specifically, this would be orders that look like this:

```ts
{
  "feeAmount": "0",
  "partiallyFillable": true,
  // ...
}
```

When adding support for these orders, we need to be mindful of existing limit order processes that are only intended for one limit order kind and not the other (such as the surplus fee computation loop, `surplus_fee` should only be computed for **fill-or-kill** limit orders).

MartinquaXD:
I think orders with this configuration (`feeAmount: 0, partiallyFillable: true`) could already be created by liquidity owners. Just wanted to point this out as these orders created by liquidity order owners should probably still be classified as liquidity orders.

vkgnosis:
I'm wondering about this. Is the exception we're currently making useful? We don't officially support partially fillable orders so it's unclear how well this is working. Do we *have* liquidity orders that are partially fillable? Do they get settled?

If we want to keep it, then the flag controlling partially fillable orders should be an enum Disabled, AllowedForLiquidity, AllowedForEveryone.

vkgnosis:
I checked with the prod mainnet DB:
```
> select o.creation_timestamp, o.class, o.sell_token, o.buy_token, o.sell_amount, o.buy_amount, o.kind, (SELECT COALESCE(SUM(t.sell_amount), 0) FROM trades t WHERE t.order_uid = o.uid) AS sum_sell, (SELECT COALESCE(SUM(t.buy_amount), 0) FROM trades t WHERE t.order_uid = o.uid) AS sum_buy FROM orders o where partially_fillable ORDER BY creation_timestamp DESC LIMIT 20;
      creation_timestamp       |   class   |                 sell_token                 |                 buy_token                  |       sell_amount       |               buy_amount                | kind |        sum_sell         |        sum_buy
-------------------------------+-----------+--------------------------------------------+--------------------------------------------+-------------------------+-----------------------------------------+------+-------------------------+-----------------------
 2023-02-27 19:27:23.467551+00 | liquidity | \xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 | \xa1d65e8fb6e87b60feccbc582f7f97804b725521 |   200000000000000000000 |                   432312756000000000000 | buy  |                       0 |                     0
[...]

PR #1305, opened 2023-03-03
Handle partially fillable limit orders in order validation

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/1289 .

- Use config option for allowing partially fillable limit orders.
- Slightly refactor code.
- Logic change: Previously, when limit orders were disabled and an order with 0 fee was placed, it was automatically converted into a market order and validated as such. This is changed to continue to treat the order as a limit order which fails validation because limit orders are disabled. I'm not sure what the motivation for the previous way of doing things is.

### Test Plan

modified tests

### DIFF
--- a/crates/orderbook/src/arguments.rs
+++ b/crates/orderbook/src/arguments.rs
@@ -144,9 +144,11 @@ pub struct Arguments {
     #[clap(long, env, default_value = "24")]
     pub solvable_orders_max_update_age_blocks: u64,
 
+    /// Note that fill or kill liquidity limit orders are always allowed.
     #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
     pub allow_placing_fill_or_kill_limit_orders: bool,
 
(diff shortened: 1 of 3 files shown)


crates/orderbook/src/arguments.rs
```diff
@@ -144,9 +144,11 @@ pub struct Arguments {
     #[clap(long, env, default_value = "24")]
     pub solvable_orders_max_update_age_blocks: u64,
 
+    /// Note that fill or kill liquidity limit orders are always allowed.
```
nlordell, line 147: Did you mean this?
```suggestion
    /// Note that fill or kill liquidity orders are always allowed.
```

vkgnosis: The original text is what I meant. But I think what you are getting at, is that the distinction of "liquidity and limit" is pointless because we have three distinct order types: market, limit, liquidity and liquidity don't count as the other two so I don't need to mention that here.

PR #1308, opened 2023-03-04
SolverRewards: Gnosis solvers score

sunce86:
Implements score calculation for gnosis solvers.
Score calculation is called in the post processing step as an additional optimisation step (which it is, kind of).

Note that for Quasimodo we don't define score computation parameters, therefore optimisation is skipped (because Quasimodo is now external solver and should submit the score independently).

Had to decouple `objective_value::Inputs` gas_amount and gas price so it can be reused.
Had to change trait `SettlementSimulating` to return gas_amount instead of bool.


crates/solver/src/settlement_post_processing/optimize_score.rs
```diff
@@ -0,0 +1,49 @@
...
+    if settlement.score.is_some() {
+        return settlement;
+    }
+
+    let gas_amount = match settlement_simulator
+        .settlement_would_succeed(settlement.clone())
+        .await
+    {
+        Ok(gas_amount) => gas_amount * 9 / 10, /* multiply with 0.9 to get more realistic gas */
+        // amount
```
MartinquaXD, line 29: Why would is 90% a more realistic gas amount here? Since we actually simulate the tx instead of using hardcoded gas estimates for each trade or interaction shouldn't this value be 100% correct?

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+        str::FromStr,
+    },
+};
+
+#[derive(Debug, Default, Clone)]
+pub struct ScoreCalculator {
+    a: f64,
+    b: f64,
+    c: f64,
+    x: f64,
```
MartinquaXD, line 31: I think either these need nicer names (e.g. `a -> gas_amount_factor`) or some comment which describes the revert probability function which relates the score computation inputs to these parameters.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+        let exponent =
+            self.x.neg() - self.a * gas_amount - self.b * gas_price - self.c * nmb_orders as f64;
+        let revert_probability = 1. / (1. + exponent.exp());
+        let score = revert_probability * (surplus + fees) - gas_amount * gas_price;
+        Some(U256::from_f64_lossy(score))
+    }
+}
+
+const DEFAULT_SCORE_PARAMETERS: &str =
+    "\
```
MartinquaXD, line 54: Would be great to add a comment linking the code, simulation, whatever came up with these value if possible.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+    Naive,0.5604082285267333,0.00285114179288399,0.06499875450001853,3.3987949311136787;Baseline,\
```
MartinquaXD, line 55: Would be nice to format all lines like: `Name,a,b,c,x;\n` to make it easier to read.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "score_parameters: {:?}", self.score_parameters)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ScoreParameters(HashMap<SolverType, ScoreCalculator>);
+
+// expected string format:
+// "Naive,2.3,3,4.5,7;Baseline,1.2,3.4,5.6,8"
```
MartinquaXD, line 86: This (with more info + nicer wording) should probably be a doc-comment on the `score_parameters` field to generate a useful `--help` text for the CLI.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    pub fn calculate(&self, inputs: objective_value::Inputs, nmb_orders: usize) -> Option<U256> {
+        let surplus = inputs.surplus_given.to_f64()?;
+        let fees = inputs.solver_fees.to_f64()?;
+        let gas_amount = inputs.gas_amount.to_f64()?;
+        let gas_price = inputs.gas_price.to_f64()?;
+
+        let exponent =
+            self.x.neg() - self.a * gas_amount - self.b * gas_price - self.c * nmb_orders as f64;
```
MartinquaXD, line 46: It's a bit surprising that all these values get negated. Is there a reason to do it like this instead of passing the already negated parameters via the CLI?

crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -1,4 +1,12 @@
+use {
+    crate::solver::score_computation::ScoreCalculator,
+    anyhow::{Context, Result},
+    ethcontract::U256,
+    shared::external_prices::ExternalPrices,
+};
+
 pub mod optimize_buffer_usage;
+pub mod optimize_score;
```
MartinquaXD, line 9: Originally didn't realize it but these modules can all be imported privately.

crates/solver/src/solver.rs
```diff
@@ -377,6 +379,12 @@ pub fn create(
                 "configured slippage",
             );
 
+            let score_calculator = score_configuration.get_calculator(solver_type);
+            tracing::debug!(
+                solver = ?solver_type, score = ?score_calculator,
+                "configured score",
+            );
```
MartinquaXD, line 386: This information already gets printed on startup as part of the CLI arguments.

crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -113,13 +129,27 @@ impl PostProcessing for PostProcessingPipeline {
...
             &self.buffer_retriever,
             &self.weth,
             self.unwrap_factor,
         )
-        .await
+        .await;
+
+        match score_calculator {
+            Some(score_calculator) => {
+                optimize_score(
```
MartinquaXD, line 143: I would probably call it `compute_score()`. I agree that it kind of fits into the `PostProcessingPipeline` but just because the other steps are currently optimizations doesn't mean that this one also has to be called an optimization.

crates/solver/src/settlement_post_processing/optimize_score.rs
```diff
@@ -0,0 +1,49 @@
...
+        Err(_) => return settlement,
```
MartinquaXD, line 30: This can only happen if some random error happened during the simulation or if a new block got mined since the last optimization step where the settlement no longer works.
In both cases I would err on the safe side and set the `score` to 0 here to avoid having to execute a reverting solution.

crates/solver/src/settlement_post_processing/optimize_score.rs
sunce86: `eth_estimateGas` returns around 10-15% more gas than actually needed (because it's an _estimation_ function). The only way to calculate the gas exactly is to use Tenderly but IMO eth_estimateGas reduced by 10% already gives pretty good values. (Tenderly calls would take more time and increase our usage by 5-15 calls per auction).

sunce86: But I do agree additional explanation should be added to the comment.

crates/solver/src/solver/score_computation.rs
sunce86: I tried but `fmt` messes everything up, and I did not find any solution other than turning it off for all strings which I didn't want to do.

crates/solver/src/solver/score_computation.rs
sunce86: Hm, it's probably because the original formula that @harisang came up with make more sense using the `-`. TBH I did not get much into the formula itself.

harisang: Fair point, @MartinquaXD . This is usually how logistic regression is formulated, and this is the only reason the parameters were computed like this. Of course what you propose is completely equivalent.

image

MartinquaXD: Since it's usually done like this in math I think it's fine to leave it as is.

crates/solver/src/settlement_post_processing/optimize_score.rs
nlordell: > returns around 10-15% more gas than actually needed (because it's an estimation function).

It is also a product of needing it to be the total gas before refunds (an EVM requirement) so it will always be an upper bound on gas spent and not the actual gas spent.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+        let revert_probability = 1. / (1. + exponent.exp());
```
harisang, line 47: This is actually the `success_probability`.

crates/solver/src/solver/score_computation.rs
harisang: The code for collecting the data and training the model can be found here:
https://github.com/cowprotocol/risk_adjusted_rewards

The data for each solver can be found here.
https://drive.google.com/drive/u/1/folders/19yoL808qkp_os3BpLIYQahI3mQrNyx5T

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,135 @@
...
+
+const DEFAULT_SCORE_PARAMETERS: &str =
+    "\
+    Naive,0.5604082285267333,0.00285114179288399,0.06499875450001853,3.3987949311136787;Baseline,\
+     -0.24391894879979226,-0.05809501139187965,-0.000013222507455295696,4.27946195371547;CowDexAg,\
+     -0.9613998308805674,-0.14435150204689684,0.13923418474574772,2.7565258390467178;OneInch,-0.\
+     32674185936325467,-0.05930446215554123,-0.33031769043234466,3.144609301500272;Paraswap,-0.\
+     7815504846264341,-0.06965336115721313,0.0701725936991023,3.2617622830143453;ZeroEx,-1.\
+     399997494341399,-0.04522233479453635,0.11066085229796373,2.7150950015915676;BalancerSor,-0.\
+     7070951919365344,-0.1841886790519467,0.34189609422313544,3.6849833670945027";
```
harisang, line 61: Just confirmed that all values for all solvers are correct!

crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -21,10 +30,11 @@ use {
 };
 
 /// Determines whether a settlement would be executed successfully.
+/// If the settlement would succeed, the gas estimate is returned.
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait SettlementSimulating: Send + Sync {
-    async fn settlement_would_succeed(&self, settlement: Settlement) -> bool;
+    async fn settlement_would_succeed(&self, settlement: Settlement) -> Result<U256>;
```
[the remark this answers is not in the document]
sunce86: Renamed to `estimate_gas`

crates/solver/src/settlement_post_processing/optimize_score.rs
sunce86: Added comment.

crates/solver/src/solver/score_computation.rs
sunce86: Thanks @harisang added.

crates/solver/src/solver/score_computation.rs
sunce86: `#[rustfmt::skip]` helped. fixed.

crates/solver/src/solver/score_computation.rs
sunce86: Moved the comment to `score_parameters` and improved.

crates/solver/src/settlement_post_processing/mod.rs
sunce86: Fixed

crates/solver/src/solver.rs
sunce86: removed

crates/solver/src/settlement_post_processing/mod.rs
sunce86: renamed

crates/solver/src/settlement_post_processing/optimize_score.rs
sunce86: I kept is as is. The reasoning is that this component is expected to either improve the solution or do nothing. We do have a simulation again in the ranking phase that would cover failure. Setting score to 0 here is one side effect that I would prefer to avoid if possible (it would mean that it is the value provided by solver and harder to debug).

crates/solver/src/solver/score_computation.rs
sunce86: Fixed.

crates/solver/src/solver/score_computation.rs
sunce86: renamed. @harisang what should I call the parameter `x`?

harisang: So, in the logistic regression literature it is called `intercept`. Not sure if that adds helps, but I guess it is more descriptive compared to `x`

sunce86: Renamed

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Each solver parameter is a comma separated list of parameters:
+    /// [solver name],[gas amount factor],[gas price factor],[number of orders
+    /// factor],[intercept parameter]
+    #[clap(long, env, default_value = DEFAULT_SCORE_PARAMETERS)]
+    score_parameters: ScoreParameters,
+}
+
+impl Arguments {
+    pub fn get_calculator(&self, solver: SolverType) -> Option<ScoreCalculator> {
+        self.score_parameters.0.get(&solver).cloned()
```
nlordell, line 93: I think it makes more sense to merge the default parameters with what is supplied via the command line. This makes overriding a single solver's score computation parameters easier.

nlordell (commented):
Looks good to me.

One suggestion around parsing of the score parameters (i.e. we should have default values baked into code and have the CLI set overrides for each solver).

crates/solver/src/solver/score_computation.rs
sunce86: Fixed.

crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -1,5 +1,13 @@
...
+    anyhow::{Context, Result},
+    ethcontract::U256,
+    shared::external_prices::ExternalPrices,
+};
+
+mod optimize_buffer_usage;
+mod optimize_score;
+mod optimize_unwrapping;
 
 use {
```
vkgnosis, line 12: Somehow the imports got split. It should be modules first then all imports.

crates/solver/src/settlement_post_processing/optimize_score.rs
```diff
@@ -0,0 +1,56 @@
...
+        settlement::Settlement,
+        solver::score_computation::ScoreCalculator,
+    },
+    ethcontract::Address,
+    gas_estimation::GasPrice1559,
+    num::{BigRational, FromPrimitive},
+    shared::{external_prices::ExternalPrices, http_solver::model::Score},
+};
+
+pub async fn compute_score(
```
vkgnosis, line 14: From the perspective of this function it feels cleaner to take Settlement by reference and return the score. Up to you.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,169 @@
...
+pub struct ScoreCalculator {
+    gas_amount_factor: f64,
+    gas_price_factor: f64,
+    nmb_orders_factor: f64,
+    intercept: f64,
+}
+
+impl ScoreCalculator {
+    pub fn new(
+        gas_amount_factor: f64,
```
vkgnosis, line 36: To me it feels better to make the fields pub and not have a constructor in this case. Up to you.

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,169 @@
...
+            Ok((
+                SolverType::from_str(solver, true).map_err(|message| anyhow::anyhow!(message))?,
+                ScoreCalculator::new(a, b, c, x),
+            ))
+        })
+        .collect::<Result<HashMap<_, _>>>()
+}
+
+#[cfg(test)]
+mod tests {
```
vkgnosis, line 161: Maybe there should be at least one test to make sure the calculation is right. With an example from the solver team to check against.

vkgnosis (approved):

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,169 @@
...
+            .into_iter()
+            .map(|(solver, default_score_parameter)| {
+                (
+                    solver,
+                    user_score_parameters
+                        .get(&solver)
+                        .cloned()
+                        .unwrap_or(default_score_parameter),
+                )
+            })
```
nlordell, line 123: My suggestion would be to, instead of parsing both, you would have a default:
```
fn default_score_parameters() -> HashMap<SolverType, ScoreCalculator> {
    // ...
}
```

And then have the parse function just override the default ones. It feals weird to encode default values as strings instead of actual Rust data structures.

nlordell (approved):

crates/solver/src/solver/score_computation.rs
nlordell: Just a nit.

crates/solver/src/settlement_post_processing/mod.rs
sunce86: Fixed.

crates/solver/src/settlement_post_processing/optimize_score.rs
sunce86: Yes, I tried to fit the function into the context of the place where it is used, but it does not make it really reusable. Fixed.

crates/solver/src/solver/score_computation.rs
sunce86: Yes, no real need to be private here.

crates/solver/src/solver/score_computation.rs
sunce86: YES! Was planning to test against one. I committed a test and now waiting for approval from @harisang

crates/solver/src/solver/score_computation.rs
```diff
@@ -0,0 +1,180 @@
...
+impl ScoreCalculator {
+    pub fn calculate(&self, inputs: &objective_value::Inputs, nmb_orders: usize) -> Option<U256> {
+        let surplus = inputs.surplus_given.to_f64()?;
+        let fees = inputs.solver_fees.to_f64()?;
+        let gas_amount = inputs.gas_amount.to_f64()?;
+        let gas_price = inputs.gas_price.to_f64()?;
+        let exponent = self.intercept.neg()
+            - self.gas_amount_factor * gas_amount
+            - self.gas_price_factor * gas_price
+            - self.nmb_orders_factor * nmb_orders as f64;
```
fhenneke, line 43: the scaling of `gas_amount` and `gas_price` is missing. otherwise the computation looks good.

(not sure if the following is valid rust)
```suggestion
        let exponent = self.intercept.neg()
            - self.gas_amount_factor * gas_amount * 1e-6
            - self.gas_price_factor * gas_price * 1e-1
            - self.nmb_orders_factor * nmb_orders as f64;
```

sunce86: Thanks, added.

crates/solver/src/solver/score_computation.rs
sunce86: Added test and confirmed the calculation with solver team.

PR #1314, opened 2023-03-06
Make sure fill or kill and partially fillable are distinguished

vkgnosis:
part of https://github.com/cowprotocol/services/issues/1289

When we wrote the fill or kill limit order code, we didn't yet know how partially fillable limit orders would be handled. Now that we are going to have partially fillable limit orders, some functions are confusingly refer to "limit orders" without qualifying that, even when they only handle one type of limit order. This PR changes a couple of identifiers and comments to make the distinction more clear.

I abbreviate "fill or kill" as "fok" because writing it out all the time is annoying to write and harder to read (assuming you know what it stands for).

### Test Plan

CI, existing tests


nlordell (approved):
`_fok_` sounds funny when you read it out loud...

PR #1315, opened 2023-03-06
Calculate executed amounts for fulfilled limit orders in the `driver`

sistemd:
Final PR in the asset flow verification chain, following up after https://github.com/cowprotocol/services/pull/1302.

As part of writing the tests for this I discovered that there's an oversight in the driver WRT limit orders. In particular, with our current code "synthetic" limit orders are calculated by the `solver` binary. The autopilot passes the raw limit orders to the driver and the driver simply forwards them to the solver engine. So the amounts are incorrect when they reach the solver engine.

For that reason, this PR has no tests. In the next PR I will add support for limit orders in the driver and add some tests to cover this functionality as well.

(Even if my analysis above is incorrect the correct approach is probably still to have the driver do the surplus fee adjustments, not the autopilot - I think @MartinquaXD might be the best person to confirm this.)

### Test Plan

Follow-up PRs.

### DIFF
--- a/crates/driver/src/domain/competition/solution/trade.rs
+++ b/crates/driver/src/domain/competition/solution/trade.rs
@@ -131,7 +131,7 @@ impl Trade {
                         sell: eth::Asset {
                             amount: sell
                                 .amount
-                                .checked_mul(executed.0)
+                                .checked_mul(executed.into())
                                 .ok_or(Error::Overflow)?
                                 .checked_div(buy.amount)
                                 .ok_or(Error::Overflow)?,
@@ -146,7 +146,7 @@ impl Trade {
                         buy: eth::Asset {
                             amount: buy
                                 .amount
-                                .checked_mul(executed.0)
+                                .checked_mul(executed.into())
                                 .ok_or(Error::Overflow)?
                                 .checked_div(sell.amount)
                                 .ok_or(Error::Overflow)?,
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -155,7 +155,75 @@ impl Trade {
...
+                        sell: eth::Asset {
+                            amount: executed
+                                .0
+                                .checked_mul(buy_price)
+                                .ok_or(Error::Overflow)?
+                                .checked_div(sell_price)
+                                .ok_or(Error::Overflow)?
+                                // Because of how "synthetic" limit orders are constructed as
+                                // explained above, we need to simply increase the executed sell
+                                // price by the surplus fee. We know that the user placed an order
```
nlordell, line 196: You increase the executed sell amount so you decrease the sell token price (i.e. you need more of it because its worth less).

nlordell (approved):
Looks great! One tiny comment nit, otherwise looks very good.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -155,7 +155,75 @@ impl Trade {
...
+                                // Because of how "synthetic" limit orders are constructed as
+                                // explained above, the solver received the sell amount
+                                // reduced by the surplus fee. That's why we're have to reduce the
+                                // executed amount by the surplus fee when calculating the
+                                // executed buy amount.
+                                .checked_sub(surplus_fee.into())
+                                .ok_or(Error::Overflow)?
+                                .checked_mul(sell_price)
+                                .ok_or(Error::Overflow)?
+                                .checked_ceil_div(&buy_price)
```
MartinquaXD, line 220: This is a slight change to the current driver code as that one uses `checked_div()` here. Using `checked_div_ceil()` is actually nicer as users would get 1 wei more rather then less (right?) but I wanted to point it out nonetheless in case someone missed it and would like to object.

MartinquaXD (approved):
Code looks correct (meaning like in the current driver) and the comments are really nice! 👍

MartinquaXD:
> (Even if my analysis above is incorrect the correct approach is probably still to have the driver do the surplus fee adjustments, not the autopilot - I think @MartinquaXD might be the best person to confirm this.)

I agree. The driver is able to do all the special handling whereas the `autopilot` can really only pre-adjust the `sell_amount`. Then the driver would still have to do the weird price adjustment computation. But if everything happens in the driver it's easier to trace the logic throughout the code.

Btw, right now we know that this price adjustment will be respected/done because we are the ones doing it. Do we open another attack vector by moving this logic to 3rd parties? I assume the solver team would have to verify that in their post submission tests. Otherwise solvers could "buy" themselves a good score by giving away the fees they are supposed to take, right?

nlordell:
This is a good point, but AFAIU, settlement verification will used the protocol-computed surplus fees.

PR #1316, opened 2023-03-06
Deprecate order meta data available balance

vkgnosis:
This field serves no purpose. We are soon going to need a new similar field to communicate the available balance for partially fillable orders. I want to give the new field a different name to avoid confusion.

Potential problems:
Is the frontend using this field? After grepping their repos for this field, I don't think so.

### Test Plan

CI

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -582,9 +582,10 @@ components:
         uid:
           $ref: "#/components/schemas/UID"
         availableBalance:
-          description: "Amount of sellToken available for the settlement contract to spend on behalf of the owner. Null if API was unable to fetch balance or if the order status isn't Open."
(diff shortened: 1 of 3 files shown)


crates/orderbook/openapi.yml
```diff
@@ -578,9 +578,10 @@ components:
         UID:
           $ref: "#/components/schemas/UID"
         availableBalance:
-          description: "Amount of sellToken available for the settlement contract to spend on behalf of the owner. Null if API was unable to fetch balance or if the order status isn't Open."
+          description: Unused that field that is currently always set to null and will be removed in the future.
```
MartinquaXD, line 581: Wording is a bit off.

MartinquaXD (approved):

PR #1318, opened 2023-03-07
Fix token equivalency error

MartinquaXD:
`PLM` noticed the error `tokens not part of solution for equivalency` suddenly appears frequently. This error currently gets triggered through a bug which produces this unhelpful message instead of a reasonable one.

This is what triggers the error:
1. HTTP solver returns solution settling a liquidity order buying native `ETH`
2. Driver converts the solution slightly and only collects UCP from tokens traded in user orders [[0](https://github.com/cowprotocol/services/blob/main/crates/solver/src/solver/http_solver/settlement.rs#L388)]
3. Driver appends settled orders to execution plan [[1](https://github.com/cowprotocol/services/blob/main/crates/solver/src/solver/http_solver/settlement.rs#L369), [2](https://github.com/cowprotocol/services/blob/main/crates/solver/src/solver/http_solver/settlement.rs#L242)]
4. When adding an order buying `ETH` to the execution plan we add a price for `ETH` to the UCP that is equal to the `WETH` price. This fails if we can't find a price for `WETH` because there was no user order buying or selling `WETH` or `ETH` in step 2. [[3](https://github.com/cowprotocol/services/blob/main/crates/solver/src/liquidity/order_converter.rs#L112-#L114)]

Since liquidity orders settle at their limit price anyway we don't need to add a token price equivalency for them. This fix should have 2 effects:
1. Support settlements that use liquidity orders buying `ETH` to settle a user order that doesn't trade `(W)ETH`
2. Previously affected settlements containing only liquidity orders should now return the error `NoUserOrders`

### Test Plan
CI


vkgnosis (approved):

nlordell (approved):

MartinquaXD:
This causes a test in the naive solver to fail sometimes. It's currently unclear to me why. Will convert this PR to a draft until I find out.

Issue #1290, opened 2023-03-01
Include Partially Fillable Limit Orders in Auctions


nlordell:
Given partially fillable limit orders that appear in the orderbook, we should add them to the auction if:
- They have a native token price
- There is sufficient balance for _partially_ filling the order

Note that we should make sure that the existing knobs we have for hedging risk of growing orderbook size (such as filtering limit orders based on limit prices using the reference native token prices) should also apply to partially fillable limit orders.

This will also likely require some internal refactoring, as we don't need surplus fees for all orders of class `"limit"`, but only the fill-or-kill kind.

vkgnosis:
Is there anything that needs to be done with regards to the risk adjusted rewards for orders? In `risk_adjusted_rewards.rs` we hardcode 0 if the order is a liquidity order or partially fillable.

nlordell:
> Is there anything that needs to be done with regards to the risk adjusted rewards for orders?

No. In theory, there should be a reward for these, but we are moving to the CIP-20 style scoring/payouts and explicitly decided not to figure out how to compute risk-adjusted rewards for these order types (since CIP-20 is likely to be in place before partially fillable limit orders).

vkgnosis:
Partially fillable limit orders are included in the auction but only if the balance is large enough to completely fill the order. See #1317 .

nlordell:
Closing in favour of #1378 which is specific to the remaining work.

PR #1319, opened 2023-03-07
Test that auction includes partially fillable limit order

vkgnosis:
part of https://github.com/cowprotocol/services/issues/1290

### Test Plan

new e2e test

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -181,7 +181,7 @@ impl SolvableOrdersCache {
             filter_invalid_signature_orders(orders, self.signature_validator.as_ref()).await;
         counter.checkpoint("invalid_signature", &orders);
 
-        let orders = filter_limit_orders_with_insufficient_sell_amount(orders);
+        let orders = filter_fok_limit_orders_with_insufficient_sell_amount(orders);
         counter.checkpoint("insufficient_sell", &orders);
 
         // If we update due to an explicit notification we can reuse existing balances
@@ -590,12 +590,16 @@ async fn filter_unsupported_tokens(
     Ok(orders)
 }
 
-fn filter_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
-    // Unwrap because solvable orders always have a surplus fee.
-    orders.retain(|order| match &order.metadata.class {
-        OrderClass::Limit(limit) => order.data.sell_amount > limit.surplus_fee.unwrap(),
-        _ => true,
-    });
+fn filter_fok_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
+    // Unwrap because solvable fok orders always have a surplus fee.
+    orders.retain(
+        |order| match (&order.metadata.class, order.data.partially_fillable) {
+            (OrderClass::Limit(limit), false) => {
+                order.data.sell_amount > limit.surplus_fee.unwrap()
+            }
+            _ => true,
+        },
+    );
     orders
 }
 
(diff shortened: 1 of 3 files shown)


crates/e2e/tests/e2e/partially_fillable.rs
```diff
@@ -0,0 +1,125 @@
...
+    tracing::info!("Waiting for order to show up in auction.");
+    let has_order = || async { services.get_auction().await.auction.orders.len() == 1 };
+    wait_for_condition(TIMEOUT, has_order).await.unwrap();
+
+    let auction = services.get_auction().await.auction;
+    let order = auction.orders.into_iter().next().unwrap();
+    assert!(order.data.partially_fillable);
+    assert_eq!(order.metadata.full_fee_amount, 0.into());
+    assert_eq!(order.metadata.solver_fee, 0.into());
+    assert_eq!(auction.rewards.get(&order.metadata.uid), None);
```
vkgnosis, line 108: Note these asserts. Speak up if any of these values should be different.

nlordell: These seem correct to me:
- `partiallyFillable` is set
- There is no "full fee amount" or "solver fee" as this is dependent on how much the order executes and cannot be known upfront (cc @sunce86 - as this may affect CIP-20 computations as we rely on these fields - posted in Slack as well for more visibility).
- There are no CIP-14 risk-adjusted rewards for these orders (we explicitly decided not to figure out what we need to do for this as we plan on being on CIP-20 before partially fillable limit orders are available)

nlordell (approved):

crates/autopilot/src/solvable_orders.rs
```diff
@@ -590,12 +590,16 @@ async fn filter_unsupported_tokens(
...
-    orders.retain(|order| match &order.metadata.class {
-        OrderClass::Limit(limit) => order.data.sell_amount > limit.surplus_fee.unwrap(),
-        _ => true,
-    });
+fn filter_fok_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
+    // Unwrap because solvable fok orders always have a surplus fee.
+    orders.retain(
+        |order| match (&order.metadata.class, order.data.partially_fillable) {
+            (OrderClass::Limit(limit), false) => {
+                order.data.sell_amount > limit.surplus_fee.unwrap()
```
MartinquaXD, line 598:
```suggestion
                order.data.sell_amount >= limit.surplus_fee.unwrap()
```

crates/e2e/tests/e2e/partially_fillable.rs
MartinquaXD: I guess we could also assert that this was actually created as a limit order as opposed to a liquidity order.

MartinquaXD (approved):

crates/autopilot/src/solvable_orders.rs
```diff
@@ -590,12 +590,16 @@ async fn filter_unsupported_tokens(
     Ok(orders)
 }
 
-fn filter_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
-    // Unwrap because solvable orders always have a surplus fee.
-    orders.retain(|order| match &order.metadata.class {
-        OrderClass::Limit(limit) => order.data.sell_amount > limit.surplus_fee.unwrap(),
-        _ => true,
-    });
+fn filter_fok_limit_orders_with_insufficient_sell_amount(mut orders: Vec<Order>) -> Vec<Order> {
```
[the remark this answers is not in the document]
vkgnosis: Do you have a suggestion for a better name? With the current name, all the doc comment can code is repeat the name. It could explain "fok" I guess. I intentionally started using "fok" instead of writing out fill_or_kill everywhere in a previous PR because I felt that makes it easier to read.

crates/autopilot/src/solvable_orders.rs
nlordell: I think it should be `>`, otherwise we have 0-sell amount orders which we don't want.

vkgnosis: This PR doesn't change this condition.

MartinquaXD: > This PR doesn't change this condition btw.

Yes. Just wanted to point out a possible mistake that was maybe overlooked earlier.

> I think it should be >, otherwise we have 0-sell amount orders which we don't want.

I mean this is difference is negligible so I think both options are okay but since solvers could theoretically settle the orders I thought it made sense to include them.
I'm fine with leaving it as is.

PR #1326, opened 2023-03-09
Detect 0x API Rate Limiting Errors

nlordell:
This PR adds rate limiting detection to the 0x API. Specifically, this needs to be done on the HTTP response status code and not on the error object response status code.

We also continue to bubble this up as a rate limiting error so that we can do exponential back-off when hit.

### Test Plan

The Rust compiler. No significant logic change.


nlordell:

PR #1327, opened 2023-03-09
Use More Descriptive Log on 1Inch API Error

nlordell:
This PR just makes a 1Inch log message more accurate.

### Test Plan

No logic changes, so the CI.

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -561,7 +561,7 @@ where
             Ok(result)
         }
         RestResponse::Err(err) => {
-            tracing::warn!("Failed to parse response from 1inch API: {:?}", response);
+            tracing::warn!("Received 1inch API error: {:?}", response);
             Err(err.into())
         }
     }


MartinquaXD:
We should also log the response body when we can't deserialize it.

crates/shared/src/oneinch_api.rs
```diff
@@ -553,15 +553,14 @@ async fn logged_query<D>(client: &Client, url: Url) -> Result<D, OneInchError>
 where
     D: DeserializeOwned,
 {
-    tracing::trace!("Query 1inch API for url {}", url);
+    tracing::trace!(%url, "Query 1inch API");
     let response = client.get(url).send().await?.text().await?;
+    tracing::trace!(%response, "Received 1Inch API response");
+
```
MartinquaXD, line 559: I think it makes slightly more sense to always log it in case of an error instead of having to opt-in into the `trace` logs. Then you can not lose any unexpected responses.
But this is fine as well.

vkgnosis (approved):

crates/shared/src/oneinch_api.rs
nlordell: Personally, I think it is too noisy. We can always enable `trace` logs by default in our configuration if we want this behaviour.

MartinquaXD: Oh, I didn't mean "always log every response" (I agree, too noisy) but rather "always log every response we weren't able to deserialize".

PR #1330, opened 2023-03-09
Handle malformed 1Inch responses

MartinquaXD:
Reports 1Inch error that can not be parsed as a RestError with the status code of the unparsable response. This can inform callers how to handle that error (e.g. rate limiting).

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -554,14 +554,23 @@ where
     D: DeserializeOwned,
 {
     tracing::trace!(%url, "Query 1inch API");
-    let response = client.get(url).send().await?.text().await?;
-    tracing::trace!(%response, "Received 1Inch API response");
+    let response = client.get(url).send().await?;
+    let status_code = response.status();
+    let response = response.text().await?;
+    tracing::trace!(%response, ?status_code, "Received 1Inch API response");
 
-    match serde_json::from_str::<RestResponse<D>>(&response)? {
-        RestResponse::Ok(result) => Ok(result),
-        RestResponse::Err(err) => {
+    match serde_json::from_str::<RestResponse<D>>(&response) {
+        Ok(RestResponse::Ok(result)) => Ok(result),
+        Ok(RestResponse::Err(err)) => {
             tracing::warn!(?err, "1inch API error");
             Err(err.into())
+        }
+        Err(err) => {
+            tracing::warn!(?err, "failed to parse response");
+            Err(OneInchError::Api(RestError {
+                status_code: status_code.as_u16().into(),
+                description: "failed to parse response".to_owned(),
+            }))
         }
     }
 }


vkgnosis (approved):

crates/shared/src/zeroex_api.rs
```diff
@@ -522,25 +521,22 @@ impl DefaultZeroExApi {
...
+            .map_err(ZeroExResponseError::Send)?
             .text()
             .await
             .map_err(ZeroExResponseError::TextFetch)?;
         tracing::trace!("Response from 0x API: {}", response_text);
 
-        if status == StatusCode::TOO_MANY_REQUESTS {
-            return Err(ZeroExResponseError::RateLimited);
-        }
-
```
vkgnosis, line 537: Wait, is this part right?

crates/shared/src/oneinch_api.rs
```diff
@@ -554,15 +554,24 @@ where
...
-    let response = client.get(url).send().await?.text().await?;
-    tracing::trace!(%response, "Received 1Inch API response");
-
-    match serde_json::from_str::<RestResponse<D>>(&response)? {
-        RestResponse::Ok(result) => Ok(result),
-        RestResponse::Err(err) => {
+    let response = client.get(url).send().await?;
+    let status_code = response.status();
+    let response = response.text().await?;
+    tracing::trace!(%response, ?status_code, "Received 1Inch API response");
```
nlordell, line 560: Do we handle `status_code == 429` in the caller?

crates/shared/src/zeroex_api.rs
MartinquaXD: Nope, rebased it away.

crates/shared/src/oneinch_api.rs
MartinquaXD: We currently handle it when converting it to a `TradeError`. Not sure if there are other places we need to check this case now.

MartinquaXD: Since the `OneInchPriceEstimator` is built on top of `dyn TraitFinding` this should be enough as the conversion from `TradeError` to `PriceEstimationError` is already `Err(RateLimited)` aware:
```rust
impl From<OneInchError> for TradeError {
    fn from(err: OneInchError) -> Self {
        match err {
            OneInchError::Api(err) if err.status_code == 429 => Self::RateLimited,
            err if err.is_insuffucient_liquidity() => Self::NoLiquidity,
            err => Self::Other(err.into()),
        }
    }
}
```

nlordell: Looks like we don't propagate a `RateLimited` variant from the 1Inch error, so we need to add that as well.

nlordell: Might make sense for `OneInchError` to have a `RateLimited` error variant.

nlordell (approved):

crates/shared/src/oneinch_api.rs
MartinquaXD: Will merge like this for now and clean up when there is no time pressure.

PR #1332, opened 2023-03-09
Activate auction rewards based on timestamp

sunce86:
Parameter `auction_rewards_activation_timestamp` defines the time when we switch to ranking the solutions by score.

Current plan is the following:
1. This PR becomes part of the next week release which is applied on Monday 13th of March.
2. 14th of March at 00.00am ranking by score is activated.
3. 14th of March at ~ 09.00am solver rewards script is executed for the past week ending with 14th of March 00.00am (ranking by objective_value)
4. 21st of March at ~ 09.00am solver rewards script is executed for the dates 14-21 March (ranking by score).

With this change, there is no need to do configuration changes (apply release) on 14th of March at 00.00am.

~One concern I have is the timezone considered by solver rewards script. @bh2smith do you know which one is used? Here I assumed Utc.~

### DIFF
--- a/crates/solver/src/settlement_ranker.rs
+++ b/crates/solver/src/settlement_ranker.rs
@@ -8,6 +8,7 @@ use {
         solver::{SimulationWithError, Solver},
     },
     anyhow::Result,
+    chrono::{DateTime, Utc},
     gas_estimation::GasPrice1559,
     model::auction::AuctionId,
     num::{rational::Ratio, BigInt, BigRational, CheckedDiv, FromPrimitive},
@@ -41,7 +42,7 @@ pub struct SettlementRanker {
     pub max_settlement_price_deviation: Option<Ratio<BigInt>>,
     pub token_list_restriction_for_price_checks: PriceCheckTokens,
     pub decimal_cutoff: u16,
-    pub enable_auction_rewards: bool,
+    pub auction_rewards_activation_timestamp: DateTime<Utc>,
 }
 
 impl SettlementRanker {
@@ -198,7 +199,7 @@ impl SettlementRanker {
         // objective value tie.
         rated_settlements.shuffle(&mut rand::thread_rng());
 
-        if self.enable_auction_rewards {
+        if Utc::now() > self.auction_rewards_activation_timestamp {
             rated_settlements.sort_by_key(|s| s.1.score.score());
 
             rated_settlements.iter_mut().rev().enumerate().for_each(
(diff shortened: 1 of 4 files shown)


bh2smith:
Everything is UTC for us.

vkgnosis (approved):

crates/solver/src/settlement_ranker.rs
```diff
@@ -198,7 +199,7 @@ impl SettlementRanker {
         // objective value tie.
         rated_settlements.shuffle(&mut rand::thread_rng());
 
-        if self.enable_auction_rewards {
+        if Utc::now() > self.auction_rewards_activation_timestamp {
```
MartinquaXD, line 202: If we reach this line for example 1 second before the deadline we would end up with another auction settled based on the old scheme after the deadline. Would that be a problem for the reward computation script?
I guess this is more a question for @bh2smith. If that is a problem we might have to pause the protocol for a while to ensure we have a clean cut over.

bh2smith: I am not even sure I understand the purpose for this. If we are already populating the DB for both order_rewards (as are currently in place) and `batch_rewards` (as we will transition into), why do we need a switch statement? None of this makes any sense to me.

MartinquaXD: A different solver could win the same auction depending on which way the solutions get ranked (pure objective value vs. `score` provided by solver). This switch causes us to change the ranking to the new implementation at the given point in time. So unless I'm missing a detail we might pick a bunch of winners before the deadline which get executed after the deadline where the `reference_score` is actually higher than the `winning_score`. The question is how big of a problem that would be. (And also if my understanding is even correct here)

sunce86: Indeed, from the solver rewards script there is no difference, both tables are populated as they should.
However, the difference comes from the fact that the new scheme rewards table is populated with
1. simulated values before deadline (we still give the objective_value winner to settle tx onchain)
2. real values after deadline

What @MartinquaXD is worried about is that we will detect a tx mined at 00:01am tuesday and look into new scheme tables and find the value which is a simulated value because it was inserted at 11.59am monday when the onchain settler was decided by objective_value and we might reward wrong solver.

sunce86: This might be avoided by taking `winner` from the `auction_transaction.tx_from` instead of `settlement_scores.winner` (at least for the first new scheme payout on 21st of March)

bh2smith: Yea. That is not great. solver and winner are expected to be the same. I am really hoping this does not happen.

PR #1343, opened 2023-03-13
Introduce Web3 Code Simulator with Tenderly Fallback

### LINKED ISSUE #1341: Faciliate Debugging of Failed Web3 Trade Simuations
We enable `web3` trade simulator, as Tenderly simulation credits are quite a bit more expensive that Ethereum RPC `eth_call`s. However, this makes debugging these cases much more difficult. This issue captures the work to make debugging these cases easier.

nlordell:
Fixes #1341 

This PR introduces a web3 trade simulator that falls back to Tenderly on failures. One pain-point of using the `Web3` simulator is that it makes debugging failed simulations very hard. This PR introduces a fallback mechanism such that only failed simulations go to Tenderly. This way, we save Tenderly credits but keep debugging easy!

### Test Plan

Manual test, was able to create a [simulation on Tenderly](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/7dec4cdd-ab30-4b9d-93c4-2d5883364803) for debugging.

### DIFF
--- a/crates/shared/src/code_simulation.rs
+++ b/crates/shared/src/code_simulation.rs
@@ -182,14 +182,14 @@ impl TryFrom<StateOverride> for StateObject {
 /// save failed simulations on Tenderly to facilitate debugging.
 pub struct Web3ThenTenderly {
     web3: Web3,
-    tenderly: TenderlyCodeSimulator,
+    tenderly: Arc<TenderlyCodeSimulator>,
 }
 
 impl Web3ThenTenderly {
     pub fn new(web3: Web3, tenderly: TenderlyCodeSimulator) -> Self {
         Self {
             web3,
-            tenderly: tenderly.save(true, true),
+            tenderly: Arc::new(tenderly.save(true, true)),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crates/shared/src/code_simulation.rs
```diff
@@ -178,6 +178,39 @@ impl TryFrom<StateOverride> for StateObject {
     }
 }
 
+/// A code simulator that uses Web3 in the general case, but will create and
+/// save failed simulations on Tenderly to facilitate debugging.
+pub struct Web3ThenTenderly {
```
cowfee, line 183: I'm of the opinion that there should be no `CodeSimulating` trait, there should only be a single code simulator type, and it should do "web3 then tenderly" by default. This would simplify the code. There's no reason to have three trait implementations just to facilitate a reasonable simulation behavior.

You probably already know this and are taking this route to stay consistent with the existing code style.

cowfee (approved):

crates/shared/src/code_simulation.rs
```diff
@@ -178,6 +178,39 @@ impl TryFrom<StateOverride> for StateObject {
...
+    async fn simulate(
+        &self,
+        call: CallRequest,
+        overrides: StateOverrides,
+    ) -> Result<Vec<u8>, SimulationError> {
+        match self.web3.simulate(call.clone(), overrides.clone()).await {
+            Ok(value) => Ok(value),
+            Err(err) => {
+                tracing::warn!(?err, "web3 code simulation failed");
+                self.tenderly.simulate(call, overrides).await
```
MartinquaXD, line 208: You could run the fallback tenderly simulation in a separate task to not delay the price estimation logic unnecessarily.

nlordell: The estimation wants the result though right? It can produce an error if this fails.

MartinquaXD: Yes, but in the case where you use the `tenderly` simulator the `web3` simulator already produced an `Err`. Why not simply return that after spawning the `tenderly` simulation task? I wouldn't expect the `tenderly` simulation to work when the `web3` simulation already failed so we might as well return that result ASAP.

MartinquaXD (approved):

PR #1344, opened 2023-03-13
Add Special Handling for `burn` in PhonyERC20

nlordell:
We noticed 1Inch and 0x routes failing to simulate that should have worked. This is because they used the `Dai.burn` function which doesn't work with phony ERC20 balances. This PR adds special handling to the `PhonyERC20` token so that it correctly supports burning from the fake balances setup by the phony ERC20 contract.

### Test Plan

Ran the `verified_oneinch_trade` on `DAI <> GNO` and got a simulation that [burns DAI with internal balances](https://dashboard.tenderly.co/public/safe/safe-apps/simulator/365c8cd7-f6fd-4c29-8c60-fa2cff4cb80b?trace=0.9.10.0.0.2.0.0.0.1.4.1.1) (notice the `0x9dc29fac` which is the selector for [`burn(address,uint256)`](https://www.4byte.directory/signatures/?bytes4_signature=0x9dc29fac)).


cowfee (approved):
Code seems OK but I'm missing domain knowledge.

MartinquaXD (approved):
Looks reasonable but I'm shaky on solidity. Will trust you on this one. 😅

PR #1345, opened 2023-03-14
`driver`: simulate non-internallized interactions as part of solution verification

cowfee:
Fixes https://github.com/cowprotocol/services/issues/1009.

The last missing piece for solution verification: simulating the settlement without internalizing any interactions.

### Test Plan

Added an automated test.

### DIFF
--- a/crates/driver/src/domain/competition/solution/mod.rs
+++ b/crates/driver/src/domain/competition/solution/mod.rs
@@ -132,7 +132,8 @@ impl Solution {
             auction,
             settlement::Internalization::Disable,
         )
-        .await?;
+        .await
+        .map_err(|_| VerificationError::FailingInternalization)?;
         // For the solution to be valid, the settlement with internalized interactions
         // must simulate, just like the non-internalized case above.
         self.simulate(eth, simulator, auction, settlement::Internalization::Enable)
@@ -280,7 +281,7 @@ impl Solution {
         {
             Ok(())
         } else {
-            Err(VerificationError::Internalization)
+            Err(VerificationError::UntrustedInternalization)
         }
     }
 
@@ -473,7 +474,9 @@ pub enum VerificationError {
     #[error(
         "invalid internalization: solution attempts to internalize tokens which are not trusted"
     )]
-    Internalization,
+    UntrustedInternalization,
+    #[error("invalid internalization: solution attempts to internalize failing interactions")]
+    FailingInternalization,
 }
 
 impl From<simulator::Error> for Error {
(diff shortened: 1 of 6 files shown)


crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -437,7 +458,9 @@ pub enum VerificationError {
     #[error(
         "invalid internalization: solution attempts to internalize tokens which are not trusted"
     )]
-    Internalization,
+    UntrustedInternalization,
+    #[error("invalid internalization: solution attempts to internalize failing interactions")]
```
MartinquaXD, line 462: Not sure if nit picking is worth it here but technically the simulation failure could originate from an interaction that is not supposed to be internalized. I think saying that the uninternalized version of the settlement failed to simulate is specific enough.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -41,14 +51,16 @@ impl Settlement {
...
+/// the case where no interactions were internalized.
 /// - Asset flow: the sum of tokens into and out of the settlement are
 /// non-negative, meaning that the solver doesn't take any tokens out of the
 /// settlement contract.
 /// - Internalization: internalized interactions only use trusted tokens.
 ///
 /// Such a settlement obeys the rules of the protocol and can be safely
-/// broadcast to the Ethereum network.
+/// broadcast to the Ethereum network without getting slashed. See the
+/// [`competition::Order::reward`] field for more about slashing.
```
MartinquaXD, line 63: I'm not sure if we can prevent 100% of all slashable offenses with these tests. For example if a solver provides a solution that has a ridiculously low clearing price for some token and now other solver provides a solution we would probably still slash that solver since it would have been possible to achieve a better price. I believe something like this recently happened with `Quasilabs`.
Maybe a more defensive wording would be good.

MartinquaXD (approved):

crates/driver/src/domain/competition/solution/mod.rs
nlordell: An example of this would be:
- Interaction 1: Swap 50 tokens over pool A _internalized_
- Interaction 2: Swap 50 tokens over pool A _NON-internalized_

In this example, a solver would split the swaps over the same pool in order to get better prices (since there would be less price impact). Each on their own simulates successfully, and here, the second interaction is the one to revert.

crates/driver/src/domain/competition/solution/settlement.rs
nlordell: b3cd7eaececcb0596d36bc41b134b996260ff26c

harisang:
Nice, definitely much needed! Thanks!

PR #1347, opened 2023-03-16
Fix decoding settlement

sunce86:
In case we index a transaction initiated by solver that ends up emitting `Settlement` event but is not a standard transaction calling `settle`, we want to add entry to `settlement_observations` table with all zeros. Currently this fails and `on_settlement_event_updater` is crash looping. 

More context how this happened: https://cowservices.slack.com/archives/C0361CDD1FZ/p1678983343723179

### DIFF
--- a/crates/autopilot/src/decoded_settlement.rs
+++ b/crates/autopilot/src/decoded_settlement.rs
@@ -383,7 +383,7 @@ pub fn decode_function_input(
 ) -> Result<Vec<Token>, DecodingError> {
     let input = input
         .strip_prefix(&function.selector())
-        .context("input does not start with function selector")?;
+        .ok_or(DecodingError::InvalidSelector)?;
     let decoded_input = function
         .decode_input(input)
         .context("decode input failed")?;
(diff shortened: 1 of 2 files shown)


fedgiac (approved):

harisang:
Since I am not sure I get exactly what is happening, what auction id would this "settlement" correspond to?

sunce86:
@harisang 
Auction: `5848467`
Transaction: https://etherscan.io/tx/0x305a86b8b56f9a0b62e4cb7316c9cf8f3b0e1986c5d14b5f6056a6fefe730462

harisang:
I see. So this means that the settlement that was supposed to happen as a result of Auction 5848467 didn't happen, as we overrode it with
https://etherscan.io/tx/0x305a86b8b56f9a0b62e4cb7316c9cf8f3b0e1986c5d14b5f6056a6fefe730462

This does look strange, but it's ok I guess.

https://api.cow.fi/mainnet/api/v1/solver_competition/by_tx_hash/0x305a86b8b56f9a0b62e4cb7316c9cf8f3b0e1986c5d14b5f6056a6fefe730462

Let's make sure that this does not happen again in the future though, as this would result in a penalty for the solver, under CIP-20

PR #1349, opened 2023-03-17
CIP20: Forward score to settlement creation

sunce86:
Forwards score provided by solvers to settlement creation.

Unfortunately, using `Option` and `mut settlement` while building the settlement prevented rust compiler to help me catch this.


bh2smith (approved):
Looks fine to me although I am missing whatever details I feel I should know about what the problem actually was before this change.

sunce86:
So this code change is actually connecting whatever score solvers submitted with how the score is handled (ranking of solutions etc). The problem was, even though the solvers submitted a score, it was not used for ranking. This PR fixes that.

MartinquaXD (approved):

PR #1351, opened 2023-03-17
CIP20: Improve score calculation logging

sunce86:
Added more logging and now calculation returns `Result` for more observability.

Also fixes https://github.com/cowprotocol/services/issues/1357


crates/solver/src/solver/score_computation.rs
```diff
@@ -32,18 +32,40 @@ pub struct ScoreCalculator {
...
+            ?gas_price,
+            ?nmb_orders,
+            ?exponent,
+            ?success_probability,
+            ?score,
+            "score calculation",
+        );
+        f64::to_u128(&score)
+            .map(Into::into)
+            .context("failed to convert score")
```
MartinquaXD, line 68: Not sure if we ever expect `score > u128` but you could do `U256::from_f64_lossy()` here to allow for that. If we want to report errors during the conversion we might have to `ensure!(result < U256::MAX, "score conversion possibly overflowed")`.

vkgnosis (approved):
Approving asuming that the logic for the return value is changed back to `U256::from_f64_lossy`.

MartinquaXD (approved):

Issue #1352, opened 2023-03-17
Cleanup CIP20 temporary code


sunce86:
Once the CIP20 is activated and finalized, remove all temporary code and parameters.

PR #1353, opened 2023-03-17
CIP20: Filter out non-positive score settlements

sunce86:
Do not consider non-positive zero score settlements for solver competition.
Settlement has non-positive score in two cases
- solver provided zero score
- solver did not provide score -> fallback to objective_value which is negative or zero (with or without discount doesn't matter)

Put under parameter so we can easily disable this if there are a lot of auctions with no positive winner score.

### DIFF
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -373,7 +373,7 @@ pub enum SolverRejectionReason {
     /// re-create simulation locally
     SimulationFailure(TransactionWithError),
 
-    /// The solution don't have a positive score. Currently this can happen
+    /// The solution doesn't have a positive score. Currently this can happen
     /// only if the objective value is negative.
     NonPositiveScore,
 }
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -372,6 +372,10 @@ pub enum SolverRejectionReason {
     /// The solution didn't pass simulation. Includes all data needed to
     /// re-create simulation locally
     SimulationFailure(TransactionWithError),
+
+    /// The solution don't have a positive score. Currently this can happen
+    /// only if the objective value is negative.
+    NonPositiveScore,
 }
 
 #[derive(Debug, Serialize)]
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -334,7 +334,8 @@ pub struct Arguments {
     #[clap(flatten)]
     pub score_params: score_computation::Arguments,
 
-    /// Should we skip settlements with non-positive score for solver competition?
+    /// Should we skip settlements with non-positive score for solver
+    /// competition?
     #[clap(long, env, default_value = "true")]
     pub skip_non_positive_score_settlements: bool,
 }
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -336,7 +336,7 @@ pub struct Arguments {
 
     /// Should we skip settlements with non-positive score for solver
     /// competition?
-    #[clap(long, env, default_value = "true")]
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
     pub skip_non_positive_score_settlements: bool,
 }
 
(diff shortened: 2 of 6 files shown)


harisang:
Just to reiterate, the purpose of this PR is to enforce the following specification CIP-20:

> Submitted scores have to be positive.

Thus, a solution with a score that is not positive should NOT be considered valid, and thus, would not be part of the ranking, and the subsequent "participation reward"

crates/solver/src/arguments.rs
```diff
@@ -333,6 +333,11 @@ pub struct Arguments {
 
     #[clap(flatten)]
     pub score_params: score_computation::Arguments,
+
+    /// Should we skip settlements with non-positive score for solver
+    /// competition?
```
MartinquaXD, line 338: Using `///` will cause this comment to show up in the `--help` text which doesn't seem intended here.

crates/solver/src/arguments.rs
```diff
@@ -333,6 +333,11 @@ pub struct Arguments {
...
+    #[clap(long, env, default_value = "true")]
```
MartinquaXD, line 339: Using `default_value = true` on its own means we can't disable the value. However, Valentin found out that this can be addressed by adding `action = clap::ArgAction::Set` which changes the syntax from `--bool-flag` to `--bool-flag <true|false>`. This should be used from now on for every new bool argument.

crates/shared/src/http_solver/model.rs
```diff
@@ -372,6 +372,10 @@ pub enum SolverRejectionReason {
     /// The solution didn't pass simulation. Includes all data needed to
     /// re-create simulation locally
     SimulationFailure(TransactionWithError),
+
+    /// The solution don't have a positive score. Currently this can happen
```
MartinquaXD, line 376:
```suggestion
    /// The solution doesn't have a positive score. Currently this can happen
```

crates/solver/src/settlement_ranker.rs
```diff
@@ -194,6 +195,27 @@ impl SettlementRanker {
             );
         }
 
+        // Filter out settlements with non-positive score.
+        if Utc::now() > self.auction_rewards_activation_timestamp // CIP20 activated
+        && self.skip_non_positive_score_settlements
```
MartinquaXD, line 200: Not sure if it makes sense to have `CIP-20` code enabled while also allowing negative scores. If this is not the case we don't need the new flag. I guess this gives us more flexibility so we don't have to decide this matter right now (we could simply always have this flag be true).
I guess I'm saying that to me it seems like we should not have this flag as allowing negative scores could introduce problems but I'm not 100% sure.

sunce86: Essentially, we are afraid that there could be a lot of consecutive auctions where winner would have negative objective value and therefore skipping the submission part, so we need a fast way to allow negative scores after all. 
The whole `if` will be dropped anyway in https://github.com/cowprotocol/services/issues/1352

crates/shared/src/http_solver/model.rs
sunce86: Fixed.

crates/solver/src/arguments.rs
sunce86: Nice catch. Did not check similar bool arguments. Fixed.

crates/solver/src/settlement_ranker.rs
MartinquaXD: I get the idea. I was just not sure how compatible that is with the solver rewards computation. Would it maybe make sense to use `score == 0, reference_score == 0` for the reward computation in those cases?

harisang: So, indeed, we don't want to allow negative scores, in the sense that the trivial solution with score zero would always beat them. And explicitly not submitting the empty solution and allowing negative scores can lead to problems, the most obvious being an auction with only one submitted solution of negative score; what is the reference score in that case? Also, negative scores can result in aggressive gaming under certain cases.

Long story short, allowing negative scores, which should only happen temporarily, and while ensuring that solvers report truthfully  it is really a last-resort solution, in case we see that we are filtering out tons of auctions. But, again, the plan is not to activate this flag at all.

sunce86: If we fallback to allow negative scores, it just means that those with negative objective values will have score 0 and considered as 0 in ranking. But as @harisang  said, we expect this to not happen at all.

MartinquaXD (approved):

sunce86:
Merging with 1 approval so I have time to test it out and include into release today.

PR #1362, opened 2023-03-21
Make Driver EIP-1559 Aware

nlordell:
Related to #1360

This PR makes the driver EIP-1559 aware internally. Previously, it would just make use of `effective_gas_price` everywhere. However, there are EIP-1559 specific requirements that require a more complete gas estimate picture.

This PR is a required step for implementing #1360 and fixing account balance checks before committing to a settlement.

### Test Plan

Unit tests continue to pass after adjusting for new gas price estimates. Note that `eth_gasPrice` RPC call is different than an EIP-1559 price estimate, which we do using the `gas_estimation::NativeGasEstimator` crate.

I also change E2E tests that assert a specific score to also assert a specific effective gas price, since the two are correlated.

### DIFF
--- a/crates/driver/src/domain/eth/gas.rs
+++ b/crates/driver/src/domain/eth/gas.rs
@@ -31,13 +31,13 @@ impl From<Gas> for U256 {
 #[derive(Debug, Clone, Copy)]
 pub struct GasPrice {
     /// The maximum total fee that should be charged.
-    pub max: MaxFeePerGas,
+    pub max: FeePerGas,
     /// The maximum priority fee (i.e. the tip to the block proposer) that
     /// can be charged.
-    pub tip: MaxPriorityFeePerGas,
+    pub tip: FeePerGas,
     /// The current base gas price that will be charged to all accounts on the
     /// next block.
-    pub base: BaseFeePerGas,
+    pub base: FeePerGas,
 }
 
 impl GasPrice {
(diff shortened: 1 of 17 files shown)


crates/driver/src/domain/eth/gas.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct GasPrice {
+    /// The maximum total fee that should be charged.
+    pub max: MaxFeePerGas,
+    /// The maximum priority fee (i.e. the tip to the block proposer) that
+    /// can be charged.
+    pub tip: MaxPriorityFeePerGas,
+    /// The current base gas price that will be charged to all accounts on the
+    /// next block.
+    pub base: BaseFeePerGas,
```
MartinquaXD, line 40: Not so sure about all these types. `MaxFeePerGas` and `BaseFeePerGas` don't have to be 2 different types as they are the same unit. And I think the same goes for `MaxPriorityFeePerGas` as well.
I would say in this context a type usually represents a unit and the field name + doc comments explains what the value + unit is supposed to represent. Especially since these 3 things only make sense as part of a `GasPrice` and don't get used on their own.
But since we don't really use these types they also don't hurt.

crates/driver/src/tests/setup/blockchain/mod.rs
```diff
@@ -15,6 +19,17 @@ pub fn web3(url: &str) -> DynWeb3 {
...
 
+/// Compute the effective gas price estimate from a node.
+pub async fn effective_gas_price(web3: &DynWeb3) -> eth::U256 {
+    let estimate = NativeGasEstimator::new(web3.transport().clone(), None)
+        .await
+        .unwrap()
+        .estimate()
+        .await
+        .unwrap();
+    eth::U256::from_f64_lossy(estimate.effective_gas_price())
```
MartinquaXD, line 30: This could be done in a lossless manner, no?
I don't expect it to matter but it looks surprising.
Also why does this return `eth::U256` when the type `EffectiveGasPrive` already exists?

MartinquaXD (approved):
LGTM. Just bike shedding a bit.

crates/driver/src/tests/setup/blockchain/mod.rs
nlordell: This is in the testing code. I wanted to not use domain types here.

crates/driver/src/domain/eth/gas.rs
nlordell: The reason I kept them as separate types is that they model slightly different things and have different behaviors in the context of EIP-1559.

I will just unify them to `FeePerGas`, as I agree they may be slightly overkill.

PR #1365, opened 2023-03-22
`driver`: Verify Solver Account Balance

nlordell:
This PR adds an additional check for solver balances before committing to a solution.

Previously, we had relied on simulations from the node failing on `insufficient funds for gas * price + value` errors. This was problematic for two reasons:
* We weren't specifying gas parameters correctly when creating access lists. This lead to the default `gas` and `gas_price` values to be chosen by the node and Geth seems to use unreasonable defaults (cc @mateipopa - this solves the issue that we saw on Görli connected to NodeReal, where we were not submitting settlements because of insufficient balances despite having a good amount of Ether and network conditions being "normal")
* These errors are node-dependent and not guaranteed by the Ethereum RPC standard. Specifically, Nethermind does not check account gas values on simulations (see https://github.com/cowprotocol/services/pull/1361#issuecomment-1477951819).

### Test Plan

Added an E2E test that verifies that the `driver` responds with a correct error code when the solver account has insufficient balance for the proposed settlement.


crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -142,15 +151,13 @@ impl Ethereum {
...
-    fn into_request(tx: eth::Tx) -> web3::types::TransactionRequest {
-        web3::types::TransactionRequest {
-            from: tx.from.into(),
-            to: Some(tx.to.into()),
-            value: Some(tx.value.into()),
-            data: Some(tx.input.into()),
-            access_list: Some(tx.access_list.into()),
-            ..Default::default()
-        }
+    pub async fn balance(&self, address: eth::Address) -> Result<eth::Ether, Error> {
```
MartinquaXD, line 154: nit: This could also be interpreted as the total balance of some random token address. Maybe mention eth in the function name or a doc comment.

crates/driver/src/tests/cases/solver_balance.rs
```diff
@@ -0,0 +1,245 @@
...
+    crate::{
+        domain::competition::{self, auction},
+        infra,
+        tests::{self, hex_address, setup},
+    },
+    itertools::Itertools,
+    serde_json::json,
+};
+
+/// Test that the /solve errors when solver balance is low.
```
MartinquaXD, line 14:
```suggestion
/// Test that the `/solve` request errors when solver balance is low.
```

crates/driver/src/tests/cases/solver_balance.rs
```diff
@@ -0,0 +1,245 @@
...
+            contracts: infra::config::file::ContractsConfig {
+                gp_v2_settlement: Some(settlement.address()),
+                weth: Some(weth.address()),
+            },
+        },
+        geth: &geth,
+    })
+    .await;
+
+    // Transfer out all of the solver's funds.
```
MartinquaXD, line 162: Would it be easily doable to verify that we require `current_balance >= settlement_gas * max_gas_price_after_n_blocks`?

MartinquaXD (approved):

crates/driver/src/infra/blockchain/mod.rs
nlordell: I'll add it as a doc-comment.

I feel like the name + type made it clear that it returned an Ether balance (specifically `balance() -> eth::Ether` instead of `balance() -> eth::Asset`).

crates/driver/src/tests/cases/solver_balance.rs
nlordell: Nice! Added this (at least, my interpretation of the suggestion) and caught a small issue - we weren't considering the `max_priority_fee_per_gas` in the computation.

MartinquaXD: Yes, basically verify the check you've been working on that the solver has to have enough eth to submit the solution even if the gas price increases by `12.5%` for 10 blocks consecutively.

nlordell: So I changed the E2E test to leave a reasonable amount of Eth (more than enough for a `settle` transaction) but not enough if a gas spike were to happen, and asserted that the balance check works there.

Actually filling blocks to 100% gas usage for 10 consecutive blocks is pretty tricky though...

PR #1366, opened 2023-03-22
Support external solvers returning partial fills

### LINKED ISSUE #1322: Accept Partial Fills From External Solvers.
This issue captures the work around accepting partially fillable order executions from external solvers.

Specifically, it would accept partially fillable orders with the existing `executedAmount` field (that indicates what absolute amount of the order was executed in its "side" token). Additionally, partial limit order fills **must** include a new `executedSurplusFee` amount (absolute amount in the order's "side" token) which represents how much surplus fee needs to be taken from each solver.

The order's actual execution will be determined by the settlements uniform clearing prices and the executed surplus fee amount (similar to how existing limit order executions and "custom prices" are determined, but with the `suplusFee` coming from the solver instead of the `autopilot`). Computing and encoding the actual order execution is outside the scope of this issue and covered in #1323.

MartinquaXD:
Fixes #1322 

Adds support for solvers to return a computed `fee` with their order execution. In the new driver `fee` is a `U256` and implicitly refers to the `sell_token` and in the current driver `fee` is actually `{token: H160, amount: U256}` since this already existed (but was unused). Of course we can easily change the expected API response if you'd like.

This is required to correctly settle partially fillable limit orders. This value gets used internally like `surplus_fee` as this value really expresses the same thing (just computed by the solver).
Ideally I would have liked to use enums to model the different kinds of fees (computed by protocol vs. solver) but the solutions I tried all ended up with even more ugly matching logic and `unreachable!()` as they the enums couldn't contain all the required information.
So in order to make it as safe as possible I made sure that all orders that need a solver computed fee have that (in the logic that handles solver responses) so we can streamline the code by using `.expect()` wherever needed.

Also since it wasn't super clear to me where a fee value was supposed to be scaled vs. taken as given I added a few doc comments that should help understand the details of this PR.

### Compatibility
For fok limit orders we were able to hide the `surplus_fee` and price adjustment shenanigans in the driver to enable solvers to settle limit orders without changing anything on their end. With partially fillable limit orders solvers definitely have to add some new logic to handle them but we at least can design the response such that we only have 1 code path for limit orders in the driver.
This means the driver can treat the new `fee` as it currently treats the `surplus_fee` and solvers would have to submit their solutions in a specific way where the `executed` amounts should **not** contain the `fee` which would look like this for a sell order:
* solver returns `executed_sell_amount`, `executed_buy_amount`, `fee` and some `clearing_prices` where `executed_sell_amount * sell_price = executed_buy_amount * buy_price` where `fee` is big enough to pay for the execution cost
* driver then crafts a tx selling `(executed_sell_amount + fee)` to buy `executed_buy_amount` with a reduced `sell_price` to capture the `fee`

### Test Plan

CI. E2E test in a follow up PR.


crates/solver/src/liquidity.rs
```diff
@@ -212,8 +212,28 @@ impl LimitOrder {
...
+        Self {
+            filled_amount: filled,
+            executed_solver_fee: solver_fee,
+        }
+    }
+}
+
 impl Settleable for LimitOrder {
-    type Execution = U256;
+    type Execution = LimitOrderExecution;
```
MartinquaXD, line 236: I think this is the basic change that needs to happen and everything else follows from there.
@nlordell does that seem sensible to you?

nlordell: I think so as well.

crates/solver/src/liquidity.rs
```diff
@@ -212,8 +212,28 @@ impl LimitOrder {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LimitOrderExecution {
+    /// The amount that the order `side` (`buy`, `sell`) should be filled by
+    /// this trade.
+    pub filled_amount: U256,
```
harisang, line 219: Are these variable names matching exactly how the solvers would report this in their json response?

MartinquaXD: No. The new value is currently being read from the `fee` field. Which already exists in our parsing logic but was not used AFAIK. If this is actually a problem we might have to introduce an entirely new name. Nothing is set in stone at this point so feel free to suggest something but if I recall correctly you also called the field `fee` in your overview document.

MartinquaXD: Actually [here](https://github.com/cowprotocol/services/blob/main/crates/shared/src/http_solver/model.rs#L261-L269) are the fields we currently read from the solver response. So turns out that I actually misremembered slightly as the existing (but currently unused) format is actually `fee: {amount: "0", token: "0xeee..eeee"}`. But I would rather turn that into simple `U256` which implicitly refers to orders `sell_token`. WDYT?

harisang: I see. Yeah `fee` is fine so let's stick with it. Regarding the format, I guess for uniformity reasons (since we send it to solvers in the fee -> {amount, token} format in the input.json), we can keep it as is.

harisang: And regarding specifying the executed amounts, depending on whether it's a sell or buy order, the solver would submit an `exec_sell_amount` or `exec_buy_amount`, accordingly, right?

MartinquaXD: Right now the parsing logic always expects both values but we only use one and and recompute the other based on the ucp.

crates/driver/src/boundary/settlement.rs
```diff
@@ -81,28 +82,45 @@ impl Settlement {
...
                     // executed amounts which seems weird to me... why is a
                     // solver specifying trades with 0 executed amounts?
-                    if eth::U256::from(trade.executed).is_zero() {
+                    if eth::U256::from(trade.execution.filled).is_zero() {
                         return Err(Error::Boundary(anyhow!("unexpected empty execution")));
                     }
 
-                    (to_boundary_order(&trade.order), trade.executed.into())
+                    let solver_fee = match trade.order.solver_determines_fee() {
+                        true => trade.execution.fee.expect("API ensures fee exists").0,
```
vkgnosis, line 95: Where is that ensured?

MartinquaXD: That happens [here](https://github.com/cowprotocol/services/pull/1366/files#diff-9f2c31c10679a7e214aef5191d3b280418861f993fc0fa3d7da78ac8f9d994f5R45-R48). If a solver provides a trade execution without a fee for a partially fillable limit order the JSON response fails to convert to the domain `Solution` type.

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -276,6 +296,7 @@ fn match_prepared_and_settled_orders(
                 executed_buy_amount: settled.exec_buy_amount,
                 executed_sell_amount: settled.exec_sell_amount,
                 exec_plan: settled.exec_plan,
+                fee: settled.fee.map(|fee| fee.amount),
```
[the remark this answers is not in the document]
MartinquaXD: Yes. I have no idea why it was added. My best guess is that only Quasimodo returns that value for some reason. Maybe @harisang has some info on that?
I'm totally fine with adding a new field to the order execution that just stores a `U256`. 👌

crates/driver/src/boundary/settlement.rs
nlordell: Would be nice if the invariant is kept local to the `competition::solution::Trade` type and module (i.e. not leak here).

I think I remember you mentioning that this was tricky to do, just food for thought.

crates/solvers/openapi.yml
```diff
@@ -568,13 +568,20 @@ components:
             - $ref: "#/components/schemas/Fulfillment"
             - $ref: "#/components/schemas/JitTrade"
         - type: object
+          required:
+            - executedAmount
```
nlordell, line 572: I don't think this is required for JIT orders, and assumed to be the full JIT order amount. (I may be remembering wrong though!).

nlordell: Nevermind, it does use this field.

crates/solver/src/solver/http_solver/settlement.rs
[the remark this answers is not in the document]
nlordell: It _looks_ like this field just echos the `fee` field that we pass in with the order. I would also err on the side of caution and introduce a new `exec_fee_amount` field for this.

crates/driver/src/boundary/settlement.rs
MartinquaXD: Mhh, I guess something like this could work:
```rust
pub struct SolverFee(pub competition::order::SellAmount);
pub struct ProtocolFee(pub competition::order::SellAmount);

pub enum Fee {
    Solver(SolverFee),
    Protocol(ProtocolFee),
}

impl OrderExecution {
    /// Panics when the trade should not or does not contain a [`SolverFee`].
    pub fn solver_fee(&self, order: &competition::Order) -> SolverFee {
        match (order.solver_determines_fee(), self.fee) {
            (true, Fee::Solver(fee)) => fee,
            _ => panic!("invariants violated"),
        }
    }

    /// Panics when the trade should not or does not contain a [`ProtocolFee`].
    pub fn protocol_fee(&self, order: &competition::Order) -> ProtocolFee {
        match (order.solver_determines_fee(), self.fee) {
            (true, Fee::Protocol(fee)) => fee,
            _ => panic!("invariants violated"),
        }
    }
}
```

nlordell: I did something similar by making the `Fulfillment` fields private and verifying the invariants at creation time.

vkgnosis (approved):
Looks fine but  I think the suggestion about not reusing an existing fee field is open.

nlordell:
> Looks fine but I think the suggestion about not reusing an existing fee field is open.

fd02061d21a1f209bd85a4922d93645451455d2c

PR #1367, opened 2023-03-22
Verify Solver Account Balances

nlordell:
Together with #1363 this supersedes #1361.

#1363 introduced a change where we compute the absolute maximum `max_fee_per_gas` we are willing to pay per solver run. With that change, we can now, deterministically, decide upfront if a solver account has sufficient Ether to submit a settlement accounting for gas spikes.

Historically, we relied on `eth_estimateGas` RPC call failing when a solver had insufficient funds, however, this behavior is node-dependent and cannot be relied on (See https://github.com/cowprotocol/services/pull/1361#issuecomment-1477951819 for more details).

In this PR we change the settlement raking to additional query solver balances and ensure that the settlement can be executed even if the `max_fee_per_gas` is charged.

The equivalent fix in the co-located driver was implemented in #1365.

### Test Plan

Unfortunately, unit testing this code is very hard 😞. However, the actual happy path will get exercised in end-to-end tests. Please carefully review the math here.

### DIFF
--- a/crates/solver/src/settlement_rater.rs
+++ b/crates/solver/src/settlement_rater.rs
@@ -8,12 +8,13 @@ use {
             settle_method,
             simulate_and_estimate_gas_at_current_block,
         },
-        solver::{SettlementWithSolver, Simulation, SimulationWithError, Solver},
+        settlement_submission::gas_limit_for_estimate,
+        solver::{SettlementWithSolver, Simulation, SimulationError, SimulationWithError, Solver},
     },
     anyhow::{Context, Result},
     contracts::GPv2Settlement,
     ethcontract::errors::ExecutionError,
-    futures::future::join_all,
+    futures::{future, future::join_all},
     gas_estimation::GasPrice1559,
     itertools::{Either, Itertools},
     model::solver_competition::Score,
@@ -26,7 +27,11 @@ use {
         external_prices::ExternalPrices,
         http_solver::model::{InternalizationStrategy, SimulatedTransaction},
     },
-    std::{borrow::Borrow, sync::Arc},
+    std::{
+        borrow::Borrow,
+        collections::{HashMap, HashSet},
+        sync::Arc,
+    },
     web3::types::AccessList,
 };
 
@@ -188,7 +193,7 @@ impl SettlementRating for SettlementRater {
             )
             .await?;
 
-        let gas_price =
+        let effective_gas_price =
             BigRational::from_float(gas_price.effective_gas_price()).expect("Invalid gas price.");
 
         let rate_settlement = |id, settlement: Settlement, gas_estimate: U256| {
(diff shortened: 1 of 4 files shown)


crates/solver/src/settlement_rater.rs
```diff
@@ -218,13 +223,37 @@ impl SettlementRating for SettlementRater {
...
             }
         };
 
+        let solver_balances = future::join_all(
+            simulations
+                .iter()
+                .filter(|result| result.gas_estimate.is_ok())
+                .map(|result| result.simulation.solver.account().address())
+                .collect::<HashSet<_>>()
+                .into_iter()
```
MartinquaXD, line 239: nit: `itertools` is already in scope so you might as well use `.unique()` here.

crates/solver/src/settlement_submission.rs
```diff
@@ -39,7 +39,13 @@ use {
     web3::types::TransactionReceipt,
 };
 
-const ESTIMATE_GAS_LIMIT_FACTOR: f64 = 1.2;
+/// Computes a gas limit from a gas estimate that account for some buffer in
```
MartinquaXD, line 42:
```suggestion
/// Computes a gas limit from a gas estimate that accounts for some buffer in
```

crates/solver/src/settlement_rater.rs
```diff
@@ -235,14 +264,34 @@ impl SettlementRating for SettlementRater {
...
-                        simulation.transaction.access_list,
-                    )),
+                    Ok(gas_estimate) => {
+                        let gas_limit = gas_limit_for_estimate(gas_estimate);
+                        let required_balance = gas_limit
+                            .saturating_mul(U256::from_f64_lossy(gas_price.max_fee_per_gas));
+                        let solver_balance = solver_balances
+                            .get(&simulation.solver.account().address())
+                            .copied()
+                            .unwrap_or_default();
```
MartinquaXD, line 274: I think more appropriate would be to return a separate error in case the web3 call to get the balance failed. Otherwise solvers might get confused then they receive `SimulationError::InsufficientBalance` when they actually had enough funds.
But I expect this to be rare enough that it might not be worth it to make the code even more noisy.

MartinquaXD (approved):
LGTM

crates/solver/src/settlement_rater.rs
nlordell: > But I expect this to be rare enough that it might not be worth it to make the code even more noisy.

This was my rationale. Also, we are moving to co-location soon so I don't want to optimize too much for solver UX here.

sunce86:
I suspect some nodes are ignoring even the `max_fee_per_gas`, for example:
https://github.com/openethereum/openethereum/blob/main/crates/ethcore/src/executive.rs#L1104-L1108
(or this might be a bug or a standard spec for `eth_estimateGas` has changed in the meantime)

Overall, I don't think the time spent into investigating this is worth it. We should just add an explicit balance check and stop relying on `estimateGas`.
@nlordell @fleupold

nlordell:
> Overall, I don't think the time spent into investigating this is worth it. We should just add an explicit balance check and stop relying on estimateGas.

This is what we started doing now. We do explicit balance checks.

PR #1368, opened 2023-03-22
Reorganize dex solver config loading

MartinquaXD:
Reorganized the directory structure responsible for loading configs for dex solvers to be more in line with the file structure of their logic.
This makes it so that config loading and the solver implementation are composed of 2 parts.
1. logic that is shared across all dex solvers
2. logic that only relates to each solver's dex API

Also adds a new argument which controls how we handle partially fillable orders required for [this](https://github.com/cowprotocol/services/pull/1358#issuecomment-1479201711).

### Test Plan
CI

### DIFF
--- a/crates/solvers/src/infra/config/dex/balancer/file.rs
+++ b/crates/solvers/src/infra/config/dex/balancer/file.rs
@@ -1,0 +1,54 @@
+use {
+    crate::{
+        domain::eth,
+        infra::{config::dex::file, contracts, dex},
+    },
+    ethereum_types::H160,
+    serde::Deserialize,
+    serde_with::serde_as,
+    std::path::Path,
+};
+
+#[serde_as]
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+struct Config {
+    /// The URL of the Balancer SOR API.
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    endpoint: reqwest::Url,
+
+    /// Optional Balancer V2 Vault contract address. If not specified, the
+    /// default Vault contract address will be used.
+    vault: Option<H160>,
+
+    /// Optional CoW Protocol Settlement contract address. If not specified,
+    /// the default Settlement contract address will be used.
+    settlement: Option<H160>,
+}
+
+/// Load the driver configuration from a TOML file.
+///
+/// # Panics
+///
+/// This method panics if the config is invalid or on I/O errors.
+pub async fn load(path: &Path) -> super::Config {
+    let config: Config = file::parse(path).await;
+
+    // Balancer SOR solver only supports mainnet.
+    let contracts = contracts::Contracts::for_chain(eth::ChainId::Mainnet);
+
+    super::Config {
+        sor: dex::balancer::Config {
+            endpoint: config.endpoint,
+            vault: config
+                .vault
+                .map(eth::ContractAddress)
+                .unwrap_or(contracts.balancer_vault),
+            settlement: config
+                .settlement
+                .map(eth::ContractAddress)
+                .unwrap_or(contracts.settlement),
+        },
+        base: file::load_base_config(path).await,
+    }
+}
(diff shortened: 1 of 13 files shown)


crates/solvers/src/infra/config/dex/mod.rs
```diff
@@ -0,0 +1,10 @@
+pub mod balancer;
+mod file;
+pub mod zeroex;
+
+use crate::domain::{dex::slippage, eth};
+
+pub struct BaseConfig {
```
nlordell, line 7: Mostly an organizational/naming nit, but to me this is not a "base configuration" but the "dex-solver configuration" (i.e. the configuration options required by the `solver::Dex` and unrelated to the actual DEX API being used).

Maybe it makes sense to have:
```
infra::config::dex
├── solver
├── balancer
└── zeroex
```

crates/solvers/src/infra/config/dex/balancer/file.rs
```diff
@@ -1,20 +1,17 @@
...
     serde::Deserialize,
     serde_with::serde_as,
     std::path::Path,
-    tokio::fs,
 };
 
 #[serde_as]
 #[derive(Deserialize)]
-#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
```
nlordell, line 14: I liked the `deny_unknown_fields` as it helps prevents misconfigurations (imagine you had an _optional_ field `foo` and you instead set `fou`, you might silently run thinking you have `foo` configured but in reality you do not).

You could instead flatten the fields here and have something like:
```rust
#[derive(Deserialize)]
struct balancer::Config {
    // Balancer related fields.
    endpoint: reqwest::Url,
    vault: Option<H160>,
    settlement: Option<H160>,
    
    #[flatten]
    solver: infra::config::dex::solver::File,
}
```

And then `impl From<infra::config::dex::solver::File> for <infra::config::dex::solver::Config>`.

The unfortunate thing is that it leaks the `File` type, but you can scope that with `pub(super)` to limit it to the `infra::config::dex`.

crates/solvers/src/infra/config/dex/file.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// The amount of eth a partially fillable order should be filled for at
+    /// least.
+    #[serde(default = "default_smallest_partial_fill")]
+    smallest_partial_fill: eth::U256,
+}
+
+fn default_endpoint() -> reqwest::Url {
+    "https://api.0x.org/swap/v1/".parse().unwrap()
+}
+
```
nlordell, line 39:
```suggestion
```

crates/solvers/src/infra/config/dex/file.rs
```diff
@@ -0,0 +1,86 @@
...
+fn default_affiliate() -> H160 {
+    contracts::Contracts::for_chain(eth::ChainId::Mainnet)
+        .settlement
+        .0
+}
+
```
nlordell, line 45:
```suggestion
```

nlordell (approved):
LGTM. A couple of nits:
- naming nit: "base config" is a bit vague - to me its really the configuration for DEX single-order solvers, so `solver::Config` makes a bit more sense
- I would prefer to keep `deny_unknown_fields` as it helps catch misconfiguration errors.

crates/solvers/src/infra/config/dex/balancer/file.rs
nlordell: Alternatively, we can add nesting and then we don't need to flatten (and would essentially just change the format).

So:
```toml
endpoint = "https://my.endpoint"

[solver]
relative-slippage = "1.2"
```

nlordell: Nice, I like how it turned out.

crates/solvers/src/infra/config/dex/mod.rs
nlordell: Makes sense. I think that the new way you organized the configs makes a lot of sense.

fedgiac (approved):
Looks good to me.

PR #1371, opened 2023-03-23
Add Example Legacy Solver Adapter Configuration

nlordell:
This PR just adds a sample Legacy solver configuration for documentation purposes.

### Test Plan

```
cargo run -p solvers -- legacy --config crates/solvers/config/example.legacy.toml
```

### DIFF
--- a/crates/solvers/config/example.legacy.toml
+++ b/crates/solvers/config/example.legacy.toml
@@ -1,0 +1,3 @@
+chain-id = "1"
+solver-name = "CoW Solver"
+endpoint = "https://solver.cow.fi"


poolpitako (approved):

crates/solvers/config/example.legacy.toml
```diff
@@ -0,0 +1,3 @@
+chain-id = "1"
+solver-name = "CoW Solver"
+endpoint = "https://solver.cow.fi"
```
fedgiac, line 3: Should we append `/solve` to the URL? This is how it's done [here](https://github.com/cowprotocol/services/a87229fb2ce8062a6ce5a8e22bf12c7bb25bba4e/main/crates/solvers/src/tests/legacy/mod.rs) and tests fails without it.

MartinquaXD (approved):

crates/solvers/config/example.legacy.toml
MartinquaXD: Appending `/solve` should not be needed. The test setup code you are referring to was written when the `DefaultHttpSolverApi` still used a hardcoded `/solve` endpoint. But this is no longer the case so we could drop it. We would just have to also slightly adjust the failing tests so they also no longer send the requests to `<URL>/solve`.

fedgiac (approved):

Issue #1355, opened 2023-03-19
Add Balancer liquidity to Gnosis Chain


sunce86:
Balancer is now deployed on Gnosis Chain: https://cowservices.slack.com/archives/C049P351GKF/p1678454130688709

Contracts are different from ethereum, so we need to import them and implement FactoryIndexing for each.

PR #1376, opened 2023-03-27
Balancer On Gnosis Chain

nlordell:
Fixes https://github.com/cowprotocol/services/issues/1355

This PR adds support for Balancer V2 on Gnosis Chain.

We needed to additionally add indexing for V3 and V4 weighted pools (since only those pools are deployed on Gnosis Chain). They work the same way as V1 as far as reading pool data and generating swap calldata, so it is mostly an exercise in generating new contract bindings with different addresses.

Note that we skipped weighted pools V2 - this is because [they are disabled because of a vulnerability](https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345), so it is not really worth indexing.

AFAIU, we don't need to change any of our code in order to work with these pool types (as the core math and fetching work in the same way, in fact they appear treated the same in the Balancer subgraph and SOR source).

### Test Plan

Ran the solver on Mainnet and Goerli with the new Weighted pool factories V2 and V3, pools were indexed as expected.

For example, on Gnosis Chain:
```
2023-03-27T07:16:06.104Z  INFO solver::run: using baseline sources baseline_sources=[BalancerV2]
2023-03-27T07:16:06.988Z DEBUG shared::sources::balancer_v2::pool_init: initialized registered pools block=27143948 pools=17
```

- [x] Double-checked with Balancer team that we don't need to do anything for these pool factories.


vkgnosis (approved):

sunce86:
Did you check if we are able to index `composable stable pool` type of pool? Like this one:
https://app.balancer.fi/#/gnosis-chain/pool/0xfedb19ec000d38d92af4b21436870f115db22725000000000000000000000010

I was under impression that this is also a new factory type.

crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -163,6 +165,8 @@ pub struct BalancerPoolFetcher {
 #[clap(rename_all = "verbatim")]
 pub enum BalancerFactoryKind {
     Weighted,
```
MartinquaXD, line 167: Should we also rename `Weighted -> WeightedV1` and `Stable -> StableV1` now?

MartinquaXD (approved):

crates/shared/src/sources/balancer_v2/pool_fetching.rs
nlordell: I kept the names as that is what they are called in the Balancer V2 repo (in order to avoid confusion). I'm also fine with changing it though - no string feelings.

MartinquaXD: I guess it also makes sense to leave it as it is. 👌

nlordell:
@vkgnosis and @MartinquaXD - After talking to the Balancer team, I removed Balancer weighted pools V2 (see PR description) and added V4 (as that is the current version). They also confirmed that there is no real changes to the math.

PR #1377, opened 2023-03-27
Fix code formatting

vkgnosis:
While reviewing https://github.com/cowprotocol/services/pull/1338 I noticed that part of this function wasn't getting formatted. The reason is a very long line with a macro which rustfmt can't handle. In this commit move from `matches!` to `if let` which fixes the issue. This leads to the comment getting line wrapped which breaks the ascii art formatting of it. To fix this I reformat the comment.

### Test Plan

CI, no logic change

### DIFF
--- a/crates/solver/src/settlement.rs
+++ b/crates/solver/src/settlement.rs
@@ -363,52 +363,72 @@ impl Settlement {
         max_settlement_price_deviation: &Ratio<BigInt>,
         tokens_to_satisfy_price_test: &PriceCheckTokens,
     ) -> bool {
-        if matches!(tokens_to_satisfy_price_test, PriceCheckTokens::Tokens(token_list) if token_list.is_empty())
-        {
-            return true;
+        if let PriceCheckTokens::Tokens(token_list) = tokens_to_satisfy_price_test {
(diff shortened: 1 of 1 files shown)


crates/solver/src/settlement.rs
```diff
@@ -398,23 +399,38 @@ impl Settlement {
...
-                // |                     clearing_price_sell_token / clearing_price_buy_token                                 |
-                // is bigger than:
                 // max_settlement_price_deviation
                 //
-                // This is equal to: |clearing_price_sell_token * external_price_buy_token - external_price_sell_token * clearing_price_buy_token|>
-                // max_settlement_price_deviation * clearing_price_buy_token * external_price_buy_token * clearing_price_sell_token
+                // This is equal to:
+                //
+                // clearing_price_sell_token * external_price_buy_token -
+                // external_price_sell_token * clearing_price_buy_token
```
MartinquaXD, line 418: The original comment took the absolute value of this.

vkgnosis: fixed

MartinquaXD (approved):

nlordell (approved):

PR #1380, opened 2023-03-27
Add Support for RFQ-T to 0x Solver

nlordell:
This PR adds support for RFQ-T liquidity to the 0x solver.

Specifically, it sets `takerAddress` and `intentOnFilling=true` for 0x quotes. We had originally set `intentOnFilling=false` as it made sense given the documentation, however the 0x team let us know that we can use this value to be `true` (and we just might have PMMs penalize us, which just means they won't give us RFQ quotes anymore - and not that they will give us invalid RFQ quotes). `takerAddress` is needed for RFQ-T support.

`enableSlippageProtection` flag was also added at the 0x team's suggestion (and it made sense to add along with the existing changes).

Note that I added this behind a feature flag so we can easily turn it off-and-on depending on whether or not we run into any issues.

For additional context: <https://cowservices.slack.com/archives/C0360881AHZ/p1677881945829199>

### Reviewer Notes

Since this implements this feature for both the co-located and legacy 0x solver, I split the PR into two commits to make reviewing easier.

### Test Plan

Adjusted tests.

Manual test with colocated driver + solver:
```
% curl -s http://localhost:11088/solver/quote -H 'Content-Type: application/json' --data '@-' <<JSON | jq
{
  "sellToken": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
  "buyToken": "0x6b175474e89094c44da98b954eedeac495271d0f",
  "kind": "sell",
  "effectiveGasPrice": "15000000000",
  "deadline": "2024-03-25T00:00:00.000Z"
}
JSON
{
  "amount": "1770171997678222400000",
  "interactions": [
    {
      "value": "0",
      "callData": "0x095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25eff0000000000000000000000000000000000000000000000000000000000000000"
    },
    {
      "value": "0",
      "callData": "0x095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25effffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
    },
    {
      "target": "0xdef1c0ded9bec7f1a1670819833240f027b25eff",
      "value": "0",
      "callData": "0x6af479b200000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000de0b6b3a76400000000000000000000000000000000000000000000000000565d76d7cdbaa3d1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002bc02aaa39b223fe8d0a0e5c4f27ead9083c756cc20001f46b175474e89094c44da98b954eedeac495271d0f000000000000000000000000000000000000000000869584cd0000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab4100000000000000000000000000000000000000000000004b1e5351ad6421897a"
    }
  ]
}
```

### DIFF
--- a/crates/shared/src/trade_finding/zeroex.rs
+++ b/crates/shared/src/trade_finding/zeroex.rs
@@ -58,7 +58,9 @@ impl Inner {
                 sell_amount,
                 buy_amount,
                 slippage_percentage: None,
+                taker_address: None,
                 excluded_sources: self.excluded_sources.clone(),
+                intent_on_filling: false,
                 enable_slippage_protection: false,
             })
             .await?;
--- a/crates/shared/src/zeroex_api.rs
+++ b/crates/shared/src/zeroex_api.rs
@@ -60,8 +60,12 @@ pub struct SwapQuery {
     pub buy_amount: Option<U256>,
     /// Limit of price slippage you are willing to accept.
     pub slippage_percentage: Option<Slippage>,
+    /// The taker address to use.
+    pub taker_address: Option<H160>,
     /// List of sources to exclude.
     pub excluded_sources: Vec<String>,
+    /// Wether or not the taker intends on filling the quote.
+    pub intent_on_filling: bool,
     /// Requests trade routes which aim to protect against high slippage and MEV
     /// attacks.
     pub enable_slippage_protection: bool,
@@ -78,6 +82,7 @@ impl SwapQuery {
         url.query_pairs_mut()
             .append_pair("sellToken", &addr2str(self.sell_token))
             .append_pair("buyToken", &addr2str(self.buy_token))
+            .append_pair("intentOnFilling", &self.intent_on_filling.to_string())
             .append_pair(
                 "enableSlippageProtection",
                 &self.enable_slippage_protection.to_string(),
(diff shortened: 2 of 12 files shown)


vkgnosis (approved):

crates/shared/src/trade_finding/zeroex.rs
```diff
@@ -58,7 +58,9 @@ impl Inner {
                 sell_amount,
                 buy_amount,
                 slippage_percentage: None,
+                taker_address: None,
                 excluded_sources: self.excluded_sources.clone(),
+                intent_on_filling: false,
```
MartinquaXD, line 63: Wasn't this supposed to be `true` or rather behind a flag now?

MartinquaXD: > The PR only changed this for the solvers, not price estimators. I will ask the 0x team if they think it would be a good idea to enable this also for price estimation

Makes sense. 👍 

> That being said, this would lead to slightly different prices between price estimation and solving, so maybe we should?

I think we don't necessarily need to have the same quoting properties for `/solve` and `/quote`.

MartinquaXD (approved):

PR #1387, opened 2023-03-29
Add flag to driver controlling partially fillable orders

vkgnosis:
We had an issue where the driver crashed on a partially fillable order. It makes more sense to control this feature in the driver than in the autopilot.

### Test Plan

compiles, e2e tests still work

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -77,6 +77,8 @@ pub struct Driver {
     logger: DriverLogger,
     web3: Web3,
     last_attempted_settlement: Option<AuctionId>,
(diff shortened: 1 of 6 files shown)


crates/solver/src/driver.rs
```diff
@@ -266,6 +269,12 @@ impl Driver {
             prices: auction.prices.clone(),
         };
 
+        if !self.process_partially_fillable_orders {
+            auction
+                .orders
+                .retain(|order| !order.data.partially_fillable);
```
nlordell, line 275: We also need to check that they are limit orders right? We still want partially fillable liquidity orders AFAIU.

sunce86 (approved):

Issue #136, opened 2022-04-07
Filter out or remove filled partially fillable orders


MartinquaXD:
Currently we allow partially fillable orders in the auction which are already filled.
Because they have been filled completely they have a token amounts of 0 which results in alerts in our solvers.
I think those orders should not be included in the auction because they don't help and only add noise.

Related [Slack thread](https://cowservices.slack.com/archives/C0375NV72SC/p1649069673085119)

PR #1388, opened 2023-03-29
Remove solvable orders zero input balance check

vkgnosis:
Reverts https://github.com/cowprotocol/services/pull/137 which was used to fix https://github.com/cowprotocol/services/issues/136 . While working on partially fillable orders I noticed I did not understand the purpose of this code. 

The mentioned PR adds a check for orders that require no input balance. It justifies that this could happen for partially fillable orders. I don't understand that because:
- We don't allow orders to be created with zero amounts.
- When fetching open orders from the database, we filter out orders that are completely filled.

Thus it is impossible that an order will end up needing 0 input balance at this point. But then why did we think this was the correct fix before? I don't know.

### Test Plan

CI

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -425,8 +425,6 @@ fn solvable_orders(
             // need a way to communicate this to the solver. We could repurpose
             // availableBalance for this.
             let needed_balance = match max_transfer_out_amount(&order) {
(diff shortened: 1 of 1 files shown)


nlordell:
> Thus it is impossible that an order will end up needing 0 input balance at this point.

It can happen if you have, for example, a **buy** order selling 1Xwei for 2Ywei after the order is half filled (as 1Xwei / 2 = 0Xwei in integer math).

PR #1389, opened 2023-03-29
Solver Computed Surplus Fees

### LINKED ISSUE #1292: Add Partially Fillable Limit Order Support to Single Order Solvers
Add support to all `SingleOrderSolver` implementations for partially fillable limit orders.

This requires research on the "best approach" for doing this, so it is a bit of an open-ended task. A naive algorithm would be to binary search on the fill amount and try to find some partial fill that works (limited to `n` iterations for example). However, some DEX aggregator APIs report price impact for a trade (such as 1Inch and 0x), this can be used for more accurately searching for the optimal fill amount for a partially fillable limit order.

nlordell:
Fixes #1292.

This PR adds surplus fee computation to the DEX solvers for partially fillable limit orders. In particular, it also adjusts the limit prices so that the UCP reflects the "actual traded amounts" without the fees.

### Test Plan

Added E2E tests for partially fillable limit order fee logic.

### DIFF
--- a/crates/solvers/src/domain/auction.rs
+++ b/crates/solvers/src/domain/auction.rs
@@ -33,6 +33,17 @@ pub struct Token {
 #[derive(Clone, Copy, Debug)]
 pub struct Price(pub eth::Ether);
 
+impl Price {
+    /// The base Ether amount for pricing.
+    const BASE: u128 = 10_u128.pow(18);
+
+    /// Computes an amount equivalent in value to the specified [`eth::Ether`]
+    /// at the given price.
+    pub fn ether_value(&self, eth: eth::Ether) -> Option<U256> {
+        eth.0.checked_mul(Self::BASE.into())?.checked_div(self.0 .0)
+    }
+}
+
 /// The estimated effective gas price that will likely be used for executing the
 /// settlement transaction.
 #[derive(Clone, Copy, Debug)]
--- a/crates/solvers/src/domain/dex/mod.rs
+++ b/crates/solvers/src/domain/dex/mod.rs
@@ -4,7 +4,7 @@
 
 use {
     crate::{
-        domain::{eth, order, solution},
+        domain::{auction, eth, order, solution},
         util,
     },
     ethereum_types::U256,
@@ -89,6 +89,9 @@ pub struct Swap {
 }
 
 impl Swap {
+    /// An approximation for the overhead of executing a trade in a settlement.
+    const SETTLEMENT_OVERHEAD: u64 = 106_391;
+
     pub fn allowance(&self) -> solution::Allowance {
         solution::Allowance {
             spender: self.allowance.spender.0,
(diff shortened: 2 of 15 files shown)


crates/solvers/src/tests/dex/partial_fill.rs
```diff
@@ -120,12 +120,11 @@ async fn tested_amounts_decrease() {
                 "sellToken": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
                 "buyToken": "0xba100000625a3754423978a60c9317c58a424e3D",
                 "sellAmount": "16000000000000000000",
-                "buyAmount": "3641580551073046209760",
-                // Let's just assume 0 fee to not further complicate the math.
+                "buyAmount": "3630944624685908136768",
```
nlordell, line 123: This value changed for there to be enough room in the limit price to account for the surplus fee (the swap used to match exactly at the limit price).

crates/solvers/src/domain/auction.rs
```diff
@@ -33,6 +33,17 @@ pub struct Token {
 #[derive(Clone, Copy, Debug)]
 pub struct Price(pub eth::Ether);
 
+impl Price {
+    /// The base Ether amount for pricing.
+    const BASE: u128 = 1_000_000_000_000_000_000;
```
vkgnosis, line 38:
```suggestion
    const BASE: u128 = 10u128.pow(18);
```

crates/solvers/src/domain/dex/mod.rs
```diff
@@ -101,20 +104,82 @@ impl Swap {
...
+            Some(
+                sell_price.ether_value(eth::Ether(
+                    self.gas
+                        .0
+                        .checked_add(Self::SETTLEMENT_OVERHEAD.into())?
+                        .checked_mul(gas.0 .0)?,
+                ))?,
+            )
+        } else {
+            None
```
vkgnosis, line 132: In the following code you always use `fee.unwrap_or_default` so it might be easier to make fee non optional immediately by turning this branch into `0`.

crates/solvers/src/domain/dex/mod.rs
```diff
@@ -101,20 +104,82 @@ impl Swap {
...
+                    .min(order.sell.amount);
+                let buy = util::math::div_ceil(
+                    sell.checked_sub(fee.unwrap_or_default())?
+                        .checked_mul(self.output.amount)?,
+                    self.input.amount,
+                )?;
+                (sell, buy)
+            }
+        };
+
```
vkgnosis, line 163: Isn't there a missing check against self.input and self.output here? I'm not that familiar with this code but this struct seems to be an arbitrary on chain swap and we're checking whether it can be used to settle this order. In that case I'd expect a check to make sure that the sell amount from the order (without fees that the contract keeps) is at least the swap's input amount. But this check doesn't seem to exist.

nlordell: Executed `sell` and `buy` amounts are computed using `self.input` and `self.output`. We then do a `order.sell * buy >= order.buy * sell` cross-product check to verify that the swap amounts are verified.

crates/solvers/src/domain/solution.rs
```diff
@@ -34,33 +34,41 @@ pub enum Trade {
 pub struct Fulfillment {
     order: order::Order,
     executed: U256,
+    fee: Option<U256>,
```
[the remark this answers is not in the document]
nlordell: Interesting. The reason I used an `Option` here is to have different branches for orders that can have solver-specified surplus fees, and orders that can't. In particular, we expect order's that can't to not specify a `fee` in their fulfillment.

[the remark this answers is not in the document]
nlordell: I refactored this a bit and added a `solution::Fee` type that models how the fee is charged to the order. It is either a `Protocol` fee, that is, the fee is charged naturally by including the order, or a `Surplus` fee which is an additional surplus fee that the solver can charge.

I used `Surplus` as the name since, from the solver's perspective, the only instance of surplus fees that it chooses are for these kinds of orders. Hope this helps clarify things (and then we have a type that documents what field represents).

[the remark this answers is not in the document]
nlordell: I created a `solution::Fee` type to hopefully more accurately model what the `Option<U256>` was trying to represent.

crates/solvers/src/domain/dex/mod.rs
nlordell: I added an additional E2E test that verifies that an order won't get settled if the input amounts don't satisfy the order's limit price after applying surplus fees.

vkgnosis: For me you don't need to add a test. I'm just confused about the semantics of Swap at this point. Doesn't block the PR.

vkgnosis (approved):

crates/solvers/src/domain/dex/mod.rs
nlordell: 419e957cc4f65ee536e45e39e0fa43bfb5e99e4c

nlordell: > but this struct seems to be an arbitrary on chain swap and we're checking whether it can be used to settle this order.

This is exactly it. It produces a `solution::Solution`, which, inside, contains a `solution::trade::Fulfillment` (i.e. a traded order). In order to create one of these, you need:
- An order
- An executed amount
- A fee strategy (protocol or solver-computed surplus fee).

Since `solution::trade::Fulfillment::new` can fail if any of these are invalid, we implicitly check that the swap amount is correct (since we use the swapped amounts to compute the order's executed amount).

crates/solvers/src/domain/solution.rs
```diff
@@ -34,33 +34,43 @@ pub enum Trade {
...
-        if (!order.partially_fillable && executed != fill)
-            || (order.partially_fillable && executed > fill)
+        let (fill, full) = match order.side {
+            order::Side::Buy => (order.buy.amount, executed),
+            order::Side::Sell => (
+                order.sell.amount,
+                executed.checked_add(fee.surplus().unwrap_or_default())?,
+            ),
+        };
+        if (!order.partially_fillable && full != fill) || (order.partially_fillable && full > fill)
```
MartinquaXD, line 55: Shouldn't this rather say `order.partially_fillable && fill > full` to make sense? But I think it works out because `fill` and `full` got mixed up during initialization.

PR #1390, opened 2023-03-29
Support Partially Fillable Orders in Baseline

nlordell:
In the context of co-location, the Baseline solver is treated differently than DEX aggregator solves (mostly because it is synchronous and fast and doesn't have to deal with many things like rate limiting, etc.).

Because of this, we can have a specialized implementation for partially fillable orders where we attempt multiple swaps at different prices in a single `solve` loop instead of doing it over multiple auctions.

### Test Plan

Added an E2E test that successfully solves a partially fillable limit order with the baseline solver 🎉.

### DIFF
--- a/crates/solvers/src/domain/dex/mod.rs
+++ b/crates/solvers/src/domain/dex/mod.rs
@@ -89,9 +89,6 @@ pub struct Swap {
 }
 
 impl Swap {
-    /// An approximation for the overhead of executing a trade in a settlement.
-    const SETTLEMENT_OVERHEAD: u64 = 106_391;
-
     pub fn allowance(&self) -> solution::Allowance {
         solution::Allowance {
             spender: self.allowance.spender.0,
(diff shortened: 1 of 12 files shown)


crates/solvers/src/domain/dex/mod.rs
```diff
@@ -107,90 +104,28 @@ impl Swap {
     pub fn into_solution(
         self,
         order: order::Order,
-        sell_price: auction::Price,
-        gas: auction::GasPrice,
+        gas_price: auction::GasPrice,
+        sell_token: auction::Price,
     ) -> Option<solution::Solution> {
-        if (order.sell.token, order.buy.token) != (self.input.token, self.output.token) {
```
nlordell, line 113: moved pretty much verbatim to `solution::Single`.

vkgnosis (approved):

crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -27,6 +28,10 @@ pub struct Baseline {
     /// - A value of 2 indicates: `A -> B -> C -> D`
     /// - etc.
     pub max_hops: usize,
+    /// The maximum number of attempts to solve a partially fillable order.
+    /// Basically we continuously halve the amount to execute until we find a
```
MartinquaXD, line 32: We can do the proper binary search for the baseline solver over the simpler approach that was needed when solving across multiple `/solve` requests.
The reason is that the binary search requires the search space to be sorted in order to produce a correct result (or generally make correct progress). This is not the case when we search for the ideal `fill_amount` across multiple `/solve` requests as the ideal `fill_amount` can move as liquidity moves around when more blocks get built. This is not the case for the baseline solver as it always operates on a single block which makes the search space stable and therefore sorted so to speak.
IMO this shouldn't block merging this PR as we are short on time but it's an improvement we should consider in the future.

MartinquaXD (approved):
LGTM. I really like the code that builds the iterator of requests. 👍

crates/solvers/src/domain/solver/baseline.rs
nlordell: > IMO this shouldn't block merging this PR as we are short on time but it's an improvement we should consider in the future.

I though of this, but unfortunately we are trying to optimize for surplus which isn't increasing and decreasing over the search space.

I agree there are other improvements that we can do, but given how the baseline solver also has quite limited liquidity I'm not sure it is worth the effort (without something like an LP book to back it up).

MartinquaXD: > I though of this, but unfortunately we are trying to optimize for surplus which isn't sorted over the search space.

That's true. Didn't think about maximizing surplus. 🤦‍♂️ 
> but given how the baseline solver also has quite limited liquidity I'm not sure it is worth the effort.

💯

PR #1391, opened 2023-03-29
Partially Fillable Order E2E Test

nlordell:
Since partially fillable orders are completely implemented for the co-located solvers, this PR adds an E2E test to verify partially fillable settlement execution works as expected.

I ran into some issues with limit order executed amount computations (I think this is what @MartinquaXD was talking about in #1366), which are now fixed.

I also did some QoL stuff. Notably:
- Added `Debug` implementations to display calldata fields in hex
- Shared `colocation` module for E2E tests

### Test Plan

🥄


crates/e2e/tests/e2e/colocation_partial_fill.rs
```diff
@@ -0,0 +1,98 @@
...
+    let sell_balance = onchain
+        .contracts()
+        .weth
+        .balance_of(trader.address())
+        .call()
+        .await
+        .unwrap();
+    assert_eq!(sell_balance, 1_999_999_999_740_430_040_u128.into());
+    let buy_balance = token.balance_of(trader.address()).call().await.unwrap();
+    assert_eq!(buy_balance, 1_662_497_915_624_478_906_u128.into());
```
vkgnosis, line 94: These asserts might be too exact. They are going to fail if a tiny detail in the solver calculation changes or due to floating point error (if floating point is used anywhere in the process). Should at least use approximate comparison (like for floating point numbers) or could make it even less strict by only asserting that some amount in some large expected range was traded.

vkgnosis (approved):

crates/e2e/tests/e2e/colocation_partial_fill.rs
nlordell: That's a great point. It is sensitive to gas costs as well which might change between hardhat versions. I will change this to use approximate comparisons.

nlordell: cf0a18f7bd551e961f940ce8c368a2ca680d9f0a

PR #1395, opened 2023-03-29
E2E Test With Legacy Solver

nlordell:
Adds an E2E test for partially fillable limit orders.

Luckily, there was a test already setup, so I just tweaked the values a bit to ensure that we always partially fill around half of the order.

### Test Plan

🥄

### DIFF
--- a/crates/e2e/tests/e2e/partially_fillable.rs
+++ b/crates/e2e/tests/e2e/partially_fillable.rs
@@ -70,9 +70,7 @@ async fn test(web3: Web3) {
     );
 
     let services = Services::new(onchain.contracts()).await;
-    services.start_autopilot(vec![
-        "--process-partially-fillable-limit-orders=true".to_string()
-    ]);
+    services.start_autopilot(vec![]);
     services
         .start_api(vec![
             "--allow-placing-partially-fillable-limit-orders=true".to_string()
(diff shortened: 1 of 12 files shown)


crates/e2e/tests/e2e/partially_fillable.rs
```diff
@@ -108,20 +108,15 @@ async fn test(web3: Web3) {
...
-
-    tracing::info!("Waiting for auction to be cleared.");
-    let auction_is_empty = || async { services.get_auction().await.auction.orders.is_empty() };
-    wait_for_condition(TIMEOUT, auction_is_empty).await.unwrap();
-    */
+    // Expecting a partial fill
+    let sell_balance = token_a.balance_of(trader_a.address()).call().await.unwrap();
+    assert_eq!(sell_balance, 249_999_832_941_266_241_166_u128.into());
+    let buy_balance = token_b.balance_of(trader_a.address()).call().await.unwrap();
+    assert_eq!(buy_balance, 199_519_711_827_096_257_754_u128.into());
```
vkgnosis, line 121: Same comment as before about assert_eq on exact values being fickle.

vkgnosis (approved):

crates/e2e/tests/e2e/partially_fillable.rs
nlordell: 9afb7cf581e0a3bb5fca8460397338546fdd5a86

MartinquaXD (approved):

PR #1399, opened 2023-03-30
CIP20: Filter out too high scores

sunce86:
Solver team decided to filter out the scores that are too high (breaking the social consensus).

Whole explanation can be read here https://cowservices.slack.com/archives/C035N0YEB6J/p1680111646093119

### DIFF
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -376,6 +376,18 @@ pub enum SolverRejectionReason {
     /// The solution doesn't have a positive score. Currently this can happen
     /// only if the objective value is negative.
     NonPositiveScore,
+
+    /// The solution has a score that is too high. This can happen if the
+    /// score is higher than the maximum score (surplus + fees)
+    #[serde(rename_all = "camelCase")]
+    TooHighScore {
+        #[serde(with = "u256_decimal")]
+        surplus: U256,
+        #[serde(with = "u256_decimal")]
+        fees: U256,
+        #[serde(with = "u256_decimal")]
+        max_score: U256,
+    },
 }
 
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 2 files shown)


crates/solver/src/settlement_ranker.rs
```diff
@@ -208,6 +210,32 @@ impl SettlementRanker {
             });
         }
 
+        // Filter out settlements with too high score.
+        if Utc::now() > self.auction_rewards_activation_timestamp {
```
MartinquaXD, line 214: I think at this point we no longer have to feature gate CIP-20 stuff, right?

crates/solver/src/settlement_ranker.rs
```diff
@@ -208,6 +210,32 @@ impl SettlementRanker {
...
+            // CIP20 activated
+            rated_settlements.retain(|(solver, settlement, _)| {
+                let surplus = big_rational_to_u256(&settlement.surplus).unwrap_or(U256::MAX);
+                let fees = big_rational_to_u256(&settlement.solver_fees).unwrap_or(U256::MAX);
+                let max_score = surplus.saturating_add(fees);
+                let valid_score = settlement.score.score() < max_score;
```
MartinquaXD, line 220: The max score itself is also valid, no?
```suggestion
                let valid_score = settlement.score.score() <= max_score;
```

crates/shared/src/http_solver/model.rs
```diff
@@ -376,6 +376,18 @@ pub enum SolverRejectionReason {
...
+    /// The solution has a score that is too high. This can happen if the
+    /// score is higher than the maximum score (surplus + fees)
+    #[serde(rename_all = "camelCase")]
+    TooHighScore {
+        #[serde(with = "u256_decimal")]
+        surplus: U256,
+        #[serde(with = "u256_decimal")]
+        fees: U256,
+        #[serde(with = "u256_decimal")]
+        max_score: U256,
```
MartinquaXD, line 389: Should we maybe also add the submitted score here just for easier debugging?

crates/solver/src/settlement_ranker.rs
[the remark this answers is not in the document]
sunce86: @harisang ?

crates/solver/src/settlement_ranker.rs
sunce86: Yes, but wanted to keep it for consistency since the option of reverting the CIP20 changes is still there (at least theoretically :) )

crates/shared/src/http_solver/model.rs
sunce86: That's a nice suggestion.Added.

MartinquaXD (approved):

harisang:
> > Solver team decided to filter out the scores that are too high (breaking the social consensus).
> 
> I don't see this conclusion in the Slack thread.

Attaching a screenshot from CIP-20 that refers to this part.
image

nlordell (approved):

PR #1401, opened 2023-03-30
Better logging for `SingleOrderSolver`

MartinquaXD:
While keeping an eye on the new partial fill support I noticed that it's not super easy to debug.
To address that I added a tracing span per order that gets solved that only contains the `order.id` and the solver name. That should give you a good clue of what's going on.

I also noticed that the component responsible for choosing the partial fill amount had a noisy log. Now that info only gets logged when it matters.

### Test Plan
Compiler, CI

### DIFF
--- a/crates/e2e/tests/e2e/partially_fillable.rs
+++ b/crates/e2e/tests/e2e/partially_fillable.rs
@@ -25,7 +25,7 @@ async fn test(web3: Web3) {
         .deploy_tokens_with_weth_uni_v2_pools(to_wei(1_000), to_wei(1_000))
         .await;
 
-    token_a.mint(trader_a.address(), to_wei(101)).await;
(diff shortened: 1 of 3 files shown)


crates/solver/src/solver/single_order_solver.rs
```diff
@@ -154,48 +212,16 @@ impl Solver for SingleOrderSolver {
 
         let mut settlements = Vec::new();
         let settle = async {
-            let name = self.inner.name();
             while let Some(order) = orders.pop_front() {
-                let Some(fill) = self.fills.order(&order, &auction.external_prices) else {
```
MartinquaXD, line 159: Moved mostly unchanged to `solve_single_order()`. Instead of using `continue` or pushing into collections it now uses `SolveResult` to signal the outcome of the request.

vkgnosis (approved):

PR #1404, opened 2023-03-31
Increase partial fill attempt after success

MartinquaXD:
It's possible that a partially fillable order can only be solved for a small portion with the given liquidity. We'll try lower and lower amounts until we eventually find that fill amount. If then some big trade happens on-chain which lowers the price on a key liquidity pool it could be that a way bigger partial fill is then possible.
In order to find out we need to increase the tried fill amount after a successful attempt. This was originally suggested by @fedgiac [here](https://github.com/cowprotocol/services/pull/1358#pullrequestreview-1354254330).

One unfortunate thing is that this also causes us to flip-flop between finding some solution and finding no solution in case no new liquidity appeared and we just happened to find the biggest possible fill amount. I hope this is not a huge problem, though.

### Test Plan
Adjusted e2e test in the `solvers` crate

### DIFF
--- a/crates/solver/src/solver/single_order_solver.rs
+++ b/crates/solver/src/solver/single_order_solver.rs
@@ -176,7 +176,13 @@ impl SingleOrderSolver {
                 )
                 .transpose()
         }) {
-            Some(Ok(settlement)) => SolveResult::Solved(settlement),
+            Some(Ok(settlement)) => {
+                if let Some(order_uid) = order.id.order_uid() {
+                    // Maybe some liquidity appeared that enables a bigger fill.
+                    self.fills.increase_next_try(order_uid);
+                }
+                SolveResult::Solved(settlement)
+            }
             Some(Err(err)) => {
                 tracing::warn!(%name, ?err, "encoding error");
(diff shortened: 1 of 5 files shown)


vkgnosis (approved):
The logic before this PR is:
- Start at max amount.
- If can't build settlement, lower amount.
- If can build settlement, keep same amount.
- If reach minimum, go back to max amount.

This PR makes the change of increasing the amount by one step if settlement builds. Another alternative would be to go back to the max amount in that case.

The flip flopping is no problem imo. In general it's always going to be the case that there might not be a solution next run and I don't see it as obviously more likely to happen with this strategy. And increasing the amount has the advantage of not getting stuck at lower amounts for a long time, which is inefficient.

fedgiac (approved):
I think the flip flopping isn't so bad. For it to happen, it means that the order either wasn't matched already in the previous run or it was matched but there was a larger amount available. It seems reasonable to expect that the second case is more common, since the only reason I can think of for the order not being matched is that some other solver matched at the same time by coincidence some unrelated orders (well, and the solver component having issues).

Issue #998, opened 2022-12-23
Send `solve` request to SeaSolver as gzip


sunce86:
Compress request body when sending `solve` request to `SeaSolver`.
Response from `SeaSolver` is already gzipped right now (need to double check).

It should be relatively easy to:
1. Set the header `Content-Encoding = gzip`
2. Compress request body

Also, the change should be configurable `per external solver`.

vkgnosis:
We should do this for all solvers. It is unnecessary configuration overhead to keep track of it per solver.  (We might have an issue or a slack thread about this already.) We would document that we are going to send requests gzipped and inform solvers about this. Then we can manually test that they support. Then switch every solver to gzip.

vkgnosis:
This could be more important than expected. Checking the api pod's network stats through `iftop` I see that 100mbit/s upload can spike to quasimodo from price estimates. This can be reduced a lot with some form of compression.

vkgnosis:
My plan for this is the following: Check with all existing http solvers whether they support gzipped request bodies. If they do, then we can unconditionally enable this and make it a requirement for solvers to support this. If they don't, then try to autodetect whether the solver supports it on start up by sending two test requests, one with gzip and one without.

PR #1405, opened 2023-03-31
Support sending http solver solve requests gzipped

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/998 .

This PR doesn't enable compressed requests anywhere, which has to be done as a follow up. One idea for that is to automatically detect whether the http solver supports gzipped requests so that we don't have to configure it.

### Test Plan

adapted unit test to check that the request is gzipped


crates/shared/src/http_solver.rs
```diff
@@ -158,8 +161,6 @@ impl HttpSolverApi for DefaultHttpSolverApi {
                 .append_pair("auction_id", auction_id.to_string().as_str());
         }
         let query = url.query().map(ToString::to_string).unwrap_or_default();
-        let body = serde_json::to_string(&model).context("failed to encode body")?;
-        tracing::trace!(%url, %body, "request");
```
vkgnosis, line 162: Removed because with the new code we don't always have a string to log. We could serialize the json twice so that we can log it but I don't think it's useful enough to keep.

MartinquaXD (approved):
Looks alright to me.

sunce86 (approved):
Is the next step to ask external solvers about their support for gzip?

vkgnosis:
In the issue linked in the first post I write a bit about the plan.

Issue #1378, opened 2023-03-27
Send All Partially Fillable Limit Orders To Solvers


nlordell:
https://github.com/cowprotocol/services/issues/1290#issuecomment-1484755755

Specifically, we currently filter out partially fillable limit orders in the same way we filter out market orders in the auction. That is, the `autopilot` heuristically decides which orders the users have balance for, and only send orders for which users have enough balance to execute the rest of the order.

This is not ideal because:
- If a user has X balance and 3 orders selling X, it isn't clear to the autopilot which is the "best" order to include in the auction, this is the job of the solvers
- If a user has X balance and a partially fillable limit order selling 2X, then it would be possible for half of the order to be matched, but it is currently being disregarded

The proposed change is to:
- [ ] include **all** partially fillable limit orders for which the user has some balance in the auction
- [ ] fetch user balances for partially fillable limit orders to send to the solver engines
- [ ] make sure that Gnosis solvers handle user balances

Related to https://github.com/cowprotocol/services/issues/1064, scoped specifically for partially fillable limit orders.

vkgnosis:
The current description doesn't specify this, but this would make sense for all orders, not just partially fillable orders. Long term we should do it this way for all orders. Short term it is simpler to only have the new behavior for partially fillable orders.

What should we do?

nlordell:
> What should we do?

Yes. However, this issue in particular is **only** for partially fillable limit orders. Eventually we want to do this for all orders, but it is a slightly larger change that solvers need to be aware of.

vkgnosis:
Related https://github.com/cowprotocol/services/issues/1408

vkgnosis:
Quick summary. In https://github.com/cowprotocol/services/pull/1406 we added a simpler version of this change to the non-colocation code. I'm hoping we implement the complete version only for colocation to save time.

Also, After trying to implement it, I feel this change does not became significantly simpler to implement only for partially fillable orders. We should implement it for all order types at the same time. It is unlikely that external solvers will quickly change their code to take balances into account so for some time we are going to need to make this choice for them with a similar algorithm as what we're doing now.

vkgnosis:
Another summary. The current state is that autopilot only filters orders on individual settlability, which is enough sell token balance for sell amount + fee for FOK, and at least 1 balance for partially fillable. The driver does the full balance assignment.

[...]

PR #1406, opened 2023-04-03
Allow using less than full balance for partially fillable orders

vkgnosis:
Related to https://github.com/cowprotocol/services/issues/1378 .

The full implementation of that issue is more complicated than expected. Instead, this PR implements a simple version of it that brings the main benefit. The main benefit is that a partially fillable order, for which the user doesn't have full sell balance, can still be settled.

This simple version continue to let the autopilot make the choice of what partially fillable orders to include and forwards the balance for the order in the auction data so that it can be used by the driver.

This is not yet (properly) implemented for the colocation code because we don't pass the available balance along there.

### Test Plan

adapted unit tests and new e2e test


crates/solver/src/liquidity/order_converter.rs
```diff
@@ -58,12 +54,49 @@ impl OrderConverter {
...
+                .checked_add(remaining.remaining(order.data.fee_amount)?)
+                .context("partially fillable")?;
+            let have = order
+                .metadata
+                .partially_fillable_balance
+                .context("no balance for partially fillable order")?;
+            tracing::trace!(%need, %have, "partially fillable order conversion");
+            if have < need {
+                solver_fee = solver_fee
+                    .checked_mul(have)
```
MartinquaXD, line 77: You could create another `Remaining` instance from `have` and `need` to compute the adjusted amounts. Might look a little nicer.

MartinquaXD (approved):
Don't know what the complications are with the proper solution but this seems like a reasonable intermediate solution.

crates/solver/src/liquidity/order_converter.rs
vkgnosis: The Remaining struct does a similar calculation but it was made for the order executed amounts and for example numberator, denominator are private. I didn't want to generalise it for this so I repeated the calculation.

crates/solver/src/liquidity/order_converter.rs
```diff
@@ -58,12 +54,49 @@ impl OrderConverter {
...
+        let mut solver_fee = remaining.remaining(order.metadata.solver_fee)?;
+        let mut sell_amount = remaining.remaining(sell_amount)?;
+        let mut buy_amount = remaining.remaining(order.data.buy_amount)?;
+
+        // Partially fillable orders are included in the auction when there is at least
+        // 1 atom balance available.
+        if order.data.partially_fillable {
+            let need = sell_amount
+                .checked_add(remaining.remaining(order.data.fee_amount)?)
+                .context("partially fillable")?;
```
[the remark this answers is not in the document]
vkgnosis: Will make it a bit nicer.

vkgnosis:
I'm merging this because it makes testing easier so you can place orders that must be partial matched because missing balance. Reviews still appreciated.

PR #1407, opened 2023-04-03
Store executed surplus fee

### LINKED ISSUE #1403: Store Solver Computed Fees in Database
Similar to how we store the `executed_surplus_fee` for fill-or-kill limit orders, we need to store the solver-computed surplus fee for partially fillable limit orders (I would suggest using the same `executed_surplus_fee` column).

This information is needed:
1. For the UI to show the fee that the user was charged
2. For CIP-20 observed value computation (see https://github.com/cowprotocol/services/issues/1324)

Note that, in theory (just like for FoK limit orders), it can be computed by decoding calldata and comparing the order's clearing price to the uniform clearing price. However, I do think there is value in storing this as a column in the database.

MartinquaXD:
Fixes #1403 

In order to correctly compute the solver rewards for settling partially fillable limit orders we need to store the fee they computed in our DB.
All of the required logic was already in place. We just had to use the `solver_fee` as the `surplus_fee` for partially fillable limit orders in the `Trade` logic.

### Test Plan
Manual test:
* temporarily adjusted test logic to not clear the DB after a test
* ensured the test solver + autopilot are correctly authenticated (to actually store the competition info)
* ran `cargo test -- --ignored partially_fillable` (e2e test that settles a partially fillable order)
* verified that we have an entry in the `order_execution` table with the expected `surplus_fee` amount in the docker testing DB

Extended one existing e2e test to verify that `executed_X_amount`s get updated in the metadata after a partial fill


sunce86 (approved):

vkgnosis:
It would be nice to add the manual test (checking against the DB) to one of the e2e tests.

vkgnosis (approved):

MartinquaXD:
> It would be nice to add the manual test to one of the e2e tests. When the order is settled we can query the api and see that the fee has been set in the response.

I was reluctant to do it as I thought I would have to query the DB but querying the API to notice the updated fields in the metadata is indeed a nicer way to do it. [e878657](https://github.com/cowprotocol/services/pull/1407/commits/e8786571701818dc623007cb988a5377931073ad)

Issue #1409, opened 2023-04-03
Fee scaling for partial fills


fhenneke:
It looks as if fees for partially fillable orders are not computed correctly from the settlement.

The test transaction [here](https://goerli.etherscan.io/tx/0xbd4734d2c727582cf162ee59202d7e75d3fc9d3c99a7c249cd81ac768f94651f) seemed to charge a fee of around 0.027ETH, while the [competition endpoint](https://barn.api.cow.fi/goerli/api/v1/solver_competition/by_tx_hash/0xbd4734d2c727582cf162ee59202d7e75d3fc9d3c99a7c249cd81ac768f94651f) reports a fee of 0.013ETH, i.e., around 50% of the actual fee.

I am not sure if this is a bug or of it still persists.

One potential cause might be the scaling applied to the fee in this function:

https://github.com/cowprotocol/services/blob/7f664e81020bf3e46558cdfb8b923a0170d41cef/crates/autopilot/src/decoded_settlement.rs#L258

nlordell:
AFAICT, this was fixed in https://github.com/cowprotocol/services/pull/1411 and we no longer scale partially fillable order executed fee amounts (instead they get recorded in the DB with the scaled signed `fee_amount` + the unscaled `surplus_fee_amount`).

We should probably confirm that this is indeed fixed before closing the issue.

PR #1411, opened 2023-04-03
Partial fill compatible observed score

### LINKED ISSUE #1324: Compute Observed Value for Partially Fillable Limit Orders
This issue captures the work for computing the observed value of a partially fillable limit order within a solution.

One issue, is that these orders don't have an executed `solverFee` amount (that indicates the exact "fee" that was passed into the solver without subsidies and including the fee "objective scaling factor"). This means that, for now, the "fee" component of the observed value is exactly its executed executed fee.

One issue with this is that partially fillable limit orders are less attractive (the fee amounts aren't artificially scaled to make matching orders more attractive and don't benefit from subsidies).

MartinquaXD:
Fixes #1324, #1409

The original implementation of the observed score assumed all orders were fill-or-kill. With the introduction of partially fillable limit orders this has to change.
Currently the only change in this PR is that instead of repurposing the `full_orders_in_tx()` function we add a new function which is tailor made for the observed score computation. This is required as the old function accumulated all fill amounts of an order into one. That means we didn't really know which fill amount actually happened in the tx we looked at.
The new function starts by looking at `order_executions` in the given tx and joins the remaining data on top of them.

### Test Plan
Added an e2e test to verify that 2 partial fills across 2 auctions result in unique `executed_amounts` and `fees`.


crates/autopilot/src/decoded_settlement.rs
```diff
@@ -251,22 +248,12 @@ fn surplus(
...
-    let fee = match order.kind {
-        model::order::OrderKind::Buy => {
-            solver_fee * u256_to_big_rational(&order.executed_amount)
-                / u256_to_big_rational(&order.buy_amount)
-        }
-        model::order::OrderKind::Sell => {
-            solver_fee * u256_to_big_rational(&order.executed_amount)
-                / u256_to_big_rational(&order.sell_amount)
-        }
-    };
```
MartinquaXD, line 267: `solver_fee` does not have to be scaled anymore since the `OrderExecution` should already contain the scaled value for partial fills.

crates/e2e/tests/e2e/setup/mod.rs
```diff
@@ -95,6 +95,7 @@ where
     let http = create_test_transport(NODE_HOST);
     let web3 = Web3::new(http);
     let resetter = Resetter::new(&web3).await;
+    services::clear_database().await;
```
MartinquaXD, line 98: Also clearing before the test as I noticed during local tests that the DB doesn't get cleared when you abort the test. That caused this particular e2e to then fail indefinitely because the indexing logic couldn't find the left over tx hashes.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -175,16 +175,21 @@ impl DecodedSettlement {
...
-                .find(|order| order.signature == trade.signature.0)
-            {
+            match orders.iter().find(|order| {
+                // I think theoretically it's possible that the same partiallye fillable order
+                // gets filled twice for the same `executed_amount` in the same
+                // auction with different `executed_solver_fee`s but I'm not
+                // sure this can be made unambiguous here and I'm not sure if it
+                // would practically ever happen.
+                order.signature == trade.signature.0
+                    && order.executed_amount == trade.executed_amount
```
vkgnosis, line 191: Is there supposed to be a one to one correspondence of traders and order-executions? I see that below we're warning if an order wasn't found. Should that be an error?
If there is a one to one correspondence then we could remove a used order-executions so that it can't be matched again. The fee is always added so it doesn't matter which one you match first as long as you eventually  match both.

crates/autopilot/src/decoded_settlement.rs
vkgnosis: Should this assumption be documented on the sovler_fee field in OrderExecution? I'm not sure, just saying.

crates/database/src/orders.rs
```diff
@@ -531,6 +531,72 @@ WHERE
...
+    tx_hash: &'a TransactionHash,
+) -> BoxStream<'a, Result<OrderExecution, sqlx::Error>> {
+    const QUERY: &str = const_format::formatcp!(
+        r#"
+-- This will fail if we ever have multiple settlements in the same transaction because this WITH
+-- query will return multiple rows. If we ever want to do this, a tx hash is no longer enough to
+-- uniquely identify a settlement so the "orders for tx hash" route needs to change in some way
+-- like taking block number and log index directly.
+WITH settlement AS (
+    SELECT block_number, log_index
```
vkgnosis, line 561: This query shares a lot of structure with the one it was created from. The `with` block and the `where` block are identical. Might be worth factoring that out with the const string concat we're using for queries in some places. Not sure.

crates/e2e/tests/e2e/partially_fillable_observed_score.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    // Expecting a partial fill because order sells 100 but user only has 50.
+    let sell_balance = token_a
+        .balance_of(trader_a.address())
+        .call()
+        .await
+        .unwrap()
+        .to_f64_lossy();
+    // Depending on how the solver works might not have sold all balance.
+    assert!((0e18..=1e18).contains(&sell_balance));
```
vkgnosis, line 135: In this test you are assuming the exact balance that was sold (50). This happens in the asserts below. To be consistent with that the asserts here (that were copied from the other test) should be changed. Maybe remove them.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: > Is there supposed to be a one to one correspondence of traders and order-executions?

Ideally yes. If this is not the case our DB is missing some data AFAIK.

> I see that below we're warning if an order wasn't found. Should that be an error?

Probably no since it's likely more helpful to have some incomplete data instead of no data due to erroring out.

> If there is a one to one correspondence then we could remove a used order-executions so that it can't be matched again. The fee is always added so it doesn't matter which one you match first as long as you eventually match both.

Sounds good, will do. This might be incorrect in case we have for some reason more trades in our DB than on-chain but l think we probably don't have to worry about that.

crates/database/src/orders.rs
MartinquaXD: Yes. I copied those parts one to one. Will try to refactor it and check what it looks like. 👌

crates/autopilot/src/decoded_settlement.rs
sunce86: Are you sure about this?
Objective value computation still does the scaling here: https://github.com/cowprotocol/services/blob/186fd4f987af034eaef5c54b7b4adf24ef6d82e1/crates/solver/src/settlement.rs#L464

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -128,15 +134,10 @@ impl TryFrom<database::orders::FullOrder> for Order {
             kind: order_kind_from(order.kind),
             sell_token: H160(order.sell_token.0),
             buy_token: H160(order.buy_token.0),
-            sell_amount: big_decimal_to_u256(&order.sell_amount).context("sell_amount")?,
+            sell_amount: big_decimal_to_u256(&order.sell_amount_before_fees)
+                .context("sell_amount")?,
```
sunce86, line 138: Why changed to `sell_amount_before_fees`? Because you renamed it in the database query? And then we again save it to `sell_amount`?

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -180,17 +181,17 @@ impl DecodedSettlement {
...
-                // auction with different `executed_solver_fee`s but I'm not
-                // sure this can be made unambiguous here and I'm not sure if it
-                // would practically ever happen.
+            match orders.iter().enumerate().find(|(i, order)| {
                 order.signature == trade.signature.0
                     && order.executed_amount == trade.executed_amount
+                    // It's possible to have multiple fills with the same `executed_amount` for the
+                    // same order with different `solver_fees`. To end up with the correct total
+                    // fees we can only use every `OrderExecution` exactly once.
+                    && used_indices.insert(*i)
```
vkgnosis, line 192: It feels more natural to build the HashShet with the orders directly and then remove from the hash set when you use one up.

crates/database/src/orders.rs
vkgnosis: I implemented this.

crates/e2e/tests/e2e/partially_fillable_observed_score.rs
vkgnosis: I implemented this.

crates/autopilot/src/decoded_settlement.rs
vkgnosis: I refactored this.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: I think I got confused while writing the custom DB query. The `sell_amount` doesn't get any special treatment so calling it `sell_amount_before_fees` just causes confusion. I think what I actually wanted to do is mention that the `executed_amount` doesn't contain the fees (in case of sell orders).

crates/database/src/orders.rs
```diff
@@ -494,43 +494,100 @@ pub async fn single_full_order(
...
+#[derive(Debug, sqlx::FromRow)]
+pub struct OrderExecution {
+    /// The `solver_fee` that got executed for this specific fill.
+    pub executed_solver_fee: Option<BigDecimal>,
+    pub sell_token: Address,
+    pub buy_token: Address,
+    pub kind: OrderKind,
+    /// The entire `sell_amount` of the order.
+    pub sell_amount: BigDecimal,
+    /// The entire `sell_amount` of the order.
```
vkgnosis, line 549:
```suggestion
    /// The entire `buy_amount` of the order.
```

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -175,18 +175,26 @@ impl DecodedSettlement {
...
         self.trades.iter().fold(0.into(), |acc, trade| {
-            match orders
-                .iter()
-                .find(|order| order.signature == trade.signature.0)
-            {
-                Some(order) => {
-                    acc + match fee(external_prices, order) {
+            match orders.iter().position(|order| {
+                order.signature == trade.signature.0
+                    && order.executed_amount == trade.executed_amount
```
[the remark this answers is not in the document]
MartinquaXD: The underlying assumption was that there should always be a 1:1 relation between trades on-chain and trades in our DB which should make that irrelevant. Maybe it makes sense to issue a warning if the DB list is not empty after the function call as that would mean that we have more trades in our DB than actually happened on-chain.

vkgnosis (approved):

sunce86 (approved):
Lg

PR #1412, opened 2023-04-04
Take partial fill into account for execution_respects_order

vkgnosis:
Felix noticed and reported on Slack that this function had a comment saying it wasn't right for partially fillable orders. This is correct. We forgot to adjust it when implementing partial fill support for single order solvers.

### Test Plan

adapted test

### DIFF
--- a/crates/solver/src/solver/single_order_solver.rs
+++ b/crates/solver/src/solver/single_order_solver.rs
@@ -486,9 +486,13 @@ pub fn execution_respects_order(
     executed_sell_amount: U256,
     executed_buy_amount: U256,
 ) -> bool {
-    // note: This would be different for partially fillable orders but LimitOrder
-    // does currently not contain the remaining fill amount.
-    executed_sell_amount <= order.sell_amount && executed_buy_amount >= order.buy_amount
+    executed_sell_amount <= order.sell_amount
+        && if order.partially_fillable {
+            order.sell_amount.full_mul(executed_buy_amount)
+                >= executed_sell_amount.full_mul(order.buy_amount)
+        } else {
+            executed_buy_amount >= order.buy_amount
+        }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/solver/src/solver/single_order_solver.rs
```diff
@@ -486,9 +484,13 @@ pub fn execution_respects_order(
     executed_sell_amount: U256,
     executed_buy_amount: U256,
 ) -> bool {
-    // note: This would be different for partially fillable orders but LimitOrder
-    // does currently not contain the remaining fill amount.
-    executed_sell_amount <= order.sell_amount && executed_buy_amount >= order.buy_amount
+    executed_sell_amount <= order.sell_amount
```
MartinquaXD, line 487: This is incorrect for fok orders, no?
I believe the correct thing would be to first check that the executed amount makes sense (depending on partial fill and the order side) and then ensure that `provided_price >= limit_price`.

MartinquaXD (commented):
> Felix noticed and reported on Slack that this function had a comment saying it wasn't right for partially fillable orders. This is correct. We forgot to adjust it when implementing partial fill support for single order solvers.

It's true that this function is not compatible with partial fills but when we use a dex solver for partial fills we actually let it solve a synthetic fok order with less volume (in that case it should never fill partially). So unless we start to use `allowPartialFills` in the `1Inch` API it shouldn't make a difference.
I'm not against this patch, though. Just wanted to provide some background.

crates/solver/src/solver/single_order_solver.rs
vkgnosis: Even if an order is marked as FOK, it has no downside to the user to sell less. This is the reasoning with which the function was originally written this way and which this PR doesn't change. If you say FOK sell 10 X for 20 Y then what you care about is getting at least 20 Y in a single trade. You don't care about selling 10 X. So if for some unexpected reason the SingleOrderSolver facilitates the trade with less sell amount then we should take it.

vkgnosis: Although I see now that this might not be right because it means the contract keeps the extra sell amount. (does it?). It would be more correct to sell the full sell amount and return even more buy amount.

vkgnosis:
>  it shouldn't make a difference

Right, I didn't consider that. I still feel it would be good to have this change so it doesn't bite us later when a single order solver becomes smarter.

crates/solver/src/solver/single_order_solver.rs
vkgnosis: We should discuss this in an issue because if this is wrong then it was already wrong. This PR doesn't change it.

MartinquaXD: > Although I see now that this might not be right because it means the contract keeps the extra sell amount. (does it?). It would be more correct to sell the full sell amount and return even more buy amount.

Yes. This was what I was thinking as well.

> Still it is unclear whether that isn't better than not trading at all if this is what the single order solver returned. It does "respect the order".

I think if we could make the even better trade work that would be nice. I just wasn't sure if this would somehow cause problems in the settlement contract because the amounts don't match.

I guess the nice thing would be to allow this better trade as legal while still forwarding the extra money to the user. But I think that could be done in a separate PR.

MartinquaXD (approved):
Approving as this change wouldn't hurt for future proofing purposes.

Issue #1400, opened 2023-03-30
Remove risk-adjusted rewards code


sunce86:
Once the CIP20 is finalised, we should remove all the code for risk adjusted solver rewards which is how rewards were calculated before CIP20 activation.

PR #1420, opened 2023-04-05
Remove CIP14 Risk adjusted rewards

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1400

Reverts https://github.com/cowprotocol/services/pull/627

~I did not remove `reward` from colocated driver structs, since I am still not sure it needs to be removed (because there is some mention of slashing based on the reward value). Anyway, will address this as part of CIP20 implementation for colocated driver.~ Removed `reward` from colocated driver structs also.

I kept the `reward` in order struct used for http solvers communication since it's a breaking change (similar to `is_mature` flag).


database/sql/V050__drop_execution_reward.sql
```diff
@@ -0,0 +1,4 @@
+-- No need to save the CIP14 reward in the database, since it is substituted with CIP20 scheme.
+
+ALTER TABLE order_execution
+DROP COLUMN reward;
```
vkgnosis, line 4: This might come up to validate historical payouts. Let's not delete it yet and ask on slack (`#backend` maybe) what we should do with this data.

vkgnosis (approved):

database/sql/V050__drop_execution_reward.sql
sunce86: Reverted dropping the column after discussion on slack.

PR #1425, opened 2023-04-06
Save competition in autopilot

sunce86:
Before colocation, the data needed for calculation of CIP20 rewards is saved from the driver, using the solver competition endpoint (at the moment just before entering the submission of the winning settlement).

After colocation, since we cannot assume the driver will do this anymore, all the data must be saved from the autopilot. 

Important consequences of this move are:
1. Each driver needs to send a performance/participation address together with the score, so the autopilot knows who are the participants.
2. [TODO MAYBE] Before colocation we knew the winning solution before submission, so we were able to filter out external token prices and save only the prices for tokens used in the winning solution. After colocation, we don't know the winner full call_data before it appears onchain (at least we still did not decide on this), so we need to save the external prices to `auction_prices` table for ALL tokens used in auction, and then filter them out after the transaction is mined.
3. Calculation of fees for partial limit orders - before colocation, those were directly received as part of the solution from the solver. After colocation, the calculation is based on the combination of onchain data and auction data.

### Test Plan
- [x] Fee calculation for a partial limit order equal to the one in solver competition.
Unit test `total_fees_test_partial_limit` confirms this.

### Todos:
- [ ] @nlordell fees for partial limit orders are calculated and saved after 15minutes, and we need order executions to be populated asap after mined transaction, consider reducing the `MAX_REORG_BLOCK_COUNT` or refactoring `OnSettlementEventUpdater` to be reorg safe.
- [ ] Fix failing CI test.


crates/autopilot/src/run_loop.rs
```diff
@@ -101,6 +104,39 @@ impl RunLoop {
...
+                            }
+                            _ => ExecutedFee::Solver(auction_order.metadata.solver_fee),
+                        };
+                        order_executions.push(OrderExecution {
+                            order_id: order.uid,
+                            executed_fee,
+                        });
+                    }
+                    None => {
+                        tracing::error!(?order.uid, "order not found in auction");
```
nlordell, line 136: I don't think this should be an error - this can be a JIT order for example.

crates/autopilot/src/run_loop.rs
```diff
@@ -101,6 +104,39 @@ impl RunLoop {
...
+                                match auction_order.data.partially_fillable {
+                                    // calculated by the solver, therefore we have to trust the
+                                    // driver here
+                                    true => {
+                                        ExecutedFee::Surplus(order.surplus_fee.unwrap_or_default())
+                                    }
+                                    // calculated by the protocol, therefore we can trust the
+                                    // auction
+                                    false => ExecutedFee::Surplus(surplus_fee.unwrap_or_default()),
+                                }
```
nlordell, line 126: I think trusting the solver is not ideal here. Also, the surplus fee can be computed from the settlement transaction, so we don't actually need to trust the driver in this case.

crates/autopilot/src/driver_model.rs
```diff
@@ -138,6 +138,7 @@ pub mod solve {
         pub id: String,
         pub score: U256,
         pub reward: Reward,
+        pub orders: Vec<Order>,
```
nlordell, line 141: Can we just send back a vector of `OrderUid`? It feels weird to send a whole `Order` struct when we make such little use of the data.

crates/autopilot/src/run_loop.rs
nlordell: > for example, when fetching full orders

👍 

> maybe when the settlement event is indexed

Yes, either this or on `Trade` events.

crates/autopilot/src/driver_model.rs
sunce86: Fixed.

crates/autopilot/src/run_loop.rs
sunce86: Set to debug instead of removing since I want to have this check (at least for now)

crates/autopilot/src/run_loop.rs
sunce86: Tried to add it to event indexing and it's super inconvenient. `Trade` event do not contain all necessary data to compute the fee, therefore additional async calls are needed which would slow down the event indexing code (which was the original reason why the `OnSettlementEventUpdater` was initially created). That said, I am thinking about implementing this in `OnSettlementEventUpdater` after all, but reducing the block delay after reading the https://github.com/cowprotocol/services/issues/1492

crates/autopilot/src/run_loop.rs
```diff
@@ -74,14 +79,81 @@ impl RunLoop {
...
+                    Some(auction_order) => {
+                        let executed_fee = match auction_order.metadata.class {
+                            OrderClass::Limit(LimitOrderClass { surplus_fee, .. }) => {
+                                match auction_order.data.partially_fillable {
+                                    // we don't know the surplus fee in advance. will be populated
+                                    // after the transaction containing the order is mined
+                                    true => ExecutedFee::Surplus(None),
+                                    // calculated by the protocol, therefore we can trust the
+                                    // auction
+                                    false => ExecutedFee::Surplus(surplus_fee),
```
sunce86, line 126: Not sure if I should take the `surplus_fee` from the auction or reuse the code from partial limit orders for calculating the `surplus_fee` from the mined transaction. Mostly because the encoding is done in the driver we don't control anymore with colocation.

crates/autopilot/src/database/on_settlement_event_updater.rs
```diff
@@ -60,6 +60,18 @@ impl super::Postgres {
             )
             .await
             .context("insert_settlement_observations")?;
+
+            // delete auction prices for auction_id and all auctions before it
+
+            // TODO dont delete all auctions, instead:
```
sunce86, line 66: Should I do this? I think I should

crates/autopilot/src/run_loop.rs
sunce86: Added the fee calculation to the `OnSettlementEventUpdater`

cowfee:
I see that the driver tests are failing, so merging is blocked. Also you're not following DDD and some docs are missing. Don't worry about this, I'll fix the driver stuff tomorrow (maybe I will open another PR to be merged into this PR). For now if everything is OK from the autopilot side you should be good to go. But please don't merge this yet.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -254,6 +254,68 @@ impl DecodedSettlement {
...
+                        let executed_sell_amount = trade
+                            .executed_amount
+                            .checked_mul(uniform_buy_price)?
+                            .checked_div(uniform_sell_price)?;
+                        adjusted_buy_price.checked_sub(executed_sell_amount)?
+                    }
+                    OrderKind::Sell => {
+                        let sell_amount = adjusted_sell_price
+                            .checked_mul(uniform_buy_price)?
+                            .checked_div(uniform_sell_price)?;
```
MartinquaXD, line 303: It seems odd to me that we can compute the "real" sell amount only with prices.
I mean we currently encode the adjusted prices simply by storing the adjusted token amounts so in that sense `adjusted_sell_price` is equal to `adjusted_buy_amount` but this is not the only way one could encode the prices so people not using the reference driver could do it differently.
In my mind what needs to be done here is something like this:
```rust
let received_buy_amount = trade.executed_amount * adjusted_sell_price / adjusted_buy_price;
let sell_amount_needed_with_ucp = received_buy_amount * uniform_buy_price / uniform_sell_price;
let fee = trade.executed_amount - sell_amount_needed_with_ucp;
```
These price conversions are tricky so I might have gotten it wrong. Maybe @harisang can check these conversion?

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -254,6 +254,68 @@ impl DecodedSettlement {
...
+                let adjusted_buy_price = self.clearing_prices.get(buy_index)?;
+
+                // the logic is opposite to the code in function `custom_price_for_limit_order`
+                let fee = match trade.flags.order_kind() {
+                    OrderKind::Buy => {
+                        let executed_sell_amount = trade
+                            .executed_amount
+                            .checked_mul(uniform_buy_price)?
+                            .checked_div(uniform_sell_price)?;
+                        adjusted_buy_price.checked_sub(executed_sell_amount)?
```
MartinquaXD, line 298: Same concern as with the `sell` version.
This would be my suggestion:
```rust
let required_sell_amount = trade.executed_amount * adjusted_buy_price / adjusted_sell_price;
let required_sell_amount_with_ucp = trade.executed_amount * uniform_buy_price / uniform_sell_price;
let fee = required_sell_amount - required_sell_amount_with_ucp;
```

crates/autopilot/src/driver_model.rs
```diff
@@ -112,19 +112,34 @@ pub mod solve {
...
+    pub struct Reward {
+        /// The address to which the CIP20 performance reward will be sent for
+        /// this solution. Usually the solver address used for
+        /// submitting the solution onchain.
+        pub performance_address: H160,
+        /// The address to which the CIP20 participation reward will be sent for
+        /// this solution. Usually the solver address used for
+        /// submitting the solution onchain.
+        pub participation_address: H160,
+    }
```
MartinquaXD, line 134: Do we actually need this info? Don't solvers get payed out to whatever address is configured at the time of payout and not at the time of solution submission?

crates/autopilot/src/run_loop.rs
```diff
@@ -74,14 +79,81 @@ impl RunLoop {
...
         // TODO: Keep going with other solutions until some deadline.
         if let Some((index, solution)) = solutions.pop() {
             // The winner has score 0 so all solutions are empty.
-            if solution.score == 0. {
+            if solution.score == 0.into() {
                 return;
             }
+
+            let auction_id = id;
+            let winner = solution.reward.performance_address;
```
MartinquaXD, line 92: I think this confuses reward payout address with the winning solver address which don't have to be the same.

crates/autopilot/src/run_loop.rs
MartinquaXD: Depends on what `executed_fee` is supposed to mean in this case.
If this is just the value to be used for objective value computations trusting the protocol computed value here is correct. If you are interested in the fee that the user actually ended up paying this value needs to be computed when we index the trade execution.
I believe the former is needed here so this code looks fine to me.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -205,7 +213,13 @@ impl Competition {
 
         let id = settlement.id;
         *self.settlement.lock().unwrap() = Some(settlement);
-        Ok((id, score))
+
+        let reward = solution::Reward {
+            performance_address: self.solver.address(),
+            participation_address: self.solver.address(),
```
MartinquaXD, line 219: I think right now every solver has a cow payout address that is different from the solver address itself.

crates/autopilot/src/database/competition.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[derive(Clone, Default)]
+pub struct Competition {
+    pub auction_id: AuctionId,
+    pub winner: H160,
+    pub winning_score: U256,
+    pub reference_score: U256,
+    /// Addresses to which the CIP20 participation rewards will be payed out.
+    /// Usually the same as the solver addresses.
+    pub participants: HashSet<H160>,
```
MartinquaXD, line 38: I think participants needs to be distinct from the payout address. For example a solver might request to change the address in the middle of the week. If in that case we had a participation reward computation logic that requires at least `n` submissions to qualify for a payout the solver would receive less than they reasonably should.

crates/autopilot/src/database/on_settlement_event_updater.rs
MartinquaXD: I think we probably don't have to do `1`. Prices that don't get deleted because we didn't do `1` will get deleted when we index the next auction so I think having additional code is not worth it.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -205,7 +213,13 @@ impl Competition {
...
+        };
+        let orders = vec![]; // TODO: implement
```
cowfee, line 221: @sunce86 Can you please open an issue and tag it with `colocation` and explain in detail what needs to be done there? Then, reference the issue in this comment as `TODO(#XYZ): calculate the orders` or whatever.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -32,6 +32,7 @@ pub use {interaction::Interaction, settlement::Settlement, trade::Trade};
 #[derive(Debug, Clone)]
 pub struct Solution {
     pub id: Id,
+    pub score: Option<Score>, // TODO remove Option
```
cowfee, line 35: Same here.

crates/autopilot/src/driver_model.rs
sunce86: This was introduced because we need to save the participants for each auction in the `auction_participants` table.
Seemed easier and more general to do it this way (solvers can dynamically decide and change the payout addresses without our action). 
But we could definitely do it differently, like having a map <solver, address> locally in autopilot if we want to remove `Reward` from the driver interface...

crates/autopilot/src/run_loop.rs
sunce86: Makes sense. Are you maybe aware how we get the submission address for solvers under colocation?

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -254,6 +254,68 @@ impl DecodedSettlement {
...
+        let solver_fee = match &order.executed_solver_fee {
+            Some(solver_fee) => u256_to_big_rational(solver_fee),
+            None => {
+                // this should be a partial limit order
+                if !trade.flags.partially_fillable() {
+                    tracing::warn!("missing solver fee for non partial fee order");
+                    return None;
+                }
+
+                // get uniform prices
```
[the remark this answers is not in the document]
sunce86: @nlordell tagging to see how to handle this. Obviously we need these prices and this code heavily depends on the way the current driver does settlement encoding. 
In order to enforce the encoding format without updating the settlement contract, we probably need to do it via CIP as Martin suggested.

crates/autopilot/src/driver_model.rs
sunce86: I substituted `Reward` with `SubmissionAddress` to simplify things. (a lot of comments came from the confusion I introduced with this new field).

crates/autopilot/src/run_loop.rs
sunce86: Same as here: https://github.com/cowprotocol/services/pull/1425#discussion_r1198807045

crates/driver/src/domain/competition/mod.rs
sunce86: Same as here: https://github.com/cowprotocol/services/pull/1425#discussion_r1198807045

crates/autopilot/src/database/competition.rs
sunce86: Same as here: https://github.com/cowprotocol/services/pull/1425#discussion_r1198807045

crates/autopilot/src/database/on_settlement_event_updater.rs
MartinquaXD: > I am not aware of this. Can you show me the code where it happens?

I thought this is what happens in the code right after this comment block. Seems like I misunderstood what's actually happening there.

crates/autopilot/src/decoded_settlement.rs
sunce86: I think you got it right. And for partial limit orders we need to rely on `executed_amount * prices` and not on full amount. Committed the changes. UT passes for `sell` case.

sunce86:
> Also you're not following DDD

@cowfee can you name a few issues and what is expected? I must admit I did not spend much time checking the new architecture and guidelines.

crates/driver/src/domain/competition/mod.rs
sunce86: I didn't know how to get all ids for all orders in the settlement with the highest score (why settlement has multiple solutions?)

crates/driver/src/domain/competition/solution/mod.rs
sunce86: I am not sure score needs to be optional at this specific place. What I was aiming with this is:
1. Between autopilot and driver, score should not be optional!
2. Between driver and solver, score should be optional. As CIP20 states, solvers can provide score discount, which means, in that case, the driver needs to calculate the score=objective_value and reduce it with the score discount. This is again, to support the transition period when we switch to colocation.

crates/autopilot/src/decoded_settlement.rs
sunce86: Commited. Related to https://github.com/cowprotocol/services/pull/1425#discussion_r1198837899

crates/autopilot/src/database/on_settlement_event_updater.rs
```diff
@@ -60,6 +63,35 @@ impl super::Postgres {
...
+            // 1. for current auction, delete all entries containing prices for tokens not
+            // included in the auction
+            // 2. for all auctions before current, if not already filtered, delete all
+            // entries (might need to add a column to auction_prices to indicate
+            // if it was filtered or not)
+            database::auction_prices::delete(&mut ex, auction_data.auction_id)
+                .await
+                .context("delete_before_auction_id")?;
```
nlordell, line 94: It feels weird that we need to delete prices here. Specifically, the driver already communicates which orders are included, so we can instead just store the prices for the orders that are included instead of all of them, then we don't need to delete auctions.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -254,6 +259,115 @@ impl DecodedSettlement {
             }
         })
     }
+
+    /// Returns the list of partial limit orders with their fees
+    pub fn order_executions(
```
nlordell, line 264: naming nit: this only returns partially fillable order executions, I think the name should reflect that.

crates/autopilot/src/decoded_settlement.rs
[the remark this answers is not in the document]
nlordell: I think this assumption is fine. Agree that, once external solvers start submitting their own transactions that we should formalize this as being part of the social consensus rules in the forum.

crates/autopilot/src/database/competition.rs
```diff
@@ -0,0 +1,136 @@
...
+        f.debug_struct("Competition")
+            .field("auction_id", &self.auction_id)
+            .field("winner", &self.winner)
+            .field("winning_score", &self.winning_score)
+            .field("reference_score", &self.reference_score)
+            .field("participants", &self.participants)
+            .field("block_deadline", &self.block_deadline)
+            .finish()
+    }
+}
```
nlordell, line 59: If we filter out prices for which there are orders, this is not needed anymore.

crates/autopilot/src/database/on_settlement_event_updater.rs
nlordell: > then we can save the auction prices before submission and there is no need for this code.

I think there already is no need for this code. We can save prices for all tokens for which there are orders (from the `order_executions` vector).

crates/autopilot/src/decoded_settlement.rs
nlordell: I see, I think this returns the list of order executions to update (since the surplus fee isn't known until there is a transaction on-chain).

crates/autopilot/src/run_loop.rs
```diff
@@ -74,14 +79,81 @@ impl RunLoop {
...
+            let reference_score = solutions
+                .last()
+                .map(|(_, response)| response.score)
+                .unwrap_or_default();
+            let mut participants = solutions
+                .iter()
+                .map(|(_, response)| response.submission_address)
+                .collect::<HashSet<_>>();
+            participants.insert(solution.submission_address); // add winner as participant
+            let prices = auction.prices.clone();
```
nlordell, line 103: Here specifically, I would:

```rust
let mut prices = BTreeMap::new();
```

And then, in the orders loop (in pseudocode):
```rust
prices.insert(auction_order.sell_token, auction.prices[auction_order.sell_token]);
prices.insert(auction_order.buy_token, auction.prices[auction_order.buy_token]);
```

crates/database/src/auction_prices.rs
```diff
@@ -38,6 +38,17 @@ pub async fn fetch(
...
+/// given auction id.
+pub async fn delete(ex: &mut PgTransaction<'_>, auction_id: AuctionId) -> Result<(), sqlx::Error> {
+    const QUERY: &str = "DELETE FROM auction_prices WHERE auction_id <= $1";
+    sqlx::query(QUERY)
+        .bind(auction_id)
+        .execute(&mut *ex)
+        .await?;
+    Ok(())
+}
+
```
nlordell, line 51: Can be removed with my suggestion ☝️

nlordell (commented):
Looks good. Just one potential improvement I still see (which is to store only the auction prices that you actually need).

cowfee:
@nlordell Your latest comment should be addressed in https://github.com/cowprotocol/services/pull/1425/commits/70c173f20e3c3411629b8a9e0010921ff44a4573. I'm not adding prices if they're missing from the original auction, should this be an error instead?

crates/autopilot/src/database/on_settlement_event_updater.rs
MartinquaXD: What's the strategy to deal with solvers that settle other orders than they promised?
For example an `EIP-1271` order could become unsettlable while the solver tries to bring the solution on-chain. And possibly to cut their losses they could try to substitute the order with some other order.
Or alternatively it could make sense for a solver to continue crunching the numbers on an auction it already won hoping that it could improve the solution while it tries to bring a solution on-chain.

I think it's fine to enforce that the solution contains the orders the solver promised but AFAIR we didn't talk about that, yet. Maybe we could open an issue with all the hard requirements we need to enforce for the new API and add to it all the little things we discover while working on the new API.

vkgnosis:
Code looks fine but I will give it another pass. What could be better is the PR description. I feel it doesn't reflect the state of the PR after discussion. That is going to be harder later to understand what happened here. Although I don't know if you Niksha have all the context to properly update the PR description without Dusan.

vkgnosis (approved):
A couple of nit picks that you don't have to fix but could.

crates/autopilot/src/decoded_settlement.rs
nlordell: @cowfee - ping on naming nit.

crates/autopilot/src/run_loop.rs
```diff
@@ -74,14 +83,88 @@ impl RunLoop {
...
+                .collect::<HashSet<_>>();
+            participants.insert(solution.submission_address); // add winner as participant
+
+            let mut prices = BTreeMap::new();
+            let block_deadline = self.current_block.borrow().number
+                + self.submission_deadline
+                + self.additional_deadline_for_rewards;
+            // save order executions for all orders in the solution, except for partial
+            // limit orders. order executions for partial limit orders will be
+            // saved after settling the order onchain.
```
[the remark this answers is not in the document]
nlordell: Also I think the comment is a little misleading - we "save" (i.e. add rows) for partially fillable orders, they just have no surplus fee.

crates/autopilot/src/run_loop.rs
```diff
@@ -74,14 +83,88 @@ impl RunLoop {
...
+                        order_executions.push(OrderExecution {
+                            order_id: *order_id,
+                            executed_fee,
+                        });
+                        if let Some(price) = auction.prices.get(&auction_order.data.sell_token) {
+                            prices.insert(auction_order.data.sell_token, *price);
+                        }
+                        if let Some(price) = auction.prices.get(&auction_order.data.buy_token) {
+                            prices.insert(auction_order.data.buy_token, *price);
+                        }
```
nlordell, line 145: Can we log an `ERROR` if the token is missing? It should really *never* happen, and if it does, something is sufficiently wrong that it should be investigated.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -228,24 +228,43 @@ impl SolverTimeout {
...
 }
 
-impl Score {
-    pub fn zero() -> Self {
-        Self::default()
+impl Risk {
+    // TODO(#1533) Improve the risk merging formula.
+    /// Combine two risk values.
+    pub fn merge(self, other: Risk) -> Self {
+        Self(self.0 + other.0)
```
nlordell, line 267: Might be worth adding a comment why we add them (it causes solvers to under-bid which is less dangerous than over-bidding)

nlordell (approved):
Approving assuming some small nits are addressed (either through code change or justified in thread):
- We add logs for when tokens are missing, as this should never happen)
- Some minor comment and naming nits.

crates/autopilot/src/decoded_settlement.rs
cowfee: https://github.com/cowprotocol/services/pull/1425/commits/95c81c85b5cbd401aa8b983c7844004a67098777

crates/driver/src/domain/competition/solution/mod.rs
cowfee: https://github.com/cowprotocol/services/pull/1425/commits/d13ffb5710b45ba70a0f4f02b2e9aaa728da5b28

crates/autopilot/src/run_loop.rs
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1425/commits/d13ffb5710b45ba70a0f4f02b2e9aaa728da5b28

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -254,6 +259,115 @@ impl DecodedSettlement {
...
+                        match self.fee(external_prices, &order, trade) {
+                            Some(fee) => Some((order.order_uid, fee)),
+                            None => {
+                                tracing::warn!("possible incomplete fee calculation");
+                                None
+                            }
+                        }
+                    }
+                    None => None,
+                }
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1425/commits/d13ffb5710b45ba70a0f4f02b2e9aaa728da5b28

crates/autopilot/src/arguments.rs
```diff
@@ -185,6 +185,18 @@ pub struct Arguments {
     /// Driver base URLs.
     #[clap(long, env, use_value_delimiter = true)]
     pub drivers: Vec<Url>,
+
+    /// The maximum number of blocks to wait for a settlement to appear on
+    /// chain.
+    #[clap(long, env, default_value = "5")]
+    pub submission_deadline: usize,
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1425/commits/d13ffb5710b45ba70a0f4f02b2e9aaa728da5b28

crates/autopilot/src/decoded_settlement.rs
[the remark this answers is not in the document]
harisang: So if i understand correctly, the simplest example for which this is an issue is just one sell limit order being settled. In that case, the solver provides clearing prices for the sell and buy token, but then the driver ends up computing different, "fake" clearing prices so as to be able to apply them in the total amount sent to the contract by the user and not the "after fee" amount that the solver uses to compute its clearing prices.

And the idea here is that the final calldata should not contain the UCP that the solver computes, but only the ones actually used, in order to save on gas. Do I get this right?

I am noting this down, and will think a bit about this.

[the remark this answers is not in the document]
sunce86: To simplify, we want to make sure that we always have this structure in the clearing prices vector:
1. List of UCPs for all orders.
2. After them, list of fake (adjusted) prices only for limit orders.
(in that order)
This is obviously respected right now when we run the driver, but it needs to be respected when external solvers run the driver.

PR #1446, opened 2023-04-20
Start documenting database schema

vkgnosis:
We should have better documentation on the current database schema. In this PR I start a new database readme. The content of the old one isn't database relevant anymore. It contained old design notes when we first planned the database.

Let me know if you're okay with the format of this document.

### Test Plan

n/a

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -30,6 +30,30 @@ This table is used for [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9b
 ### onchain_order_invalidations
 ### onchain_placed_orders
 ### order_execution
+
+Summary:  
+Contains metainformation for trades required to for reward computations that can not be recoverd from the block chain and are not stored in a persistent manner somewhere else.
+
+ Column      | Type    | Nullable | Default
+-------------|---------|----------|--------
+ order\_uid   | bytea   | not null |
+ auction\_id  | bigint  | not null |
+ reward      | double  | not null |
+ surplus\_fee | numeric | nullable |
+ solver\_fee  | numeric | nullable |
+
+Indexes:  
+- "order\_rewards\_pkey" PRIMARY KEY, btree (`order_uid`, `auction_id`)
+  This table was originally called "order\_rewards" and renamed since then but the old index remained
+
+Description:  
+order\_uid: which order this trade execution is related to  
+auction\_id: in which auction this trade was initiated  
+reward: revert adjusted solver rewards, deprecated in favor of [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f)  
+surplus\_fee: dynamic fee computed by the protocol that should get taken from the surplus of a trade, this value only applies and is set for fill-or-kill limit orders.  
+solver\_fee: value that is used for objective value computations. This either contains a fee equal to the execution cost of this trade computed by a solver (only applies to partially fillable limit orders) or the solver\_fee computed by the backend (which may include subsidies; applies to every other order) or it contains.  
+
+
 ### order_quotes
 ### orders
 
--- a/database/README.md
+++ b/database/README.md
@@ -32,13 +32,13 @@ This table is used for [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9b
 ### order_execution
 
 Summary:  
-Contains metainformation for trades required to for reward computations that can not be recoverd from the block chain and are not stored in a persistent manner somewhere else.
+Contains metainformation for trades, required for reward computations that cannot be recovered from the blockchain and are not stored in a persistent manner somewhere else.
 
- Column      | Type    | Nullable | Default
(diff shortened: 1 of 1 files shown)


database/README.md
```diff
@@ -1,14 +1,42 @@
...
+   Column    |  Type  | Nullable | Default
+-------------|--------|----------|---------
+ auction_id  | bigint | not null |
+ participant | bytea  | not null |
+Indexes:
+- "auction_participants_pkey" PRIMARY KEY, btree (auction_id, participant)
+
+This table is used for [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f). It stores which solvers (identified by ethereum address) participated in which auctions (identified by auction id).
+
+TODO: CIP-20 makes no mention of a need to know about all participants. Why do we store this?
```
vkgnosis, line 20: open question

vkgnosis: Is that documented more officially? It isn't part of CIP-20.

MartinquaXD: From the CIP:
```
The protocol commits to spend 20_000_000 COW tokens per year i.e. around 383307 COW per week on rewards. If at the end of an accounting period (currently: a week) the total amount of performance based rewards to be paid to solvers is smaller than the amount committed to by the protocol, the remaining sum is distributed to solver teams which consistently provide solutions. Those rewards will be distributed pro-rate to solvers based on the number of auctions they submitted a valid solution for over the accounting period.
```

rfritsch-ethz:
I think something like this would be really helpful! @vkgnosis 

Unfortunately, I can't help adding more content, since I don't know it :)
But here are some things that are not clear to me, and might be worth adding in the file:

 - why does the auctions table only contain a single row?
 - what is the log_index column, e.g. in the trade and settlement tables?
 - is every trade part of a settlement? if so, how should the tables be joined?
 - fee amounts are in sell token units?
 - what is the owner and receiver of an order? in particular, is one of them always the user address?
 - what is the difference between the buy and sell amounts in the trades and orders tables?
 - why do trades without a corresponding order (same uid) exist?

cowfee (commented):
Looks good. Might make sense to also have some inline documentation along with each column. Of course this might not be needed for all columns, and can be left out for some of them. It could be part of the markdown table which describes the DB table.

bh2smith:
Might also be good to tag some relevant parties that are known to be able to provide feedback on this 

For example the solvers (@harisang , @fhenneke, ... ) and analytics (@bh2smith, @gentrexha, @olgafetisova ) folks.

database/README.md
```diff
@@ -1,14 +1,188 @@
...
+### fixed_bytes_test
+### flyway_schema_history
+### interactions
+### invalidations
+### onchain_order_invalidations
+### onchain_placed_orders
+### order_execution
+
+Summary:  
+Contains metainformation for trades required to for reward computations that can not be recoverd from the block chain and are not stored in a persistent manner somewhere else.
```
harisang, line 35:
```suggestion
Contains metainformation for trades, required for reward computations that cannot be recovered from the blockchain and are not stored in a persistent manner somewhere else.
```

database/README.md
```diff
@@ -1,14 +1,188 @@
...
+Indexes:  
+- "order\_rewards\_pkey" PRIMARY KEY, btree (`order_uid`, `auction_id`)
+  This table was originally called "order\_rewards" and renamed since then but the old index remained
+
+Description:  
+order\_uid: which order this trade execution is related to  
+auction\_id: in which auction this trade was initiated  
+reward: revert adjusted solver rewards, deprecated in favor of [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f)  
+surplus\_fee: dynamic fee computed by the protocol that should get taken from the surplus of a trade, this value only applies and is set for fill-or-kill limit orders.  
+solver\_fee: value that is used for objective value computations. This either contains a fee equal to the execution cost of this trade computed by a solver (only applies to partially fillable limit orders) or the solver\_fee computed by the backend (which may include subsidies; applies to every other order) or it contains.
```
harisang, line 54: Comment seems to not end properly (maybe typo?)

database/README.md
```diff
@@ -1,14 +1,188 @@
...
+ owner: where the sell\_token will be taken from  
+ creation\_timestamp: when the order was created  
+ sell\_token: address of the token that will be sold  
+ buy\_token: address of the token that will be bought  
+ sell\_amount: amount in sell\_token that should at most be sold  
+ buy\_amount: amount of buy\_token that should at least be bought  
+ valid\_to: point in time when the order can no longer be settled  
+ fee\_amount: amount in sell\_token the owner agreed upfront as a fee to be taken for the trade  
+ kind: whether the order is a buy or a sell order  
+ partially\_fillable: determines if the order can be executed in multiple smaller trades or if it everything has to be executed at once
```
harisang, line 141:
```suggestion
 partially\_fillable: determines if the order can be executed in multiple smaller trades or if everything has to be executed at once  
```

PR #1448, opened 2023-04-20
Silence alert for dust orders

nlordell:
#1439 introduced alert noise when encountering partially fillable orders that scale to 0 amounts.

This PR modifies the auction building process to filter out these orders before hand (there is no point in including them in the auction if they can't be executed anyway).

Note that this is achieved by modifying the existing `remaining_amount::Remaining` logic to scale by both remaining execution and available balance. One implementation note is that we **do not** put these in the same ratio, but instead scale in two steps. This is to prevent filtering out large orders with large partial balance, and would affect [existing orders](https://explorer.cow.fi/orders/0xbf386c0e1913c11edfdf6d267ea716b72fc3a3bd6002eef116d1c42b86b17b3a10ccd4136471c7c266a9fc4569622989fb4cab9964b6c2f4?tab=overview)).

Note that one other approach would be to use a `BigRational` instead of `Ratio<U256>`. The reason we don't do this is the same for why were using `Ratio<U256>` for scaling for remaining execution in the first place: the smart contract scales amounts by this ratio when computing executed amounts, so we could potentially revert - this helps guard against those cases.

### Test Plan

Added some new unit tests to:
- Verify the balance scaling logic for both partially fillable and fill-or-kill orders
- Added a test to document why we split `execution` scaling and `balance` scaling into two separate ratios.

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -209,11 +209,11 @@ impl SolvableOrdersCache {
             new_balances.insert(query, balance);
         }
 
-        let orders = filter_dust_orders(orders, &new_balances);
-        counter.checkpoint("dust_order", &orders);
-
         let orders = orders_with_balance(orders, &new_balances, self.ethflow_contract_address);
         counter.checkpoint("insufficient_balance", &orders);
+
+        let orders = filter_dust_orders(orders, &new_balances, self.ethflow_contract_address);
+        counter.checkpoint("dust_order", &orders);
 
         // create auction
         let (orders, prices) = get_orders_with_native_prices(
@@ -397,18 +397,28 @@ fn orders_with_balance(
 
 /// Filters out dust orders i.e. partially fillable orders that, when scaled
 /// have a 0 buy or sell amount.
-fn filter_dust_orders(mut orders: Vec<Order>, balances: &Balances) -> Vec<Order> {
+fn filter_dust_orders(
+    mut orders: Vec<Order>,
+    balances: &Balances,
+    ethflow_contract: Option<H160>,
+) -> Vec<Order> {
     orders.retain(|order| {
         if !order.data.partially_fillable {
             return true;
         }
 
-        let Some(balance) = balances.get(&Query::from_order(order)) else {
+        let balance = if Some(order.metadata.owner) == ethflow_contract {
+            // For EthFlow orders, assume that there is always enough balance.
+            U256::MAX
+        } else if let Some(balance) = balances.get(&Query::from_order(order)) {
+            *balance
+        } else {
             return false;
         };
+
         let Ok(remaining) = remaining_amounts::Remaining::from_order_with_balance(
             order,
-            *balance
+            balance
         ) else {
             return false;
         };
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -19,6 +19,7 @@ use {
         bad_token::BadTokenDetecting,
         current_block::CurrentBlockStream,
         price_estimation::native_price_cache::CachingNativePriceEstimator,
+        remaining_amounts,
         signature_validator::{SignatureCheck, SignatureValidating},
     },
     std::{
@@ -208,6 +209,9 @@ impl SolvableOrdersCache {
             new_balances.insert(query, balance);
         }
 
+        let orders = filter_dust_orders(orders, &new_balances);
+        counter.checkpoint("dust_order", &orders);
+
         let orders = orders_with_balance(orders, &new_balances, self.ethflow_contract_address);
         counter.checkpoint("insufficient_balance", &orders);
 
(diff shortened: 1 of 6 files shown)


crates/solver/src/liquidity/order_converter.rs
```diff
@@ -69,44 +72,12 @@ impl OrderConverter {
 
         // The reported fee amount that is used for objective computation is the
         // order's full full amount scaled by a constant factor.
-        let mut solver_fee = remaining.remaining(order.metadata.solver_fee)?;
-        let mut sell_amount = remaining.remaining(sell_amount)?;
-        let mut buy_amount = remaining.remaining(order.data.buy_amount)?;
-
-        // Partially fillable orders are included in the auction when there is at least
-        // 1 atom balance available.
-        if order.data.partially_fillable {
```
nlordell, line 78: This logic was moved to `shared::remaining_amounts` module so it can be shared here and with `autopilot::solvable_orders` module.

crates/shared/src/remaining_amounts.rs
```diff
@@ -222,9 +270,109 @@ mod tests {
...
+            assert_eq!(remaining.remaining(800.into()).unwrap(), 200.into());
+        }
+    }
+
+    #[test]
+    fn support_scaling_for_large_orders_with_partial_balance() {
+        let u = <U256 as From<u128>>::from;
+        let order = Order {
+            data: OrderData {
+                sell_amount: u(10).pow(u(30)),
```
MartinquaXD, line 353: This way you don't need `u` at all.
```suggestion
                sell_amount: U256::exp10(30),
```

crates/shared/src/remaining_amounts.rs
```diff
@@ -12,31 +13,18 @@ use {
 /// Works like the smart contract by taking intermediate overflows into account
 /// for partially fillable orders.
 pub struct Remaining {
-    numerator: U256,
-    denominator: U256,
+    execution: Ratio<U256>,
+    balance: Ratio<U256>,
```
MartinquaXD, line 17: Small doc-comments for these fields would be nice.

crates/shared/src/remaining_amounts.rs
```diff
@@ -48,19 +36,80 @@ impl Remaining {
...
+    }
+
+    pub fn zero() -> Ratio<U256> {
+        Ratio::new_raw(0.into(), 1.into())
+    }
+
+    pub fn scalar_mul(ratio: &Ratio<U256>, scalar: U256) -> Option<U256> {
+        scalar
+            .checked_mul(*ratio.numer())
+            .and_then(|product| product.checked_div(*ratio.denom()))
```
MartinquaXD, line 112: I think this `and_then()` could probably be dropped by using `?` instead making the whole thing slightly easier to read.

crates/shared/src/remaining_amounts.rs
```diff
@@ -48,19 +36,80 @@ impl Remaining {
...
-        total
-            .checked_mul(self.numerator)
-            .and_then(|product| product.checked_div(self.denominator))
-            .context("overflow")
+        ratio::scalar_mul(
+            &self.balance,
+            ratio::scalar_mul(&self.execution, total)
+                .context("overflow scaling for order execution")?,
+        )
+        .context("overflow scaling for available balance")
```
MartinquaXD, line 94: The smart contract does not concern itself with the balance of the user for the scaling so I think this should probably not throw on overflows, right?
In that case the scaling based on balance could be done with `Ratio<U512>` and only error if the end result is greater than `U256::MAX`.

crates/shared/src/remaining_amounts.rs
nlordell: Ah right, left-over copy-pasta.

crates/shared/src/remaining_amounts.rs
nlordell: Why do you give me so much work @MartinquaXD  😛. Yes, you are right - this should not error on overflows in intermediate calculations. Will patch.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -391,6 +395,36 @@ fn orders_with_balance(
...
+/// have a 0 buy or sell amount.
+fn filter_dust_orders(mut orders: Vec<Order>, balances: &Balances) -> Vec<Order> {
+    orders.retain(|order| {
+        if !order.data.partially_fillable {
+            return true;
+        }
+
+        let Some(balance) = balances.get(&Query::from_order(order)) else {
+            return false;
+        };
```
vkgnosis, line 408: Doesn't this make the insufficient balance check that we have after the call to this function obsolete? And it doesn't have the special ethflow logic from that check.

nlordell: > Doesn't this make the insufficient balance check that we have after the call to this function obsolete?

No, this just checks that balance scaling doesn't yield a different amount, and only for partially fillable orders. I intentionally wanted to keep the filtering separate so that they are logged as being filtered for different reasons. I changed the filter to be applied _after_ the balance filtering to be more accurate.

> And it doesn't have the special ethflow logic from that check.

Good catch. Fixed.

nlordell:
@vkgnosis and @MartinquaXD - all comments should be addressed now.

MartinquaXD (approved):

vkgnosis (approved):

PR #1455, opened 2023-04-24
Don't require a `dyn Solver` to simulate and rate a Settlement

MartinquaXD:
We recently noticed that the `1Inch` solver does not compute a fee for partially fillable limit orders.
To fix that I want to be able to simulate a settlement in the context of the `SingleOrderSolver`. This will allow us to get the exact gas used for a given settlement and will enable a bunch of other nice things like knowing before settlement merging which solutions work and which are the best. This would allow us to trivially merge the best solutions across all dex solvers to have a single more competitive gnosis solver (which also respects the rules that every solver should only submit a single solution)

Because the `SingleOrderSolver` does not store a `dyn Solver` we can't pass that to the `SettlementRater`.
I saw 2 options:
1. Adjust the `SettlementRater` to not take a `dyn Solver`
2. Build some sort of wrapper or new type that implements the `Solver` trait so I can use the `SettlementRater` as is.

Although `1` would have been very easy it seemed quite hacky so I went with `2`. (Feel free to comment if you disagree with the choice)

### Test Plan
Existing tests, CI

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -383,7 +383,7 @@ impl Driver {
             auction: competition_auction,
             solutions: rated_settlements
                 .iter()
-                .map(|(solver, rated_settlement, _)| SolverSettlement {
+                .map(|(solver, rated_settlement)| SolverSettlement {
                     solver: solver.name().to_string(),
                     solver_address: solver.account().address(),
                     objective: Objective {
@@ -429,7 +429,7 @@ impl Driver {
         };
 
         let mut settlement_transaction_attempted = false;
-        if let Some((winning_solver, winning_settlement, _)) = rated_settlements.pop() {
+        if let Some((winning_solver, winning_settlement)) = rated_settlements.pop() {
             tracing::info!(
                 "winning settlement id {} by solver {}: {:?}",
                 winning_settlement.id,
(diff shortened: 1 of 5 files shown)


crates/solver/src/driver.rs
```diff
@@ -383,7 +383,7 @@ impl Driver {
             auction: competition_auction,
             solutions: rated_settlements
                 .iter()
-                .map(|(solver, rated_settlement, _)| SolverSettlement {
+                .map(|(solver, rated_settlement)| SolverSettlement {
```
nlordell, line 386: Actually kind of surprised that this really wasn't used anywhere except for logging... Nice cleanup 🧹

crates/solver/src/solver.rs
```diff
@@ -175,9 +175,17 @@ pub type Solvers = Vec<Arc<dyn Solver>>;
...
+    /// Identifier used for metrics and logging.
+    pub name: String,
+    /// Address used for simulating settlements of that solver.
+    pub account: Account,
+}
+
 pub struct Simulation {
     pub settlement: Settlement,
-    pub solver: Arc<dyn Solver>,
+    pub solver: SolverInfo,
```
nlordell, line 188: Net improvement IMO.

nlordell (approved):

vkgnosis (approved):

Issue #1428, opened 2023-04-13
Collect more data on who provides the winning price estimate


MartinquaXD:
### General Idea

It would be great to have some sort of estimate about how confident we are that we actually found the best quote for a given trade. If we detect a lot of quotes where we with high confidence didn't return the best price we could issue an alert.

What I'm thinking about would be to let the `CompetitionPriceEstimator` keep track of which estimator wins how often per `(sell_token, buy_token, order_kind)`. If, for example we see that `1Inch` provides the best quote for `(usdc, usdt, sell)` 90% of the time and some rate limiting error prevents it from submitting a price this time we could be 90% confident that we didn't give the user the best price.

If enough of these high confidence missed quotes pile up we could say with high confidence that on average our end users are having a degraded experience.

### Notes

This is just an idea and probably requires a bunch of research. For example it's not clear that `(sell_token, buy_token, order_kind)` is all the data needed to predict the winner. It could be that different estimators perform better for low and high volume trades.

Also if the system works really well we could even consider to not send requests to estimators where we are 95% sure it would lose. There are for example tokens that can only be estimated by a single API (looking at you yearn). For those tokens it doesn't make sense to spam the other estimators.

We should only record the outcome for an estimator if it actually provided a quote (not when it's returning internal server errors, or rate limiting errors).

nlordell:
Very neat idea.

PR #1456, opened 2023-04-24
Competition price estimator metrics

MartinquaXD:
First work on #1428 

Implements the logic for collecting stats on which estimator regularly wins which kind of trade. Also we already start to make predictions (if we have enough data). But since we don't know how accurate our predictions would be we don't act upon them yet and instead only collect metrics. I expect the metrics introduced in this PR are probably not fine grained enough to properly evaluate the idea but it's a first step.

Because the price estimators currently only have their name as an identifier I used that in our stats as well. But because we store possibly every estimator name once per trade we can might use a lot of memory so I decided to only store ref counted Strings instead.

### Test Plan
CI, manual test for metrics (todo)

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -115,11 +115,16 @@ impl RacingCompetitionPriceEstimator {
     ) -> Self {
         assert!(!inner.is_empty());
         Self {
-            competition: (successful_results_for_early_return.get() >= inner.len())
-                .then_some(Default::default()),
             inner,
             successful_results_for_early_return,
+            competition: None,
         }
+    }
+
+    /// Enables predicting the winning quote and gathering of related metrics.
+    pub fn with_preditions(mut self) -> Self {
+        self.competition = Some(Default::default());
+        self
     }
 }
 
@@ -230,6 +235,12 @@ impl CompetitionPriceEstimator {
         Self {
             inner: RacingCompetitionPriceEstimator::new(inner, number_of_estimators),
         }
+    }
+
+    /// Enables predicting the winning quote and gathering of related metrics.
+    pub fn with_preditions(mut self) -> Self {
+        self.inner.competition = Some(Default::default());
+        self
     }
 }
 
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -8,8 +8,95 @@ use {
     },
     futures::stream::StreamExt,
     model::order::OrderKind,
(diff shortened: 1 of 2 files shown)


crates/shared/src/price_estimation/competition.rs
```diff
@@ -79,11 +183,33 @@ impl PriceEstimating for RacingCompetitionPriceEstimator {
...
+                "winning price estimate"
+            );
+
+            // Collect stats for winner predictions.
+            if let (Some(competition), Ok(_)) = (&self.competition, &result) {
+                let trade = Trade::from(query);
+                if let Some(Some(prediction)) = predictions.get(&trade) {
+                    let label = match prediction.winner == estimator {
+                        true => "correct",
+                        false => "incorrect",
```
sunce86, line 200: As you mentioned, more fine grained metric would be useful. But I would add those even in this PR since we will probably get some random result for correctness (let's say 50%), and won't know what to conclude from it. We probably need to know the correctness per trade.

sunce86 (commented):
Nice.

crates/shared/src/price_estimation/competition.rs
MartinquaXD: > We probably need to know the correctness per trade.

Yeah, I was also thinking about this but I was not sure how well grafana would handle this as we would generate a huge amount of metrics because of all the different quote combinations.

MartinquaXD: Added fine grained metrics in [0ffa112](https://github.com/cowprotocol/services/pull/1456/commits/0ffa112f20b102cad763c2d2b8f79a57a3e9bcc1) and in case the amounts of metrics cause problems I also added a CLI flag to control the feature in [13636e7](https://github.com/cowprotocol/services/pull/1456/commits/13636e7eab52649b5df53a86629566fb509d358b).

nlordell (approved):
Neat idea!

Is there a potential for bias in the `HistoricalWinners` in the context of a `RacingCompetitionPriceEstimator` in that estimates that are quick to finish will be over-represented in the counts?

MartinquaXD:
> Is there a potential for bias in the HistoricalWinners in the context of a RacingCompetitionPriceEstimator in that estimates that are quick to finish will be over-represented in the counts?

Yes. That's why I only enabled it in the optimal price estimator for now: [code](https://github.com/cowprotocol/services/pull/1456/files#diff-ec9ccc954f4d59ba8ad726c9586dd802dafae91e98f6a25cdc436b5bf2490ee1R331-R336)

sunce86 (approved):

Issue #1453, opened 2023-04-24
Improve score reporting for Gnosis solvers


harisang:
Given that we have started using MEV Blocker, this means that reverts are much cheaper, as far as I understand. In particular, it seems that we are able to do cheap cancellations and only use `21,000` units of gas. If that is the case, then we should improve the score reporting of Gnosis solvers. In particular, we should switch to the following updated score function:

`score = success_probability * (surplus + fees - simulated_gas * gas_price) - (1 - success_probability) * 21,000 * gas_price`

PR #1458, opened 2023-04-26
Adjust score calculator for reverts

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1453

I thought about putting the change under the configuration parameter, but it seems like it's not needed since we are going to use MEVBlocker permanently. Also, new parameter would need to be coupled somehow with the transactions submission strategy parameters (every time we need to update the submission configuration we would potentially need to also remember to update the configuration for score calculator).

@harisang can you double check that the unit test change is correct? Now the score is slightly increased.


harisang (approved):
Yes, looks correct!

MartinquaXD (approved):

PR #1459, opened 2023-04-27
Allow missing fees on cheap chains

MartinquaXD:
@fleupold requested a hotfix to support an OTC trade on GC. This currently fails because solvers that could facilitate that trade (Naive, Quasimodo) are not participating in the solver competition because their solutions get discarded due to missing fees for partially fillable limit orders.
This PR adds a new CLI argument that either triggers a partially fillable trade without a fee to be discarded or sets the fee to 0.

### Test Plan
compiler

### DIFF
--- a/crates/solver/src/arguments.rs
+++ b/crates/solver/src/arguments.rs
@@ -361,7 +361,7 @@ pub struct Arguments {
     /// filllable limit orders or set the fee to 0. This can make sense on
     /// chains where we are not so concerned about the fee (e.g. gc,
     /// goerli).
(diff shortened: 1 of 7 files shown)


nlordell (approved):
Feels hacky - but fine by me.

MartinquaXD:
> Feels hacky

Absolutely. But I don't see a non-hacky but quick solution. Ideally we would just patch the solvers to compute the fee but that takes time. We are just lucky that `Karpatkey` wants to do the trade on a cheap chain.
Totally fine with reverting this ASAP.

PR #1442, opened 2023-04-19
Add Rate Limiting Support to Single Order Solvers

nlordell:
Follow up to #1310

This PR introduces rate limit handling for the single order solvers. This is introduced as single order solvers occasionally enter "rate limiting spiral of death". Specifically, our orderbook is very large, and in the event we start getting rate limited, we tend to spam every single order to the DEX API as we run through the entire orderbook to try to solve (this is because round-tripping when getting rate limited is very quick).

To combat this, we now have back-off strategies in our single order solvers.

### Test Plan

CI and existing unit tests.

### Release Notes

Configured rate limiting strategy in staging.

PR #1463, opened 2023-05-01
Stop Rate Limited Single Order Solver

nlordell:
In #1442 we introduced rate limiting logic to single order solvers. This was done to allow single order solvers to recover in the face of too many orders (single order solvers were entering a "death spiral", where at some point within an auction, it would get rate limited and continue single order solving for the remaining orders in the book, making the problem even worse).

The fix helped in that we are no longer sending requests to external APIs for some time when rate limited, however, it caused issues with the single order solver success metrics. Specifically, when a solver gets rate limited, it will fail all orders until the rate limit backoff timeout is done. This causes solvers that are being rate limited at some point in an auction to have inflated error numbers (since the single order solve finishes very quickly, and the solver will run through the entire orderbook like before, but without actually making requests to the 3rd party service that is rate limiting us).

This PR changes the single order solver logic to always back-off the remainder of an auction.

### Test Plan

Added new unit test to verify logic.

### DIFF
--- a/crates/solver/src/solver/single_order_solver.rs
+++ b/crates/solver/src/solver/single_order_solver.rs
@@ -204,17 +204,17 @@ impl SingleOrderSolver {
             }
             Err(SettlementError::RateLimited) => {
                 self.metrics.single_order_solver_failed(name);
-                tracing::warn!(%name, "rate limited");
-                return SolveResult::Failed;
+                tracing::warn!("rate limited");
+                return SolveResult::RateLimited;
             }
             Err(SettlementError::Retryable(err)) => {
                 self.metrics.single_order_solver_failed(name);
-                tracing::warn!(%name, ?err, "retryable error");
+                tracing::warn!(?err, "retryable error");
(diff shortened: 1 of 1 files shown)


crates/solver/src/solver/single_order_solver.rs
```diff
@@ -204,17 +204,17 @@ impl SingleOrderSolver {
...
                 self.metrics.single_order_solver_failed(name);
-                tracing::warn!(%name, "rate limited");
-                return SolveResult::Failed;
+                tracing::warn!("rate limited");
+                return SolveResult::RateLimited;
             }
             Err(SettlementError::Retryable(err)) => {
                 self.metrics.single_order_solver_failed(name);
-                tracing::warn!(%name, ?err, "retryable error");
+                tracing::warn!(?err, "retryable error");
```
nlordell, line 212: We already include the solver name in the span, no need to duplicate it here.

vkgnosis (approved):

PR #1466, opened 2023-05-03
Add metrics for dropped quotes

MartinquaXD:
Related to #1428

This PR adds another CLI argument determining how confident we want to be that we provide the best price. This PR still doesn't actually drop any requests. We just simulate how many requests we would actually drop and how often we would still report the best price.
Hopefully this experiment shows that we can drop lots of price estimation requests if we only require a very high confidence of providing the best price. But it's even possible that we could drop a bunch of requests if we set the required confidence to 100%.

In the final implementation we should still add a chance to send requests to solvers we normally not send requests just in case they win a requests every now and again.

### Test Plan
CI

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -163,6 +163,12 @@ pub struct Arguments {
     /// trade and enables metrics accordingly.
     #[clap(long, env, action = clap::ArgAction::Set, default_value = "false")]
     pub enable_quote_predictions: bool,
+
+    /// Determines how confident we want to be that we quote the best price.
+    /// If the value is < 1 we may skip sending quote requests to price
+    /// estimators that perform poorly for the given trade.
+    #[clap(long, env, default_value = "1")]
+    pub quote_prediction_confidence: f64,
 }
 
 impl Display for Arguments {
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation.rs
```diff
@@ -163,6 +163,13 @@ pub struct Arguments {
     /// trade and enables metrics accordingly.
     #[clap(long, env, action = clap::ArgAction::Set, default_value = "false")]
     pub enable_quote_predictions: bool,
+
+    /// Determines how confident we want to be that we quote the best price.
+    /// If the value is < 1 we may skip sending quote requests to price
+    /// estimators that have returned some winning quote in the past for a given
+    /// trade.
```
vkgnosis, line 170: I don't understand this sentence.

vkgnosis (approved):

nlordell (approved):

crates/shared/src/price_estimation.rs
MartinquaXD: I updated the `--help` text to explain what the feature is supposed to do in it's final form which is hopefully clearer. This exact behavior (`>= 1` meaning always asking every estimator) needs to be implemented in a follow up PR to ensure that we can completely disable this optimization in case it causes problems.

PR #1469, opened 2023-05-04
Single order solver simulation

MartinquaXD:
Right now the single order solvers might not produce accurate fees for partially fillable limit orders or overall `gas_used` for the entire settlement. This is due to us using all sorts of heuristics for swaps or trusting the `gas_used` of some dex API responses.
But these can be very inaccurate (e.g. 1Inch reports 0 `gas_used` with the trade verification flag disabled) so it makes sense to rather simulate the settlement and take that gas estimate instead.

This refactoring also makes it so that we can more aggressively merge simulations as we already know before merging which settlements would revert on their own. This will also make it trivial to build a single aggregating solver that merges the best solutions from all the single order solvers.

### Test Plan
CI, manual test


crates/solver/src/solver/zeroex_solver.rs
```diff
@@ -154,57 +153,18 @@ impl SingleOrderSolving for ZeroExSolver {
             .await
             .context("get_approval")?;
 
-        let create_settlement = || {
```
MartinquaXD, line 157: I dropped this logic and its test since the `SettlementRanker` already discards solutions with negative scores.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -405,17 +405,16 @@ impl SettlementEncoder {
         Ok(execution)
     }
 
-    pub fn append_to_execution_plan(&mut self, interaction: impl Interaction + 'static) {
+    pub fn append_to_execution_plan(&mut self, interaction: Arc<dyn Interaction>) {
```
MartinquaXD, line 408: The majority of changes in this diff originates from here which was needed to to make the interactions in the settlement clonable.

vkgnosis: Not directly related to this PR. It feels like we shouldn't store `dyn Interaction` in most cases. Instead we should store `EncodedInteraction`, which is a simple cloneable type. You could consider implementing something like this as a separate PR to make this one simpler.

MartinquaXD: Makes sense. Will look into that.

crates/solver/src/solver/single_order_solver.rs
```diff
@@ -308,14 +346,29 @@ impl Solver for SingleOrderSolver {
         )
         .await;
 
+        let mut settlements: Vec<_> =
+            futures::future::join_all(settlements.into_iter().enumerate().map(
+                |(index, intermediate)| self.finalize_settlement(intermediate, &auction, index),
+            ))
+            .await
```
vkgnosis, line 353: Before this code, we ensure that creating the single order settlements doesn't take too long by making it part of the timeout. This code isn't part of the timeout but it can take a long time because it uses the network. That can make the solver take too long. I would combine the finalizing into the loop above. Can probably be expressed with `Stream`s where we have a pipeline order -> single order settlement -> finalize . When the timeout is hit or the order stream ends  we merge.

crates/solver/src/settlement/settlement_encoder.rs
vkgnosis: This is a problem with existing code so it's not blocking for this PR btw. Just wanted to mention it.

crates/solver/src/solver/single_order_solver.rs
```diff
@@ -274,48 +314,86 @@ impl Solver for SingleOrderSolver {
...
+            loop {
+                tokio::select! {
+                    // if `solve` stops early there are still settlements to be finalized
+                    _ = &mut solve => (),
+                    // all possible settlements have been finalized
+                    _ = &mut finalize => break,
+                    // we reached the timeout and should return the results ASAP
+                    _ = &mut timeout => break,
+                }
+            }
```
vkgnosis, line 387: I would change it like this. Then you can remove the previous borrow checker happy braces so that code is less indented.

```suggestion
        {
            let solve = solve.fuse();
            futures::pin_mut!(solve);
            futures::pin_mut!(timeout);
            futures::pin_mut!(finalize);
            loop {
                tokio::select! {
                    // if `solve` stops early there are still settlements to be finalized
                    _ = &mut solve => (),
                    // all possible settlements have been finalized
                    _ = &mut finalize => break,
                    // we reached the timeout and should return the results ASAP
                    _ = &mut timeout => break,
                }
            }
        }
```

vkgnosis (approved):

crates/solver/src/solver/single_order_solver.rs
MartinquaXD: Neat, thanks! 👍

MartinquaXD:
Merging with 1 review as the manual test looked good and I would like to have this feature in the next release.

crates/solver/src/settlement/settlement_encoder.rs
nlordell: The reason we store `dyn Interaction` is that we need each interaction to be able to decide how it interacts with others. For example WETH interaction "collapsing" logic is part of the `Interaction` trait and can't be done with an `EncodedInteraction`.

Issue #1464, opened 2023-05-03
Return multiple solutions from solvers to the driver


cowfee:
It would be more ergonomic for solvers which use the driver to be able to return multiple solutions when their driver calls `/solve`. The driver would then validate solutions **and filter out** the invalid ones. It would then score the solutions and publish the best-scoring one. This allows some solvers to completely omit logic for validation and scoring. Other solvers might still prefer to include their own validation and/or scoring logic so that they can optimize their solutions.

The API between the driver and solver needs to change. The meaning of "solution ID" and "settlement ID" also needs to change. This is what the flow will look like:

- The driver calls the solver `/solve` endpoint.
- This endpoint returns multiple solutions. Every solution has a solution ID which is unique within the response.
- The driver picks a solution, generates the _settlement ID_ (which we previously referred to as "solution ID," but the term "solution ID" has a slightly different meaning now, due to the bullet point above), and embeds the settlement ID in the settlement.
- If it wins the competition, the driver publishes the settlement.

The score and validity of each solution needs to be logged from the driver. Tenderly simulation links should be included to ease debugging. The solvers can query the autopilot and the driver logs to understand exactly what happened and why.

harisang:
Comment/clarification from an outsider. The above would imply that the driver component is actually the one that is aware what constitutes a "good/competitive" solution, while the solver component is oblivious to that and acts more like a router that attempts to generate feasible solutions.

I am not saying this is good or bad, but just want to understand if we want to think of the driver as being the one evaluating the quality of each generated solution, while the solver just generates candidate executions without being able to internally evaluate which one to choose.

nlordell:
> The above would imply that the driver component is actually the one that is aware what constitutes a "good/competitive" solution

In our design, the driver is aware of what is:
- A valid solution
- How to estimate an observed value similar to how we do now.

I agree that under CIP-20, the second point may be not exactly what we want, but that is a larger change and orthogonal to this feature IMO. With CIP-20 in mind, I would expect the driver to do the same validation it does today (i.e. preventing solvers from foot-guns) and just pick the solution with the highest reported score from the solver engines instead of estimating what the settlement observed value will be (like it does today).

Issue #1468, opened 2023-05-04
Verify and score quotes in the `driver`


cowfee:
The quoting process relies on the solving process. Generating a quote from the driver involves requesting the solver to generate a solution to a "synthetic order," which is a fake order that trades the assets that the user wants to get a quote for.

The solver can return multiple solutions. Before using these solutions for anything, the driver needs to verify that they are valid (i.e. that they can actually execute) and it needs to pick the best one among them (according to some criteria). During the quoting process, verifying is difficult because the synthetic order is fake: it isn't signed and the `from` value is also fake, i.e., it's not guaranteed to have any funds. So this synthetic order, in this form, can't be simulated. OTOH, picking the best quote is easy, since the best quote is simply a valid quote which gives the user the best price.

Therefore, the problem statement comes down to: how do we verify that the solutions generated for quotes are correct? I.e., how do we simulate the solutions that solve synthetic orders?

MartinquaXD:
Quote verification has to happen in the `autopilot` / `orderbook` anyway so we might as well not implement it in the `driver`.

PR #696, opened 2022-11-01
Use Buffered transport everywhere

nlordell:
Partially implement #688.

This PR changes the default `web3::Transport` implementation that is used in the services to be the `transport::Buffered` type. This transport implementation receives JSON RPC requests over a channel and automatically batches them based on runtime configuration parameters. The main motivation for using this are two-fold:
- First, this makes coding of "batched" requests easier (you basically just need to `futures::join` concurrent RPC requests)
- Second, it allows for runtime configuration of maximum batch size (where previously it used a hard-coded `const` value of `100`) as well as number of concurrent batches (which we didn't have previously).

Additionally, I removed manual uses of the `transport::Buffered` in the code since the underlying transport is already buffered, and double-buffering RPC requests doesn't really serve a purpose.

Note that we currently don't specify any configuration parameters to the buffered transport and just use the default of 100 (so it mostly works like it used to).

In follow up PRs I will:
1. Remove the manual batching code of various components (which should simplify a lot of code!)
2. Introduce CLI configuration parameters for configuring the buffered transport

### Test Plan

CI, ran the services locally without issue and saw automatic batching happening:
```
[...]

PR #1471, opened 2023-05-04
`driver`: allow multiple solutions to be returned from the solvers

cowfee:
Fixes https://github.com/cowprotocol/services/issues/1464.

Allow solvers to return multiple solutions to the driver. Some details of how this affects the quoting process still need to be ironed out. This is captured by https://github.com/cowprotocol/services/issues/1468.

There will be a follow-up PR to this one to update the openapi spec.

### Test Plan

Automated tests.


crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -10,201 +10,212 @@ use {
...
-        })
+            })
+            .collect()
     }
 }
 
+#[derive(Debug, Deserialize)]
+pub struct Solutions {
+    solutions: Vec<Solution>,
+}
```
cowfee, line 211: Are we OK with this format? I picked this format because I think it might be easier to change the API in the future if we return an object rather than returning an array directly. But if others disagree I'd be happy to change this to just a plain old JSON array.

cowfee:
Before this can be merged, I also have to update the solvers because our E2E tests are passing. That's mostly a mechanical change, though, so the PR is still OK to review.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -39,79 +40,130 @@ pub struct Competition {
...
+            if solution.is_empty() {
+                tracing::info!(id = ?solution.id, "discarding solution: empty");
+                false
+            } else {
+                true
+            }
+        });
+
+        // Verify the solutions, encoding them into settlements.
+        let settlements = join_all(solutions.map(|solution| async move {
```
vkgnosis, line 88: This unbounded join_all should probably be bounded. Or at least we need to think about what the expected number of returned solutions is. If we get 1000 solutions we don't want to simulate them all concurrently because the node can't handle that.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -39,79 +40,130 @@ pub struct Competition {
...
+            tracing::trace!(id = ?solution.id, "verifying");
+            (
+                solution.id,
+                solution.verify(&self.eth, &self.simulator, auction).await,
```
vkgnosis, line 92: Not directly related to this PR. I notice now that this can be a problem with the auction deadline. We forward the deadline to the solver. When we get the result, we spend extra time verifying the solution(s) and scoring them. This might push us out of the deadline. This is exasperated with more solutions. Not sure what the best way to handle it is.

vkgnosis (approved):

crates/driver/src/domain/competition/mod.rs
cowfee: The deadline is reduced before being sent to the solver for exactly this reason. Perhaps the best option would be to make it so that the amount to reduce by can be specified in the config, and additionally, we should log how long the verification is taking. This way the solvers can see how long verification is taking, and since they know the average number of solution they'll return, they can set the config value accordingly.

crates/driver/src/domain/competition/mod.rs
cowfee: I would test this first.

MartinquaXD: Didn't @nlordell take care of this problem by always using `transport::Buffered` under the hood? ([PR](https://github.com/cowprotocol/services/pull/696))

crates/driver/src/domain/competition/mod.rs
```diff
@@ -39,79 +40,130 @@ pub struct Competition {
...
+                    .tap_err(|err| {
+                        tracing::info!(?err, id = ?settlement.solution.id, "discarding solution: failed scoring")
+                    })
+                    .ok()
+                    .map(|score| (score, settlement))
+            });
+
+        // Trace the scores.
+        let scores = scores.map(|(score, settlement)| {
+            tracing::info!(id = ?settlement.solution.id, score = f64::from(score.clone()), "solution scored");
```
MartinquaXD, line 127: nit: Might be easier to read if it were part of the previous `filter_map`
```rust
.filter_map(|(result, settlement)| match result {
    Ok(score) => {
        tracing::info!(id = ?settlement.solution.id, score = f64::from(score.clone()), "solution scored");
        Some((score, settlement))
    },
    Err(err) => {
        tracing::info!(?err, id = ?settlement.solution.id, "discarding solution: failed scoring");
        None
    }
}
```

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -55,8 +87,8 @@ impl Settlement {
     }
 }
 
-/// A settlement which has been verified to respect some the rules of the
-/// protocol. In particular:
+/// A settlement which has been verified to respect certain fundamental the
```
MartinquaXD, line 90:
```suggestion
/// A settlement which has been verified to respect certain fundamental
```

crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -10,201 +10,212 @@ use {
...
+            .map(|solution| {
+                Ok(competition::Solution {
+                    id: solution.id.into(),
+                    trades: solution
+                        .trades
+                        .into_iter()
+                        .map(|trade| match trade {
+                            Trade::Fulfillment(fulfillment) => {
+                                let order = auction
                             .orders
```
MartinquaXD, line 32: Looks like `cargo fmt` struggled with this block.

crates/driver/src/tests/cases/multiple_solutions.rs
```diff
@@ -0,0 +1,65 @@
...
+        .solution(Solution::Valid)
+        // Additional calldata increases the TX size, so the fees are higher resulting in a lower
+        // score.
+        .solution(Solution::AdditionalCalldata { bytes: 5 })
+        .done()
+        .await;
+
+    let solve = test.solve().await;
+
+    solve.ok().score(-46008923437586.0);
```
MartinquaXD, line 33: Not related to this PR but it's not really possible to tell by reading the test that the order without the additional bytes is picked or that they actually differ in their score.

crates/driver/src/tests/setup/new/mod.rs
```diff
@@ -114,10 +114,22 @@ pub struct Setup {
...
     now: infra::time::Now,
     config_file: Option<PathBuf>,
+    solutions: Vec<Solution>,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum Solution {
+    /// Set up the solver to return a valid solution.
+    Valid,
+    /// Set up the solver to return a valid solution, with additional random
```
MartinquaXD, line 127: nit: The additional data is hardcoded.
```suggestion
    /// Set up the solver to return a valid solution, with additional
```

crates/driver/src/domain/competition/mod.rs
vkgnosis: With the right buffered transport configuration you're right that it will limit the concurrency. Still feels a bit fragile because I had to look that up and it would change if the transport config changes. I'd feel better with a limit here but it isn't that important.

Feel free to create an issue for it and not do it in this PR.

Another thought is that for multiple solutions a streaming approach is more fitting. Then Driver can start simulating immediately without waiting for all solutions and can cut off when the deadline is near. That's out of scope for this PR too but might be nice to have for the future.

MartinquaXD (approved):
Looks good overall. Just a few nits.

crates/driver/src/domain/competition/mod.rs
nlordell: Making this configurable would be a nice-to-have.

cowfee: https://github.com/cowprotocol/services/issues/1474

crates/driver/src/domain/competition/mod.rs
cowfee: https://github.com/cowprotocol/services/issues/1473

Issue #1461, opened 2023-04-27
Prepare http solver auction instance before determining the deadline


MartinquaXD:
The `Yearn` team mentioned that sometimes their solver times out because the deadline is not always constant.
Overall I think it's up to the team to fit inside the time window but I agree that it would be helpful if the time to solve is always constant.
Right now this is not the case because the `HttpSolver` has to build an `Instance` (`instance.json`) before sending it to the solver URLs. This process does not always take the same amount of time and it would probably be good if we could do that before determining the auction deadline.
This would have 2 possible outcomes:
1. if we want to give gnosis solvers (that don't need that additional work) the same amount of time as before the time between settlements will increase by ~1s
2. gnosis solvers would have slightly less time to compute a solution

I think both outcomes are fine because the variance is not so big right now but I wanted to bring it up regardless.

vkgnosis:
I'm unsure if this is worth doing. As you said, solvers need to be able to handle the deadline. They will always need some code to make sure that they return within the deadline even if that means cutting some computation short. Whether the deadline changes per call to the solver or is a hardcoded constant doesn't change this. But it is more robust to use the deadline from the request. If we rely on hardcoding 30 seconds then it would be a huge hassle to change this later. Also with colocation the deadline definitely has some variation.

Gelato-B2B:
Good day. 

My team and I are building a solution on top of CoW - https://docs.google.com/presentation/d/1npKvDYhcinaqmlCT6zjjIW1Ykg66tURY-kebeD0dKMY/edit?usp=sharing 

We believe that there is value in giving solvers a well defined time limit to supply a solution and let them know immediatly who the winner is after the limit is reached. This would  allow for a better way to plug in all sorts of RFQ based systems such as what we are building. 

For example, in our case market makers are asked to provide the best price they can give in response to CoW order. As market makers have to pre commit to a certain price, they will have to add some margin to cover their risks which results in cow users getting worse prices than what they could potentially get. Narrowing the timeframe between a) maker having to pre commit to a certain price and b) maker receiving confirtmation on whether he got to settle the trade  would basically mean better prices for CoW users in the end. 

At least this is how we view it. I wonder, does CoW currently support any type of RFQ based liquidity? If not, is there a general interest in plugging in such liquidity to optimize pathfidning?

MartinquaXD:
Becomes irrelevant with solver-driver colocation. Auction pre-processing will be the responsibility of the solver so we'll no longer be able to pick deadlines with that in mind.

PR #1476, opened 2023-05-05
Move order pre processing into solvers

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/1378 .

Currently we do order pre processing in the driver run loop. There we decide what orders are assigned what balance and normalize them (convert from `model::Order` to  `solver::LimitOrder`). This is in tension with letting solvers do the balance assignment. Previously all solvers expected pre-filtered `LimitOrder`s, now some solvers (external http solvers) expect raw orders with the extra information needed for balance assignment.

This PR moves the pre processing from the driver run loop into the individual solvers. The order type in `solver::Auction` is changed to `model::Order` from `solver::LimitOrder`. The logic is still the same as before (same balance filtering). It has only moved and is now run by every `solver::Solver`. In another PR we can then change the balance filtering for external http solvers.

I also had to change the `SingleOrderSolving` trait slightly. It doesn't make sense to pass the full `solver::Auction` in, so now  we only pass the pieces that were needed (external prices, gas price). This is done in a separate commit in the same PR.

### Test Plan

existing tests still pass

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -8,10 +8,9 @@ use {
         auction_preprocessing,
         driver_logger::DriverLogger,
         in_flight_orders::InFlightOrders,
-        liquidity::order_converter::OrderConverter,
         liquidity_collector::{LiquidityCollecting, LiquidityCollector},
         metrics::SolverMetrics,
-        order_balance_filter::OrderBalanceFilter,
+        order_balance_filter,
         orderbook::OrderBookApi,
         settlement::{PriceCheckTokens, Settlement},
         settlement_ranker::SettlementRanker,
@@ -40,6 +39,7 @@ use {
     num::{rational::Ratio, BigInt, ToPrimitive},
     primitive_types::{H160, U256},
     shared::{
+        account_balances::BalanceFetching,
         current_block::CurrentBlockStream,
         ethrpc::Web3,
         external_prices::ExternalPrices,
@@ -70,7 +70,6 @@ pub struct Driver {
     solution_submitter: SolutionSubmitter,
     run_id: u64,
     api: OrderBookApi,
-    order_converter: Arc<OrderConverter>,
     in_flight_orders: InFlightOrders,
     settlement_ranker: SettlementRanker,
     logger: DriverLogger,
@@ -78,7 +77,7 @@ pub struct Driver {
     last_attempted_settlement: Option<AuctionId>,
     process_partially_fillable_liquidity_orders: bool,
     process_partially_fillable_limit_orders: bool,
-    order_balance_filter: OrderBalanceFilter,
+    balance_fetcher: Arc<dyn BalanceFetching>,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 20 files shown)


crates/solver/src/driver.rs
```diff
@@ -283,52 +280,25 @@ impl Driver {
...
-        let previous_orders: Vec<OrderUid> = auction
-            .orders
-            .iter()
-            .map(|order| order.metadata.uid)
-            .collect();
         let balance_start = Instant::now();
-        let orders = self.order_balance_filter.filter(auction.orders).await;
+        let balances =
+            order_balance_filter::fetch_balances(self.balance_fetcher.as_ref(), &auction.orders)
+                .await;
```
MartinquaXD, line 286: Fetching the balances before determining the auction deadline basically takes care of https://github.com/cowprotocol/services/issues/1461 as well. 👍

MartinquaXD (approved):
Looks alright to me.

crates/solver/src/solver.rs
```diff
@@ -118,7 +120,7 @@ pub struct Auction {
     pub run: u64,
 
     /// The GPv2 orders to match.
-    pub orders: Vec<LimitOrder>,
+    pub orders: Vec<Order>,
```
nlordell, line 123: This is a big change, but I support it (`LimitOrder` being both "liquidity" and "protocol order" was a bit messy IMO).

crates/solver/src/driver.rs
```diff
@@ -283,52 +280,25 @@ impl Driver {
...
         tracing::debug!(
-            "filtering orders based on balance took {}s",
+            "fetching order balances took {}s",
```
nlordell, line 288: I am a bit worried this will take too long in production. We need to fetch something around 1000 balances at the moment. Can you run this locally connected to the production orderbook to verify that there is no issue with this taking too long?

I think this is something that came up before, but I can't remember the reasoning behind having the `solver` re-fetch the balances instead of including them as part of each order in an auction (since the autopilot already fetches these are part of its auction updating loop).

nlordell (commented):
Changes look good. Just have an open question.

Also, this needs to be ported to the co-located driver right? Can we create an issue to track that?

crates/solver/src/driver.rs
vkgnosis: This is not new in this PR. It already runs in production. It came up in the PR that introduced it [here](https://github.com/cowprotocol/services/pull/1434#pullrequestreview-1391989453) . According to the log message it always takes less than 1 sec i prod.

vkgnosis:
> Also, this needs to be ported to the co-located driver right? Can we create an issue to track that?

https://github.com/cowprotocol/services/issues/1408

crates/solver/src/solver.rs
vkgnosis: I agree it's a big change. I give justification for it in the PR description.

nlordell (approved):

crates/solver/src/driver.rs
nlordell: One thing that I just realizes complicates things, is that balance fetching now has to take pre- and post- interactions into account in both the driver and the autopilot. Not a huge deal, but something that is important to remember.

PR #1479, opened 2023-05-06
Additional Autopilot Logs

nlordell:
In debugging timing/race conditions issues with auction building and event indexing (specifically, orders appearing in auctions after trades have already happened) - I added a few new logs to help with this.

Specifically, we now log when we update the `autopilot` solvable orders cache, as well as when we store the new auction in the database to be served by the orderbook API. This should help in debugging the aforementioned race conditions.

### Test Plan

CI.

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -233,10 +233,13 @@ impl SolvableOrdersCache {
         };
         counter.record(&auction.orders);
 
-        if self.store_in_db {
+        let id = if self.store_in_db {
             let id = self.database.replace_current_auction(&auction).await?;
-            tracing::info!(auction = %id, %block, "new auction");
-        }
+            tracing::info!(auction = %id, %block, "stored new auction in database");
+            Some(id)
+        } else {
+            None
+        };
 
         *self.cache.lock().unwrap() = Inner {
             auction: Some(auction),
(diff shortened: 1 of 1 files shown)


crates/autopilot/src/solvable_orders.rs
```diff
@@ -234,7 +234,8 @@ impl SolvableOrdersCache {
         counter.record(&auction.orders);
 
         if self.store_in_db {
-            let _id = self.database.replace_current_auction(&auction).await?;
+            let id = self.database.replace_current_auction(&auction).await?;
+            tracing::info!(auction = %id, %block, "new auction");
```
MartinquaXD, line 238: Without reading up on the code where the log is produced I would not have known that `new auction` only refers to the DB.
Maybe `stored new auction in DB` and `stored new auction in autopilot cache` would be slightly more helpful.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -248,6 +249,7 @@ impl SolvableOrdersCache {
             balances: new_balances,
         };
 
+        tracing::debug!(%block, "updated current cached auction");
```
MartinquaXD, line 252: For improved grep-ability I would also add the auction `id` here.
Then you'll see all the useful information by simply searching for the auction id you are investigating.

vkgnosis (approved):

crates/autopilot/src/solvable_orders.rs
nlordell: Unfortunately, we don't always have the `id` available here. I can log an `Option<ID>` though.

MartinquaXD (approved):

PR #1482, opened 2023-05-08
Read write post interactions in db

### LINKED ISSUE #1481: Read/Write post interactions in the DB
We already store pre-interactions in the DB (for EthFlow orders) but we now also have to store post-interactions.
I see 2 approaches for this:
1. rename `interactions` table to `pre_interactions`
2. introduce a `type` enum (`pre`, `post`) in the current `interactions` table

I will go for `2`.

MartinquaXD:
Fixes #1481

Adds DB migration to allow differentiating pre- and post-interactions.
Adjust some DB functions to work with pre- and post-interactions.
Reads `post_interactions` from DB and puts them into `OrderMetadata` and propagates it to some other places
Note that we now only have the required helper functions to store interactions in the DB but there is still no code path that actually calls that code for post-interactions.

### Test Plan
Adjusted existing tests for slightly changed DB helper functions
Copied test for verifying roundtrip of pre-interactions for post-interactions


cowfee:
OTOH merging into a feature branch can make the code more stale over time and more difficult to merge later. I'd be OK with merging to `main` TBH but I'm also OK with using a feature branch.

crates/database/src/orders.rs
```diff
@@ -404,6 +409,7 @@ pub struct FullOrder {
     pub buy_token_balance: BuyTokenDestination,
     pub presignature_pending: bool,
     pub pre_interactions: Vec<(Address, BigDecimal, Vec<u8>)>,
+    pub post_interactions: Vec<(Address, BigDecimal, Vec<u8>)>,
     pub ethflow_data: Option<(Option<TransactionHash>, i64)>,
```
cowfee, line 413: I know this isn't on you, but it would be really nice to get rid of these tuples and either name these fields or give strong types to the tuple elements. It's hard to tell at a glance what the `BigDecimal` and `i64` are for. And since this is a public signature being able to deduce as much as possible at a glance is important.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -24,9 +24,14 @@ pub struct Order {
     pub kind: Kind,
     pub app_data: AppData,
     pub partial: Partial,
-    /// The onchain calls necessary to fulfill this order. These are set by the
-    /// user and included in the settlement transaction.
-    pub interactions: Vec<eth::Interaction>,
+    /// The onchain calls supposed to run before sending the user funds to the
+    /// settlement contract.
```
cowfee, line 28:
```suggestion
    /// The onchain calls to run before sending user funds to the settlement contract.
```

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -24,9 +24,14 @@ pub struct Order {
...
+    /// These are set by the user and included in the settlement transaction.
+    pub pre_interactions: Vec<eth::Interaction>,
+    /// The onchain calls supposed to run after sending bought tokens from the
+    /// settlement contract to the user.
```
cowfee, line 32:
```suggestion
    /// The onchain calls to run after sending tokens from the settlement contract
    /// to the user.
```

crates/driver/src/domain/quote.rs
```diff
@@ -103,7 +103,9 @@ impl Order {
                 kind: competition::order::Kind::Market,
                 app_data: Default::default(),
                 partial: competition::order::Partial::No,
-                interactions: Default::default(),
+                // TODO add actual pre- and post-interactions
```
cowfee, line 106: I'm not sure what this implies. Should we include any interactions in the quote? As always, I'd really appreciate opening an issue for the TODO with a detailed description of what needs to be done and why (and how, if it applies), unless this TODO is something really simple that you're going to do in a follow-up.

cowfee (changes requested):
This breaks driver tests. I know they're not running in CI which is a problem, I'll add them next thing. But in the meantime please run and fix them manually.

crates/driver/src/domain/quote.rs
MartinquaXD: An order with pre- or post-interactions should result in higher fees than the same order without these interactions as they required gas to run.
That means this somehow has to be factored into the quoting process. However, after bit more thought it's no longer clear to me who should be responsible for accounting for these gas costs. It could be the driver which would require us to pass the interactions as my comment suggests.
But it could also be centralized in the orderbook. The orderbook could estimate the gas costs of the pre- and post-interactions and just add that to the gas estimation of each quote it receives.

cowfee: So the user will specify the pre- and post- interactions when generating the quote?

cowfee: If the users are specifying interactions during quoting (they are, correct?) then this does indeed seem like the right place to implement this.

cowfee: https://github.com/cowprotocol/services/issues/1491

crates/database/src/orders.rs
MartinquaXD: Unless I misunderstood the comment on the big SQL query this unfortunate type is due to limitations in `sqlx`. I created a type alias and attached the related comments to it for now.

vkgnosis (approved):

crates/database/src/orders.rs
nlordell: I _think_ you should be able to use a struct with `derive(sqlx::FromRow)` right? (May be wrong though).

MartinquaXD: This currently doesn't work and there is an open [issue](https://github.com/launchbadge/sqlx/issues/298) about it.

cowfee (approved):

nlordell:
> Would it make sense to merge this PR to a feature branch instead of main?

Personally, I'm not a huge fan of feature branches. I would rather have a feature flag that can be turned on and off at runtime and merge directly into main. In terms of DB migration, I think we can always "undo" DB changes with additional migrations if needed (not to mention that the DB format for this feature seems relatively well specified and unlikely to change - famous last words and all...)

PR #1485, opened 2023-05-09
Encode post interactions

### LINKED ISSUE #1484: Add post interactions to `SettlementEncoder`

MartinquaXD:
Fixes #1484 

In order to actually execute post-interactions the `SettlementEncoder` needs to be aware of them. This is mostly a copy-and-paste job since it already knows about pre-interactions for EthFlow orders.

### Test Plan

Added unit test to verify that merging treats post interactions correctly
Added unit test that verifies that `add_trade()` correctly adds pre- and post-interactions to the settlement.

### DIFF
--- a/crates/solver/src/settlement/settlement_encoder.rs
+++ b/crates/solver/src/settlement/settlement_encoder.rs
@@ -2,7 +2,7 @@ use {
     super::{trade_surplus_in_native_token_with_prices, ExternalPrices, Trade, TradeExecution},
     crate::{encoding::EncodedSettlement, interactions::UnwrapWethInteraction},
     anyhow::{bail, ensure, Context as _, Result},
-    itertools::{Either, Itertools},
+    itertools::Either,
     model::{
         interaction::InteractionData,
         order::{Order, OrderClass, OrderKind},
@@ -597,11 +597,8 @@ impl SettlementEncoder {
             clearing_prices,
             trades,
             interactions: [
-                // In the following it is assumed that all different interactions
-                // are only required once to be executed.
                 self.pre_interactions
                     .into_iter()
-                    .unique()
                     .flat_map(|interaction| interaction.encode())
                     .collect(),
                 iter::empty()
@@ -626,7 +623,6 @@ impl SettlementEncoder {
                     .collect(),
                 self.post_interactions
                     .into_iter()
-                    .unique()
                     .flat_map(|interaction| interaction.encode())
                     .collect(),
             ],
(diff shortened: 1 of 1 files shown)


crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -593,11 +597,8 @@ impl SettlementEncoder {
             clearing_prices,
             trades,
             interactions: [
-                // In the following it is assumed that all different interactions
-                // are only required once to be executed.
```
MartinquaXD, line 597: While writing a unit test for encoding the interactions I stumbled over this assumption. It didn't seem necessary to me so I dropped it but if it's actually required I can adjust my test instead.
Could this be a workaround required for multiple Ethflow orders in a settlement?
cc @fedgiac

fedgiac: It is a workaround: it's going to work without but it's going to be gas inefficient if there are multiple eth-flow orders in the same settlement.
Eth-flow orders need a call to [`wrapAll`](https://etherscan.io/address/0xD02De8Da0B71E1B59489794F423FaBBa2AdC4d93#code) to wrap ETH. However, it's enough to call it once after all traded orders have been created.

MartinquaXD: In that case it might be worthwhile to collapse all `wrapAll` interactions from the EthFlow contract into one and allow duplicates for anything else.
On the other hand there are probably only very few settlements which have multiple EthFlow orders where it would make a difference.

sunce86 (approved):

cowfee (approved):

vkgnosis (approved):

crates/solver/src/settlement/settlement_encoder.rs
nlordell: This is an optimization that applies to EthFlow orders only. In the general sense, pre-interactions aren't "collapsible", so IMO we shouldn't do this in the general case.

Maybe, since this is technically a regression, can we create a GitHub issue to track it? I would prefer changing this code to handle the "general" pre-interaction case.

I agree that there are likely few settlements where this will make a difference, so I don't think this would make a meaningful difference to gas costs. @MartinquaXD - maybe worth asking the Data team during the "office hours" today, how many settlements have more than one EthFlow order settled at the same time. This should give us an exact Ξ value of gas money saved since EthFlow launch.

nlordell (approved):
LGTM.

MartinquaXD:
Merging without the `wrapAll()` optimization since the impact is estimated to be negligible. Opened an issue and a research request for the analytics team so we don't forget about the optimization and can reimplement it in case it actually helps significantly.

PR #1488, opened 2023-05-09
Promote log to debug

sunce86:
I want to remove the temporary `log_filter` configuration we introduced for CIP20 observations: https://github.com/cowprotocol/infrastructure/blob/73a3e20ad270a9af279bd9a09dde08d9ea825831/services/Pulumi.yaml#L19

But before that, let's promote the `OnSettlementEventUpdater ` logs from `trace` to `debug` since I always want to have this log for debugging purposes.

Edit: also improving the logs for score calculator (related to issue with Baseline solver fallback to objective value).

### DIFF
--- a/crates/solver/src/settlement_post_processing/mod.rs
+++ b/crates/solver/src/settlement_post_processing/mod.rs
@@ -146,7 +146,7 @@ impl PostProcessing for PostProcessingPipeline {
                 .await
                 .map_or_else(
                     |err| {
-                        tracing::warn!("Failed to compute score: {}", err);
+                        tracing::warn!(?err, "failed to compute score");
                         None
                     },
                     Some,
--- a/crates/solver/src/settlement_post_processing/mod.rs
+++ b/crates/solver/src/settlement_post_processing/mod.rs
@@ -143,7 +143,14 @@ impl PostProcessing for PostProcessingPipeline {
                     prices,
                     &solver_account.address(),
                 )
-                .await,
+                .await
+                .map_or_else(
+                    |err| {
+                        tracing::warn!("Failed to compute score: {}", err);
+                        None
+                    },
+                    Some,
+                ),
                 ..optimized_solution
             },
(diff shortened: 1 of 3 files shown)


crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -143,7 +143,14 @@ impl PostProcessing for PostProcessingPipeline {
                     prices,
                     &solver_account.address(),
                 )
-                .await,
+                .await
+                .map_or_else(
+                    |err| {
+                        tracing::warn!("Failed to compute score: {}", err);
```
MartinquaXD, line 149: Usually this results in nicer logging.
```suggestion
                        tracing::warn!(?err, "failed to compute score");
```

MartinquaXD (approved):

crates/solver/src/settlement_post_processing/optimize_score.rs
```diff
@@ -42,12 +47,12 @@ pub async fn compute_score(
         .calculate(&inputs, nmb_orders)
         .map(Score::Score);
 
-    tracing::trace!(
+    tracing::debug!(
         "solver {} objective value {}, score {:?}",
         solver,
         inputs.objective_value(),
         score
     );
```
cowfee, line 55: Same here like Martin said above, something like
```suggestion
    tracing::debug!(
        %solver, ?score, objective_value = %inputs.objective_value(),
        "computed score"
    );
```

cowfee (approved):

crates/solver/src/settlement_post_processing/mod.rs
sunce86: fixed

crates/solver/src/settlement_post_processing/optimize_score.rs
sunce86: fixed

Issue #1489, opened 2023-05-09
Unify `settlement::Verified` and `Settlement`


cowfee:
The domain is revealing itself more clearly. It's time to update the code to match it.

An [unverified settlement](https://github.com/cowprotocol/services/blob/cb1f39da9dd43c9e9328b6099a482f49fe5e3b4d/crates/driver/src/domain/competition/solution/settlement.rs#L47) is not useful for anything. The transaction created in the encoding process is [used for simulation](https://github.com/cowprotocol/services/blob/cb1f39da9dd43c9e9328b6099a482f49fe5e3b4d/crates/driver/src/domain/competition/solution/mod.rs#L201-L202) but the `Settlement` abstraction is incoherent and only used as an `inner` field of `settlement::Verified`. Possibly encoding a solution into a tx should be exposed as a function to the `solution` module, and verification should be an invariant of the `Settlement` type.

Seems like a `settlement::Tx` type might be warranted, and the `simulate` method should maybe reside on this type. `settlement::Tx::encode` would create the settlement transaction, which could later be turned into a `Settlement`, after additional validation.

Issue #1477, opened 2023-05-05
Merge solutions in the driver


cowfee:
Since the driver is responsible for scoring and verifying all solutions returned by a solver, the driver should also be responsible for merging together those valid solutions which can be merged. This should be done after the invalid solutions have been filtered out and before the scoring.

The `solution` field should probably be removed from `Settlement` and instead a field such as `solutions: Vec<solution::Id>` should be added. This is because multiple solutions can be encoded into one settlement due to merging, and the settlement doesn't really need to know about the solution data. It only cares about the IDs, and even that is just for observability.

For now, the merging process will rely on the legacy code from the boundary. The work required to refactor this code is defined in https://github.com/cowprotocol/services/issues/1478.

MartinquaXD:
> This should be done after the invalid solutions have been filtered out and before the scoring.

In order to filter out invalid solutions we already have to filter them which gives us scoring for free. Also having a score attached to valid settlements will give us a good idea which settlements might be worth merging.
But of course we still have to ensure that newly created merged settlements are also scored.

cowfee:
> In order to filter out invalid solutions we already have to filter them which gives us scoring for free.

@MartinquaXD Filtering (verifying) does not give scoring for free.

MartinquaXD:
Maybe my wording was bad. I meant that going from an unverified solution to verified solution is a lot of work because we have to do simulations which takes time. But going from a verified solution to a scored solution is basically free as it doesn't even do any heavy logic or network calls.

Issue #1480, opened 2023-05-08
Optimize settlement merging


cowfee:
Settlement merging will first be done by simply merging the settlements in a randomized order. It is possible to improve this process:
- Use a [dynamic programming](https://en.wikipedia.org/wiki/Dynamic_programming) approach, keeping a map of `Set<solution ID> -> final merged settlement>` ([memoization](https://en.wikipedia.org/wiki/Memoization)) and recursively attempting to calculate every valid settlement and then picking the best-scoring one.
- Use a greedy approach where the settlement merging process assumes the following heuristic: it's best to merge the higher-scoring settlements. So, before merging, score all the settlements, and then try to merge them in order of highest score first. Since this is a dumb greedy algorithm, this can be very suboptimal (merging three lower-scoring solutions can be better than merging two higher-scoring solutions, but this algorithm would do the latter). It should be investigated whether or not this yields better results in practice compared to the randomization approach.
- A combination of the above: maybe use dynamic programming if the number of settlements is sufficiently small, and switch to a different approach otherwise.

Issue #1478, opened 2023-05-05
Don't rely on legacy code for solution merging in `driver`


cowfee:
Ideally the `Settlement` type and all of the code related to it should be in the `driver`. This includes the code for merging settlements. The legacy code is [here](https://github.com/cowprotocol/services/blob/753eccd64befe24c44103cf58b39bb47df529f67/crates/solver/src/settlement/settlement_encoder.rs#L628-L696) for reference. Below is a description:

The clearing prices between two solutions A and B should be merged as follows:
- For those tokens which have clearing prices in both A and B, all of the clearing prices in A can be scaled (using the same factor) to give the clearing prices in B (and, of course, vice versa). Meaning, the common clearing prices are equal up to a scaling factor. If this is not the case, the solutions cannot be merged.
- If there is a clearing price in A which isn't in B (or vice versa), then simply use that clearing price in the resulting merged clearing prices set without modification.

The `tokens` field should be removed as business rules should define it to be equal to the tokens in clearing prices.

The `trades` and `unwraps` should be merged as a set union. If `trades` are not disjoint (i.e., the two solutions solve some of the same orders), then the two solutions can't be merged, because it's impossible to know how to merge the onchain interactions correctly.

The idea of merging clearing prices seems (at first glance) to represent an important piece of domain logic.

It's important to note that once the merging code moves into the driver, the merging should be done on the `Solution` type. It's currently done on the `Settlement` type because of the boundary integration.

cowfee:
What's the deal with merging `trades`? In the legacy code, if the settlements being merged have any orders in common, they are considered to be unmergable. This seems incorrect to me. At least, if the trades have the same interactions, the settlements should be mergable? Or is this a minute detail not worth bothering with?

MartinquaXD:
> At least, if the trades have the same interactions, the settlements should be mergable? Or is this a minute detail not worth bothering with?

It's not always given that you know which custom interactions belong to which settlement. Imagine 2 orders that both sell token A for token B. We could have a solution which has 1 interaction per order getting the liquidity from some pool but we could also have a solution that only does a single interaction which acquired enough liquidity for both trades.
This can complicate things.

And a more specific reason why this merging currently doesn't happen in the legacy code is that it only merges solutions of single order solvers (`1Inch, `0x`, ...) meaning that 2 solutions that settle the same order with the same interactions are identical because they can not differ in the execution of another order because only a single order will ever be part of those solutions.

PR #1490, opened 2023-05-09
`driver`: settlement merging

cowfee:
Fixes https://github.com/cowprotocol/services/issues/1477.
Fixes https://github.com/cowprotocol/services/issues/1489.

Implement settlement merging in the driver. I opened a few additional issues to follow up on in the future: https://github.com/cowprotocol/services/issues/1478, https://github.com/cowprotocol/services/issues/1480, https://github.com/cowprotocol/services/issues/1489.

### Test Plan

Automated tests.


crates/driver/src/domain/competition/mod.rs
```diff
@@ -95,36 +97,100 @@ impl Competition {
...
+                {
+                    Ok(m) => {
+                        tracing::debug!(
+                            settlement_1 = ?settlement.solutions(),
+                            settlement_2 = ?other.solutions(),
+                            "merged settlements"
+                        );
+                        *other = m;
+                        merged = true;
+                        break;
```
MartinquaXD, line 140: Always merging a candidate settlement into some other settlement at most once would mean that we don't look at a lot of possible candidates.
This is probably fine for now when we want to get **something** that works but this might need to be improved in the future.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -95,36 +97,100 @@ impl Competition {
...
+                            settlement_2 = ?other.solutions(),
+                            "settlements can't be merged"
+                        );
+                    }
+                }
+            }
+            // If [`settlement`] can't be merged into any other settlement, this is its
+            // final, most optimal form. Push it to the results.
+            if !merged {
+                results.push(settlement);
```
MartinquaXD, line 155: nit: I think it might be easier to read to drop `merged` and instead move the push to the `break` statement.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -133,6 +143,75 @@ impl Verified {
...
+            .clone()
+            .merge(other.inner.boundary.clone())?;
+
+        let gas = Gas {
+            // After two settlements have been merged, the only verification that
+            // needs to be done is simulating the settlement (which also yields the gas
+            // needed by the final merged settlement). All other verified rules still
+            // hold after the settlements have been merged. See also [`Solution::verify`].
+            estimate: simulator
+                .gas(settlement.tx(contract, Internalization::Enable))
```
MartinquaXD, line 190: I think this will fail if one of the settlements sends native ETH to some smart contract wallet because that requires the correct access lists to be used.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -133,6 +143,75 @@ impl Verified {
...
+                id: self.inner.id,
+                solutions: self
+                    .inner
+                    .solutions
+                    .union(&other.inner.solutions)
+                    .copied()
+                    .collect(),
+                boundary: settlement,
+            },
+            access_list: self.access_list.clone().merge(other.access_list.clone()),
```
MartinquaXD, line 211: It might not be enough to merge the access lists. Merging multiple settlements might result in different storage slots being read than in any of the individual settlements.
Not sure how often this would be a problem, though.

crates/driver/src/tests/cases/merge_settlements.rs
```diff
@@ -0,0 +1,102 @@
...
+        .order(Order {
+            name: "second order",
+            amount: 400000000000000000u64.into(),
+            sell_token: "A",
+            buy_token: "B",
+            ..Default::default()
+        })
+        // These two solutions result in different clearing prices and can't be merged.
+        .solution(Solution::Valid, &["first order"])
+        .solution(Solution::AdditionalCalldata { bytes: 3 }, &["second order"])
```
MartinquaXD, line 87: Is it required for this test that the second solution gets some extra data?

crates/driver/src/domain/competition/mod.rs
cowfee: As mentioned in the PR description, this is noted in https://github.com/cowprotocol/services/issues/1480.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -133,6 +143,75 @@ impl Verified {
...
+        // The solver must be the same for both settlements.
+        if self.inner.boundary.solver != other.inner.boundary.solver {
+            return Err(super::Error::DifferentSolvers);
+        }
+
+        // If the solutions being merged are not disjoint, the settlements can't be
+        // merged.
+        if !self.inner.solutions.is_disjoint(&other.inner.solutions) {
+            return Err(super::Error::DuplicateSolutions);
+        }
```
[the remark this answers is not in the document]
cowfee: I'm OK with leaving this in. It makes the code and the intention more explicit. The legacy code will need to be refactored away at some point anyway.

crates/driver/src/domain/competition/mod.rs
cowfee: That's not the same thing. The `push` needs to happen if _no_ merge happened. OTOH what you're proposing would result in `push` happening when a merge happens, which is the opposite from what we want.

cowfee: Hmm, odd that this confused you. I thought the comments would make this really easy to follow. Am I missing something?

MartinquaXD: Yeah, you are right. I remember stumbling over the same thing the first time I tried to review the PR and came to the conclusion that my suggestion is not equivalent. When I wanted to continue the review after github stopped having issues I tripped over the same thing but forgot about that insight. 🤦‍♂️

cowfee:
So it seems like everything from `Settlement::verify` should also be done during the merge process. So this requires changes from https://github.com/cowprotocol/services/issues/1489 to be implemented.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -95,36 +97,100 @@ impl Competition {
...
+            .collect_vec();
+
+        // TODO(#1483): parallelize this
+
+        // Merge the settlements in random order.
+        settlements.shuffle(&mut rand::thread_rng());
+        // The merged settlements in their final form.
+        let mut results = Vec::new();
+        while !settlements.is_empty() {
+            let settlement = settlements.pop().unwrap();
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/ca1ca88a0bffa0035548a599d72f7a230adbeb7e

crates/driver/src/domain/competition/mod.rs
```diff
@@ -95,36 +97,100 @@ impl Competition {
...
+                            settlement_2 = ?other.solutions(),
+                            "merged settlements"
+                        );
+                        *other = m;
+                        merged = true;
+                        break;
+                    }
+                    Err(err) => {
+                        tracing::debug!(
+                            err = ?err,
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/ca1ca88a0bffa0035548a599d72f7a230adbeb7e

crates/driver/src/domain/competition/mod.rs
```diff
@@ -95,36 +97,100 @@ impl Competition {
...
+        let settlements = results;
+
+        // Score the settlements.
+        let scores = join_all(settlements.into_iter().map(|settlement| async move {
+            tracing::trace!(
+                solutions = ?settlement.solutions(),
+                settlement_id = ?settlement.id(),
+                "scoring settlement"
+            );
+            (settlement.score(&self.eth, auction).await, settlement)
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/ca1ca88a0bffa0035548a599d72f7a230adbeb7e

crates/driver/src/domain/competition/solution/settlement.rs
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/8c0d5f3549b5455cb1c08f98989bf8a332963d1a

crates/driver/src/domain/competition/solution/settlement.rs
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/8c0d5f3549b5455cb1c08f98989bf8a332963d1a

crates/driver/src/domain/competition/solution/settlement.rs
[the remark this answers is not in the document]
cowfee: Ended up removing it :) https://github.com/cowprotocol/services/pull/1490/commits/8c0d5f3549b5455cb1c08f98989bf8a332963d1a

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -122,14 +122,14 @@ impl Solution {
...
-        // guaranteed by all nodes.
-        let balance = eth.balance(self.solver.address()).await?;
-        if balance < gas.required_balance() {
+        // Ensure that the solver has sufficient balance for the settlement to be mined.
+        if eth.balance(self.solver.address()).await? < gas.required_balance() {
             return Err(Error::InsufficientBalance);
         }
 
+        // TODO Does this need to be done as part of merged settlement verification as
+        // well? I think it's not worth bothering with this.
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/8c0d5f3549b5455cb1c08f98989bf8a332963d1a

crates/driver/src/tests/cases/merge_settlements.rs
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/8b4ce5b494118078b6061d466de43ced4b834060

cowfee:
This is now ready for review again.

vkgnosis (approved):

crates/driver/src/boundary/settlement.rs
```diff
@@ -164,35 +155,40 @@ impl Settlement {
...
     }
 
-    pub fn tx(self) -> eth::Tx {
-        let encoded_settlement = self.inner.encode(match self.internalization {
+    pub fn tx(
+        &self,
+        id: settlement::Id,
+        contract: &contracts::GPv2Settlement,
+        internalization: Internalization,
+    ) -> eth::Tx {
```
nlordell, line 167: This split makes sense.

crates/driver/src/domain/competition/mod.rs
```diff
@@ -84,47 +86,102 @@ impl Competition {
...
+            // Try to merge [`settlement`] into some other settlement.
+            for other in settlements.iter_mut() {
+                match other.merge(&settlement, &self.eth, &self.simulator).await {
+                    Ok(m) => {
+                        tracing::debug!(
+                            settlement_1 = ?settlement.solutions(),
+                            settlement_2 = ?other.solutions(),
+                            "merged solutions"
+                        );
+                        *other = m;
```
nlordell, line 130: This is neat, so `settlements` vector keeps the "most merged" settlements until they can't be merged anymore and pushed into `results` vector.

One worry I have is that driver will actually miss solving deadlines because of this. Can we add a timeout to the "merge loop" so that we only keep on trying to merge for as long as we have time, and then just bail out and pick the best we've got once we've hit the deadline? (Can be done in a follow up).

crates/driver/src/domain/competition/mod.rs
```diff
@@ -84,47 +86,102 @@ impl Competition {
...
         let scores = scores.map(|(score, settlement)| {
-            tracing::info!(id = ?settlement.solution.id, score = f64::from(score.clone()), "solution scored");
+            tracing::info!(
+                solutions = ?settlement.solutions(),
+                settlement_id = ?settlement.id,
+                score = f64::from(score.clone()),
+                "settlement scored"
+            );
             (score, settlement)
         });
```
nlordell, line 186: Not a huge fan of this because `map` is lazy. Changes to the code 👇 can have unintended consequences where the tracing doesn't happen anymore. (i.e. its relying on the fact that `max_by_key` will collect the iterator which isn't exactly clear from a "code intent" point of view IMO).

nlordell (approved):
Looks great, just some small nits.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -1,12 +1,314 @@
...
+        // gas). However, publishing a settlement with interactions that violate
+        // these rules constitutes a punishable offense for the solver, even if
+        // the interactions are internalized. To ensure that this doesn't happen, check
+        // that the settlement simulates even when internalizations are disabled.
+        if solutions
+            .values()
+            .flat_map(|solution| solution.interactions.iter())
+            .any(|interaction| interaction.internalize())
+        {
+            Self::simulate(
```
MartinquaXD, line 163: I think there might be a simple way to avoid a bunch of node requests and thus reducing latency if it's needed in the future by passing in an `Option<AccessList>` into `Self::simulate()`.
In the second simulation (with `Internalization::Disable`) you already know which access list is required to make the tx work for smart contract wallets so you could easily pass that access list into the second simulation.
This idea assumes that `eth_createAccessList` ignores useless access list entries attached to the tx (which I'm not 100% sure is the case).

MartinquaXD (approved):

crates/driver/src/domain/competition/mod.rs
cowfee: https://github.com/cowprotocol/services/issues/1498. Also, I added your summary as a comment here: https://github.com/cowprotocol/services/pull/1490/commits/f0393f9a60770bd7758ea51e743ca7d080fb8ecd

crates/driver/src/domain/competition/mod.rs
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/bd3e26de4e7594553c0bb8ed7e68b8fbfd21342d

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -464,10 +289,10 @@ pub enum VerificationError {
...
-}
-
-impl From<simulator::Error> for Error {
-    fn from(value: simulator::Error) -> Self {
-        VerificationError::from(value).into()
-    }
+    #[error("insufficient solver account Ether balance")]
+    InsufficientBalance,
+    #[error("settlement solutions are not disjoint during merging")]
+    DuplicateSolutions,
```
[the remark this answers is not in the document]
cowfee: https://github.com/cowprotocol/services/pull/1490/commits/d55fe7ed2883fb19da1f79efb7606e859e1cd971

crates/driver/src/domain/competition/solution/settlement.rs
cowfee: Like this? https://github.com/cowprotocol/services/pull/1490/commits/a0963a71f0c9180816fd7cf6518ef84b56d3a809

MartinquaXD: Yes. This is even nicer since we don't have to depend on `eth_createAccessList` dropping unnecessary storage items for us. 👍

PR #1493, opened 2023-05-10
DB docs

MartinquaXD:
Add DB docs to help people become effective with our DB scheme.

closes https://github.com/cowprotocol/services/pull/1446


vkgnosis (approved):
Nit: I wouldn't do the `<details>` this way. It feels annoying to click them all because usually you want to read all of them. If some of the text in one of those tags is really long, it could go below the table.

I didn't go through all the content but I don't want to make too many comments now. I'll double check it at some point and then can make follow up PR to fix things if needed.

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+
+This table is used for [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f). It stores which solvers (identified by ethereum address) participated in which auctions (identified by auction id). CIP-20 specifies that "solver teams which consistently provide solutions" get rewarded.
+
+   Column     |  Type  | Nullable | Details
+--------------|--------|----------|--------
+ auction\_id  | bigint | not null | id of the auction
+ participant  | bytea  | not null | solver that submitted a **valid** solution for the auction
+
+Indexes:  
+- PRIMARY KEY: btree(`auction_id`, `participant`)
```
cowfee, line 21: Documenting the indexes for each table seems like a huge maintenance burden. Especially considering that these indexes are mostly an optimization. I'd leave them out TBH.

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+auction\_id | bigint  | not null | in which auction this price was provided
+token       | bytea   | not null | address of the token the price refers to
+price       | numeric | not null | the atoms of ETH that can be bought with 1 atom of the token
+
+Indexes:  
+- PRIMARY KEY: btree(`auction_uid`, `token`)  
+
+### auction\_transaction
+
+Because the the transaction hash of a given settlement depends on the gas parameters it ultimately gets submitted with onchain we can't know the hash before it got submitted. That's why we store the transaction sender (`tx_from`) and next nonce of the winning solver. With that information we can later associate the auction with the transaction that brought it onchain by cross-referencing the `tx_from` and `tx_nonce`.
```
cowfee, line 38:
```suggestion
Because the transaction hash of a given settlement depends on the gas parameters it ultimately gets submitted with onchain we can't know the hash before it got submitted. That's why we store the transaction sender (`tx_from`) and next nonce of the winning solver. With that information we can later associate the auction with the transaction that brought it onchain by cross-referencing the `tx_from` and `tx_nonce`.  
```

cowfee:
It just occurred to me - what about the [`COMMENT`](https://www.postgresql.org/docs/current/sql-comment.html) SQL command? Have you considered maybe using it instead of having a separate document for the DB structure?

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+ log\_index    | bigint | not null | index in which the log was emitted
+ order\_uid    | byteai | not null | order that got invalidated
+
+Indexes:  
+- PRIMARY KEY: btree(`block_number, log_index`)  
+- invalidations\_order\_uid: btree(`order_uid`, `block_number`, `log_index`)  
+
+### onchain\_order\_invalidations
+
+Stores data of [`OrderInvalidation`](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapOnchainOrders.sol#L46-L49) events emited by the `ICoWSwapOnchainOrders` interface.
```
cowfee, line 123:
```suggestion
Stores data of [`OrderInvalidation`](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapOnchainOrders.sol#L46-L49) events emitted by the `ICoWSwapOnchainOrders` interface.
```
(A small but useful tip: if you're writing lots of text, install a spellchecker in your IDE. It's a lifesaver.)

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+
+### onchain\_placed\_orders
+
+Stores data of [`OrderPlacement`](https://github.com/cowprotocol/ethflowcontract/blob/main/src/interfaces/ICoWSwapOnchainOrders.sol#L23-L44) events emited by the `ICoWSwapOnchainOrders` interface plus some metadata.
+
+ Column           | Type                                | Nullable | Details
+------------------|-------------------------------------|----------|--------
+ uid              | bytea                               | not null | order that got created also known as order\_uid
+ sender           | bytea                               | not null | user that created the order with the smart contract
+ is\_reorged      | boolean                             | not null | if the backend detects that an block creating an order got reorged it gets invalidated with this flag
```
cowfee, line 143:
```suggestion
 is\_reorged      | boolean                             | not null | if the backend detects that a block creating an order got reorged it gets invalidated with this flag
```

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+- order\_owner: hash(`owner`)
+- order\_quoting\_parameters: btree(`sell_token`, `buy_token`, `sell_amount`)
+- order\_valid\_to: btree(`valid_to`)
+- user\_order\_creation\_timestamp: btree(`owner`, `creation_timestamp` DESC)
+- user\_valid\_to: btree(`valid_to`)
+- version\_idx: btree(`settlement_contract`)
+
+### order\_quotes
+
+Quotes that an order was created with. This quotes get stored persistently and can be used to evaluate how accurate the quoted fee predicted the execution cost that actually happened on-chain.
```
cowfee, line 177:
```suggestion
Quotes that an order was created with. These quotes get stored persistently and can be used to evaluate how accurate the quoted fee predicted the execution cost that actually happened on-chain.
```

cowfee, line 177:
```suggestion
Quotes that an order was created with. These quotes get stored persistently and can be used to evaluate how accurate the quoted fee predicted the execution cost that actually happened on-chain.
```

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+ signed        | boolean | not null | specifies if an a signature was given or revoked
+
+Indexes:  
+- PRIMARY KEY: btreebtree(`block_number`, `log_index`)  
+- most\_recent\_with\_orderuid: btree (`order_uid`, `block_number` DESC, `log_index` DESC)  
+- presignature\_owner: hash(`owner`)
+
+### quotes (and quotes\_id\_seq counter)
+
+Stores quotes in order to determine whether it makes sense to allow a user to creat an order with a given `fee_amount`. Quotes are short lived and get deleted when they expire. `id`s are unique and increase monotonically.
```
cowfee, line 245:
```suggestion
Stores quotes in order to determine whether it makes sense to allow a user to create an order with a given `fee_amount`. Quotes are short lived and get deleted when they expire. `id`s are unique and increase monotonically.
```

cowfee, line 245:
```suggestion
Stores quotes in order to determine whether it makes sense to allow a user to create an order with a given `fee_amount`. Quotes are short lived and get deleted when they expire. `id`s are unique and increase monotonically.
```

database/README.md
```diff
@@ -1,14 +1,414 @@
...
+----------|--------
+ erc20    | Bought tokens will be added to the ERC20 token balance of that user
+ internal | Bought tokens will be added to the balancer vault internal balance of the user ([docs](https://docs.cow.fi/smart-contracts/vault-relayer/balancer-internal-balances))
+
+#### orderclass
+
+ Value     | Meaning
+-----------|--------
+ market    | Short lived order that may receive surplus. Users agree to a static fee upfront by signing it.
+ liquiidty | These orders must be traded at their limit price and may not receive any surplus. Violating this is a slashable offence.
```
cowfee, line 412:
```suggestion
 liquidty | These orders must be traded at their limit price and may not receive any surplus. Violating this is a slashable offence.
```

cowfee (approved):

database/README.md
vkgnosis: I'd like to keep the indexes. They are important design choices.

MartinquaXD:
> It just occurred to me - what about the [COMMENT](https://www.postgresql.org/docs/current/sql-comment.html) SQL command? Have you considered maybe using it instead of having a separate document for the DB structure?

Didn't know this is a thing. Seems pretty cool.
I'm a bit hesitant to try it out. Either we have to add (and especially update) the comments with DB migrations or one would have to manually modify the databases which both seem undesirable.

nlordell (approved):

PR #1497, opened 2023-05-11
Be explicit about fillable amount error

vkgnosis:
Noticed while on call debugging. This is cleaner and doesn't rely on trace logs being enabled because we already log when an order is excluded here. Noticed while on call debugging.

I'm not implementing the full error trait through ThisError because it isn't needed here and is more work.

### Test Plan

none

### DIFF
--- a/crates/solver/src/solver/single_order_solver/fills.rs
+++ b/crates/solver/src/solver/single_order_solver/fills.rs
@@ -25,6 +25,21 @@ pub struct Fills {
     smallest_fill: BigRational,
 }
 
+/// The reason `Fills::order` failed.
+#[derive(Debug)]
+pub enum Error {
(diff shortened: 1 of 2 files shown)


crates/solver/src/solver/single_order_solver/fills.rs
```diff
@@ -25,6 +25,21 @@ pub struct Fills {
     smallest_fill: BigRational,
 }
 
+/// The reason `Fills::order` couldnt' failed.
```
MartinquaXD, line 28:
```suggestion
/// The reason `Fills::order` failed.
```

MartinquaXD (approved):
Nice!

nlordell (approved):
nice

Issue #1105, opened 2023-01-19
Add driver tests to CI


sistemd:
Run the driver tests as part of our CI.

PR #1503, opened 2023-05-11
CI improvements

cowfee:
Fixes https://github.com/cowprotocol/services/issues/1105.

Add driver tests to CI. Start using `docker-compose` in CI instead of translating the compose setup into a GitHub workflow. This has a few benefits: it ensures that the local environment setup is on-par with the CI environment setup, it removes the burden of maintaining two different environment setup flows, and it tests the docker-compose setup in CI so it's impossible to commit code with a broken docker-compose.

There is one issue: because dev-geth is part of this repo, it needs to be built from scratch every time the dev-geth compose service is started. This adds up to 10 minutes to the CI pipeline. The solution to this problem is to extract dev-geth into its own repository, with its own release schedule and push the docker image to the registry. This way the dev-geth code will not need to get rebuilt all the time. I opened [an issue](https://github.com/cowprotocol/services/issues/1505) for this.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -6,7 +6,7 @@ on:
   push:
     branches: [main]
 jobs:
-  rust-misc:
+  lint:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -25,78 +25,87 @@ jobs:
...
-          ls -al
-      - name: Cache Hardhat
-        id: cache-hardhat
-        uses: actions/cache@v2
+          profile: minimal
+          toolchain: nightly
+          components: rustfmt
+      # Start the build process in the background. The following cargo test command will automatically
+      # wait for the build process to be done before proceeding.
+      - run: cargo build -p orderbook -p database --tests &
```
MartinquaXD, line 58: Would it be possible to only build all the targets once and still run the tests in parallel afterwards?
Maybe we can create a docker image with the built binaries in one job and let other jobs use that image to run the tests?

This should also get rid of a lot of duplicated setup code.

.github/workflows/pull-request.yaml
```diff
@@ -25,78 +25,87 @@ jobs:
...
+    env:
+      # Shrink artifact size by not including debug info. Makes build faster and shrinks cache.
+      CARGO_PROFILE_DEV_DEBUG: 0
+      CARGO_PROFILE_TEST_DEBUG: 0
     steps:
       - uses: actions/checkout@v2
-      - run: npm install @apidevtools/swagger-cli
-      - run: node_modules/.bin/swagger-cli validate crates/orderbook/openapi.yml
-      - run: node_modules/.bin/swagger-cli validate crates/driver/openapi.yml
-      - run: node_modules/.bin/swagger-cli validate crates/solvers/openapi.yml
```
MartinquaXD, line 102: I think we should probably keep the openapi test

MartinquaXD (commented):
I like the idea of using the `docker-compose` file for the CI setup. 👍

MartinquaXD:
Also it looks like the CI pipeline still expects the old test to report a result for some reason. 🤷‍♂️ 
Screenshot 2023-05-11 at 22 37 21

.github/workflows/pull-request.yaml
cowfee: I removed this accidentally :smile:

.github/workflows/pull-request.yaml
cowfee: So, it is possible in theory by using [workflow artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts), and I did try this. However, it's just another instance of [software](https://github.com/actions/upload-artifact/issues/199), so no go :(

.github/workflows/pull-request.yaml
cowfee: https://github.com/cowprotocol/services/pull/1503/commits/22fa247c1df91a99a1deeda3b38c53ba73866d7d

MartinquaXD (approved):

cowfee:
> Also it looks like the CI pipeline still expects the old tests to report results for some reason. man_shrugging Not sure if this is actually a problem, though. Screenshot 2023-05-11 at 22 37 21

Fixed this as well.

MartinquaXD:
It's only installing the nightly version of rustfmt. This is needed for the lint job but not for the test jobs as they don't run with the nightly toolchain.

vkgnosis:
Ah I got confused by that. Currently we install nightly rustfmt in all of the jobs even though it's only needed for lint.

vkgnosis (approved):

nlordell (approved):

PR #1510, opened 2023-05-16
Add strategy to submission error

sunce86:
Adds strategy name to `SubmissionError` type.

This is needed for metrics. To be more specific, now we can add an alert if Flashbots (MEVBlocker) is consistently timing out.

### DIFF
--- a/crates/solver/src/order_balance_filter.rs
+++ b/crates/solver/src/order_balance_filter.rs
@@ -34,6 +34,8 @@ impl BalancedOrder {
 }
 
 /// Fetch the balances for `balance_orders`.
+///
+/// If a balance can't be fetched, it isn't part of the result.
 pub async fn fetch_balances(
     fetcher: &dyn BalanceFetching,
     orders: &[Order],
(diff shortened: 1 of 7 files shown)


crates/solver/src/settlement_submission.rs
```diff
@@ -425,69 +461,110 @@ impl SubmissionError {
...
+    }
+}
+
+impl std::error::Error for SubmissionError {}
+
+impl SubmissionError {
+    /// Returns the outcome for use with metrics.
+    pub fn as_outcome(&self) -> SettlementSubmissionOutcome {
+        self.inner.as_outcome()
+    }
```
MartinquaXD, line 542: I think these methods could be dropped as they just expose functions on `inner` which is already `pub`. Just `into_anyhow()` seems to add some value by avoiding ambiguity.

crates/solver/src/settlement_submission.rs
```diff
@@ -425,69 +461,110 @@ impl SubmissionError {
...
+    pub strategy: &'static str,
+}
+
+impl std::fmt::Display for SubmissionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+impl std::error::Error for SubmissionError {}
```
MartinquaXD, line 536: The `std::error::Error` implementation doesn't seem to be required for compilation.

crates/solver/src/settlement_submission.rs
```diff
@@ -357,20 +384,26 @@ impl SolutionSubmitter {
...
+                strategy: strategy.label(),
+            }),
+        }
     }
 }
 
-/// An error during settlement submission.
 #[derive(Debug)]
-pub enum SubmissionError {
+pub enum SubmissionErrorClass {
```
MartinquaXD, line 406: Nit: Not sure renaming is super appropriate here.
1. This enum is the actual error and the new struct is an error + meta information so I would rather give that one a different name.
2. Way bigger diff

crates/solver/src/driver_logger.rs
```diff
@@ -129,7 +130,7 @@ impl DriverLogger {
                 // of race conditions as hard errors.
                 tracing::warn!(settlement_id, ?err, "Failed to submit settlement",);
                 self.metrics
-                    .settlement_submitted(err.as_outcome(), solver_name);
+                    .settlement_submitted(err.as_outcome(), solver_name, err.strategy);
```
MartinquaXD, line 133: Looks like this could take the new error + metadata type directly without passing its member fields as parameters.

nlordell (commented):
One thing that feels weird about this is that depending on which strategy "finishes first", we can potentially miss logging metrics. Can we instead write metrics for each strategy as it executes?

crates/solver/src/settlement_submission.rs
sunce86: Not used anymore.

crates/solver/src/settlement_submission.rs
sunce86: Irrelevant after latest changes.

crates/solver/src/driver_logger.rs
sunce86: Irrelevant after latest changes.

MartinquaXD (approved):

vkgnosis (approved):

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -348,6 +348,10 @@ impl<'a> Submitter<'a> {
         fallback_result
             .transpose()
             .unwrap_or(Err(SubmissionError::Timeout))
+            .map_err(|err| {
+                track_strategy_outcome(&format!("{name}"), err.as_outcome().label());
```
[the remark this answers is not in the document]
sunce86: Yes.

PR #1511, opened 2023-05-16
Return remaining balance back out

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/1378 

We are going to need to assign balances for FOK orders while letting solvers assign balances for partially fillable orders. To do this, we need to know what balances are remaining after assigning FOK orders.

I also simplify the function slightly so that it iterates over all orders in one loop without grouping them by Query first.

### Test Plan

existing and new test


crates/solver/src/order_balance_filter.rs
```diff
@@ -57,71 +59,68 @@ pub fn max_balance(orders: &[Order]) -> HashMap<Query, U256> {
...
+        if order.data.partially_fillable {
+            let balance_for_order = std::cmp::min(needed_balance, *remaining_balance);
+            if balance_for_order == 0.into() {
+                continue;
             }
+            result.push(BalancedOrder {
+                order,
+                available_sell_token_balance: balance_for_order,
+            });
+            *remaining_balance -= balance_for_order;
```
MartinquaXD, line 114: Not sure I'm understanding the PR description correctly. Doesn't it say that we **don't** want to assign balances for partially fillable orders. And that we need to know the **remaining** balance after assigning all the FOK orders?
These lines seem to do the opposite.

MartinquaXD (commented):
The simplification of `balance_orders()` looks good.

crates/solver/src/order_balance_filter.rs
vkgnosis: This PR is a refactor without changing logic. I need this refactor for the next PR, which will make use of the remaining balances.

MartinquaXD (approved):

PR #1512, opened 2023-05-16
Verified quotes

MartinquaXD:
This PR introduces the undocumented `verified` quote quality that could be used for quoting orders with custom interactions. But this could easily be dropped in case we actually decide to cleanly split price estimation from quoting.
Another alternative would be to automatically generate a verified quote if we detect user interactions on a quote request (TBD).
I also prepared a new `verified` price estimator which we will need regardless of the exact implementation we want to do.

Currently setting a trade verification strategy not only enabled support for the `verified` quote quality but also forces every estimator of the `optimal` price estimation logic to be verified. This wasn't changed by this PR but it's a bit weird to not be able to differentiate these 2 features at the moment so a follow up PR should handle that.

### Test Plan
manual test that `priceQuality: "verified"` gets handled correctly and that it returns an error if no trade verification strategy was configured

### DIFF
--- a/crates/shared/src/price_estimation/factory.rs
+++ b/crates/shared/src/price_estimation/factory.rs
@@ -59,8 +59,8 @@ pub struct PriceEstimatorFactory<'a> {
 
 #[derive(Clone)]
 struct EstimatorEntry {
-    optimal: Arc<dyn PriceEstimating>,
-    fast: Arc<dyn PriceEstimating>,
+    verified: Option<Arc<dyn PriceEstimating>>,
+    unverified: Arc<dyn PriceEstimating>,
     native: Arc<dyn PriceEstimating>,
 }
 
@@ -180,11 +180,9 @@ impl<'a> PriceEstimatorFactory<'a> {
             .as_ref()
             .and_then(|trade_verifier| estimator.verified(trade_verifier));
 
-        let fast = instrument(estimator, name);
-        let optimal = match verified {
-            Some(verified) => instrument(verified, format!("{name}_verified")),
-            None => fast.clone(),
-        };
+        let unverified = instrument(estimator, name);
+        let verified =
+            verified.map(|verified| instrument(verified, format!("{name}_verified")) as _);
 
         // Eagerly create the native price estimator, even if we don't use it.
         // It just simplifies price estimator creation code and only costs a few
@@ -196,8 +194,8 @@ impl<'a> PriceEstimatorFactory<'a> {
         let native = instrument(T::init(self, name, params)?, name);
 
         Ok(EstimatorEntry {
-            optimal,
-            fast,
+            verified,
+            unverified,
             native,
         })
     }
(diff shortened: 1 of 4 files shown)


MartinquaXD (commented):
I just noticed that this split of verifying the quote and verifying user interactions in the future will not work if the pre-interactions are responsible for moving funds into the `from` address in the first place. :/

MartinquaXD:
> For example, why is verified the only estimator that is optional?


Unfortunately we currently cant create a `verified` version of all price estimator implementations (`baseline` and `balancer` are missing) so we can't have it non-optional. This is eluded to [here](https://github.com/cowprotocol/services/pull/1512/files#diff-ec9ccc954f4d59ba8ad726c9586dd802dafae91e98f6a25cdc436b5bf2490ee1R443-R445).

crates/shared/src/price_estimation/factory.rs
```diff
@@ -203,6 +201,7 @@ impl<'a> PriceEstimatorFactory<'a> {
             optimal,
             fast,
```
nlordell, line 202: I think this means we no longer need a distinction between `optimal` and `fast`, but rather `verified` and `unverified` right? I would change the entry to only have those two (instead of having the three `optimal`, `fast` and `verified`).

Also, I refactored the `PriceEstimatorCreating` trait so that it can be used for co-located driver estimators, and so we don't have separate code paths for traditional and driver-based estimator creation. See #1521

crates/shared/src/price_estimation/factory.rs
```diff
@@ -268,15 +267,22 @@ impl<'a> PriceEstimatorFactory<'a> {
...
+                let estimator = estimator.clone().verified(trade_verifier.clone());
+                instrument(estimator, format!("{}_verified", driver.name)) as _
+            });
+
             let entry = EstimatorEntry {
                 optimal: estimator.clone(),
                 fast: estimator.clone(),
                 native: estimator,
+                verified,
             };
```
nlordell, line 286: Specifically, this code can be shared with `create_estimator_entry`. (see comment https://github.com/cowprotocol/services/pull/1512/files#r1198810411)

nlordell (commented):
Changes look good. Some small nits.

crates/shared/src/price_estimation/factory.rs
MartinquaXD: > I think this means we no longer need a distinction between optimal and fast, but rather verified and unverified right?

I think this is not true (at least for now).
1. Verifying all quotes that previously were `optimal` is too much for prod. So we still need a way to get the cheaper to compute `optimal` quotes.
2. Not all estimators that currently have a verified implementation so `verified` would contain fewer estimators than `optimal`. This means `verified` is (as of now) **not** just a stricter version of `optimal`; it's slightly different.

nlordell: > This means verified is (as of now) not just a stricter version of optimal; it's slightly different.

This I agree completely, my point is that now `optimal` and `fast` estimators are the same:

```rust
        let optimal = fast.clone();
```

So, instead of having `optimal` and `fast` estimators for each entry, we can have:

```rust
#[derive(Clone)]
struct EstimatorEntry {
    unverified: Arc<dyn PriceEstimating>,
    verified: Option<Arc<dyn PriceEstimating>>,
    native: Arc<dyn PriceEstimating>,
}
```

And just construct `fast` and `optimal` `RacingPriceEstimator`s differently.

nlordell (approved):

PR #1519, opened 2023-05-18
Do not consider single order solver failures when there is no match

fedgiac:
I've seen alerts the Paraswap solver when there wasn't enough liquidity to fill an order.
My understanding from [this issue](https://github.com/cowprotocol/services/issues/1003) is that this isn't expected behavior.

This PR makes all the Paraswap errors for "insufficient liquidity" count as a success (this is arguable: I was also considering having a different label, e.g., `skipped`).
It also makes the 1inch solver use this code path rather than custom logic inside its implementation.

### Test Plan

I don't know how to test this PR properly. Any suggestion is welcome.


MartinquaXD (approved):

crates/solver/src/solver/oneinch_solver.rs
```diff
@@ -129,11 +129,6 @@ impl OneInchSolver {
         tracing::debug!("querying 1Inch swap api with {:?}", query);
         let swap = match self.client.get_swap(query).await {
```
nlordell, line 130: I think this can now be simplified to:
```suggestion
        let swap = match self.client.get_swap(query).await?;
```

nlordell (approved):

PR #1523, opened 2023-05-19
Fix Gnosis solvers score computation since revert cost is now zero

harisang:
This PR readjusts the score computation for Gnosis solvers, since we now have zero cost whenever a transaction fails. This is due to MEV Blocker and the soft cancellations we can now do.

See also relevant discussion on [Slack](https://cowservices.slack.com/archives/C0375NV72SC/p1684508285452079).


harisang:

MartinquaXD (approved):

nlordell (approved):

PR #1524, opened 2023-05-22
Add More Error Context on Invalid Signatures

nlordell:
This PR adds additional information for signature errors to facilitate debugging from external integrations. Namely:
- We include the signing message when the signer does not match the owner. This should help debugging issues with signing schemes for ECDSA signatures.
- We include the order hash for invalid EIP-1271 signatures. This, along with the specified signature bytes, should allow us to recreate the `isValidSignature` in Tenderly for example call to see what is up.

### Test Plan

Rust compiler, adjusted unit tests to check for new error variants.


vkgnosis (approved):

crates/orderbook/src/api/post_order.rs
```diff
@@ -121,10 +122,20 @@ impl IntoWarpReply for ValidationErrorWrapper {
...
+                        "recovered signer {signer} from signing hash {message:?} does not match \
+                         from address"
+                    ),
+                ),
+                StatusCode::BAD_REQUEST,
+            ),
+            ValidationError::InvalidEip1271Signature(hash) => with_status(
+                error(
+                    "InvalidEip1271Signature",
+                    format!("signature for hash {hash:?} is not valid"),
```
MartinquaXD, line 138: Should this error message maybe convey that the hash was computed from the order data or do we expect people to know this?

PR #1528, opened 2023-05-24
Remove app data hash based fee subsidy

vkgnosis:
Related to https://github.com/cowprotocol/services/issues/1465 . With the meaning of app data changing, it doesn't make sense to award a subsidy to an arbitrary app data hash value because the value is a hash of something else and not freely chooseable.

If we wanted to keep app id (which is not app data hash) based subsidies then we would need to do that through the full app data. But this feature in general makes little sense because anyone could pick an app id that has a subsidy even when they're not actually coming from that app. So it makes more sense to remove. I don't think we're actively using this anyway. In the infra configuration I see one value which has last been mentioned on Slack more than a year ago.

Follow ups:
- Remove this configuration from the infra repo.

### Test Plan

CI

### DIFF
--- a/crates/autopilot/src/database/onchain_order_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events.rs
@@ -550,7 +550,6 @@ async fn get_quote(
         kind: order_data.kind,
         // Original quote was made from user account, and not necessarily from owner.
         from: order_placement.sender,
-        app_data: order_data.app_data,
     };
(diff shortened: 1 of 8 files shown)


vkgnosis:
I forgot that for the quote endpoint we need to return the app data back out to facilitate easy signing. Will need to change this slightly to only contain the "app id based fee subsidy" part.

nlordell (approved):

PR #1531, opened 2023-05-24
Simulate DEX Solver Swaps For Accurate Solver Fees

nlordell:
https://github.com/cowprotocol/services/pull/1469 introduced simulations to the single order solvers so that we can produce accurate fees when solving for partially fillable limit orders. This PR ports some of that logic to our `solvers` binary for the DEX aggregator solvers.

### Test Plan

This PR is still in draft as it broke the integration tests (it added new configuration options). A bit more setup is required for automated tests (specifically, we need to setup a node and deploy contracts to verify that the simulation works).


vkgnosis:
In the description you write that you are porting a subset of the other. So how come `crates/contracts/solidity/Trader.sol` needs additional logic. I would expect that to not need changing because the other PR was able to be implemented without adding to this file.

nlordell:
> the other PR was able to be implemented without adding to this file.

The main difference is that the other implementation was done in the `solver` binary, which has access to the full order for simulation (and can build valid settlements). Here, we are simulating in the `solvers` binary which, being a solver engine, gets "reduced" order data, and does not have sufficient data to build a full settlement from the order. As such, we need to use some state override simulation techniques to work around this limitation.

crates/contracts/solidity/Trader.sol
```diff
@@ -87,6 +97,75 @@ contract Trader {
...
+        trades[0] = Trade({
+            sellTokenIndex: 0,
+            buyTokenIndex: 1,
+            receiver: address(0),
+            sellAmount: sell.amount,
+            buyAmount: buy.amount,
+            validTo: type(uint32).max,
+            appData: bytes32(0),
+            feeAmount: 0,
+            flags: 0x40, // EIP-1271
```
MartinquaXD, line 141: Why do you use `EIP-1271` here?

MartinquaXD (commented):
Overall I like the idea (wanted to do the same thing on the `autopilot` and `orderbook` side) but I think building the settlement in solidity is not needed.

sunce86:
@nlordell  What is the status of this PR? 

Could be beneficial for https://github.com/cowprotocol/services/issues/1899 which needs `gas_used`.

nlordell:
PR needs some love. I can look into reviving it before I go.

nlordell:
Superseded.

PR #1532, opened 2023-05-25
Create app-data-hash crate

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/1465 .

Note that this crate doesn't concern itself with the full app data format. It treats it as raw bytes. I've done it this way to make it easy to reuse. We will have more specific validation logic in the order creation code.

I am using the tiny-keccak crate because we already depend on it through web3.

### Test Plan

CI

### DIFF
--- a/crates/app-data-hash/src/lib.rs
+++ b/crates/app-data-hash/src/lib.rs
@@ -1,0 +1,76 @@
+//! App data refers to extra information that is associated with orders. This
+//! information is not validated by the contract but it is used by other parts
+//! of the system. For example, a user could specify that they want their order
+//! to be COW only, which is something only the backend understands. Or what
+//! their intended slippage when creating the order with the frontend was, which
+//! adjusts the signed prices.
+//!
+//! On the smart contract level app data is freely choosable 32 bytes of signed
+//! order data. This isn't enough space for some purposes so we interpret those
+//! bytes as a hash of the full app data of arbitrary length. The full app data
+//! is thus signed by the user when they sign the order.
+//!
+//! This crate specifies how the hash is calculated. It takes the keccak256 hash
+//! of the input bytes. Additionally, it provides a canonical way to calculate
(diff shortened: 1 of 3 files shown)


nlordell (approved):

crates/app-data-hash/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+//! order data. This isn't enough space for some purposes so we interpret those
+//! bytes as a hash of the full app data of arbitrary length. The full app data
+//! is thus signed by the user when they sign the order.
+//!
+//! This crate specifies how the hash is calculated. It takes the keccak256 hash
+//! of the input bytes. Additionally, it provides a canonical way to calculate
+//! an IPFS CID from the hash. This allows full app data to be uploaded to IPFS.
+//!
+//! Note that not all app data hashes were created this way. As of 2023-05-25 we
+//! are planning to move to the scheme implemented by this crate but order have
```
cowfee, line 18:
```suggestion
//! are planning to move to the scheme implemented by this crate but orders have
```

cowfee (commented):
Why does this deserve its own crate?

vkgnosis:
"deserve" isn't the right word. It's nice for it to be in it's own crate. That's better for code organization and compile times. I'd like more of "shared" to be in separate crates. For this crate there is a good chance some other piece of rust code is going to want to use it so I preemptively made it a crate. For example, an analytics script retrieving app data from IPFS or a custom order placement script.

nlordell:
> I don't like the idea of doing "shared code" preemptively but it doesn't matter.

I actually think this fits into `model` crate quite well (given what is already in there), or just in the `orderbook` crate to not prematurely create a new crate without knowing if the data team will actually use this crate (seeing as I think they don't have any Rust projects at the moment).

My personal opinion is that we don't _need_ a new crate for this yset, but there are valid arguments for creating a separate crate for this already, so I'm definitely fine with it.

PR #1537, opened 2023-05-26
Validate full app data on order creation

vkgnosis:
Part of https://github.com/cowprotocol/services/issues/1465 .

Implements validation of app data according to the new scheme discussed in the issue. Validation only happens if the new format is used. If the old format (plain hash) is used, everything works as before. This allows us to test the new approach with the frontend and other code.

### Test Plan

CI, new unit test, existing e2e tests make sure orders can still be created


crates/shared/src/app_data.rs
```diff
@@ -0,0 +1,103 @@
...
+                backend: Default::default(),
+            });
+        }
+
+        if full_app_data.len() > self.size_limit {
+            return Err(anyhow!(
+                "app data has size {} which is larger than limit {}",
+                full_app_data.len(),
+                self.size_limit
+            ));
```
MartinquaXD, line 46: nit: Should we specify in which format the message is supposed to be smaller than the size limit?
For example the user might escape all the `"` or unicode characters which would mean the raw string they send us could be way bigger than the data we actually store on disk.
I think just looking at the raw bytes of the sent string is fine (anything else would probably be a lot work for little improvement) but making the error message more explicit would be low hanging fruit.

crates/shared/src/order_validation.rs
```diff
@@ -459,6 +467,22 @@ impl OrderValidating for OrderValidator {
...
+        // compatibility.
+        if let Some(full_app_data) = &order.full_app_data {
+            let validated = self
+                .app_data_validator
+                .validate(full_app_data.as_bytes())
+                .map_err(ValidationError::InvalidAppData)?;
+            if validated.hash != order.data.app_data {
+                return Err(ValidationError::AppDataHashMismatch {
+                    actual: validated.hash,
+                });
```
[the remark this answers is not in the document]
vkgnosis: An API user already knows that value because they sent the request. But I guess it could clear up some confusion when they didn't know about the app data field at all. I'll add it.

crates/shared/src/app_data.rs
vkgnosis: I'll add this somewhere. Either this message or the openapi docs.

crates/model/src/order.rs
```diff
@@ -357,6 +357,7 @@ pub struct OrderCreation {
     #[serde(flatten)]
     pub signature: Signature,
     pub quote_id: Option<QuoteId>,
+    pub full_app_data: Option<String>,
```
nlordell, line 360: While we expect the full app data to be a string, shouldn't we already deserialize instead of it being a string in the model here? We can use `serde` deserializers for that (meaning that you can't have an `OrderCreation::full_app_data` that is invalid).

Also, I'm not a fan of `full_app_data` as a name. I don't have any good suggestions beside `appData` (which would be a breaking change, but I think that this will have to be a breaking change anyway).

nlordell: Specifically an argument for not using a `String` here, the OpenAPI spec specifically says this is an `object`, but the current type doesn't encode that at all.

nlordell (commented):
Looks good to me. Two nits:
- I think we should have an `AppData` type that guarantees coherent `full_app_data` values instead of carrying strings in our model.
- I'm really not a fan of the name `full_app_data` at it seems imprecise (why is it full, how is "full" app data different than the "app_data")

In terms of making this API change backwards compatible, can we instead have:

```rust
enum AppData {
    /// **DEPRECATED**. Legacy app data specified by hash.
    Hash(H256),
    /// App data value. This is the pre-image of the order's `app_data` field that is used for signing.
    Value(AppData),
}
```

Then we can add an optional `appDataHash` field that is used for validation (what the current `appData` + `fullAppData` is used for today).

nlordell:
Also, this will eventually lead to a breaking API change (this PR should not introduce a breaking API change itself, but I imagine it will eventually become a breaking change). This should be co-ordinated with the front-end team (cc @anxolin).

crates/model/src/order.rs
vkgnosis: Openai says that this is a string. I also added a FullAppData object to document the json schema. But it is a string.

We can't deserialize here because we need to store the exact signed bytes in the database.

vkgnosis: "full app data" is the name I used in the design issue. There is "app data hash" and "full app data". We shouldn't make a breaking change for the old name but I'm open for suggestions on changing the new name. Note that currently this isn't a breaking change because the field is optional and I would like to keep it a non breaking change when possible: Order created with all zero app data don't need to set full app data.

crates/orderbook/openapi.yml
```diff
@@ -563,9 +564,25 @@ components:
...
+              description: |
+                A string encoding a json object like `"{\"hello\":\"world\"}"`. If this field is
+                set, then `appData` must be set to the keccak-256 hash of this field's utf-8 encoded
+                bytes. If this field is not set, then `appData` must be all zeroes. The last
+                condition is not yet enforced.
+
+                The json object can contain arbitrary application specific data (json key values).
+                The optional key `backend` is special. It has to conform to the schema documented in
+                `FullAppData`. The total byte size of `fullAppData` in utf-8 encoded bytes is
+                limited to 1000.
```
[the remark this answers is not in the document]
vkgnosis: The limit of 1000 doesn't have a lot of thought behind it. I just picked a number large enough for the current front end app data.

crates/model/src/order.rs
vkgnosis: > * I'm really not a fan of the name `full_app_data` at it seems imprecise (why is it full, how is "full" app data different than the "app_data")

This is explained in the api documentation. But the yes the field names are slightly misleading.

MartinquaXD: I believe at some point Nick suggested to use the [JCS](https://www.rfc-editor.org/rfc/rfc8785) format. This would result in a deterministic JSON format. That could be stored as a `jsonb` in the DB and used to recreate the originally signed data.
This would probably mean that we have to introduce `serde_jcs` and a bunch of awkward constraints only to avoid this string in the API here. Also since we will presumably store the backend relevant app data information in a DB table being able to store the original signed data as `jsonb` is kind of pointless.

vkgnosis: I see what you mean now. Note that the orderbook already validates all these things and returns appropriate errors. Only validated json makes it into the DB. They just don't happen on the top level Deserialize derive. Where this would make a difference is if we had other rust code making use of model. That code would have an easier time because it can use the more type safe representation.

I'm hesitant to put too much domain logic in the model crate. We already have too much of that. I'd like it to be closer to the json level. It's similar to what Niksha advocates for with DTO objects. We can still a struct like you propose that is optimized for use in Orderbook (for example) but I'm not sure it should be in model.

Regardless of what representation we use in Rust code, this is a json string. Would it be ok to move this discussion to later, out of this PR? I'd like to have this logic in and I feel the intricacies of the Rust representation are not so important compared to that.

crates/model/src/order.rs
```diff
@@ -357,6 +357,7 @@ pub struct OrderCreation {
...
 }
```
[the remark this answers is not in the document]
vkgnosis: I see. We can differentiate between old app data field and new app data field based on the content of the string. The hash isn't a valid json object and vice versa. I dislike a bit that we use the same field with different meaning. This wouldn't work for example if both formats were json objects. On the other hand for this special case it is case kind of neat.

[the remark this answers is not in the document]
vkgnosis: I do like your suggestion that conceptually we want an enum here with the variants "app data hash only" (backward compat), "full app data only" (hash is implicit), "full app data and hash" (hash is verified for better debugging).

crates/model/src/order.rs
nlordell: > Would it be ok to move this discussion to later, out of this PR?

nlordell: OK - lets table this suggestion for now and keep `OrderCreation` more DTO like. I do think that if we start manipulating the `AppData` (i.e. once we start parsing it to verify pre- and post- interactions, special order types, etc.) then we should introduce a proper `AppData` abstraction.

MartinquaXD (approved):

crates/model/src/order.rs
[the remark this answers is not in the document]
cowfee: Yes

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
-                Orders can optionally include a quote ID. This way the order can be linked to a quote
-                and enable providing more metadata when analyzing order slippage.
-              type: integer
-              nullable: true
-          required:
-            - signingScheme
-            - signature
+                Short:
+
+                If you do not care about app data, set this field to `\"{}\"` and make sure that the order you signed for this request had its app data field set to `0xb48d38f93eaa084033fc5970bf96e559c33c4cdc07d889ab00b4d63f9590739d`.
```
cowfee, line 608: I think you meant to say this:

```suggestion
                If you do not care about app data, set this field to `\"{}\"` and make sure that the order you signed for this request had its appDataHash field set to `0xb48d38f93eaa084033fc5970bf96e559c33c4cdc07d889ab00b4d63f9590739d`.
```

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
+
+                Long:
+
+                A string encoding a json object like `"{\"hello\":\"world\"}"`.
+
+                This field determines the smart contract order's app data field, which is assumed to be set to the keccak-256 hash of the utf-8 encoded bytes of this string. You must ensure that the signature that is part of this request indeed signed a smart contract order with the app data field set appropriately. If this isn't the case, signature verification will fail. For easier debugging you can additionally set the `appDataHash` field.
```
cowfee, line 614: Same here I think.

```suggestion
                This field determines the smart contract order's app data field, which is assumed to be set to the keccak-256 hash of the utf-8 encoded bytes of this string. You must ensure that the signature that is part of this request indeed signed a smart contract order with the appDataHash field set appropriately. If this isn't the case, signature verification will fail. For easier debugging you can additionally set the `appDataHash` field.
```

nlordell (approved):

cowfee (approved):

crates/orderbook/openapi.yml
vkgnosis: It depends what the library the api user is using calls it. In our official library it is called "app data". https://github.com/cowprotocol/contracts/blob/5c25837f585d79dc5450fb80f66c41f596c3198d/src/ts/order.ts#L49

crates/shared/src/order_validation.rs
```diff
@@ -424,9 +434,36 @@ impl OrderValidating for OrderValidator {
...
+            OrderCreationAppData::Hash { hash } => *hash,
+            OrderCreationAppData::Full { full } => validate(full)?.hash,
+        };
+
         let owner = order.verify_owner(domain_separator)?;
         let signing_scheme = order.signature.scheme();
-        let data = order.data();
+        let data = OrderData {
+            app_data: app_data_hash,
+            ..order.data()
```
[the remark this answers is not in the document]
vkgnosis: It's not needed here. I did this to be more explicit. But yeah I will think about this.

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
+
+                This field must be the encoding of a valid json object. The json object can contain arbitrary application specific data (json key values). The optional key `backend` is special. It has to conform to the schema documented in `BackendAppData`.
+
+                The total byte size of this field's utf-8 encoded bytes is limited to 1000.
+              type: string
+        appDataHash:
+          description: |
+           May be set for debugging purposes. If set, this field is compared to what the backend internally calculates as the app data hash based on the contents of `appData`. If the hash does not match, an error is returned. If this field is set, then `appData` must be a string encoding of a json object.
```
anxolin, line 622: nit: I would start the description by mentioning what it is, rather what you can do with it

crates/orderbook/openapi.yml
anxolin, line 608: I don't think we should encourage people not not use this by saying they can add `{}`. Ideally, integrations should create one that would include their appCode at least. 

Also, If they don't care, why not leaving it blank if you don't care about it? `0x0` and undefined appData?

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
-          required:
-            - signingScheme
-            - signature
+                Short:
+
+                If you do not care about app data, set this field to `\"{}\"` and make sure that the order you signed for this request had its app data field set to `0xb48d38f93eaa084033fc5970bf96e559c33c4cdc07d889ab00b4d63f9590739d`.
+
+                Long:
+
+                A string encoding a json object like `"{\"hello\":\"world\"}"`.
```
anxolin, line 612: I don't think you need to add these scapes characters, might confuse the reader

Also, I don't think we should mention they should be able to add any JSON. Ideally we should encourage to follow the schema in https://github.com/cowprotocol/app-data

crates/orderbook/openapi.yml
anxolin, line 614: nit: `appData` field instead of `app data` since it's the actual name in the contract for the field

crates/orderbook/openapi.yml
anxolin: curious, why is `appDataHash` optional? 
doesn't the user need to provide the signature already that should already be using the matching hash?

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
-                If set, the backend enforces that this address matches what is decoded as the signer of
-                the signature. This helps catch errors with invalid signature encodings as the backend
-                might otherwise silently work with an unexpected address that for example does not have
-                any balance.
-              allOf:
-                - $ref: "#/components/schemas/Address"
-              nullable: true
-            quoteId:
+      type: object
+      properties:
```
anxolin, line 534: this adds a bit of repetition. Is there a way to keep it more DRY? (or is this being auto-generated?)

crates/orderbook/openapi.yml
```diff
@@ -540,32 +530,113 @@ components:
...
+            Orders can optionally include a quote ID. This way the order can be linked to a quote
+            and enable providing more metadata when analyzing order slippage.
+          type: integer
+          nullable: true
+        appData:
+          description: |
+            This field comes in two forms for backward compatibility. The first form (app data hash) will eventually stop being accepted.
+          anyOf:
+            - $ref: "#/components/schemas/AppDataHash"
+            - title: Full App Data
```
anxolin, line 604: If `appDataHash` and `appData` are connected, shouldn't it be an object with the two properties, the data and the hash? (for the second type)

anxolin (commented):
Thanks for tagging me. I will tag also @alfetopito just to get his opinion. 

I feel this whole change should have a spec and plan in notion to follow, because as I mentioned I'm concerned with the dependencies of this. 

@olgafetisova @bh2smith should also be participating on this document to make sure the changes are as "backward compatible as possible", also we need to make sure the docs are easy to understand and follow. For this, I believe we should make it easy for anyone doing an integration to identify themselves and attach a document with an `appCode` (or other information)

Maybe this will help you https://www.notion.so/cownation/appData-Orders-Metadata-020d5f4e18bd45248a26ec5b16dba6df?pvs=4

* also we had a deprecated one https://www.notion.so/cownation/App-data-Old-way-to-create-c133f1d27acc40838aef5cb8ad54a66d?pvs=4 

I believe this change should be treated as a coordinated project, and not as a backend task 🙏

crates/orderbook/openapi.yml
vkgnosis: The api user has three ways to use to provide app data when creating an order:
1. The way it always worked. Providing a hash only. This continues to happen through the field we've always had `appData` for backward compatibility.
2. The new way. By providing the full app data json in the field `appData`.
3. The new way with extra debugging. By providing full app data and this new field `appDataHash`.

If the api user uses approach 2 and makes a mistake when calculating the app data hash locally, then they will get a generic error that signature verification failed. Because the smart contract order they signed had a different app data hash than what the backend used.

In order to make this easier to debug, the api user can provide the hash that they signed too. This allows the backend to produce a better error message.

crates/orderbook/openapi.yml
vkgnosis: This came up in [this](https://github.com/cowprotocol/services/pull/1537#discussion_r1219592815) answer above already, which I don't want to repeat.

crates/orderbook/openapi.yml
vkgnosis: The only way to not repeat is to create a new object that contains all the fields that are shared between `OrderCreation` and `OrderData` and referencing that object from both use cases. I didn't do that because it didn't feel like the right semantics. It would only exist to hold the shared fields without modelling anything. Not sure I can explain this right.

crates/orderbook/openapi.yml
vkgnosis: The reason I did it this way is for backward compatibility on the v1 create order route.

vkgnosis:
> I feel this whole change should have a spec and plan in notion to follow, because as I mentioned I'm concerned with the dependencies of this.

The "revamp app data issue" is the spec. Although I didn't update it completely with some choices I made throughout the PRs. I can write a plan in Notion too. I want to emphasize again that there is a distinction between app data from the backend's point of view and the frontend's, which I mentioned in a comment above [here](https://github.com/cowprotocol/services/pull/1537#discussion_r1219592815). I think it would be good to keep these separate so the components are more decoupled.

As an example of complete decoupling, you could imagine that the frontend doesn't use backend's app data at all and instead keeps it's own database of order-uid and full app data. We don't have to go that far, but we could. The way the new app data scheme is designed, you can store arbitrary content and you can easily store the content in IPFS, but you don't have to.

crates/orderbook/openapi.yml
alfetopito: > But from a backend API point of view, it has no meaning

That's the thing, the appData is not meant to have meaning for the backend (exclusively)
It's meant for adding information that doesn't fit into what backend or the SC expects.

I think at least the `appCode` is already valuable information for Analytics to discern source of orders?
cc @bh2smith to confirm whether this is used ATM.

> The frontend happens to follow some app data specification but we're not forcing everyone to follow it.

We on the contrary would like everyone to follow it, not for frontend's sake, but for everyone interacting with the protocol.
This way you know the format and what type of data will be stored, how to read it, etc

> Either way should be easy to use for an api user that doesn't care about app data

I also agree with Anxo that having a null condition rather than `{}` is preferred

vkgnosis: I've rewritten the "revamp app data" issue to talk about frontend vs backend app data more.

crates/orderbook/openapi.yml
anxolin: it's just so you don't need to C&P descriptions, examples, etc. Keeping it DRY will reduce the errors for example if you do one change in one field but don't do it in another. They are extremely similar. 

>  It would only exist to hold the shared fields without modelling anything. Not sure I can explain this right.

That wouldn't be a problem, it's just an abstraction that refactor both types. 

Anyways,is not that important to me, I just mentioned because this might be more error prone.

crates/orderbook/openapi.yml
anxolin: > I don't know how we use the frontend's app data at the moment. But from a backend API point of view, it has no meaning

@vkgnosis I have to say, I mentioned this before, and I never like when we make this differentiation Backend/Frontend, and you call this front-end appData. We really made an effort to make this part very explicit.

Let me explain clarify my point.

When we defined this, even though it was a Frontend initiative (we were the first creating orders), we brought @bh2smith @alfetopito @alfetopito and @nlordell into the discussion. We discussed over a document until we got an agreement. This is a soft contract (not SC enforced), but is highly beneficial of encouraging people to use a common patter/specification for the meta-data

We agreed on a format that can be future proof:
- Allows to add new types of metadata
- It's versioned: both the document, and the metadatas
- Its definition is formalized in https://json-schema.org/
- Its in GIT https://github.com/cowprotocol/app-data so you can see the schemas, and its evolution

We provided tutorials: https://docs.cow.fi/front-end/creating-app-ids/create-the-order-meta-data-file

We created Notion pages: 
- https://www.notion.so/cownation/appData-Orders-Metadata-020d5f4e18bd45248a26ec5b16dba6df?pvs=4
- https://www.notion.so/cownation/App-data-Old-way-to-create-c133f1d27acc40838aef5cb8ad54a66d?pvs=4

The Data team did an integration and is now consumed in Dune.

Some external parts use this format. 

It is used for the affiliate program (now deprecated), not only in CoW Swap, but external parties

It is used the UTM codes https://www.notion.so/cownation/Create-UTM-codes-for-on-chain-analytics-cd1ec4ce33964da48a374c8d17b00913?pvs=4

I belive is not a Frontend standard, its something we put a lot of effort into having some agreement on this format, is something we should advertise even though it was not meaningful for the backend until pre-interactions was not implemented.

This can be changed, and it's good to see how we improve it (the hashing change is a good idea), but I want to make sure we agree this is not Frontend and whatever we do needs to be:
- Carefully planned
- Carefully Agreed by all teams
- Shoudn't break any of the apps, documentation, etc (all this should be part of the plan)

vkgnosis: Thanks for clarifying that. I didn't know about the history.

crates/orderbook/openapi.yml
vkgnosis: You're right that they already have it. To me it felt cleaner to not be forced to provide it because it can be computed from the full app data. It is redundant. The advantage of being forced to provide it is that we make our code and the api docs simpler. Will think about it.

crates/orderbook/openapi.yml
anxolin: No worries @vkgnosis I understand, happy I could provide more context.

bh2smith:
So this has been merged and already live then in staging? Are there already orders created with this? its going to cause problems with our dune sync. All new appHashes will not be found on IPFS and we will upload null to Dune (as if there was no content). I hope this isn't already live, because it would be nice if we could prepare our changes before.

PR #714, opened 2022-11-03
Periodically calculating surplus fees in a background task

ennmichael:
Fixes #issue.

Describe context of what this change does, why it is needed and how it is accomplished

### Test Plan

Explain how you and a reviewer have/intend to test this change.

### Release notes

Optionally add notes, suggestions and warnings for the releaser. They will be included in the release changelog.

PR #436, opened 2022-08-10
Add auction database functions to orderbook and autopilot

vkgnosis:
They will be used soon in another PR that makes the autopilot store auctions in the database and the orderbook read them from the database.

### Test Plan

CI, actual db logic is already tested in database crate

PR #1541, opened 2023-05-30
Deduplicate database::FullOrder model::Order conversion

vkgnosis:
We have two almost identical versions of `fn full_order_into_model_order(order: FullOrder) -> Result<Order>`. One is in orderbook and one is shared. The one in shared is used by autopilot. They have these differences:
- Shared doesn't calculate the order status.
- Orderbook always uses `full_fee_amount` as `solver_fee` whereas Shared has a special case to use `fee_amount` as `solver_fee` for liquidity orders.

The new combined implementation takes the order status logic from orderbook and the full fee amount logic from Shared.

### Test Plan

existing tests

### DIFF
--- a/crates/orderbook/src/database/orders.rs
+++ b/crates/orderbook/src/database/orders.rs
@@ -3,51 +3,28 @@ use {
     anyhow::{Context as _, Result},
     async_trait::async_trait,
     chrono::{DateTime, Utc},
-    database::{
-        byte_array::ByteArray,
-        orders::{FullOrder, OrderKind as DbOrderKind},
-    },
+    database::byte_array::ByteArray,
     ethcontract::H256,
     futures::{stream::TryStreamExt, FutureExt, StreamExt},
     model::{
-        app_id::AppDataHash,
-        order::{
-            EthflowData,
-            Interactions,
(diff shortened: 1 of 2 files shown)


nlordell:
Makes sense. For some reason I vaguely remember this being intentional (as order loading has different semantics in both services), but I'm not 100% sure about it.

vkgnosis:
The PR that created the second version is https://github.com/cowprotocol/services/pull/714 but I didn't find an explanation there.

MartinquaXD:
> The PR that created the second version is https://github.com/cowprotocol/services/pull/714 but I didn't find an explanation there.

That PR only moved that code from `orderbook` to `shared`. The explanation why there were 2 functions is [here](https://github.com/cowprotocol/services/pull/436#discussion_r942305375).

MartinquaXD (approved):
I'm fine with going back to 1 full order function again.
I guess we could split them again if we ever actually refactor the different order types.
Whatever you prefer.

vkgnosis:
Of course my own PR :) . Should have explained it in a comment. Reading that again, it does make long term sense to split it. But we should actually do the thing where we have split types.

Issue #1506, opened 2023-05-15
Error normalizing limit order


vkgnosis:
We got this error log
> ERROR (pod: xdai-solver-staging-648bf6df67-stc5x):
2023-05-12T16:40:47.232Z ERROR auction{id=7719016 run=10680}: solver::driver: error normalizing limit order err=order with 0 amounts order=0x47cc22cee7babe5497e920925bdc49a307837b0eaf9d016a16f21f3961f4b2757b2e78d4dfaaba045a167a70da285e30e8fca196bb9aca00

It might be related to changes in https://github.com/cowprotocol/services/pull/1439 and https://github.com/cowprotocol/services/pull/1448 .

In our code we have a comment
> This should never happen unless we are getting malformed orders from the API.

Not sure how valid this old comment still is so not sure whether this is a bug in autopilot or driver.

MartinquaXD:
This particular order:
1. was part of a solution
2. was then in-flight
3. produced the error
4. got filtered out as a dust order afterwards
[logs](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2023-05-12T16:39:33.668Z',to:'2023-05-12T16:41:30.228Z'))&_a=(columns:!(log),filters:!(),grid:(columns:('@timestamp':(width:164))),index:ea511870-d9b3-11ed-a9d0-a17451f01cc1,interval:auto,query:(language:kuery,query:'log:%20%220x47cc22cee7babe5497e920925bdc49a307837b0eaf9d016a16f21f3961f4b2757b2e78d4dfaaba045a167a70da285e30e8fca196bb9aca00%22'),sort:!(!('@timestamp',desc))))
This leads me to believe that some orders that pass the dust order in the autopilot would no longer pass the same check when they get adjusted for the partial execution that is currently in-flight on the driver side.

fedgiac:
And [again](https://cowservices.slack.com/archives/C0371SB243E/p1684434036353879). ([0x927d16269a43acfe67f8bab718fcaef36e27b54ed9b8f4650f593df5b894df9d0d9a5678e73e5bbc0ee09faf8e550b196c76fdad645a6264](https://explorer.cow.fi/orders/0x4caeeb93b4bbab7b979b34a2f547765e07e5b5c902a9baf1b43569f08b88beca6f364ec121020e246993855666a536821d6876216468bc70?tab=overview))

PR #1542, opened 2023-05-31
Demote order conversion alert

MartinquaXD:
Closes: #1506

We decided that it's not worth to alert on this error. The reason is that after putting an order in the auction in the `autopilot` the circumstances might change enough that it will result in the order being dropped on the `driver` side due to no error.
This could happen because a partial fill for that order is already in-flight which would reduce the fillable amount in the `driver` too much. Also it could be that the user's balance changes when we see a new block which results in the same thing.
The only way to catch that would be to basically do the same conversion and check the `OrderConverter` does before hand to filter those orders out. But if we are fine with discarding those orders without an alert in this pre-test there is no good reason we shouldn't just demote the `error` to a `debug` log directly in the `OrderConverter` and not introduce that pre-test.

### Test Plan
CI

### DIFF
--- a/crates/solver/src/solver.rs
+++ b/crates/solver/src/solver.rs
@@ -609,10 +609,8 @@ fn balance_and_convert_orders(
         .into_iter()
         .filter_map(|order| match converter.normalize_limit_order(order) {
             Ok(order) => Some(order),
-            // This should never happen unless we are getting malformed
-            // orders from the API - so raise an alert if this happens.
             Err(err) => {
-                tracing::error!(?err, "error normalizing limit order");
+                tracing::debug!(?err, "error normalizing limit order");
                 None
             }
         })


vkgnosis (approved):

crates/solver/src/solver.rs
```diff
@@ -609,10 +609,8 @@ fn balance_and_convert_orders(
         .into_iter()
         .filter_map(|order| match converter.normalize_limit_order(order) {
             Ok(order) => Some(order),
-            // This should never happen unless we are getting malformed
-            // orders from the API - so raise an alert if this happens.
             Err(err) => {
-                tracing::error!(?err, "error normalizing limit order");
+                tracing::debug!(?err, "error normalizing limit order");
```
nlordell, line 613: Should this be a `warn`?
```suggestion
                tracing::warn!(?err, "error normalizing limit order");
```

PR #1544, opened 2023-05-31
Implement hacky support for soft cancellations

MartinquaXD:
[Context](https://cowservices.slack.com/archives/C0361CDD1FZ/p1684746958813839) on slack.
Simply enabling soft cancellations in our MevBlocker proxy is not sufficient to avoid subtle bugs in our submission logic with that feature. We also have to add actual support for it.

The basic problem seems to be this:
1. solver wins auction
2. we start to submit the tx
3. tx start to revert so we cancel the tx
4. cancellation does not get mined (I assume due to the inner workings of MevBlocker + soft cancellations)
5. same solver wins again
6. we want to submit the next solution but we remember that the previous cancellation has not been mined yet and the gas price didn't increase enough that "normal" nodes would replace the tx so we just don't submit anything for 2 minutes (or however long it takes for the gas price to increase enough)

Since `MevBlocker` supports soft cancellations we can pretend there is no pending tx waiting to be mined.
So when the same solver wins then next auction again we will not wait for the gas price to increase by 12.5% before starting to submit the winning tx.

### Test Plan
CI

### DIFF
--- a/crates/driver/src/infra/config/file/mod.rs
+++ b/crates/driver/src/infra/config/file/mod.rs
@@ -89,6 +89,15 @@ enum Mempool {
         /// Flashbots above regular gas price estimation.
         #[serde(default = "default_max_additional_flashbots_tip")]
         max_additional_tip: f64,
+        /// Configures whether the submission logic is allowed to assume the
+        /// submission nodes implement soft cancellations. With soft
+        /// cancellations you don't have to submit a no-op transaction to
+        /// replace another transaction in the mempool. Instead you can just
+        /// issue another transaction with the same sender and nonce and
+        /// the submission node will no longer propagate the previous
+        /// transactions
+        #[serde(default = "default_soft_cancellations_flag")]
+        use_soft_cancellations: bool,
     },
 }
 
@@ -114,6 +123,10 @@ fn default_max_confirm_time_secs() -> u64 {
 
 fn default_max_additional_flashbots_tip() -> f64 {
     3.0
+}
+
+fn default_soft_cancellations_flag() -> bool {
+    false
 }
 
 #[serde_as]
--- a/crates/driver/src/infra/config/file/mod.rs
+++ b/crates/driver/src/infra/config/file/mod.rs
@@ -91,11 +91,9 @@ enum Mempool {
         max_additional_tip: f64,
         /// Configures whether the submission logic is allowed to assume the
         /// submission nodes implement soft cancellations. With soft
-        /// cancellations you don't have to submit a no-op transaction to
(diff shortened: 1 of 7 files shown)


vkgnosis (approved):

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -269,14 +271,34 @@ impl<'a> Submitter<'a> {
...
+                                // associated gas fee is at least 12.5% higher than on the pending
+                                // tx.
+                                // The remaining submission logic assumes the regular replacement
+                                // conditions so it would wait for the gas price to increase before
+                                // submitting another tx from the same `sender` if there is still a
+                                // tx pending. This would block the whole driver.
+                                // Because this is unnecessary with soft cancellations we can
+                                // simply "forget" the submitted tx by clearing `transactions`.
+                                tracing::debug!("clear list of pending tx hashes due to soft cancellation");
+                                transactions.clear();
```
nlordell, line 297: Just to double-check, `transactions` is per `sender` right?

nlordell (approved):
Approved assuming the answer to my question ☝️ is "yes".

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -89,6 +89,15 @@ enum Mempool {
         /// Flashbots above regular gas price estimation.
         #[serde(default = "default_max_additional_flashbots_tip")]
         max_additional_tip: f64,
+        /// Configures whether the submission logic is allowed to assume the
+        /// submission nodes implement soft cancellations. With soft
+        /// cancellations you don't have to submit a no-op transaction to
+        /// replace another transaction in the mempool. Instead you can just
+        /// issue another transaction with the same sender and nonce and
+        /// the submission node will no longer propagate the previous
+        /// transactions
```
cowfee, line 98:
```suggestion
        /// transactions.
```

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -89,6 +89,15 @@ enum Mempool {
         /// Flashbots above regular gas price estimation.
         #[serde(default = "default_max_additional_flashbots_tip")]
         max_additional_tip: f64,
+        /// Configures whether the submission logic is allowed to assume the
+        /// submission nodes implement soft cancellations. With soft
+        /// cancellations you don't have to submit a no-op transaction to
+        /// replace another transaction in the mempool. Instead you can just
```
cowfee, line 95:
```suggestion
        /// cancellations the driver doesn't have to submit a no-op transaction to
        /// replace another transaction in the mempool. Instead it can just
```

PR #1545, opened 2023-05-31
Fix whitespace mistakes in database readme

vkgnosis:
Was editing the file for a different reason and wanted to do this in separate PR.

### Test Plan

n/a

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -4,9 +4,9 @@ Code that directly interfaces with the database through SQL queries lives in the
 
 With a live database information for all tables can be retrieved with the `\d` command and information for a specific table with `\d MyTable`.
 
-Some tables only store data emitted via smart contract events. Because we only have a single deployment of the [`GPv2Settlement`](https://github.com/cowprotocol/contracts/blob/main/src/contracts/GPv2Settlement.sol) settlement contract shared across staging and production environments events related to staging **and** production orders and settlements will be present in **both** the staging **and** production databases.  
-[CoWSwapEthFlow](https://github.com/cowprotocol/ethflowcontract/blob/main/src/CoWSwapEthFlow.sol) we actually deployed twice so events related to the staging environment should only show up in the staging DB and likewise for production.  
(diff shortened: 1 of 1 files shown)


vkgnosis:
I see. Didn't know that trailing space has semantic meaning in markdown. The way I'm used to making a new line in markdown is to have a full empty line. Are you ok with changing the trailing spaces to empty lines? It's a small preference for me.

MartinquaXD:
Those options are not equivalent for the 2 markdown flavors, though. A full empty line will produce a new paragraph whereas 2 spaces will only produce a line break.
I was not able to produce identical layout in github markdown and "regular" markdown without those 2 spaces. But I don't care too much and most people will probably only look at the markdown rendered on github anyway so there is probably no need for the greater compatibility.
Will try to find a markdown viewer that supports the github markdown flavor to use in the future.

MartinquaXD (approved):

vkgnosis:
Didn't mean to merge without finishing discussion but auto merge was on. I was under the impression that usually a normal line ending is *not* a new line in the rendered view. You *need* to indicate that you want a real new line with a full empty line.

MartinquaXD:
AFAICT only the github markdown renderer behaves like you described for line breaks.
Another thing that is compatible across github markdown and regular markdown is to use a backslash for line breaks.
But if the `\` is followed by an empty line the `\` will actually be rendered so it's easier to make mistakes with that.

vkgnosis:
Markdown doesn't have one specification that is accepted everywhere. The main spec I know is [CommonMark](https://commonmark.org/) . The relevant links there are:

- https://spec.commonmark.org/0.30/#paragraphs
- https://spec.commonmark.org/0.30/#hard-line-break
- https://spec.commonmark.org/0.30/#soft-line-breaks

An empty line separates paragraphs. Two spaces at the end of a line are a hard line break. A normal end of line is a soft line break. How paragraph is visualized is up to the tool but a paragraph almost always ends is at least a line break. Semantically my intention is usually to write paragraphs. I don't want to manually place line breaks. According to this spec then I should write a full paragraph in one line (to not induce soft line breaks) and end it with an empty line.

PR #1556, opened 2023-06-02
Refactor OrderCreation to allow for new app data scheme

vkgnosis:
In https://github.com/cowprotocol/services/pull/1537#discussion_r1212850311 we discussed how the new app data format should be added in a backward compatible way. This PR implements Nick's suggestion:

> My suggestion would be to have `appData` be either a "full" app data (where a sidecar `app_data_hash` can be specified to assist with validation) or a hash. So all of these would be valid:
> 
> ```js
> { "appData": "0x...." }
> { "appData": "{\"hello\":\"world\"}", "appDataHash": "0x...." }
> { "appData": "{\"hello\":\"world\"}" }
> ```

This change happens in `crates/model/src/order.rs`. The other changes deal with making the refactor compile. I didn't convert all e2e tests yet. I will do that after review to avoid unnecessary work in case of changes.

In order to implement the new format I made the following changes:
- Stop storing OrderData in OrderCreation. It is no longer true that all fields from OrderData exist in the same way in OrderCreation. In OrderData app data is always a hash. In OrderCreation app data can be in several forms. We can go from OrderCreation to OrderData but not the other way around.
- Remove conversions between OrderCreation and Order. It is unclear how they should work because they need to make guesses about some fields. It is better to do this explicitly. For an example of this see `crates/shared/src/order_validation.rs` line 575, which requires very specific values that weren't documented in `Order::from_order_creation`.
- Derive Default for OrderCreation. If you need a valid signature, use `.sign`.
- Temporarily error in post_order if the new app data is used.

After this PR, I am going to rebase #1537 and remove the temporary error in post_order. (Note to self: Remember to change openapi spec.)

### Test Plan

existing tests pass


crates/model/src/order.rs
```diff
@@ -347,19 +318,71 @@ impl OrderData {
...
     pub from: Option<H160>,
     #[serde(flatten)]
     pub signature: Signature,
     pub quote_id: Option<QuoteId>,
+    #[serde(flatten)]
+    pub app_data: OrderCreationAppData,
 }
 
 impl OrderCreation {
+    pub fn order_data(&self) -> OrderData {
```
nlordell, line 353: naming nit: `order.order_data()` feels redundant, I would just prefer `order.data()`.

crates/model/src/order.rs
```diff
@@ -388,28 +411,51 @@ impl OrderCreation {
...
+    }
+}
+
+impl OrderCreationAppData {
+    /// Returns the app data hash. Does not validate expectation.
+    pub fn hash(&self) -> AppDataHash {
+        match self {
+            Self::Hash { hash } => *hash,
+            Self::Full { full } | Self::Both { full, .. } => {
+                AppDataHash(app_data_hash::hash_full_app_data(full.as_bytes()))
```
nlordell, line 457: Should this error for `Self::Both` if `app_data_hash::hash_full_app_data(full) != hash`?

I.e., is there a reason why the validation logic can't move here?

vkgnosis: The expectation is a quirk of our order creation validation. If a user provides an expectation, we may validate it. I find it weird to handle that logic here.

nlordell (commented):
Nice! I think this a really good change. One comment about moving the validation logic into the `OrderCreationAppData -> AppDataHash` transformation.

crates/model/src/order.rs
nlordell: But then the `OrderCreationAppData -> AppDataHash` has some ambiguous semantics which feels a bit weird.

nlordell: In my mind, the abstraction is an `OrderCreation` that we get from a user, and that data can sometimes not correctly convert to an `OrderData` (the validation). In that sense, having `OrderCreation::order_data` and `OrderCreationAppData::hash` potentially return errors reflects that well.

vkgnosis: The semantics of this method are to return the smart contract order's app data field for this order.

If you want to move this validation here then how do you justify not moving all of the order validation code into here? We could also validate that valid_to is not in the past and so on. But that doesn't make sense to me because it is one more level detached from the structural representation of the api types that the model crate is supposed to provide.

nlordell (approved):

crates/model/src/order.rs
nlordell: I think that is a different kind of validation though - validating the app data is checking that the `OrderCreation` data is consistent (like checking signatures for example, which is in this module AFAIR), while the `validTo` check is validating "order rules" - an expired order is a valid `OrderData`, it just can't be used and is thus rejected by the API

nlordell:
Approved as I don't want to bikeshed too much on the "where we validate the app data matches the hash" comment, and I do think the PR is already good as is.

vkgnosis:
I'll probably merge this with the current state of the `hash` function. We might naturally change this again when we introduce the more type safe app data struct.

PR #1564, opened 2023-06-06
Add Flag For Specifying only Buy Orders for 0x Estimator

nlordell:
This PR adjusts the 0x price estimator to only respond to buy orders when a "buy only" flag is set. This allows us to reduce load on the 0x API for the requests that it is more important for serving.

### Test Plan

Added a unit test, also did a manual test running the orderbook locally with 0x sell orders disabled and saw that quotes were failing on "unsupported order type".

### DIFF
--- a/crates/shared/src/price_estimation/zeroex.rs
+++ b/crates/shared/src/price_estimation/zeroex.rs
@@ -3,7 +3,6 @@ use {
         trade_finder::{TradeEstimator, TradeVerifier},
         PriceEstimateResult,
         PriceEstimating,
-        PriceEstimationError,
         Query,
     },
     crate::{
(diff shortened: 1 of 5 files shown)


crates/shared/src/price_estimation/zeroex.rs
```diff
@@ -40,7 +57,28 @@ impl PriceEstimating for ZeroExPriceEstimator {
         &'a self,
         queries: &'a [Query],
     ) -> futures::stream::BoxStream<'_, (usize, PriceEstimateResult)> {
-        self.0.estimates(queries)
+        if !self.buy_only {
+            self.inner.estimates(queries)
+        } else {
+            async_stream::stream! {
+                let (sell, buy) = queries
```
MartinquaXD, line 64: Would it maybe be slightly nicer to drag the `bool` all the way through to the `ZeroexTradeFinder`? Then it could simply return an error for sell quotes on the lowest level and you wouldn't have to implement this additional stream logic.

MartinquaXD (commented):
Fine for me if this hotfix is super urgent but I believe there is a slightly nicer option.

MartinquaXD (approved):

Issue #1475, opened 2023-05-05
`driver` tracing and metrics


cowfee:
The driver needs to expose some metrics and tracing for observability. There's already some logging, but this should probably be more structured.

#### Requirements

- Log every auction.
- Log all solutions returned from the solver engines.
- Whenever a solution is dropped, log the solution ID and the reason why it was dropped.
- Log the winning solution.
- Log when solutions are successfully merged.
- Log as much as possible in `infra::mempool` and `boundary::mempool`. This can be modeled after the logging in the legacy codebase, to understand what's useful to log and what isn't.
- Metric for the number of solutions submitted to the protocol.
- Metric for the number of winning solutions.
- Solution failure metric labeled by failure reason, to track how often solutions are being dropped and why.

These are the fundamental requirements to be able to go into production. They will be refined over time. Some additional nice-to-haves would be metrics related to solution merging.

Note that some logging and metrics are already inherited from the legacy code by virtue of the boundary modules.

#### DDD

I imagine a module in `infra` which exposes hooks which are meaningful to the domain. For example, something like `struct infra::observe::AuctionSolved(auction::Id, Solution)` or `infra::observe::auction_solved(auction::Id, Solution)` or similar.

nlordell:
I think we should add metrics around transaction submission as well. I think we inherit a bunch from using the `boundary` module, but it might make sense to have more metrics for the driver-specific logic there as well.

Generally speaking, transaction submission is an area where you want to err on the side of "too much logging/metrics" rather than too few (given that many issues appear in weird, not easily reproducible, race-condition-y ways because of the decentralised nature of the blockchain).

nlordell:
Other than that, things look good. Note that we inherit metrics and logging from liquidity indexing.

PR #1565, opened 2023-06-06
`driver`: observability

cowfee:
Progress on https://github.com/cowprotocol/services/issues/1475.

Adds an observability module. Currently, this module only implements logging, no metrics yet.

Introduces a new `util::Bytes` type which is debug-formatted as a hex string, and starts using it instead of `Vec<u8>`, which reduces the need for manual `Debug` impls.

Note that the only two places where logging is done without using the `observe` module are the `boundary` (not worth introducing `observe` there at this moment, but `observe` should be implemented once the boundary code gets removed) and the `tests`.

### Test Plan

Unit tests still pass and the logs look pretty.

### DIFF
--- a/crates/driver/src/infra/observe/mod.rs
+++ b/crates/driver/src/infra/observe/mod.rs
@@ -4,6 +4,7 @@
 //! and update the metrics, if the event is worth measuring.
 
 use {
+    super::Mempool,
     crate::{
         boundary,
         domain::{
@@ -160,8 +161,8 @@ pub fn solver_response(endpoint: &Url, res: Result<&str, &http::Error>) {
 }
 
 /// Observe that a mempool failed to send a transaction.
-pub fn mempool_failed(err: &boundary::Error) {
-    tracing::warn!(?err, "sending transaction via mempool failed");
+pub fn mempool_failed(mempool: &Mempool, err: &boundary::Error) {
+    tracing::warn!(?err, ?mempool, "sending transaction via mempool failed");
 }
 
 /// Observe that an invalid DTO was received.
--- a/crates/driver/src/infra/mempool/mod.rs
+++ b/crates/driver/src/infra/mempool/mod.rs
@@ -22,7 +22,7 @@ pub async fn execute(
         async move {
             let result = mempool.execute(solver, settlement).await;
             if let Err(err) = result.as_ref() {
-                observe::mempool_failed(err);
+                observe::mempool_failed(mempool, err);
             }
             result
         }


crates/driver/src/infra/observe/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+
+// TODO The idea is that some of the functions below will also update the
+// metrics. This should be fairly easy to do, just go down the list and think
+// about if the thing is worth measuring or not. This will be done in a
+// follow-up PR.
+
+/// Observe a received auction.
+pub fn auction(auction: &Auction) {
+    tracing::info!(?auction, "received auction");
+}
```
MartinquaXD, line 36: Not a big fan of adding this additional ceremony for very simple logs.
I see that this might be nice in case an event needs extremely verbose logging or also want to increment metrics at the same time but for very simple logs (which most of these are) I would prefer to do them inline.
As a side note logging stuff here will also obscure where in which file/component the event was triggered from.
IMO having the code location helps a lot with quickly navigating logs especially for people less familiar with the code base:
<pre>
<code>
2023-06-07T07:16:15.112Z DEBUG caching_native_price_estimator: <b>shared::price_estimation::competition</b>: winning price estimate query=Query
</code>
</pre>

crates/driver/src/util/bytes.rs
```diff
@@ -0,0 +1,30 @@
+/// A thin wrapper around a vector of bytes. Provides hex debug formatting.
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
+pub struct Bytes<T>(pub T);
```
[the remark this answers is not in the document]
vkgnosis: It's common to not have bounds on the raw struct and only on the functions and traits that actually need the bound. Not sure about this case because it's purely a wrapper type but in general that's sensible.

vkgnosis, line 3: I don't like storing these in structs to avoid having a manual Debug implementation. Now using the fields as intended is harder because you need `.0`, `.into()`, etc or some trait that exists for the raw bytes isn't implemented on the wrapper. I'm not extremely against it. For small structs like Uid and AppData a manual impl is better. For large structs I can see this have more use. An alternative is using a crate that derives Debug and allows you to annotate individual fields with custom formatters like in serde.

crates/driver/src/infra/observe/mod.rs
vkgnosis: I agree. I feel logging is meant to be low level, directly in the code, with convenient access to local variables. Having it directly in the module is important to control the log level too through `tracing`, although there might be other ways to accomplish that.

For metrics I see the point in having a dedicated observer more because metrics are more of an intentional output of the system. Whereas logging is whatever we find useful for debugging at the moment.

I'm willing to try it in this style too but while keeping in mind that we should re-evaluate it after a while.

nlordell (commented):
I feel like the proposed changes work for this PR in particular. I do worry that it doesn't scale well as the code in the `driver` grows.

It does make logging more explicit though (which helps with bad logging practices), but my fear is that it will bias us to write fewer logs to a fault.

nlordell (approved):

vkgnosis (approved):
approving but I still have an open discussion about the Bytes wrapper. Github makes it easy to miss as a comment.

crates/driver/src/util/bytes.rs
cowfee: This is similar to your [`ethcontract::Bytes`](https://docs.rs/ethcontract/latest/ethcontract/tokens/struct.Bytes.html), actually:

> Distinguishes a list of u8 from bytes.

Binary data is not the same thing as a list of (small) numbers. `Vec<u8>` is a list of numbers.

I agree that having to write `.0` is annoying. I personally don't mind it that much and am OK with it.

PR #1574, opened 2023-06-08
Store full app data in separate table

vkgnosis:
We want to store full app data in indexed by the contract app data that refers to it. This is needed to so that we can have an API route for retrieving full app data by contract app data. This encodes on the database level that a contract app data can only have one associated full app data. This was previously not encoded because we were storing the full app data inline in the orders table.

We do not need to worry about hash collisions because both schemes use cryptographic hashes. A hash collision would be problematic because it would prevent creating an order with valid app data because the contract app data would be the same as a different already existing full app data.

### Test Plan

CI tests that order insertion with and without full app data still works

### DIFF
--- a/crates/database/src/app_data.rs
+++ b/crates/database/src/app_data.rs
@@ -1,0 +1,84 @@
+use {crate::AppId, sqlx::PgConnection};
+
+/// Tries to associate the contract app data with the full app data.
+///
+/// If this contract app data already existed then the existing full app data is
+/// returned.
+///
+/// If this contract did not already exist, it is inserted and the passed in
+/// full app data is returned.
+pub async fn insert(
+    ex: &mut PgConnection,
+    contract_app_data: &AppId,
+    full_app_data: &[u8],
+) -> Result<Vec<u8>, sqlx::Error> {
+    const QUERY: &str = r#"
+WITH insertion AS (
+    INSERT INTO app_data (contract_app_data, full_app_data)
+    VALUES ($1, $2)
+    -- returns null on conflict
+    ON CONFLICT DO NOTHING
+    -- returns inserted value without conflict
+    RETURNING full_app_data
+)
+SELECT COALESCE(
+    (SELECT full_app_data FROM insertion),
+    (SELECT full_app_data FROM app_data WHERE contract_app_data = $1)
+)
+;"#;
+    sqlx::query_scalar(QUERY)
+        .bind(contract_app_data)
+        .bind(full_app_data)
+        .fetch_one(ex)
+        .await
+}
+
+pub async fn fetch(
+    ex: &mut PgConnection,
+    contract_app_data: &AppId,
+) -> Result<Option<Vec<u8>>, sqlx::Error> {
+    const QUERY: &str = r#"
+SELECT full_app_data
+FROM app_data
+WHERE contract_app_data = $1
+;"#;
+    sqlx::query_scalar(QUERY)
+        .bind(contract_app_data)
+        .fetch_optional(ex)
+        .await
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::byte_array::ByteArray, sqlx::Connection};
+
+    #[tokio::test]
+    #[ignore]
+    async fn postgres_app_data() {
+        let mut db = PgConnection::connect("postgresql://").await.unwrap();
+        let mut tx = db.begin().await.unwrap();
+        crate::clear_DANGER_(&mut tx).await.unwrap();
+        tx.commit().await.unwrap();
+
+        let contract = ByteArray([0u8; 32]);
+        // fetch non existant app data
+        let result = fetch(&mut db, &contract).await.unwrap();
+        assert!(result.is_none());
+
+        let full = vec![1u8];
+        let result = insert(&mut db, &contract, &full).await.unwrap();
+        assert_eq!(result, full);
+
+        // now exists
+        let result = fetch(&mut db, &contract).await.unwrap();
+        assert_eq!(result, Some(full.clone()));
+
+        // insert again with same app data
+        let result = insert(&mut db, &contract, &full).await.unwrap();
+        assert_eq!(result, full);
+
+        // insert again with different app data fails
+        let result = insert(&mut db, &contract, &[4, 2]).await.unwrap();
+        assert_eq!(result, full);
+    }
+}
--- a/crates/orderbook/src/database/orders.rs
+++ b/crates/orderbook/src/database/orders.rs
@@ -85,6 +85,8 @@ pub struct SolvableOrders {
 pub enum InsertionError {
     DuplicatedRecord,
     DbError(sqlx::Error),
+    /// Full app data to be inserted doesn't match existing.
+    AppDataMismatch(Vec<u8>),
 }
 
 impl From<sqlx::Error> for InsertionError {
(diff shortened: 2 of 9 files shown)


crates/database/src/app_data.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// If this contract app data already existed then the existing full app data is
+/// returned.
+///
+/// If this contract did not already exist, it is inserted and the passed in
+/// full app data is returned.
+pub async fn insert(
+    ex: &mut PgConnection,
+    contract_app_data: &AppId,
+    full_app_data: &[u8],
```
cowfee, line 13: Is it "contract app data", "app data hash", od "app id"? Would be nice to have one name for this concept instead of three, and always refer to it using that name.

crates/database/src/app_data.rs
```diff
@@ -0,0 +1,83 @@
...
+mod tests {
+    use {super::*, crate::byte_array::ByteArray, sqlx::Connection};
+
+    #[tokio::test]
+    #[ignore]
+    async fn postgres_app_data() {
+        let mut db = PgConnection::connect("postgresql://").await.unwrap();
+        let mut tx = db.begin().await.unwrap();
+        crate::clear_DANGER_(&mut tx).await.unwrap();
+        tx.commit().await.unwrap();
```
MartinquaXD, line 61: I'm a bit surprised to see the `commit` here. Not that is makes a difference it just looks like all the other DB tests never commit the tx.
Does **not** committing the tx maybe lead to a deadlock or some other problem?

crates/orderbook/src/orderbook.rs
```diff
@@ -147,11 +164,13 @@ impl From<ValidationError> for ReplaceOrderError {
     }
 }
 
+/*
 impl From<InsertionError> for ReplaceOrderError {
     fn from(err: InsertionError) -> Self {
         Self::Add(err.into())
     }
 }
+*/
```
cowfee, line 173: Why leave commented code? Why not just remove it? Are you planning to do something with it later?

crates/database/src/app_data.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        let contract = ByteArray([0u8; 32]);
+        // fetch non existant app data
+        let result = fetch(&mut db, &contract).await.unwrap();
+        assert!(result.is_none());
+
+        let full = vec![1u8];
+        insert(&mut db, &contract, &full).await.unwrap();
```
MartinquaXD, line 69: nit: This could also assert that `insert` immediately returns the new value if you rely on that behavior.

cowfee (approved):

crates/orderbook/src/database/orders.rs
```diff
@@ -179,18 +176,24 @@ impl OrderStoring for Postgres {
...
+            let contract_app_data = &ByteArray(order.data.app_data.0);
+            let full_app_data = full_app_data.as_bytes();
+            let existing =
+                database::app_data::insert(&mut ex, contract_app_data, full_app_data).await?;
+            if full_app_data != existing {
+                return Err(InsertionError::AppDataMismatch(existing));
+            }
+        }
+
+        ex.commit().await.unwrap();
```
MartinquaXD, line 195: The function already returns a `Result`. Maybe a failing `commit` should return an error as well instead of being unwrapped.

crates/orderbook/src/orderbook.rs
```diff
@@ -147,11 +164,13 @@ impl From<ValidationError> for ReplaceOrderError {
     }
 }
 
+/*
```
MartinquaXD, line 167: Only commented out. Maybe left over?

database/README.md
```diff
@@ -8,6 +8,20 @@ Some tables only store data emitted via smart contract events. Because we only h
...
+
+See [here](https://github.com/cowprotocol/services/issues/1465) for more details. In this table the contract app data is either the old unixfs based scheme, or the new keccak scheme. The new scheme can be validated by keccak-256 hashing the full app data, which should produce the contract app data. The old scheme cannot be validated.
+
+Column             | Type  | Nullable | Details
+-------------------|-------|----------|-------
+ contract_app_data | bytea | not null | 32 bytes. Referenced by `orders.app_data`.
+ full_app_data     | bytea | not null | Is utf-8 but not stored as string because the raw bytes are important for hashing.
+
+Indexes:
+- "app_data_pkey" PRIMARY KEY, btree (contract_app_data)
```
MartinquaXD, line 23: nit: Unescaped `_` can result in _italics_ in some markdown renderers.
Screenshot 2023-06-08 at 21 50 34

MartinquaXD (commented):
PR makes sense to me. Just a few nits.

crates/database/src/app_data.rs
vkgnosis: The type AppId should be renamed in this crate. I started using "contract app data" consistently in writing since working on the "revamp app data" issue.

crates/orderbook/src/orderbook.rs
vkgnosis: Should be removed. Accidentally left in.

crates/database/src/app_data.rs
vkgnosis: This test is special in a way I didn't document. Transactions fail when one of their query execution fails. In this test I'm testing a failing query and I keep using the database afterwards. If that happened in a transaction then the queries after the first failure would all fail because the transaction as a whole failed.

crates/orderbook/src/database/orders.rs
vkgnosis: Good catch. These shouldn't unwrap at all. I don't know what I was thinking when I added unwrap here.

crates/orderbook/src/database/orders.rs
```diff
@@ -179,18 +176,24 @@ impl OrderStoring for Postgres {
...
-                    insert_order(&order, transaction).await?;
-                    if let Some(quote) = quote {
-                        insert_quote(&order.metadata.uid, &quote, transaction).await?;
-                    }
-                    Ok(())
-                }
-                .boxed()
-            })
-            .await
+        let mut ex = connection.begin().await.unwrap();
```
nlordell, line 179: This should also not `unwrap` correct?

crates/orderbook/src/orderbook.rs
```diff
@@ -91,13 +92,29 @@ pub enum AddOrderError {
...
+    fn from_insertion(err: InsertionError, order: &Order) -> Self {
         match err {
             InsertionError::DuplicatedRecord => AddOrderError::DuplicatedOrder,
             InsertionError::DbError(err) => AddOrderError::Database(err),
+            InsertionError::AppDataMismatch(existing) => AddOrderError::AppDataMismatch {
+                contract_app_data: order.data.app_data,
+                // Unwrap because this error can only occur if full app data was set.
+                provided: order.metadata.full_app_data.clone().unwrap(),
+                // Unwrap because we only store utf-8 full app data.
+                existing: String::from_utf8(existing).unwrap(),
```
nlordell, line 116: I wouldn't unwrap here, as this assumes a specific format for `app_data` which is technically "outside the scope of this module". You can just use `String::from_utf8_lossy` to make this "fail-proof".

nlordell (approved):
LGTM. Some small nits, but nothing blocking approval (the `unwrap` issue was already acknowledged and assume will be addressed before merging).

crates/orderbook/src/orderbook.rs
vkgnosis: The problem with that is that we could get silent changes to app data in the conversion which would later confuse us. I could add an error log but not sure if that is better than unwrap for something that shouldn't happen.

nlordell: > The problem with that is that we could get silent changes to app data in the conversion which would later confuse us

This is just error context that we return in the API right? If anything I find this makes encoding issues more visible?

vkgnosis: Went with the log.

Issue #1205, opened 2023-02-13
Smart Contract Buy Eth Support without Access Lists


nlordell:
Related to #50 which enabled Smart Contracts to receive proceeds from orders with the marker `BUY_ETH_ADDRESS` using access lists. More context for why it was necessary is included in the issue.

This issue specifically captures the work to support Smart Contracts to receive Ether **without** relying on access list support (as it is not always available - for example when doing relayed settlement transactions).

Access lists support for Smart Contract receivers (added in #846) is a mechanism that allows us to "warm up" the receiver account and accessed storage slots so that the on-chain `payable(receiver).transfer` call does not exceed the 2300 gas stipend. The other mechanism for doing the same thing is to warm up the address as part of a pre-interaction. This works because pre-interactions forward all remaining gas to the contract, and so won't run into the same issues as `transfer` with its limited gas stipend. Essentially, the interaction we want to include for warming up Smart Contract Ether receiver addresses is:

```js
{
  target: receiver, // the receiver address
  value: 0, // 0-value
  callData: "0x", // empty calldata bytes
}
```

Since this call will happen from the Settlement contract's calling context, some extra security considerations should be taken. Specifically, I think it **may** make sense to trampoline this call over another intermediary contract in order to "un-sudo" the call. Specifically, I propose we use the [`MultiSendCallOnly`](https://github.com/safe-global/safe-contracts/blob/main/contracts/libraries/MultiSendCallOnly.sol) which was deployed by the Safe team on all networks that we support (as well as being audited and known to be Safe - pun intended).

PR #1575, opened 2023-06-08
Implement quote verification

MartinquaXD:
Implements stricter quote verification logic.
Previously the `TradeVerifier` logic tried the best it could to simulate trades where we didn't know the `from` address or were lacking balances. This can cause a lot of subtle issues but since we have stricter assumptions for validating quotes (compared to price estimates) we can take advantage of them to make the simulation more robust and more similar to reality.

The new verification logic assumes the `trader` has the required `sell_token` (or at least gets them via pre-interactions). With this very big assumption we can basically simulate the trade exactly as it was suggested by the `TradeFinder`.
That means we first have to build an encoded settlement from the `QuoteQuery` and `Trade`. This has a few interesting details:
1. We allow for the actual trade to be worse than the `Trade` promised (lower buy_amount or higher sell amount depending on the order `kind`). The advantage is that when some API proposes trades that don't respect the price they promise we can compute the actual price after the simulation by looking at key balances throughout the simulation. The same idea was used in the previous simulation logic but the new implementation has to be a bit different since we have to trigger the balance checks inside the settlement (triggered by custom interactions in the settlement).
2. `ETH` is not an `ERC20` token but the `driver` abstracts that fact away. At the moment we don't have this luxury for the trade simulation which is why we have to add the same workarounds for paying `ETH` as the `driver` would. Most importantly this means we have to manually unwrap the expected amount of `WETH`.
Another consideration is that we also want to be able to pay `ETH` to smart contracts. The driver supports that by generating the required access lists. In the trade simulation we can do it a little easier by simply adding a transaction sending 0 `ETH` to the `receiver` ([more context](https://github.com/cowprotocol/services/issues/1205)).
3. Since the smart contract needs to check that the order is actually allowed to be executed we overwrite the `trader` address with a smart contract that implements `isValidSignature()` for `EIP-1271` that always returns true.

Then we set up the simulation with a bunch of state overrides (to impersonate `solver` and `trader`) and start the simulation.
1. The simulation tx gets triggered by the `solver` as it also would in the real case. That means the solver has to pay for the gas so the `trader` doesn't actually have to have any `ETH` for the simulation to work.
2. However, the `trader` might not have the required `WETH` (if they sell that) or approvals set for the trade. Because we don't want to force actual users to wrap `ETH` or approve our contract before they actually see a price they like we have to do that before the simulation. Due to security constraints this has to happen in the context of the `trader` address so the `Solver` has to call a helper function on the `Trader` that takes care of things.
3. Now the `Solver` can simply execute the settlement while monitoring the gas usage. We monitor the gas usage because the API might not only report bad prices but also bad gas estimates. This means that the API doesn't even have to report any gas estimate at all (which also enables a very important optimization for the `1Inch` API).
4. Lastly we just have to parse the output of the `Solver` helper function to get the requested balances and gas used and craft an `Estimate` from that. The `gas_used` reported by the simulation will be slightly inflated so we also have to discount any gas that was used simply for simulations purposes but not for the actual trade itself.

TLDR;
The new simulation expects `from` to have the required balances (or get them during pre-interactions) but has these nice features:
* simulates actual settlement
* supports paying `ETH` (also to smart contracts)
* does not overwrite the authentication logic (quotes from deny-listed solvers will always fail)
* uses correct `from` and `solver` address (tokens with custom logic for certain addresses will simulate as in reality)
* traders can get quotes without having the contract approved
* traders can get quotes for selling `WETH` even when they currently only hold `ETH`
* more accurate gas used as no heuristics for settlement overhead are required
* works with all supported token sources/destinations and considers their gas costs

### Test Plan
Currently just manual tests

### Release
The change started by this PR is still incomplete and breaks the current trade verification support. I think this is okay since we currently do these simulations in staging and will replace them soon anyway.
Let me know if you disagree.


MartinquaXD:
@nlordell I tried to streamline the gas estimation logic in [2b06659](https://github.com/cowprotocol/services/pull/1575/commits/2b06659c8771c80d8abe563353bcaed8fcc1dd49). The hope is that I can make costs of measuring gas more accurate by reducing the incurred overhead (use `gasleft()` instead of `doMeteredCall()`) and by accounting for the biggest additional costs in solidity itself (at this point hopefully just function call overhead for `store_balance()`).
Do you think this is an improvement or should we stick with the heuristic I have at the moment?

vkgnosis (approved):
Good detailed description. I can't follow all the EVM stuff but the Rust code makes sense.

crates/contracts/solidity/PhonyERC20.sol
```diff
@@ -1,173 +0,0 @@
-// SPDX-License-Identifier: MIT
```
nlordell, line 1: Goodbye... 😢

crates/contracts/solidity/PhonyERC20.sol
```diff
@@ -1,173 +0,0 @@
...
-pragma solidity ^0.8.17;
-
-import { Caller } from "./libraries/Caller.sol";
-import { SafeERC20 } from "./libraries/SafeERC20.sol";
-
-/// @dev A phony ERC20 implementation, that replaces the bytecode of an existing
-/// on-chain contract and allows minting to arbitrary addresses. This can be
-/// used to fund an account when one with a known balance can't be found.
-contract PhonyERC20 {
```
nlordell, line 10: Goodbye... 😢

crates/contracts/solidity/Solver.sol
```diff
@@ -0,0 +1,84 @@
...
+            _queriedBalances.push(IERC20(token).balanceOf(owner));
+        }
+        // Account for costs of gas used outside of metered section.
+        uint256 measurementOverhead;
+        if (_queriedBalances.length == 1) {
+            // reading cold storage costs more which results in higher overhead for the first call
+            measurementOverhead = 13355;
+        } else {
+            measurementOverhead = 4460;
+        }
```
nlordell, line 81: You can make this slightly more succinct with:
```suggestion
        _queriedBalances.push(
            token == 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE
                ? owner.balance
                : IERC20(token).balanceOf(owner)
        );
        // Account for costs of gas used outside of metered section. Noting that
        // reading cold storage costs more which results in higher overhead for the first call
        uint256 measurementOverhead = _queriedBalances.length == 1 ? 13355 : 4460;
```

Personally, I kind of prefer the `if/else` for the first one as I find it easier to read, but for the second one (since they are simple constants) I find the ternary operator a bit nicer.

Fine as is though!

crates/contracts/solidity/Trader.sol
```diff
@@ -13,78 +13,72 @@ contract Trader {
...
-            appData: bytes32(0),
-            feeAmount: 0,
-            flags: 0x40, // EIP-1271
-            executedAmount: 0,
-            signature: abi.encodePacked(address(this))
-        });
+    // The `Trader` contract gets deployed on the `from` address of the quote.
+    // Since the `from` address might be a safe or other smart contract we still
+    // need to make the `Trader` behave as the original `from` would have in
+    // case some custom interactions rely on that behvavior.
```
nlordell, line 30:
```suggestion
    // case some custom interactions rely on that behavior.
```

crates/contracts/solidity/Trader.sol
```diff
@@ -13,78 +13,72 @@ contract Trader {
...
+    fallback() external payable {
+        _fallback();
+    }
+    receive() external payable {
+        _fallback();
+    }
+    function _fallback() private {
+        bytes memory rdata = TRADER_IMPL.doDelegatecall(msg.data);
+        assembly { return(add(rdata, 32), mload(rdata)) }
+    }
```
nlordell, line 42: This might push receiving ETH over the 2300 gas stipend (which would cause buying ETH to fail in simulation). I would actually on `receive()` just do nothing (in fact, SC wallets that don't "do nothing" aren't properly supported by the Settlement contract anyway).

Concretely, I would change this to:
```suggestion
    fallback() external payable {
        bytes memory rdata = TRADER_IMPL.doDelegatecall(msg.data);
        assembly { return(add(rdata, 32), mload(rdata)) }
    }
    receive() external payable {}
```

crates/contracts/solidity/Trader.sol
```diff
@@ -13,78 +13,72 @@ contract Trader {
...
-            traderBalances[i] += IERC20(tokens[i]).balanceOf(address(this)).toInt();
-            settlementBalances[i] += IERC20(tokens[i]).balanceOf(address(SETTLEMENT)).toInt();
+        if (sellToken == nativeToken) {
+            uint256 availableBalance = IERC20(sellToken).balanceOf(address(this));
+            if (availableBalance < sellAmount) {
+                // Simulate wrapping the missing `ETH` so the user doesn't have to spend gas
+                // on that just to get a quote. If they are happy with the quote and want to
+                // create an order they will actually have to do the wrapping, though.
+                (bool sent, ) = nativeToken.call{value: sellAmount - availableBalance}("");
+                require(sent, "'from' address does not have enough (W)ETH balance");
```
nlordell, line 67: I _think_ you should be able to do:
```suggestion
                nativeToken.deposit{value: sellAmount - availableBalance}();
```

This will automatically revert on failure.

crates/contracts/solidity/Trader.sol
```diff
@@ -13,78 +13,72 @@ contract Trader {
...
+            }
         }
+
+        uint256 currentAllowance = IERC20(sellToken).allowance(address(this), address(SETTLEMENT.vaultRelayer()));
+        if (currentAllowance < sellAmount) {
+            // Simulate an approval to the settlement contract so the user doesn't have to
+            // spend gas on that just to get a quote. If they are happy with the quote and
+            // want to create an order they will actually have to do the approvals, though.
+            IERC20(sellToken).safeApprove(address(SETTLEMENT.vaultRelayer()), 0);
```
nlordell, line 76: Would include a comment as to why we need to reset the allowance.

crates/contracts/solidity/Trader.sol
```diff
@@ -13,78 +13,72 @@ contract Trader {
...
+            IERC20(sellToken).safeApprove(address(SETTLEMENT.vaultRelayer()), type(uint256).max);
+        }
+
+        (bool sent2, ) = receiver.call{value: 0}("");
+        require(sent2, "warming the storage for sending ETH to smart contract addresses failed");
```
nlordell, line 81: Cleaver. I don't think you need to check the `success` here (as the transfer would fail if done as part of a settlement).

A concrete reason for this is that an order can send ERC20 tokens to a contract that doesn't necessarily support receiving Ether, and unless the settlement is transferring Ether to the receiver, we don't want to quote simulation to fail:

```suggestion
        // warming the storage for sending ETH to smart contract addresses
        receiver.call{value: 0}("");
```

crates/shared/src/trade_finding.rs
```diff
@@ -38,6 +38,7 @@ pub struct Trade {
     pub out_amount: U256,
     pub gas_estimate: u64,
     pub interactions: Vec<Interaction>,
+    // TODO add solver address for simulations
```
nlordell, line 41: What are you using ATM?

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -80,24 +84,49 @@ impl TradeEstimator {
     }
 
     async fn estimate(&self, query: Query) -> Result<Estimate, PriceEstimationError> {
+        let owner = query.from.unwrap_or_default();
```
nlordell, line 87: Should we just error out early? Personally I don't think its worth doing a proper trade verification if `query.from` is `None` as it will almost certainly fail, and uses valuable API requests to DEX aggregator APIs and our node.

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -119,126 +148,211 @@ impl Inner {
...
+        OrderKind::Sell => vec![trade.out_amount, query.in_amount],
+        OrderKind::Buy => vec![query.in_amount, trade.out_amount],
+    };
+
+    // Configure the most disadvantageous trade possible. Should the trader not
+    // receive the amount promised by the [`Trade`] the simulation will still work
+    // and we can compute the actual [`Trade::out_amount`] afterwards.
+    let (sell_amount, buy_amount) = match query.kind {
+        OrderKind::Sell => (query.in_amount, 0.into()),
+        OrderKind::Buy => (U256::MAX, query.in_amount),
```
nlordell, line 183: This will potentially overflows in the contract. I suggest `1 << 160` or something similar.

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -119,126 +148,211 @@ impl Inner {
...
+    };
+    let query_balance = solver.methods().store_balance(token, owner);
+    let query_balance = Bytes(query_balance.tx.data.unwrap().0);
+    let interaction = (solver.address(), 0.into(), query_balance);
+    settlement.interactions[1].insert(0, interaction.clone());
+    match query.kind {
+        // query `receiver` balance right after paying out funds (first post-interaction)
+        OrderKind::Sell => settlement.interactions[2].insert(0, interaction),
+        // query `settlement` balance right before paying out funds (last regular interaction)
+        OrderKind::Buy => settlement.interactions[1].push(interaction),
```
nlordell, line 245: This is also correct, but I'm curious as to why we don't `settlement.interactions[2].insert(0, interaction)` in both cases? This feels like the most appropriate place to do it (and, for silly tokens that can modify balances from other tokens, like the `sell_token`, would get captured).

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -119,126 +148,211 @@ impl Inner {
...
+            settlement,
+            native_token,
         }
     }
 
-    async fn verify(&self, query: Query, trade: Trade) -> Result<Estimate> {
-        let trader = dummy_contract!(Trader, query.from.unwrap_or(Self::DEFAULT_TRADER));
-        let sell_token_code = self.code_fetcher.code(query.sell_token).await?;
+    async fn verify(&self, query: QuoteQuery, trade: Trade) -> Result<Estimate> {
+        // TODO: add solver address to [`Trade`]; for now we simply use Quasilab's
```
nlordell, line 269: so-nit-picky-its-not-worth-mentioning: I would use the legacy solver address 😛.

crates/shared/src/price_estimation/trade_finder.rs
```diff
@@ -119,126 +148,211 @@ impl Inner {
...
+        // query `receiver` balance right after paying out funds (first post-interaction)
+        OrderKind::Sell => settlement.interactions[2].insert(0, interaction),
+        // query `settlement` balance right before paying out funds (last regular interaction)
+        OrderKind::Buy => settlement.interactions[1].push(interaction),
+    }
+    settlement
+}
+
 impl TradeVerifier {
     const DEFAULT_GAS: u64 = 8_000_000;
```
nlordell, line 251: Lets use the block gas limit. This caused issues before:
```suggestion
    const DEFAULT_GAS: u64 = 30_000_000;
```

nlordell: Actually nevermind. This isn't part of this PR and using the block gas limit has its own problems.

nlordell (commented):
Looks great!

Just some small nits. One concern is that this would break Gnosis Chain and Goerli staging services because of the hardcoded solver address.

nlordell:
> The change started by this PR is still incomplete

What is specifically missing?

crates/shared/src/price_estimation/trade_finder.rs
MartinquaXD: After thinking more about it I think we could indeed query the balances at the same point in time.
What I was thinking about is the case where `buy_token == sell_token` which **theoretically** could be handled by the contract if a solver finds an arbitrage opportunity that pays out in the same token you put in. In that case I thought it would be more accurate to query at the end of the regular interactions but I now realize that the whole approach would not work in that case.
If we were to handle such a case the price for the token would be negative which doesn't really make sense for quotes.
Will simplify that case to always query the balance after paying out the `buy_token`.

MartinquaXD:
> What is specifically missing?

All the logic that is needed to correctly pipe through all the data that is now needed by the new approach.
The address of the solver in the response, a few fields that are now needed in the `QuoteQuery` and we need to somehow let the verification logic know that the quote is not actually for `WETH` but for native `ETH` (in case the user requested that).

nlordell (approved):

crates/shared/src/price_estimation/trade_finder.rs
MartinquaXD: Now I remember why I did it this way. If we error out some unit tests are complaining and I didn't want to blow up the diff even further because the PR in itself is already broken.
Will continue to simply `warn` for now be leave the PR unmerged to see if the remaining changes are simply enough to be merged together with this PR to not temporarily break verification.

PR #1576, opened 2023-06-09
Prevent stable to stable internalization on gnosis solvers

MartinquaXD:
Sometimes solvers propose stable to stable trades that produce good prices but require enormous gas overhead. Normally these would be discarded but due to naive buffer usage rules it's technically allowed to internalize these trades which gets rid of their high gas cost. Because these trades are not economically viable and ultimately will be paid out of the solver's pocket we want to prevent those internalizations. Because our current internalization is very crude (all or nothing) we disable this optimization for all solutions that contain some stable-to-stable order.

Context: [1](https://cowservices.slack.com/archives/C0361CDD1FZ/p1685707904853019), [2](https://cowservices.slack.com/archives/C036JAGRQ04/p1686241248700439)

### Test Plan
unit test

### DIFF
--- a/crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
+++ b/crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
@@ -79,6 +79,23 @@ mod tests {
         primitive_types::H160,
     };
 
+    fn trade(sell_token: H160, buy_token: H160) -> Trade {
+        Trade {
+            order: Order {
+                data: OrderData {
+                    sell_token,
+                    buy_token,
+                    sell_amount: 1.into(),
+                    buy_amount: 1.into(),
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+            executed_amount: 1.into(),
+            ..Default::default()
+        }
+    }
+
     #[test]
     fn test_is_only_selling_trusted_tokens() {
         let good_token = H160::from_low_u64_be(1);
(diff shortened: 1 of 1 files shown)


crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
```diff
@@ -71,17 +93,56 @@ mod tests {
...
+        assert!(some_stable_to_stable_trade(&settlement));
+        let settlement = Settlement::with_default_prices(vec![trade(usdc, usdt)]);
+        assert!(some_stable_to_stable_trade(&settlement));
+        let settlement = Settlement::with_default_prices(vec![
+            trade(usdc, usdt),
+            trade(non_stable_a, non_stable_b),
+        ]);
+        assert!(some_stable_to_stable_trade(&settlement));
+        let settlement = Settlement::with_default_prices(vec![trade(non_stable_a, non_stable_b)]);
+        assert!(!some_stable_to_stable_trade(&settlement));
```
fhenneke, line 146: Could you add a test for a  stable to non_stable trade as well? (That is where code I write would be most likely to fail.)

fhenneke:
I cannot really comment on the code but the general approach looks good to me and is what we had in mind.

crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
MartinquaXD: [299dd17](https://github.com/cowprotocol/services/pull/1576/commits/299dd179e50efddef7ac8a5b48fe1f033c7c1d3f)

cowfee (approved):

crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
```diff
@@ -41,6 +53,23 @@ fn is_only_selling_trusted_tokens(
...
+        H160(hex_literal::hex!(
+            "6B175474E89094C44Da98b954EedeAC495271d0F" // DAI
+        )),
+        H160(hex_literal::hex!(
+            "dAC17F958D2ee523a2206206994597C13D831ec7" // USDT
+        )),
+    ];
+    settlement.traded_orders().any(|o| {
+        stable_coins.contains(&o.data.sell_token) && stable_coins.contains(&o.data.buy_token)
+    })
```
nlordell, line 70: Can we additionally add a condition that:
1. It involves stable-stable coins
2. The internalized interaction would trade at more than 1:1?

The 2. condition allows us to internalize in the general case to keep on saving gas money. Stable-stables are a lot of our trades.

nlordell:
Icky... but it is an "easy way" to implement this.

nlordell (approved):

crates/solver/src/settlement_post_processing/optimize_buffer_usage.rs
nlordell: According to [this](https://cowservices.slack.com/archives/C0375NV72SC/p1686552438506019?thread_ts=1686315504.555289&cid=C0375NV72SC) the benefits of the second condition will be minimal, so I would say its not worth implementing.

PR #1580, opened 2023-06-13
Hook up new simulation logic

MartinquaXD:
Closes #1512 

I now have a rough game plan for verified quotes.
Instead of introducing a new price quality `verified` (#1512) we implicitly verify all quotes where `from` is set. All the other pieces of data (besides custom interactions) already exist in the request so we can easily bundle the data that is only needed for verification into a struct which is optional. Only if the struct is `Some` we do verifications.
Also with this new idea I would rather make every individual price estimator decide whether it supports trade verification (and return an error if it doesn't) than creating a new price estimator hierarchy that doesn't contain the ones that don't support quote verification.

Things left over which should be doable in separate PRs:
1. Make remaining estimators aware of new verification logic (`Balancer`, `Naive`, `Baseline`, `Quasimodo`, `Yearn`).
2. Make public solver address configurable with price estimators
3. Handle quotes for selling/buying native `ETH` (currently results in quotes for `WETH` which might cause custom interactions to fail)
4. Turn verification warnings into hard errors when feature is done

### Test Plan
Manual tests

### DIFF
--- a/crates/autopilot/src/database/onchain_order_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events.rs
@@ -47,7 +47,6 @@ use {
             is_order_outside_market_price,
             onchain_order_placement_error_from,
         },
-        price_estimation::Verification,
     },
     std::{
         collections::{HashMap, HashSet},
@@ -478,8 +477,7 @@ async fn parse_general_onchain_order_placement_data<'a>(
             }
             let (order_data, owner, signing_scheme, order_uid) = detailed_order_data?;
 
-            let quote_result =
-                get_quote(quoter, order_data, signing_scheme, &event, &quote_id).await;
+            let quote_result = get_quote(quoter, order_data, signing_scheme, &quote_id).await;
             let order_data = convert_onchain_order_placement(
                 &event,
                 event_timestamp,
@@ -537,7 +535,6 @@ async fn get_quote(
     quoter: &dyn OrderQuoting,
     order_data: OrderData,
     signing_scheme: SigningScheme,
-    order_placement: &ContractOrderPlacement,
     quote_id: &i64,
 ) -> Result<Quote, OnchainOrderPlacementError> {
     let quote_signing_scheme = convert_signing_scheme_into_quote_signing_scheme(
@@ -558,16 +555,14 @@ async fn get_quote(
         buy_amount: order_data.buy_amount,
         fee_amount: order_data.fee_amount,
         kind: order_data.kind,
-        verification: Some(Verification {
-            // Original quote was made from user account, and not necessarily from owner.
-            from: order_placement.sender,
-            receiver: order_data.receiver.unwrap_or(order_placement.sender),
-            sell_token_source: SellTokenSource::Erc20,
-            buy_token_destination: BuyTokenDestination::Erc20,
-            // TODO do we have to get custom interactions here?
-            pre_interactions: vec![],
-            post_interactions: vec![],
-        }),
+        // Verified quotes always have prices that are at most as good as unverified quotes but can
+        // be lower.
+        // If the best quote we can find or compute on the fly for this order suggests a worse
+        // price than the order was created with it will cause a `QuoteNotFound` or other error.
+        // Orders indexed with errors are not eligible for automatic refunding.
+        // Because we want to be generous with refunding EthFlow orders we therefore don't request a
+        // verified quote here on purpose.
+        verification: None,
     };
     get_quote_and_check_fee(
         quoter,
--- a/crates/autopilot/src/database/onchain_order_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events.rs
@@ -47,6 +47,7 @@ use {
             is_order_outside_market_price,
             onchain_order_placement_error_from,
         },
+        price_estimation::Verification,
     },
     std::{
         collections::{HashMap, HashSet},
@@ -557,8 +558,16 @@ async fn get_quote(
         buy_amount: order_data.buy_amount,
         fee_amount: order_data.fee_amount,
         kind: order_data.kind,
-        // Original quote was made from user account, and not necessarily from owner.
-        from: order_placement.sender,
+        verification: Some(Verification {
+            // Original quote was made from user account, and not necessarily from owner.
+            from: order_placement.sender,
+            receiver: order_data.receiver.unwrap_or(order_placement.sender),
+            sell_token_source: SellTokenSource::Erc20,
+            buy_token_destination: BuyTokenDestination::Erc20,
+            // TODO do we have to get custom interactions here?
+            pre_interactions: vec![],
+            post_interactions: vec![],
+        }),
     };
     get_quote_and_check_fee(
         quoter,
--- a/crates/shared/src/order_quoting.rs
+++ b/crates/shared/src/order_quoting.rs
@@ -135,24 +135,12 @@ pub struct QuoteParameters {
     pub sell_token: H160,
     pub buy_token: H160,
     pub side: OrderQuoteSide,
-    pub from: H160,
+    pub verification: Option<Verification>,
     pub signing_scheme: QuoteSigningScheme,
 }
 
 impl QuoteParameters {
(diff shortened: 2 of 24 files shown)


crates/shared/src/order_quoting.rs
```diff
@@ -660,11 +660,25 @@ impl From<&OrderQuoteRequest> for PreOrderData {
 
 impl From<&OrderQuoteRequest> for QuoteParameters {
     fn from(request: &OrderQuoteRequest) -> Self {
+        let verification = if request.from == Default::default() {
+            None
+        } else {
```
vkgnosis, line 665: Does it make sense to keep special logic for the from address being zero? The code around this is confusing to me because we don't document that zero from address has special meaning. `OrderQuoteRequest` is created by a user of the post quote api, in this API the from field is mandatory and we do not document any special behavior with zero.

crates/shared/src/order_validation.rs
```diff
@@ -510,14 +510,27 @@ impl OrderValidating for OrderValidator {
             .await
             .map_err(ValidationError::Partial)?;
 
+        let verification = match order.from {
+            Some(from) => Some(Verification {
+                from,
```
vkgnosis, line 515: `OrderCreation::from` is documented as 

> If set, the backend enforces that this address matches what is decoded as the signer of the signature. This helps catch errors with invalid signature encodings as the backend might otherwise silently work with an unexpected address that for example does not have any balance.

It doesn't make sense to me, to use it in this way here.

MartinquaXD: Does it not make sense to you that `from` is the address that needs to have the `sell_token` or that we always verify a quote when `from` is set and verified to be the `owner` of the order?
Should I extend the opanapi docs to say that we verify quotes when that field is set?

vkgnosis: The openapi docs do not mention that this changes anything about quote verification. That is confusing. I'm not sure this concept makes sense to users though. Why not always verify with the order's owner? Would also be good to document `pub verification: Option<Verification>` to explain what impact None vs Some has.

MartinquaXD: Ahh, I didn't notice that order `owner` and `from` don't have to be the same for this kind of request. 🤦‍♂️ 
If `from` is `None` we could still recover the actual `owner` from the signature but the verification logic would not kick in because `from` is `None`.
Good catch!
Will fix that and add more documentation.

crates/autopilot/src/database/onchain_order_events.rs
```diff
@@ -557,8 +558,16 @@ async fn get_quote(
...
-        // Original quote was made from user account, and not necessarily from owner.
-        from: order_placement.sender,
+        verification: Some(Verification {
+            // Original quote was made from user account, and not necessarily from owner.
+            from: order_placement.sender,
+            receiver: order_data.receiver.unwrap_or(order_placement.sender),
+            sell_token_source: SellTokenSource::Erc20,
+            buy_token_destination: BuyTokenDestination::Erc20,
+            // TODO do we have to get custom interactions here?
+            pre_interactions: vec![],
```
nlordell, line 568: Good question - my intuition is "no".

In theory, to correctly verify the EthFlow transaction exactly as it will execute, we should do:
```
from: owner, // i.e. the EthFLow contract
// ...
pre_interactions: vec![ethflow_pre_interaction()], // from <https://github.com/cowprotocol/services/blob/615cb7cc1ee7f45a35670948868810032882d9d2/crates/autopilot/src/database/onchain_order_events/ethflow_events.rs#L124>
```

However! the way the simulation currently works (where it wraps ETH for the `from` address automatically), it is actually not needed.

I would keep this as is for simplicity sake (and double check my assessment is not incorrect).

crates/shared/src/order_quoting.rs
```diff
@@ -660,11 +662,32 @@ impl From<&OrderQuoteRequest> for PreOrderData {
...
+            Some(Verification {
+                from: request.from,
+                receiver: request.receiver.unwrap_or(request.from),
+                sell_token_source: request.sell_token_balance,
+                buy_token_destination: request.buy_token_balance,
+                // TODO get from request
+                pre_interactions: vec![],
+                post_interactions: vec![],
+            })
+        };
```
nlordell, line 684: I would remove this bit. Specifically, what I would do instead is in the `/quote` endpoint handler do a `if quote.quality == ::Verified and quote.from.is_zero() { err!() }` instead of having this "0 has special semantics" leak here.

nlordell (approved):
LGTM, just one nit about "where" we should disallow quote verification for `from = address(0)`. Personally, I think this makes sense at the API route handling and not at the `PriceEstimating` implementation level.

crates/shared/src/order_quoting.rs
MartinquaXD: > Instead of introducing a new price quality verified (https://github.com/cowprotocol/services/pull/1512) we implicitly verify all quotes where from is set.

I was thinking about not introducing the `Verified` price quality after all (the PR is still not merged as I was not 100% sure how the whole thing would fit together).
If we indeed want to introduce the `Verified` price quality I agree that this special case should be dropped. But I thought it would be nice this way that the frontend wouldn't have to change for now.
LMK what you think about automatic verified quotes vs. new price quality.

nlordell: > LMK what you think about automatic verified quotes vs. new price quality.

Would this would break the existing API? As in, do you want to do a `token.balanceOf(from)` and decide whether or not to verify (which adds some latency)? I.e. what is your condition for deciding to verify quotes?

MartinquaXD: With this PR we would always verify quotes when we can (i.e. we have a non-0 `from`).
This would add latency but hopefully not too much. I can also "feature gate" it behind the `verified` price quality so we don't really change anything immediately.

crates/autopilot/src/database/onchain_order_events.rs
MartinquaXD: When we index this event the user already sent the funds to the `EthFlow` contract, right?
In that case we actually have to do it correctly because the user might no longer have enough `WETH` for our unwrap trick during the simulation.

nlordell: Dang, you are right.

MartinquaXD: That also begs the question how the quote flow for ethflow orders has to work.
When a user requests a quote for `ETH -> some token`. We have to make a choice whether to simulate from the user address or ethflow address.
If we simulate from the user address the quote we try to look up when indexing the ethflow event will have a different owner and we won't find it.
If we simulate from the ethflow contract we would only have to fake its `ETH` balance. That way the quote requested by the user and looked up when indexing the event should be identical so we can find it.
I guess the most reasonable approach is to drop the EthFlow verification in this PR and address that specifically in a separate PR.

MartinquaXD: Ahh, this is particularly annoying.
Either we only have to look up the quote because it already exists since the user requested it before creating the order. Then the owner should be the user (so we compute subsidies correctly) and we need no ethflow interactions.
Or the quote doesn't exist and we create one on the fly then the owner has to be the ethflow contract to pass the simulation at that point in time.

nlordell: > If we simulate from the user address the quote we try to look up when indexing the ethflow event will have a different owner and we won't find it.

We can just "mint" ETH (i.e. specifying a `balance` in the state override for the `from` address) for the `from` address. I think this is the easiest way around this. It also means we can always simulate (W)ETH -> Token quotes which is cool.

MartinquaXD: Had a call with Nick which concluded that we won't be requesting verified quotes when indexing onchain-orders.
The UX of that mechanism is already worse than native orders and verified quotes could make it even worse (also very complicated to get exactly right). Explained in [985658f](https://github.com/cowprotocol/services/pull/1580/commits/985658fff06e2cfca32cbe73d64b2ef9a077292d).

PR #1581, opened 2023-06-13
Add route for full app data fetching

vkgnosis:
Part of the new app data design. Allows fetching full app data by contract app data hash. You could already get an order's full app data through its order uid but now you can also do so without knowing the order.

### Test Plan

CI

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -375,7 +375,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/VersionResponse"
-  /api/v1/app_data/by_hash/{app_data_hash}:
+  /api/v1/app_data/{app_data_hash}:
     get:
       summary: Retrieve full app data from contract app data hash.
       parameters:
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -375,6 +375,22 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/VersionResponse"
+  /api/v1/app_data/by_hash/{app_data_hash}:
+    get:
+      summary: Retrieve full app data from contract app data hash.
+      parameters:
+        - in: path
+          name: app_data_hash
+          schema:
+            $ref: "#/components/schemas/AppDataHash"
+          required: true
+      responses:
+        200:
+          description: Full app data
+          content:
+            application/json: {}
+        404:
+          description: App data hash doesn't exist.
 components:
   schemas:
     TransactionHash:
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -390,7 +390,7 @@ paths:
           content:
             application/json: {}
         404:
-          description: App data hash doesn't exist.
+          description: No full app data stored for this hash.
 components:
   schemas:
     TransactionHash:
--- a/crates/orderbook/src/api/get_app_data.rs
+++ b/crates/orderbook/src/api/get_app_data.rs
@@ -1,0 +1,43 @@
+use {
+    crate::database::Postgres,
+    anyhow::Result,
+    model::app_id::AppDataHash,
+    reqwest::StatusCode,
+    std::convert::Infallible,
+    warp::{
+        reply::{with_header, with_status},
+        Filter,
+        Rejection,
+        Reply,
+    },
+};
+
+pub fn request() -> impl Filter<Extract = (AppDataHash,), Error = Rejection> + Clone {
+    warp::path!("v1" / "app_data" / "by_hash" / AppDataHash).and(warp::get())
+}
+
+pub fn get(
+    database: Postgres,
+) -> impl Filter<Extract = (Box<dyn Reply>,), Error = Rejection> + Clone {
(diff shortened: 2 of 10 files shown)


crates/orderbook/openapi.yml
```diff
@@ -375,6 +375,22 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/VersionResponse"
+  /api/v1/app_data/by_hash/{app_data_hash}:
```
nlordell, line 378: API naming nit:

```suggestion
  /api/v1/app-data/{app_data_hash}:
```

I think "by hash" is superfluous (as it is obviously the way to identify app-hashes).

crates/orderbook/src/api/get_app_data.rs
```diff
@@ -0,0 +1,43 @@
...
+    database: Postgres,
+) -> impl Filter<Extract = (Box<dyn Reply>,), Error = Rejection> + Clone {
+    request().and_then(move |contract_app_data: AppDataHash| {
+        let database = database.clone();
+        async move {
+            let result = database.get_full_app_data(&contract_app_data).await;
+            Result::<_, Infallible>::Ok(match result {
+                Ok(Some(response)) => {
+                    let response = with_status(response, StatusCode::OK);
+                    let response = with_header(response, "Content-Type", "application/json");
```
MartinquaXD, line 29: Is the header `Content-Type application/json` still correct if you only return a string?
If not, `Content-Type: text/html; charset=utf-8` would probably be desired.

MartinquaXD (approved):
Looks good overall.

crates/orderbook/openapi.yml
nlordell: Typically, in "REST" terms (although we are very loosely REST-y) you typically organize routes as `GET resource/${id}/sub-resouces...`. So, for fetching "app-data", you would identify it by its ID (for which only hashes really make sense IMO).

If you were, say, to fetch APP data by order ID, you could organize it like: `GET order/${orderUid}/app-data`. Or, another example, fetch all app-datas by transaction: `GET transaction/${transactionHash}/app-datas`, etc.

For example, I'm also not a fan of the `solver_competitions/...` routes either, and would probably have done it like:
```
GET auction/${auctionId}/solver_competition
GET transaction/${txHash}/solver_competition
```

Regarding the kebab case vs snake case, all our other routes use snake case so we should keep it that way for consistency. (so disregard my suggestion of `app_data` -> `app-data`).

None of this is important and has no technical reason behind it, like I said, its just bike-shedding and nit-picking (I approved the PR as is).

crates/orderbook/src/api/get_app_data.rs
nlordell: Good point. The more I think of it, I don't think we should limit app-data to being JSON. For example, in the `contracts` repo we use this for fee withdrawals:

https://github.com/cowprotocol/contracts/blob/5c25837f585d79dc5450fb80f66c41f596c3198d/src/tasks/dump.ts#L61

This would be a small change to the specification, but it would require changes to:

https://github.com/cowprotocol/services/blob/615cb7cc1ee7f45a35670948868810032882d9d2/crates/shared/src/app_data.rs#L38-L46

Where it would accept anything as long as the maximum size is respected and just return a `Default::default()` if it fails to parse JSON or the field is missing.

vkgnosis: The header is correct because we enforce that app data is json. If we don't want to enforce that then indeed the content type shouldn't be json. If you want to be as general as possible then it should not even be string. It should be raw bytes.

To me it makes sense to enforce json. Our backend app data scheme is (going to be) defined on json. Why would we accept app data that doesn't match that? Imagine you allow arbitrary bytes or strings and an api user sets a document that is supposed to contain backend app but fails serde_json deserialization for some reason. Maybe a typo making it invalid json or maybe a number that is too large for u64 (assuming serde_json would error in this case). Then we would silently accept this and internally treat it as default app data.

vkgnosis:
This PR isn't important for anything else right now, so I don't mind discussing some details.

crates/orderbook/src/api/get_app_data.rs
nlordell: You have a good point. I agree that enforcing JSON to avoid API usage error makes sense (and this is the intended usage anyway). Also, this is what we specified initially, and I am derailing a bit 😛.

In the future we may also want to support a "Solidity friendly" app-data specification so that it can be computed on-chain (which would likely be EIP-712 struct hashing based, so "raw bytes"), but this is out of scope for now.

crates/orderbook/openapi.yml
```diff
@@ -375,6 +375,22 @@ paths:
...
+          schema:
+            $ref: "#/components/schemas/AppDataHash"
+          required: true
+      responses:
+        200:
+          description: Full app data
+          content:
+            application/json: {}
+        404:
+          description: App data hash doesn't exist.
```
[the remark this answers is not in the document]
vkgnosis: added

crates/orderbook/openapi.yml
vkgnosis: changed

PR #1587, opened 2023-06-14
Fetch full app data from IPFS when not specified

vkgnosis:
We want to continue to support IPFS based app data even when on order creation only the contract app data hash is specified. This PR changes how orders with only a contract app data hash are handled.

Before:
Order is accepted and we may or may not already have corresponding full app data in the database.

Now:
First try to find existing app data in database. If it doesn't exist try to find it on IPFS with the old and new method of CID generation. If it still doesn't exist then continue as before without full app data. If it does exist, then don't try to verify the hash of the found full app data because this might be the old way of storing on IPFS, which we cannot validate. Or it might be hardcoded app data that doesn't correspond to any IPFS data. Do verify that the schema is as expected.

The last step is a breaking change. Before this PR we would accept an order that has a contract app data hash that resolves to an IPFS document that isn't json (or not utf-8). We would accept it because we didn't have the full app data to validate. With this PR, when we find the full app data, we will validate the schema.

### Test Plan

CI e2e test, after merge I will configure our pinata ipfs gateway and do a manual test

### DIFF
--- a/crates/orderbook/src/ipfs.rs
+++ b/crates/orderbook/src/ipfs.rs
@@ -1,6 +1,5 @@
 use {
     anyhow::{anyhow, Context, Result},
-    futures::future::Either,
     model::app_id::AppDataHash,
     reqwest::{Client, StatusCode},
     url::Url,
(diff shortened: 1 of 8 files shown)


crates/shared/src/order_validation.rs
```diff
@@ -71,11 +71,18 @@ pub trait OrderValidating: Send + Sync {
...
+    /// contract app data hash without the full app data. In this case
+    /// `full_app_data_override` is used as the full app data and the contract
+    /// app data hash is not validated against it (the hash doesn't have to
+    /// match). The full app data is still otherwise validated.
     async fn validate_and_construct_order(
         &self,
         order: OrderCreation,
         domain_separator: &DomainSeparator,
         settlement_contract: H160,
+        full_app_data_override: Option<String>,
```
nlordell, line 85: Instead of an extra field, would it be slightly cleaner to have:

```
        payload.app_data = match payload.app_data {
            OrderCreationAppData::Hash { hash } => if let Some(fetched) = self.find_full_app_data(&hash).await {
                OrderCreationAppData::Full { .. }
            } else {
                OrderCreationAppData::Hash
            },
            other => other,
        };
```

crates/orderbook/src/ipfs.rs
```diff
@@ -0,0 +1,162 @@
...
+    };
+    match futures::future::select(std::pin::pin!(fetch(old)), std::pin::pin!(fetch(new))).await {
+        Either::Left((complete, pending)) | Either::Right((complete, pending)) => {
+            if complete.is_some() {
+                complete
+            } else {
+                pending.await
+            }
+        }
+    }
```
MartinquaXD, line 103: Looks like this could be replaced with [`select_ok`](https://docs.rs/futures/latest/futures/future/fn.select_ok.html). (with slight modifications)

MartinquaXD (approved):
Looks good!

crates/shared/src/order_validation.rs
vkgnosis: The problem with turning Hash into Full is that for full we calculate the hash in the verification logic. But the hash won't match for orders that weren't created with the new app data scheme.

vkgnosis: I do intuitively think this part can be done cleaner but I'm not sure yet what the best way is.

crates/orderbook/src/ipfs.rs
vkgnosis: done

crates/shared/src/order_validation.rs
nlordell: Fine by me.

PR #1588, opened 2023-06-14
Configure Account Balance Fetching Implementation

nlordell:
Follow up to #1586

This PR adds the required configuration options to setup a simulation based balance fetcher in the services. Note that this still needs to be done for the `driver`.

### Test Plan

Ran locally and verified the balance simulation logs when trying to create an order for a token for which I had no balance (but some allowance):

```
2023-06-14T15:03:05.197Z TRACE request{id=7}: shared::account_balances::simulation: simulated balances query=Query { owner: 0xdbc682d0ec19bd276fd6c2af4e0189b584bb31e1, token: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, source: Erc20 } amount=Some(1000000000) simulation=Simulation { token_balance: 0, allowance: 115792089237316195423570985008687907853269984665640564039457584007912951809973, effective_balance: 0, can_transfer: false }
```

### DIFF
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -159,7 +159,7 @@ pub struct Arguments {
     pub node_url: Url,
 
     /// An Ethereum node URL that supports `eth_call`s with state overrides to
-    /// be used exclusively for trade simulations.
+    /// be used for simulations.
     #[clap(long, env)]
     pub simulation_node_url: Option<Url>,
 
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -3,6 +3,7 @@
 
 use {
     crate::{
+        account_balances,
         current_block,
         ethrpc,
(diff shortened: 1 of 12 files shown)


vkgnosis (approved):

crates/shared/src/arguments.rs
```diff
@@ -147,13 +148,21 @@ pub struct Arguments {
...
+
     #[clap(flatten)]
     pub logging: LoggingArguments,
 
     /// The Ethereum node URL to connect to.
     #[clap(long, env, default_value = "http://localhost:8545")]
     pub node_url: Url,
 
+    /// An Ethereum node URL that supports `eth_call`s with state overrides to
+    /// be used exclusively for trade simulations.
```
MartinquaXD, line 162: nit:
> ...to be used exclusively for trade simulations.

Is no longer correct now that balances can also be queried with simulations.

MartinquaXD (approved):
LGTM

PR #1600, opened 2023-06-19
Hook Up Pre- and Post- Interactions

nlordell:
This PR adds the remaining glue code to enable pre-interactions. Namely:
- Parse custom interactions from app data
- Store custom interactions in the database

### Test Plan

Added some unit tests for the DB operations, and E2E test introduced in #1604.

### Release Notes

**This should not be enabled on Mainnet/Gnosis Chain until we properly trampoline the user interactions!**.

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -103,47 +103,6 @@ pub struct Order {
     pub surplus_fee_timestamp: Option<DateTime<Utc>>,
 }
 
-pub async fn insert_or_overwrite_interactions(
-    ex: &mut PgConnection,
-    uid_and_interaction: &[(OrderUid, Interaction)],
-) -> Result<(), sqlx::Error> {
-    for (order_uid, interaction) in uid_and_interaction.iter() {
-        insert_or_overwrite_interaction(ex, interaction, order_uid).await?;
-    }
-    Ok(())
-}
-
-pub async fn insert_or_overwrite_interaction(
-    ex: &mut PgConnection,
-    interaction: &Interaction,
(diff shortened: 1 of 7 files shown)


crates/database/src/orders.rs
```diff
@@ -277,6 +253,69 @@ pub fn is_duplicate_record_error(err: &sqlx::Error) -> bool {
...
+    interactions: &[Interaction],
+) -> Result<(), sqlx::Error> {
+    for interaction in interactions {
+        insert_interaction(ex, order, interaction).await?;
+    }
+    Ok(())
+}
+
+pub async fn insert_interaction(
+    ex: &mut PgConnection,
```
vkgnosis, line 292: This file also has `fn insert_or_overwrite_interaction`. It feels like they should be next to eachother.

crates/orderbook/src/database/orders.rs
```diff
@@ -140,7 +170,12 @@ async fn insert_order(order: &Order, ex: &mut PgConnection) -> Result<(), Insert
             } else {
                 InsertionError::DbError(err)
             }
-        })
+        })?;
+    database::orders::insert_interactions(ex, &order.uid, &interactions)
+        .await
```
vkgnosis, line 175: You probably want to insert the order and interactions in the same database transaction.

vkgnosis (approved):

MartinquaXD (approved):

crates/database/src/orders.rs
nlordell: 614890ad82304118ca91bb21742d95c540fbded6

PR #1604, opened 2023-06-19
Add E2E Test for Pre-Interaction

nlordell:
Follow up to #1600

This PR adds an E2E test to verify settling an order with a mandatory pre-interaction works. In particular, the E2E test creates an order with an EIP-2612 `permit` pre-interaction that sets the user allowance required for swapping.

### Notes

This PR introduces a flag for optimistically assuming that all orders with pre-interactions have sufficient balance and allowance. Why? Unfortunately, Hardhat does not support `eth_call` with state overrides, or `trace_callMany` - making our ability to simulate balances account for pre-interactions limited. This flag helps us work around this **but should not be enabled in production**.

### Test Plan

This is the test!


crates/shared/src/account_balances/arguments.rs
```diff
@@ -26,6 +26,15 @@ pub struct Arguments {
     /// `Tenderly` or `Web3ThenTenderly`.
     #[clap(long, env, value_enum)]
     pub account_balances_simulator: Option<CodeSimulatorKind>,
+
+    /// Whether or not to optimistically treat account balance queries with
+    /// pre-interactions as if sufficient token balance and allowance is always
+    /// available. Useful for partially supporting pre-interactions in
+    /// environments where the required simulation infrastructure is not
+    /// available (such as in E2E tests).
+    /// pre-interactions will
```
vkgnosis, line 35: Sentence cuts off

vkgnosis (approved):

crates/shared/src/account_balances/arguments.rs
nlordell: 🤦

crates/e2e/tests/e2e/setup/onchain_components.rs
```diff
@@ -170,6 +185,76 @@ impl Deref for MintableToken {
...
+            Bytes(signature.r.0),
+            Bytes(signature.s.0),
+        );
+
+        // double check the signatures are correct...
+        permit
+            .clone()
+            .call()
+            .await
+            .expect("permit signature issue; good luck figuring this one out!");
```
MartinquaXD, line 240: 😄

crates/e2e/tests/e2e/setup/onchain_components.rs
```diff
@@ -170,6 +185,76 @@ impl Deref for MintableToken {
...
+    contract: CowProtocolToken,
+    holder: Account,
+}
+
+impl CowToken {
+    pub async fn fund(&self, to: H160, amount: U256) {
+        tx!(self.holder, self.contract.transfer(to, amount));
+    }
+
+    pub async fn permit(&self, owner: &TestAccount, spender: H160, value: U256) -> InteractionData {
```
MartinquaXD, line 199: Is there a reason these functions are specific to the cow token and can not be generalized in some way?

MartinquaXD (commented):
Not sure about the special cases introduced for the cow token. I don't have enough context to form an opinion on whether or not this is needed.

crates/e2e/tests/e2e/setup/onchain_components.rs
nlordell: So these are E2E test setup-specific functions for the CoW token. This is because the CoW token uses the same bytecode as on mainnet and (obviously) can't mint tokens arbitrarily to different addresses. The way the CoW token is setup is that it gets created and the entire initial total supply gets allocated to an address (the `self.holder`). Then the holder can distribute these tokens with the `fund` method for unit tests.

`permit` is only supported by the CoW token in our E2E tests (our other ERC20 token implementations do not support that method).

crates/shared/src/account_balances/arguments.rs
```diff
@@ -26,6 +26,14 @@ pub struct Arguments {
...
     #[clap(long, env, value_enum)]
     pub account_balances_simulator: Option<CodeSimulatorKind>,
+
+    /// Whether or not to optimistically treat account balance queries with
+    /// pre-interactions as if sufficient token balance and allowance is always
+    /// available. Useful for partially supporting pre-interactions in
+    /// environments where the required simulation infrastructure is not
+    /// available (such as in E2E tests).
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "false")]
+    pub account_balances_optimistic_pre_interaction_handling: bool,
```
cowfee, line 36: Are you sure there's no better name than `account_balances_optimistic_pre_interaction_handling`? How about `check_pre_interaction_balances` which defaults to true? Maybe I'm completely wrong. I'm just not a huge fan of this name.

cowfee (approved):

crates/shared/src/account_balances/arguments.rs
nlordell: I hope this never gets used, hence the bad name :P

nlordell: Joking aside, this flag controls whether or not we skill balance checks altogether for orders with pre-interactions and just assume that the balance will be there.

I also thought the name was long, but couldn't come up with something more concise.

MartinquaXD (approved):

PR #1605, opened 2023-06-20
Trampoline Custom Interactions over MultiSendCallOnly

nlordell:
This PR ensures custom interactions are always trampolined instead of being called by the Settlement contract directly.

Specifically, if we called interactions from the settlement contract directly, a user could easily add a `WETH.approve(me, type(uint256).max)` custom interaction to their order in order to set an unlimited approval from WETH to their account and take fees collected in the settlement contract.

In order to work-around this, we demote the call to an unprivileged context by "trampolining" the Ethereum call over an intermediary contract. We chose the Safe's `MultiSendCallOnly` contract because:
- It is already deployed on all networks we care about
- It is gas efficient
- Does exactly what we want it to do

### Reviewer Notes

This PR looks bigger than it actually is, piping the `multisend` contract everywhere was a big hassle which bloats the PR (except for in the driver - where it was mostly trivial... the `Ethereum::contracts` abstraction is boss). The interesting things to review are:
- The E2E test changes: d6ff5d7f01c1d4ecd299ef60f0825ebe4e41fce4
- The changes to the settlement encoder: https://github.com/cowprotocol/services/pull/1605/files#diff-29424a20e1cfaff71b8f0b551df98a9e03ca2d328a8cb354c294256b427fbc5b

### Test Plan

Added unit tests to verify multisend encoding and E2E test to check malicious pre-interactions don't work.


crates/contracts/src/multisend.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Clone, Copy, Debug, Default)]
+#[repr(u8)]
+pub enum Operation {
+    #[default]
+    Call = 0,
+    DelegateCall = 1,
+}
+
+pub fn encode(transactions: &[Transaction]) -> Bytes<Vec<u8>> {
```
MartinquaXD, line 26: Would it make sense to link the trampoline contract code here so readers can verify that this encoding is correct?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -593,15 +642,58 @@ impl SettlementEncoder {
             },
         );
 
+        // We trampoline the custom user interactions over the
+        // `MultiSendCallOnly` contract, ensuring that we aren't calling user-
+        // specified calls from a privileged context (i.e. the Settlement
+        // contract). We also collapse all `CoWSwapEthFlow::wrapAll` calls into
+        // one for gas efficiency.
```
vkgnosis, line 649: The last part could have been a separate PR but not worth changing now.

vkgnosis (approved):
Rust code looks fine. I can't evaluate the smart contract security side of things.

crates/solver/src/arguments.rs
```diff
@@ -365,6 +365,12 @@ pub struct Arguments {
     #[clap(long, env)]
     pub ethflow_contract: Option<H160>,
 
+    /// Address of the `MultiCallSendOnly` contract used for trampolining custom
+    /// order interactions. If not specified, will contract deployed on the
```
MartinquaXD, line 369:
```suggestion
    /// order interactions. If not specified, will use the contract deployed on the
```

crates/solver/src/settlement_submission.rs
```diff
@@ -370,6 +376,38 @@ impl SolutionSubmitter {
     }
 }
 
+/// A settlement that is ready to be used by submitters.
+#[derive(Clone, Debug)]
+pub struct SubmitterSettlement {
+    inner: Settlement,
+    encoded: EncodedSettlement,
```
MartinquaXD, line 383: Why does the submitter suddenly has to know about both settlements?

crates/e2e/tests/e2e/pre_interaction.rs
```diff
@@ -101,4 +114,26 @@ async fn pre_interaction(web3: Web3) {
...
+    // Check malicious custom interactions did not work.
+    let allowance = cow
+        .allowance(
+            onchain.contracts().gp_settlement.address(),
+            trader.address(),
+        )
+        .call()
+        .await
+        .unwrap();
+    assert_eq!(allowance, U256::zero());
```
MartinquaXD, line 127: This seems to suggest that custom interactions can revert without reverting the whole transaction.
I guess this is because the multisend trampoline call calls these interactions with a `try{} catch(e){}`?
Would be nice to document that somewhere.

MartinquaXD (commented):
One test seems to suggest that custom interactions can

crates/contracts/src/multisend.rs
nlordell: I added links in the module doc-comment: https://github.com/cowprotocol/services/blob/f24fa615d7f499e48b5f4995499da2045706c704/crates/contracts/src/multisend.rs#L4-L6

I can move them here so its more obvious to find.

crates/solver/src/settlement_submission.rs
nlordell: It uses the `EncodedSettlement` in most places, but in some places (specifically access list generation) it needs the `Settlement`. Its a much larger refactor to pull this out and I'm afraid of touching the logic too much...

crates/e2e/tests/e2e/pre_interaction.rs
nlordell: > This seems to suggest that custom interactions can revert without reverting the whole transaction.

No, this is not the case. It is just the allowance interaction is not made from the settlement contract index.

I will add something like this to document why this works:

```rust
    let allowance = cow
        .allowance(
            onchain.contracts().multisend.address(),
            trader.address(),
        )
        .call()
        .await
        .unwrap();
    assert_eq!(allowance, U256::max_value());
```

crates/contracts/src/multisend.rs
nlordell: 21d71f13528f0934c11ba865db39342e781e95ed

crates/solver/src/arguments.rs
nlordell: 21d71f13528f0934c11ba865db39342e781e95ed

crates/solver/src/settlement_submission.rs
MartinquaXD: I'm still struggling to see which new code path introduced the requirement to have both versions. Didn't we also compute access lists before?
Instead of having the encoded and unencoded version maybe it's nicer to only have the unencoded version and the data needed to encode the settlement. This would just be the `Contracts` which you already pass into `SubmitterSettlement::new()`.
This seems at least slightly nicer to me because there is only a single source of truth. Whereas with the other approach you could theoretically store the encoded and unencoded version of 2 entirely different settlements.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -593,15 +642,58 @@ impl SettlementEncoder {
...
+        let group_interactions = |interactions: Vec<InteractionData>| {
```
cowfee, line 650: I think I'd rename this to `custom_interactions`. Not entirely sure what the idea behind `group_interactions` was.

cowfee (approved):
LGTM.

crates/solver/src/settlement/settlement_encoder.rs
nlordell: Left over of a refactor, never renamed the closure.

crates/solver/src/settlement_submission.rs
nlordell: > I'm still struggling to see which new code path introduced the requirement to have both versions.

Both versions were always required. It is just that previously, there existed a function `f :: Settlement -> EncodedSettlement`, but now that doesn't exist anymore.

> Instead of having the encoded and unencoded version maybe it's nicer to only have the unencoded version and the data needed to encode the settlement.

Good suggestion.

nlordell: 223a88a6fdef1475d294c6b19961cc0c8e6fe43b

MartinquaXD (approved):

crates/solver/src/settlement/settlement_encoder.rs
nlordell: 2e254aff6ea549cf7ff4cd65cb60a581dc0ab76f

PR #1606, opened 2023-06-20
Properly join URLs

cowfee:
We have suffered various issues (especially now with colocation) because of how we join URLs and sometimes they're supposed to have a trailing slash and sometimes they don't have to. This PR makes the URL joining consistent across the codebase, such that a trailing slash on the URL or a leading slash on the path is never required.

### Test Plan

CI.


vkgnosis (approved):
I like having one (two) url joining functions that implement this.

On a URL spec level this implementation isn't correct because "http://a.a" and "http://a.a///////" really are different urls that could be handled differently by an api but I don't expect anyone to do that. And if they do it is still going to be easier to fix with this central function.

nlordell (approved):
LGTM.

Issue #1555, opened 2023-06-02
Prices service


cowfee:
## Preface

In the CoW Protocol today, quotes are used for two very distinct use cases: to provide price estimates to the user indicating how much of one token they can swap for another token, and to provide a guarantee to the user that their trade will go through at a certain price or otherwise fail. The first use case (referred to as _price estimation_ from here on) is purely a UX feature, allowing our web page to display prices to the user interactively. OTOH, the second use case (referred to as _quoting_ from here on) is crucial to the correctness of the protocol, since the quoted prices are used when placing orders, and these orders will only go through if the prices are actually achievable.

Since these two distinct features are convoluted into the same `/quote` mechanism, this leads to problems like we've seen with the USDC depeg where incorrect _price estimates_ were used as _quotes_, leading users to place failing orders and cause a bad experience for them and real monetary costs for us.

For this (and other) reasons, we have decided that we will provide separate _price estimation_ and _quoting_ mechanisms.

## Price Estimation

The price estimation request will contain `sellToken`, `buyToken`, and exactly one of `sellAmount` or `buyAmount` (or simply `amount` and `side` depending on what we prefer, the two options are equivalent). The response will contain `sellAmount`, `buyAmount`, and `fee`. The `fee` will be denominated in the sell token.

Side note: this is different from our current quote API, which supports different selectors for how the fee would affect the estimate: `sellAmountBeforeFee`, `sellAmountAfterFee`, and `buyAmountAfterFee`. The CoW Swap FE only uses two: `buyAmountAfterFee` is used for buy orders, while `sellAmountBeforeFee` is used for sell orders. All these selectors can be applied by the client after calling into the API. The fee, which is expressed in the sell token, can also be converted into the buy token by the client by using the buy/sell ratio.

In the process of price estimation, the prices service will query estimates from a number of different estimators (e.g. baseline, uniswap, our own solvers, etc.) with a deadline, and the median estimate will be returned as the final estimate for the end user.

## Quoting

A quote represents a guarantee from the protocol that **an order** will go through at **a price** that is no worse than the quoted price. Therefore, the quote request will be the order itself, without exactly one of `sellAmount` or `buyAmount`, and the response will be the `sellAmount` or `buyAmount` that the protocol can guarantee.

[...]

PR #1610, opened 2023-06-22
Prices service

cowfee:
Getting started on https://github.com/cowprotocol/services/issues/1555.

Add a new crate for the prices service. Implement the `core` logic and a simple `boundary` integration.

I want to try out the names `core` (instead of `domain`) and `external` (instead of `infra`). The reasons are:

- I don't really like `domain` because the code we've been writing so far doesn't strictly adhere to DDD. DDD is adamant about things like Entities, Value Objects, Aggregates, Aggregate Roots, etc. which are all useful in their own right but IMO we don't need them and trying to adhere to them would be going too far. I also don't want to bore the team with these things.
- I think the name `external` is more intuitive than `infra`. (As in, this module contains components which are external to our system.) Note that if we decide to apply this renaming in other crates as well, the `api` module would become a peer to the `external` module.

Curious to hear what others think about this.

What's left to do:
- Implement an API, which should be pretty simple.
- Implement the configuration layer, which is probably not going to be that simple :smile:.


crates/prices/src/core/swap.rs
```diff
@@ -0,0 +1,83 @@
...
+/// token.
+#[derive(Debug, Clone, Copy)]
+pub struct Swap {
+    /// The token to swap from.
+    pub from: FromToken,
+    /// The token to swap into.
+    pub to: ToToken,
+    /// The amount to swap.
+    pub amount: FromAmount,
+}
```
MartinquaXD, line 13: This abstraction reads like it's only intended for sell orders.
For buy orders you only know how much `ToToken` you want to buy and the price estimator is supposed to tell you how much `FromToken` you need for that. That means you couldn't really populate the `amount` field with a reasonable value for buy orders.
Please correct me if I misunderstood this struct.

crates/prices/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+mod boundary;
+mod core;
+
+pub async fn main() {
+    println!("Hello, world!")
```
MartinquaXD, line 5: 👋

crates/prices/src/core/estimate.rs
```diff
@@ -0,0 +1,84 @@
...
+    let fee = median_estimate(
+        Swap {
+            from: eth::ETH_TOKEN.into(),
+            to: swap.to,
+            amount: estimate.gas.0.into(),
+        },
+        deadline,
+        estimators,
+    )
+    .await?;
```
MartinquaXD, line 37: These need to be joined otherwise an `estimate` call with `deadline: 3s` could take 6s in total.
Also it's nice to reduce the latency for when users request a price.

MartinquaXD:
> I think the name external is more intuitive than infra. 

Agree.

> I don't really like domain because the code we've been writing so far doesn't strictly adhere to DDD.

I think we should pick the name that makes the most sense in our circumstances whether or not we adhere to predefined concepts of DDD or not. With that said I think `core` probably conveys more meaning for people who are not familiar with DDD. 👍

crates/prices/src/core/estimate.rs
```diff
@@ -0,0 +1,84 @@
...
+    .await
+    .into_iter()
+    .filter_map(|estimate| {
+        // TODO Observe the error
+        estimate.ok()
+    })
+    .collect();
+
+    // Pick the median estimate.
+    estimates.sort_by_key(|estimate| estimate.to);
```
MartinquaXD, line 67: The sorting should needs to account for quote side (`sell` or `buy`).
For `sell` orders higher out amounts are more desirable whereas for `buy` orders smaller in amounts are desirable.

crates/prices/src/core/swap.rs
cowfee: You're right, this is an oversight. For some reason I thought this covers both buy and sell orders.

crates/prices/src/core/estimate.rs
cowfee: This got joined in one of the following PRs.

crates/prices/src/core/swap.rs
cowfee: The result of a `Swap` are an estimated amount and fee, both expressed in the `to` token.

For an order which wants to _sell_ 100 token A for token B, the swap would look like this:
```
Swap { from: A, to: B, amount: 100 }
```
The fee would be denominated in token B (our buy token). However, since we know the conversion rate between A (our sell token) and B, we can convert that fee into token A. (Note: since we're working with estimates these kinds of conversions are completely acceptable.) $\Box$

For an order which wants to _buy_ 100 token B for token A, the swap would look like:
```
Swap { from: B, to: A, amount: 100 }
```
Which gives an amount in token A (our sell token) which can be swapped for 100 of token B (our buy token). It also gives a fee in token A. $\Box$

So, this `Swap` abstraction does exactly what we want (estimate the token price if we want to swap a specific amount), while being quite a bit simpler than the `Order` abstraction used by the protocol. The slight drawback is that the client has to convert the fee into the buy token themselves. Alternatively, the prices service could easily return the fee in _both_ tokens, so the client doesn't have to do anything.

crates/prices/src/core/estimate.rs
cowfee: With the way the `Swap` abstraction is designed, this is unnecessary. This is only one of the simplifications offered by the `Swap` type.

crates/prices/src/core/swap.rs
cowfee: > The prices service could easily return the fee in both tokens, so the client doesn't have to do anything.

This is probably the way to go, actually.

nlordell: In principle this sounds good, but it will result us undercharging fees for illiquid tokens.

nlordell:
Cleanup.

PR #1612, opened 2023-06-22
API for the prices service

cowfee:
Implements the API for the prices service.

### DIFF
--- a/crates/prices/src/api/routes/estimate.rs
+++ b/crates/prices/src/api/routes/estimate.rs
@@ -1,0 +1,63 @@
+use {
+    crate::{
+        api::State,
+        core::{self, eth},
+        util::serialize,
+    },
+    serde::{Deserialize, Serialize},
+    serde_with::serde_as,
+};
+
+pub(in crate::api) fn estimate(router: axum::Router<State>) -> axum::Router<State> {
+    router.route("/estimate", axum::routing::post(route))
+}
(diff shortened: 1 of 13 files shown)


crates/prices/src/api/routes/estimate.rs
```diff
@@ -0,0 +1,63 @@
...
+async fn route(
+    state: axum::extract::State<State>,
+    swap: axum::Json<Swap>,
+) -> Result<axum::Json<Estimate>, hyper::StatusCode> {
+    let estimators = state.estimators();
+    let deadline = state.deadline();
+    core::estimate(swap.0.into(), deadline, estimators)
+        .await
+        .map(|estimate| axum::Json(estimate.into()))
+        .map_err(|_| hyper::StatusCode::BAD_REQUEST)
```
MartinquaXD, line 24: `BAD_REQUEST` seems odd here. At this point we already know that the request had the correct format, right?

cowfee: Which error code should be returned here?

MartinquaXD: I'd probably go for `500` to indicate that the issues is on the `prices` side.
On the other hand I just realized this is an internal service that only we will use anyway so bike shedding about error codes is unnecessary. 👌

PR #1617, opened 2023-06-25
One Binary To Rule Them All

nlordell:
> ...One binary to find them, one binary to bring them all, and in the darkness bind them.

This PR adds a `cow` binary as a launcher combining all services binaries into one. The rationale here is to reduce the size of our Docker images to reduce the amount of disk pressure we have on our K8s nodes.

The change was designed in a way that should not require any adjustments to our infrastructure (the `cow` binary is invoked in the image's `ENTRYPOINT`, and the sub-commands are the same as the binary names).

### Test Plan

Build various Docker images:

```
% docker images ghcr.io/cowprotocol/services
REPOSITORY                     TAG          IMAGE ID       CREATED          SIZE
ghcr.io/cowprotocol/services   one-binary   888773fe38be   2 minutes ago    465MB
ghcr.io/cowprotocol/services   arm          c6f294590022   29 minutes ago   1.62GB
ghcr.io/cowprotocol/services   main-x64     47a0a6ba0a35   3 days ago       1.56GB
```

On my Aarch64 machine, its around 28% of the original size 🎉.

You can also double check that the new binary works:

```
% cargo run -p cow -- autopilot --help
```


crates/alerter/src/lib.rs
```diff
@@ -0,0 +1,442 @@
+// This application observes the order book api and tries to determine if the
```
nlordell, line 1: Code just moved from main. Interesting changes marked below.

crates/alerter/src/lib.rs
```diff
@@ -0,0 +1,441 @@
...
+pub async fn start(args: impl Iterator<Item = String>) {
+    let args = Arguments::parse_from(args);
+    shared::tracing::initialize("alerter=debug", tracing::Level::ERROR.into());
+    shared::exit_process_on_panic::set_panic_hook();
+    global_metrics::setup_metrics_registry(Some("gp_v2_alerter".to_string()), None);
+    tracing::info!("running alerter with {:#?}", args);
+    run(args).await;
+}
+
+async fn run(args: Arguments) {
```
nlordell, line 406: Only changes were these lines.

vkgnosis:
> This PR adds a `cow` binary as a launcher combining all services binaries into one. The rationale here is to reduce the size of our Docker images to reduce the amount of disk pressure we have on our K8s nodes.

I would have liked to discuss in an issue first but we can discuss it here too. The image goes from 1.5 GB to 500 MB . That's nice but it is surprising to me that this is a big deal for our cluster. How does that disk pressure manifest? If this is important then there are probably other ways we can try to reduce binary size too.

Maybe more important than docker image size is the it takes to run cargo check and cargo test during development. I suspect this PR increases that significantly because there is now  a new binary that depends on all the other binaries  as libraries. And I see two ways to fix that:
1. Don't have the combined binary in the default workspace.
2. Remove the non combined binaries.

2 is interesting because it probably reduces total build time. So I'd like to do that.

nlordell:
> 2 is interesting because it probably reduces total build time. So I'd like to do that.

Nice suggestion - will take a stab at this.

nlordell:
> the it takes to run cargo check and cargo test during development

Agree, it now spends extra time linking `cow` (building it is very quick though).

nlordell:
Cleaning up old unmerged PRs.

PR #1619, opened 2023-06-26
Prices service zeroex config

cowfee:
Progress on #1555.

Allows the price service to configure and run the zeroex estimator. Makes some changes to the `core` types. Remove the need for two separate `Estimate` types. Add new `core::Price` type that makes more sense than what I had before.

### Test Plan

Manual test.

output


crates/prices/src/core/estimate.rs
```diff
@@ -9,79 +10,86 @@ use {
...
+        // Fetch the estimate for ETH price, to convert the gas fees into [`ToToken`].
+        median_estimate(
+            Swap {
+                // TODO Automatically wrap ETH, and specify the chain ID in the config
+                // from: eth::ETH_TOKEN.into(),
+                from: eth::H160::from_str("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2")
+                    .unwrap()
+                    .into(),
+                to: swap.to,
+                amount: eth::U256::from(1000000000000000u64).into(),
```
MartinquaXD, line 45: This amount needs to be configurable to account for the difference in price between eth on mainnet and xdai on gc.

crates/prices/src/core/estimate.rs
```diff
@@ -9,79 +10,86 @@ use {
...
+                estimate.ok()
+            })
+            .unzip();
 
-    // Pick the median estimate.
-    estimates.sort_by_key(|estimate| estimate.amount);
-    estimates.get(estimates.len() / 2).copied().ok_or(Error)
+    // Pick the median estimated price and gas fee.
+    prices.sort_unstable_by_key(|&price| price * swap.amount);
+    gas_fees.sort_unstable();
```
MartinquaXD, line 76: I think we should not pick the median `price` independently from the median `gas_fees` because they relate to the same trade route and are therefore not independent from each other.

crates/prices/src/core/price.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl Price {
+    pub fn new(from: swap::FromAmount, to: swap::ToAmount) -> Self {
+        Self { from, to }
+    }
+}
+
+/// Apply the price to the [`swap::FromAmount`].
+impl Mul<swap::FromAmount> for Price {
```
MartinquaXD, line 20: Not sure if overloading the `*` operator is a great idea.
And if we do we `Output` should be `Result` because multiplications and divisions can panic which needs to be handled in some way.

crates/prices/src/lib.rs
```diff
@@ -1,17 +1,33 @@
...
     let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
+
+    let mut estimators: Vec<Box<dyn core::Estimator>> = Vec::new();
+    if let Some(zeroex) = config.zeroex {
+        if zeroex.enable {
+            estimators.push(Box::new(
+                boundary::Zeroex {
+                    api_key: zeroex.api_key,
+                    endpoint: zeroex.endpoint,
+                    timeout: std::time::Duration::from_millis(config.timeout_ms),
```
MartinquaXD, line 22: What's the reason to move `timeouts` into each individual estimator instead of having it in a single component that queries all of them?
Seems a bit error prone.

crates/prices/src/cli.rs
```diff
@@ -0,0 +1,13 @@
+use std::{net::SocketAddr, path::PathBuf};
+
+#[derive(Debug, clap::Parser)]
+pub struct Args {
+    /// The address to bind the driver to.
+    #[clap(long, env, default_value = "0.0.0.0:11098")]
+    pub addr: SocketAddr,
+
+    /// Path to the driver configuration file. This file should be in TOML
+    /// format.
```
MartinquaXD, line 10: doc comments refer to the `driver`.

crates/prices/src/config.rs
```diff
@@ -0,0 +1,29 @@
...
+        .unwrap_or_else(|e| panic!("I/O error while reading {path:?}: {e:?}"));
+    toml::de::from_str(&file)
+        .unwrap_or_else(|e| panic!("Configuration error while reading {path:?}: {e:?}"))
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct Config {
+    pub timeout_ms: u64,
+    pub zeroex: Option<Zeroex>,
```
MartinquaXD, line 20: Why is the `zeroex` config optional if it already has an `enable` flag which sounds like it manages the same thing?

crates/prices/src/core/estimate.rs
cowfee: Yeah, I had second thoughts about this as well. I think you're right.

PR #1620, opened 2023-06-26
Add receiver to order model

devanoneth:
This adds the receiver address to the `OrderModel` which is sent to solvers. This will allow solvers to combine outputs which go to the same receiver address.

### DIFF
--- a/crates/shared/src/http_solver/model.rs
+++ b/crates/shared/src/http_solver/model.rs
@@ -43,6 +43,7 @@ pub struct OrderModel {
     pub fee: TokenAmount,
     pub cost: TokenAmount,
     pub is_liquidity_order: bool,
+    pub receiver: H160,
     /// [DEPRECATED] All orders are always mature
     pub is_mature: bool,
     #[serde(default)]
@@ -527,6 +528,7 @@ mod tests {
         let native_token = H160([0xee; 20]);
         let buy_token = H160::from_low_u64_be(1337);
         let sell_token = H160::from_low_u64_be(43110);
+        let receiver = H160::from_low_u64_be(4369);
         let order_model = OrderModel {
             id: Some(OrderUid::default()),
             sell_token,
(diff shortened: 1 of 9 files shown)


devanoneth:

devanoneth:
recheck

nlordell:
Thanks for the PR! Just a question for us to get a bit more context on the motivation behind the change:

> This adds the receiver address to the `OrderModel` which is sent to solvers. This will allow solvers to combine outputs which go to the same receiver address.

Do you mean combine the "transfer outs" at the end of the settlement? These are done by the Settlement contract and we did not add a mechanism to collapse the transfer outs for multiple users into one. Or am I misunderstanding where this is useful?

devanoneth:
Hi @nlordell, this is more in the context of DeFi actions e.g. depositing into a LP pool. There is an upcoming [Aave Treasury trade which should be performed on Cowswap](https://app.aave.com/governance/proposal/249/) (this one was cancelled but we understand another one should be proposed soon). In this scenario, a solver could instead deposit the WETH and BAL together, in one tx, fulfilling both orders, not just in one solution, but actually in one tx which would reduce slippage incurred. This solution of course would also help any similar trades in future (e.g. depositing into Uniswap pools, Curve pools, or anywhere where two orders could be "squashed" into one deposit tx). The reason we need the receiver address is to know that these two orders can be fulfilled in one tx because the LP tokens are going to the same destination address. With this knowledge, the solver can treat these two separate orders as one LP deposit.

devanoneth:
Hi @nlordell, this is more in the context of DeFi actions e.g. depositing into a Uniswap USDC/WETH pool. In such a scenario, a solver could instead deposit the WETH and USDC together, in one tx, fulfilling both orders, not just in one solution, but actually in one tx which would reduce slippage incurred. This solution of course would also help any similar trades in future e.g. depositing into Balancer pools, Curve pools, or anywhere where two (or more) inputs could be "squashed" into one deposit tx. 

The reason we need the receiver address is to know that these two orders can be fulfilled in one tx because the LP tokens are going to the same destination address. With this knowledge, the solver can treat these two separate orders as one LP deposit.

PR #1621, opened 2023-06-26
API endpoint for total surplus

cowfee:
Implement API endpoint to calculate total surplus.

### Test Plan

Tried to test manually but for some reason I get this error:
```
2023-06-26T14:56:32.229Z ERROR shared::tracing: thread 'main' panicked at 'error innitializing Uniswap V3 pool fetcher: error
 sending request for url (https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v3): operation timed out
```
It seems like this API is down at the moment?

### DIFF
--- a/crates/orderbook/src/api/get_total_surplus.rs
+++ b/crates/orderbook/src/api/get_total_surplus.rs
@@ -1,0 +1,22 @@
+use {
+    primitive_types::H160,
+    serde_json::json,
+    std::convert::Infallible,
+    warp::{http::StatusCode, reply::with_status, Filter, Rejection},
+};
+
+// TODO This endpoint is not documented for now because we don't want to
+// stabilize it quite yet. Once we're sure this is what we want, we should add
+// it to the openapi spec.
+pub fn get() -> impl Filter<Extract = (super::ApiReply,), Error = Rejection> + Clone {
+    warp::path!("v1" / "users" / H160 / "total_surplus")
+        .and(warp::get())
+        .and_then(move |_| async {
+            Result::<_, Infallible>::Ok(with_status(
+                warp::reply::json(&json!({
+                    "totalSurplus": 42_133_700_000_000_000_000_u128.to_string(),
+                })),
+                StatusCode::OK,
+            ))
+        })
+}
--- a/crates/orderbook/src/api.rs
+++ b/crates/orderbook/src/api.rs
@@ -6,6 +6,7 @@ mod get_native_price;
 mod get_order_by_uid;
 mod get_orders_by_tx;
 mod get_solver_competition;
+mod get_total_surplus;
 mod get_trades;
 mod get_user_orders;
 mod post_order;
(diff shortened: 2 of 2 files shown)


vkgnosis (approved):
Needs documentation in openapi.yaml. Route looks fine.

vkgnosis:
> It seems like this API is down at the moment?

There is a baseline sources argument that defaults to include uni-v3 and uni-v3 initialization depends on some external service that is problematic. You can specify that argument to exclude uni-v3.

MartinquaXD:
> Needs documentation in openapi.yaml

Do we really want to publicly document this endpoint?
If we do there are certain expectations of stability that comes with that. Not sure if we want to guarantee that, though.
Alternatively we could document it but mention that we don't guarantee it to be stable.

crates/orderbook/src/api/get_total_surplus.rs
```diff
@@ -0,0 +1,19 @@
...
+    warp::{http::StatusCode, reply::with_status, Filter, Rejection},
+};
+
+pub fn get() -> impl Filter<Extract = (super::ApiReply,), Error = Rejection> + Clone {
+    warp::path!("v1" / "users" / H160 / "total-surplus")
+        .and(warp::get())
+        .and_then(move |_| async {
+            Result::<_, Infallible>::Ok(with_status(
+                warp::reply::json(&json!({
+                    "totalSurplus": 42_000_000_000_000u64
```
nlordell, line 14:
```suggestion
                    "totalSurplus": 42_133_700_000_000_000_000_u128.to_string(),
```

nlordell (commented):
LGTM, can you also add it to the `openapi.yaml` specification?

nlordell:
> Do we really want to publicly document this endpoint?

Good point... maybe we shouldn't.

nlordell (approved):

crates/orderbook/src/api/get_total_surplus.rs
nlordell: This will be a `U256` in the future, so this should be a `string` and not a `number`.

crates/orderbook/src/api/get_total_surplus.rs
```diff
@@ -0,0 +1,19 @@
+use {
+    primitive_types::H160,
+    serde_json::json,
+    std::convert::Infallible,
+    warp::{http::StatusCode, reply::with_status, Filter, Rejection},
+};
+
+pub fn get() -> impl Filter<Extract = (super::ApiReply,), Error = Rejection> + Clone {
+    warp::path!("v1" / "users" / H160 / "total-surplus")
```
MartinquaXD, line 9: Looks like we use underscores for the other endpoints.
```suggestion
    warp::path!("v1" / "users" / H160 / "total_surplus")
```

nlordell: Ah yes, nice catch.

vkgnosis:
It should be in openapi even just as docs for the frontend. We can say it is unstable.

PR #1633, opened 2023-06-30
`driver`: return calldata in response of `/settle`

### LINKED ISSUE #1632: `driver`: Reveal calldata in response of `/settle`
The `driver` should return both internalized and uninternalized calldata in the response of `/settle`. This is mandated by CIP-20. The driver is still responsible for publishing its own settlements, this does not change.

cowfee:
Fixes #1632.

Return the internalized and uninternalized calldata in the `/settle` response body.

### Test Plan

Updated automated tests.

### DIFF
--- a/crates/autopilot/src/driver_api.rs
+++ b/crates/autopilot/src/driver_api.rs
@@ -1,5 +1,5 @@
 use {
-    crate::driver_model::{execute, solve},
+    crate::driver_model::{settle, solve},
     anyhow::{anyhow, Context, Result},
     reqwest::Client,
     shared::http_client::response_body_with_size_limit,
@@ -30,12 +30,11 @@ impl Driver {
         self.request_response("solve", Some(request)).await
     }
 
-    pub async fn execute(
+    pub async fn settle(
         &self,
         solution_id: &str,
-        _request: &execute::Request,
-    ) -> Result<execute::Response> {
-        // TODO: should be execute
+        _request: &settle::Request,
+    ) -> Result<settle::Response> {
         self.request_response(&format!("settle/{solution_id}"), Option::<&()>::None)
             .await
     }
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -131,7 +131,7 @@ pub mod solve {
     }
 }
 
-pub mod execute {
+pub mod settle {
     use {
         derivative::Derivative,
         model::{bytes_hex, order::OrderUid, u256_decimal},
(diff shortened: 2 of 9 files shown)


crates/autopilot/src/driver_api.rs
```diff
@@ -1,5 +1,5 @@
 use {
-    crate::driver_model::{execute, solve},
+    crate::driver_model::{settle, solve},
```
cowfee, line 2: Slightly unrelated, but I noticed that this name is wrong and had to fix it. There is no `/execute`.

crates/driver/src/infra/observe/mod.rs
```diff
@@ -143,10 +143,14 @@ pub fn settling(solver: &solver::Name, id: settlement::Id) {
...
+pub fn settled(
+    solver: &solver::Name,
+    id: settlement::Id,
+    result: &Result<competition::Calldata, competition::Error>,
+) {
     match result {
-        Ok(()) => {
-            tracing::info!(%solver, ?id, "settled");
+        Ok(calldata) => {
+            tracing::info!(%solver, ?id, ?calldata, "settled");
```
cowfee, line 153: Note that it was very difficult to forget to log this, due to how the `observe` module is written. I'm sure I would have forgotten otherwise.

crates/autopilot/src/driver_model.rs
```diff
@@ -155,15 +155,17 @@ pub mod execute {
     #[derive(Clone, Debug, Default, Deserialize)]
     #[serde(rename_all = "camelCase", deny_unknown_fields)]
     pub struct Response {
-        pub account: H160,
-        pub nonce: u64,
-        #[serde_as(as = "BTreeMap<_, DisplayFromStr>")]
-        pub clearing_prices: BTreeMap<H160, U256>,
-        pub trades: Vec<Trade>,
-        pub internalized_interactions: Vec<InternalizedInteraction>,
```
nlordell, line 163: I think we still need to keep these around. I think the information of the expected input and output amounts gets lost with just the `Calldata` return type as is.

cowfee: These were never actually returned from the driver. Can you specify exactly what the response of `/settle` should be?

cowfee: Also note that, the way it's written now, the `/settle` endpoint will only return after the settlement has been confirmed. I believe this is actually wrong, since the autopilot waits for settlement confirmation? This needs to be well-defined.

nlordell: Let me double-check.

MartinquaXD (approved):
Change looks alright to me.
Not sure about the discussion around not sending more detailed data, though.

crates/autopilot/src/driver_model.rs
nlordell: >  I believe this is actually wrong,

I think so too, I think only the `autopilot` needs to wait for it to mine for returning. That being said, the `driver` should still spawn a background task for the mining (it wants to potentially cancel, resubmit with higher gas price, etc.)

nlordell: > Can you specify exactly what the response of /settle should be?

Sorry, the input and output amounts actually don't matter. The format you defined is correct. My bad.

nlordell (approved):

crates/autopilot/src/driver_model.rs
sunce86: What if there is custom driver implementation where the `settle` response is sent back to autopilot just after the settlement is mined? In that case, receiving `call_data` as part of `settle` response is too late for autopilot, since autopilot must save `call_data` to DB before settlement is mined.

PR #1634, opened 2023-07-02
`driver`: encode auction ID instead of settlement ID

cowfee:
Encode the auction ID in the settlement calldata instead of the settlement ID. Get rid of the concept of settlement IDs. Update the `/settle` endpoint to not require any parameters. In the autopilot, wait for a settlement with the auction ID to be published by the winning address.

### Test Plan

Automated tests. Eventually will be tested on staging as well, after other driver infra bugs are handled.


crates/autopilot/src/run_loop.rs
```diff
@@ -175,14 +172,11 @@ impl RunLoop {
                 return;
             }
 
-            tracing::info!("executing with solver {}", index);
-            match self
-                .execute(auction, id, &self.drivers[index], &solution)
-                .await
-            {
+            tracing::info!("settling with solver {}", index);
```
MartinquaXD, line 175: Not related to this PR but I noticed that only logging the index of the solver is pretty bad for debugging.
Would be nice to have the name here.
Also this and the next log should have the `index` as an tracing argument instead of being baked into the string.

crates/driver/src/infra/observe/mod.rs
```diff
@@ -131,33 +132,33 @@ pub fn score(solver: &solver::Name, settlement: &Settlement, score: &solution::S
...
+        auction_id = ?settlement.auction_id,
         score = score.0.to_f64_lossy(),
         "settlement scored"
     );
 }
 
 /// Observe that the settlement process is about to start.
-pub fn settling(solver: &solver::Name, id: settlement::Id) {
-    tracing::trace!(%solver, ?id, "settling");
+pub fn settling(solver: &solver::Name, auction_id: Option<auction::Id>) {
```
MartinquaXD, line 142: I think `auction_id` should probably non-optional in `settling()` and `settled()`.
If we try to settle a solution without an `auction_id` probably some invariant has been violated.

MartinquaXD (approved):
Good simplification.

crates/autopilot/src/run_loop.rs
```diff
@@ -334,7 +324,7 @@ impl RunLoop {
         let current = self.current_block.borrow().number;
         let start = current.saturating_sub(start_offset);
         let deadline = current.saturating_add(max_wait_time_blocks);
-        tracing::debug!(%current, %start, %deadline, ?tag, "waiting for tag");
+        tracing::debug!(%current, %start, %deadline, ?id, %submission_address, "waiting for settlement");
```
nlordell, line 327: nit:

```suggestion
        tracing::debug!(%current, %start, %deadline, ?id, ?submission_address, "waiting for settlement");
```

The `ethereum_types::H160` display implementation does not print the whole address.

nlordell (approved):

crates/driver/src/infra/observe/mod.rs
cowfee: Yes, in case this is `None` the endpoint will return an error. But I think that this should still be logged.

cowfee: But it definitely shouldn't be optional in `observe::settled`.

crates/autopilot/src/run_loop.rs
cowfee: Agreed, this is pretty bad.

Issue #1583, opened 2023-06-13
Store Winning Quote Source in Database


nlordell:
This issue captures the work of storing the quoter's solver address in the database in the `order_quotes` table.

Specifically, under [this proposal](https://forum.cow.fi/t/cip-draft-incentives-mechanism-for-the-price-estimation-competition/1718) we plan on implementing rewards for quoting orders. For this to work, we need to tie each order

This issue captures the work to:
1. Extend the `quotes` database with a `solver` address (the solver's submission address).
2. Include the `solver` in the `order_quotes` table that are tied to orders.

MartinquaXD:
Just wanted to point out that we continually update the `surplus_fee` of fok limit orders. That means we could do an unlimited number of quote requests but can only store a single `solver_address` because we only have a single `order_quote`.
Another consideration is that there are lots of duplicated fok limit orders so we went with the approach of doing 1 quote to update all identical orders.
@harisang, @bh2smith how big of an issue would these points be for your analysis

bh2smith:
I would have initially expected to have a single quote linked to a single order, but in the event of an order being touched multiple times by different quotes it seems more like we want to link quotes to trades (which is also fine by me). Note that we are, from what I recall, only rewarding quotes for market orders, although I guess there are "Fill or Kill" (aka FoK) market orders. 

I don't have a preference here as long as what we wind up with is something coherent.

nlordell:
@MartinquaXD - I talked with @harisang and at first, we would only be considering quotes for market orders. With that in mind, the restriction for the FoK limit orders would not be a big issue.

In the medium term, we expect FoK limit orders to also have their fees computed solver-side, which means that this would no longer become a problem.

sunce86:
Closed with https://github.com/cowprotocol/services/pull/1637

PR #1637, opened 2023-07-03
Store solver address with quote

MartinquaXD:
Stores the solver that provided a given quote in preparation for CIP-27.

This change requires us to specify a solver address with a price estimator. I decided to pass a list of arguments delimited with `|` and `,` since this is already common in the code base. Note that this can go away when we switch over to driver price estimators. Also I removed the default arguments for (native) price estimators since it's now unclear which address should be attached to the `Baseline` price estimator.
With the argument parsing in place the PR just has to pipe the address throughout the price estimation and quoting logic.

### Test Plan
Extended unit tests so at least writing to and reading from the DB is covered.

Still unclear how to nicely test that creating an order with an existing quote will correctly copy the solver address of the quote to the `order_quote`. (I'm pretty confident it works but testing this e2e would still be nice)


harisang:
Referencing issue #1583

crates/autopilot/src/arguments.rs
```diff
@@ -82,13 +82,7 @@ pub struct Arguments {
...
     /// native token.
-    #[clap(
-        long,
-        env,
-        default_value = "Baseline",
-        value_enum,
-        use_value_delimiter = true
-    )]
+    #[clap(long, env, value_enum, use_value_delimiter = true)]
     pub native_price_estimators: Vec<shared::price_estimation::PriceEstimatorType>,
```
nlordell, line 86: There shouldn't be solver addresses associated with native price queries. Rewards are just tied to actual quotes and not the native token prices used for the estimates.

MartinquaXD: Rewards are also not tied to limit orders. My plan was to have a consistent data model (i.e. store the address for all orders) and then let analytics simply filter out all orders which should not be rewarded.

nlordell: The consequence of this change is that `cargo run -p orderbook` no longer works by default. You have to remember to specify both `--price-estimator` and `--native-price-estimator`.

I agree with the "My plan was to have a consistent data model" argument, but maybe to make testing easier we can have `Baseline|0x0000000000000000000000000000000000000000` as the default argument value for the flags?

nlordell (approved):

nlordell:
Note that CI is failing because `cargo +nightly fmt` changed behaviour for `let .. = .. else { .. }` statements.

crates/shared/src/price_estimation.rs
```diff
@@ -41,16 +40,15 @@ use {
...
-    OneInch,
-    Yearn,
-    BalancerSor,
+    Baseline(H160),
+    Paraswap(H160),
+    ZeroEx(H160),
+    Quasimodo(H160),
+    OneInch(H160),
+    Yearn(H160),
+    BalancerSor(H160),
```
cowfee, line 51: Every type of estimator will have an address (right?), so maybe this should be split into `PriceEstimator = Address + PriceEsimatorKind`.

cowfee (approved):

crates/shared/src/price_estimation.rs
cowfee: It's also obnoxious to have to match on this value all the time.

harisang: The intention is indeed to only allow solvers to become estimators, and so every estimator will have an address. Note however that this means we should create an account for Quasimodo again, as it is the only "solver" that is actually only used as an estimator. (we had delisted it as a solver when we did the pks rotation).

crates/autopilot/src/arguments.rs
nlordell: Oh nice. I actually like the fix of defaulting to the 0 address.

crates/shared/src/price_estimation.rs
nlordell: Ok, so the `--price-estimators` field will be deprecated at some point.

PR #1638, opened 2023-07-04
Deprecate `mandatory` field

cowfee:
Deprecate the `mandatory` flag since it was only used by Quasimodo and it's not useful enough to keep around for this very specific use case. Ensure that all auction IDs are encoded as strings.

### Test Plan

Automated tests. Will eventually be tested on staging as well.

### DIFF
--- a/crates/driver/src/domain/competition/auction.rs
+++ b/crates/driver/src/domain/competition/auction.rs
@@ -11,7 +11,8 @@ use {
 /// solving them.
 #[derive(Debug)]
 pub struct Auction {
-    pub id: Id,
+    /// [`None`] if this auction applies to a quote.
+    pub id: Option<Id>,
     pub tokens: Vec<Token>,
     pub orders: Vec<competition::Order>,
     pub gas_price: eth::GasPrice,
--- a/crates/driver/src/domain/competition/auction.rs
+++ b/crates/driver/src/domain/competition/auction.rs
@@ -103,7 +103,7 @@ impl TryFrom<i64> for Id {
     type Error = InvalidId;
 
     fn try_from(value: i64) -> Result<Self, Self::Error> {
-        if value >= 0 {
+        if (Self::MIN..=Self::MAX).contains(&value) {
             Ok(Self(value))
         } else {
             Err(InvalidId)
--- a/crates/driver/src/domain/competition/auction.rs
+++ b/crates/driver/src/domain/competition/auction.rs
@@ -3,7 +3,7 @@ use {
         competition::{self, solution},
         eth,
     },
-    std::{num::ParseIntError, str::FromStr},
+    thiserror::Error,
 };
 
 /// An auction is a set of orders that can be solved. The solvers calculate
@@ -11,10 +11,7 @@ use {
 /// solving them.
 #[derive(Debug)]
 pub struct Auction {
-    // TODO Make this non-optional and simplify things
-    /// [`None`] if the auction is used for quoting, [`Some`] if the auction is
-    /// used for competition.
-    pub id: Option<Id>,
+    pub id: Id,
     pub tokens: Vec<Token>,
     pub orders: Vec<competition::Order>,
     pub gas_price: eth::GasPrice,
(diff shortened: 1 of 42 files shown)


crates/driver/src/domain/quote.rs
```diff
@@ -93,8 +95,12 @@ impl Order {
     }
 
     fn fake_auction(&self, gas_price: eth::GasPrice) -> competition::Auction {
+        let random_id = rand::thread_rng()
+            .gen_range(auction::Id::MIN..=auction::Id::MAX)
+            .try_into()
+            .unwrap();
         competition::Auction {
-            id: None,
+            id: random_id,
```
nlordell, line 103: Instead of a random one, why not a `0` ID. This can be useful for solvers to disambiguate between quotes and auctions internally (and not include PMMs in quotes for example).

crates/shared/src/price_estimation/http.rs
```diff
@@ -129,7 +129,7 @@ impl HttpPriceEstimator {
                     token: self.native_token,
                 },
                 is_liquidity_order: false,
-                mandatory: true,
+                mandatory: false,
```
nlordell, line 132: Why change this here?

crates/solvers/src/boundary/legacy.rs
```diff
@@ -172,9 +172,7 @@ fn to_boundary_auction(
                 cost: gas.gp_order_cost(),
                 is_liquidity_order: order.class == order::Class::Liquidity,
                 is_mature: true,
-                // Auctions for a /quote request don't have an id and always contain exactly a
-                // single user order that is mandatory to be matched.
-                mandatory: auction.id.is_none(),
+                mandatory: false,
```
nlordell, line 175: OK, the purpose of the PR is to make it so there is no real difference between auctions and quotes.

nlordell (approved):
Overall, I like the change (I think solver engines should treat auctions and quotes in the same way). Practically speaking, it could be that some solver engines want to, for example, not consider PMM liquidity when quoting (to not sign PMM orders that aren't intended to get filled).

That being said, I think there are "more correct" ways of working around this (for example, PMMs can sign orders that are only valid if `tx.origin` is the solver that they trust), so I think this is fine. Will want to double check with @harisang 

Also not, that the `id` field can potentially be reused for a quote ID in the future (which would not work with `id: Option<Id>` anyway).

crates/driver/src/domain/competition/auction.rs
```diff
@@ -90,17 +86,28 @@ impl From<Deadline> for chrono::DateTime<chrono::Utc> {
...
 }
 
-impl From<u64> for Id {
-    fn from(inner: u64) -> Self {
-        Self(inner)
+impl TryFrom<i64> for Id {
+    type Error = InvalidId;
+
+    fn try_from(value: i64) -> Result<Self, Self::Error> {
+        if value >= 0 {
```
MartinquaXD, line 106: Since this PR introduces these constants they should probably be used.
On the other hand why was the `Id` changed from `i64` to `u64` in the first place?
Is this because ids are an `i64` on the postgres level? I'm not sure we have to concern ourselves with that since they won't be negative anyway.
```suggestion
        if (Self::MIN..=Self::MAX).contains(&value) {
```

crates/shared/src/price_estimation/http.rs
cowfee: Because I've deprecated the field, so it should always be set to `false`.

crates/driver/src/domain/competition/auction.rs
MartinquaXD: > possibly JSON doesn't support them either

[Yep](https://stackoverflow.com/a/34989371). Man, JS is weird. 😅

vkgnosis: Autopilot sends an i64 because on the Postgres level it is an i64. It is intended to always be positive and it won't reach i64::MAX so which of the two Rust types you use doesn't matter so much.

The Json standard doesn't say how numbers work. They are just a sequence of digits. The browser Javascript engines always treat them as f64 as Martin linked. If you send Json betwen Rust code using serde derives on a struct with an u64 then that works.

vkgnosis (approved):

PR #1639, opened 2023-07-04
Filter Out Invalid Uniswap V2 Pools

nlordell:
This PR fixes an issue where the `driver` was sending invalid UniV2 pools to the solver engine. The engine expects ~~`uint96`~~ `uint112` reserves, but it is possible to set up a pools that has more reserves than that (effectively bricking it).

### Test Plan

Ran locally on Goerli and swapped with WETH <> `0xdc31ee1784292379fbb2964b3b9c4124d8f89c60` pool.

### DIFF
--- a/crates/driver/src/boundary/liquidity/uniswap/v2.rs
+++ b/crates/driver/src/boundary/liquidity/uniswap/v2.rs
@@ -43,9 +43,9 @@ pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Option<liquid
         .downcast_ref::<uniswap_v2::Inner>()
         .expect("downcast uniswap settlement handler");
 
-    // Trading on Uniswap V2 pools where the reserves overflows `uint96`s does
+    // Trading on Uniswap V2 pools where the reserves overflows `uint112`s does
     // not work, so filter these pools out.
-    let max = (1_u128 << 96) - 1;
+    let max = 2_u128.pow(112) - 1;
     if pool.reserves.0 > max || pool.reserves.1 > max {
         return None;
     }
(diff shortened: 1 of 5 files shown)


crates/driver/src/boundary/liquidity/uniswap/v2.rs
```diff
@@ -43,7 +43,14 @@ pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> liquidity::Li
         .downcast_ref::<uniswap_v2::Inner>()
         .expect("downcast uniswap settlement handler");
 
-    liquidity::Liquidity {
+    // Trading on Uniswap V2 pools where the reserves overflows `uint96`s does
+    // not work, so filter these pools out.
+    let max = (1_u128 << 96) - 1;
```
vkgnosis, line 48: Slight preference to using `pow` rather than `<<`.

vkgnosis (approved):
Do we not have the same problem in the pre colocation code?

nlordell:
> Do we not have the same problem in the pre colocation code?

No, we filter this out at a different stage:

https://github.com/cowprotocol/services/blob/f73f0ddb8a2fad22f6b6d1b19226eeb1ca6b9d86/crates/shared/src/sources/uniswap_v2/pool_fetching.rs#L24C16-L24C33

Although, I realize the bit width I used was wrong.

crates/driver/src/boundary/liquidity/uniswap/v2.rs
nlordell: Changed.

MartinquaXD (approved):

PR #1641, opened 2023-07-04
Sli order events

### LINKED ISSUE #1582: Time To Happy Moo SLI Implementation
This issue captures the additional work for adding measurements required for computing a "time to happy moo" SLI.

We decided to implement this by creating an "event log" per order in the database, thus allowing us to re-construct the "time to happy moo" for a particular order.

1. Create a new `order_events` table for storing order event logs. This should be a simple table with:
   ```
   order_uid | timestamp | event
   ```
  - We should consider whether or not this should be a separate DB or schema, to avoid performance issues with our orderbook database.
2. Store events for each **market** order:
   - At order creation, store a `order_uid, now(), 'created'` log
   - During auction cutting, store a `order_uid, now(), $outcome` log, where the outcome can be any one of the outcomes for an order during an auction cut:
     - `'ready'`: the order is included in an auction and sent to solvers
     - `'filtered'`: the order is not included in an auction and is not being considered. This is for cases where an order was filtered because of protocol reasons and not because it was invalid. It **should count towards the time to happy MOO**.
     - `'invalid'`: the order is invalid, so it cannot be settled on-chain. This **should not count towards the time to happy MOO**.
   - On settlement competition updating, (i.e. when the driver posts the settlement competition to the `orderbook`)
     - `'executing'`: the order was included in the winning settlement and is being executed on-chain
     - `'considered'`: the order was included in a settlement that did not win
   - When a trade event happens, store a `order_uid, now(), 'traded'` log, this will correspond to the user hearing the "MOOOO" in the UI (as it is tied to the `order.status` field that is provided by the backend which uses this trade event).
   - At order cancellation, store a `order_uid, now(), 'cancelled'` log
3. Periodically clean up table (remove all events older than 60 days for example?) - exact parameters for clean up TBD.
   
### Open Questions ❓ 

- Should we also add an `auction_id` column to the logs?
  - My intuition is that this might be useful for debugging but *not* needed for the SLI, so I would keep it out this issue.
- Should we also add an `block_number` column to the logs?
  - Similar to above. We could, however, pivot the SLI to "number of blocks to happy MOO" where this would be useful, I would keep it out of this issue for now, but would be slightly more inclined to introduce this from the beginning.
  
### Sample Data 📈 

As presented in this issue, you can assume some sample data that looks like:

```sql
CREATE TYPE OrderEvent AS ENUM (
  'created',
  'ready',
  'filtered',
  'invalid',
  'executing',
  'considered',
  'traded',
  'cancelled'
);

CREATE TABLE order_events (
    order_uid bytea NOT NULL,
    timestamp timestamptz NOT NULL,
    event OrderEvent NOT NULL
);
```

<details><summary>Some sample data:</summary>

```sql
-- Order matched right away
INSERT INTO order_events
VALUES
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', '2023-01-01 00:00:00', 'created'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', '2023-01-01 00:00:02', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', '2023-01-01 00:00:07', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', '2023-01-01 00:00:10', 'executing'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', '2023-01-01 00:00:19', 'traded');

-- Order that got cancelled after some time
INSERT INTO order_events
VALUES
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', '2023-01-01 00:00:00', 'created'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', '2023-01-01 00:00:02', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', '2023-01-01 00:00:07', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', '2023-01-01 00:00:12', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', '2023-01-01 00:00:14', 'cancelled');

-- Order that became invalid, but then eventually traded, not that only
-- a portion of its lifetime should contribute to the "time to happy MOO".
INSERT INTO order_events
VALUES
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:00:00', 'created'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:00:02', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:00:07', 'invalid'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:00:12', 'invalid'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:01:12', 'invalid'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:01:17', 'ready'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:01:23', 'executing'),
  ('\x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002', '2023-01-01 00:01:29', 'traded'),
```

_Disclaimer_: SQL statements were not tested, but should work (or work with minor alterations).

</details>

MartinquaXD:
Fixes #1582 

This PR implements storing timestamped events throughout the life cycle of every order.
In the autopilot all events get uploaded in a background task to not block the protocol unnecessarily. This is probably overkill in some places but I wanted to have a simple API that gets the job done.

### Test Plan
It seems we currently don't have the tools required to test DB state when running an e2e test so I modified an existing e2e test that simply settles an order and checked my local DB afterwards. Looked good enough to merge but should be followed up with a PR implementing and actual e2e test for that.

Screenshot 2023-07-04 at 15 18 58

### DIFF
--- a/crates/autopilot/src/database/order_events.rs
+++ b/crates/autopilot/src/database/order_events.rs
@@ -12,7 +12,7 @@ use {
 
 impl super::Postgres {
     /// Stores the given order event for all passed [`OrderUid`]s in a
-    /// background thread to not block the process.
+    /// background task to not block the thread.
     pub fn store_order_events(&self, uids: Vec<OrderUid>, label: OrderEventLabel) {
         if uids.is_empty() {
             return;
--- a/crates/autopilot/src/database/order_events.rs
+++ b/crates/autopilot/src/database/order_events.rs
@@ -1,6 +1,5 @@
 pub use database::order_events::OrderEventLabel;
 use {
-    anyhow::{Context, Result},
     chrono::{DateTime, Utc},
     database::{
         byte_array::ByteArray,
@@ -21,17 +20,21 @@ impl super::Postgres {
 
         let now = Utc::now();
         let db = self.0.clone();
-        let insert = async move {
-            let mut ex = db.acquire().await.context("acquire")?;
-            store_order_events(&mut ex, &uids, now, label)
-                .await
-                .context("store events")
-        };
         tokio::task::spawn(
             async move {
-                if let Err(err) = insert.await {
-                    tracing::warn!(?label, ?err, "failed to store order events");
-                }
+                let mut ex = match db.acquire().await {
+                    Ok(ex) => ex,
+                    Err(err) => {
+                        tracing::warn!(
+                            ?label,
+                            ?uids,
+                            ?err,
+                            "failed to acquire DB connection; could not store order events"
+                        );
+                        return;
+                    }
+                };
+                store_order_events(&mut ex, &uids, now, label).await;
             }
             .instrument(tracing::Span::current()),
         );
@@ -43,9 +46,9 @@ async fn store_order_events(
     uids: &[OrderUid],
     timestamp: DateTime<Utc>,
     label: OrderEventLabel,
-) -> Result<()> {
+) {
     for uid in uids {
-        order_events::insert_order_event(
+        if let Err(err) = order_events::insert_order_event(
             ex,
             &OrderEvent {
                 order_uid: ByteArray(uid.0),
@@ -53,7 +56,9 @@ async fn store_order_events(
                 label,
             },
         )
-        .await?;
+        .await
+        {
+            tracing::warn!(?label, ?uid, ?err, "failed to store order event");
+        }
     }
-    Ok(())
 }
--- a/crates/autopilot/src/database/order_events.rs
+++ b/crates/autopilot/src/database/order_events.rs
@@ -1,64 +1,41 @@
 pub use database::order_events::OrderEventLabel;
 use {
+    anyhow::{Context, Result},
     chrono::{DateTime, Utc},
     database::{
         byte_array::ByteArray,
         order_events::{self, OrderEvent},
     },
     model::order::OrderUid,
-    sqlx::PgConnection,
-    tracing::Instrument,
 };
 
 impl super::Postgres {
-    /// Stores the given order event for all passed [`OrderUid`]s in a
-    /// background task to not block the thread.
-    pub fn store_order_events(&self, uids: Vec<OrderUid>, label: OrderEventLabel) {
-        if uids.is_empty() {
-            return;
+    /// Inserts the given events with the current timestamp into the DB.
+    /// If this function encounters an error it will only be printed. More
+    /// elaborate error handling is not necessary because this is just
+    /// debugging information.
+    pub async fn store_order_events(&self, events: &[(OrderUid, OrderEventLabel)]) {
+        if let Err(err) = store_order_events(self, events, Utc::now()).await {
+            tracing::warn!(?err, "failed to insert order events");
         }
-
-        let now = Utc::now();
-        let db = self.0.clone();
-        tokio::task::spawn(
-            async move {
-                let mut ex = match db.acquire().await {
-                    Ok(ex) => ex,
-                    Err(err) => {
-                        tracing::warn!(
-                            ?label,
-                            ?uids,
-                            ?err,
-                            "failed to acquire DB connection; could not store order events"
-                        );
-                        return;
-                    }
-                };
-                store_order_events(&mut ex, &uids, now, label).await;
-            }
-            .instrument(tracing::Span::current()),
-        );
     }
 }
 
 async fn store_order_events(
-    ex: &mut PgConnection,
-    uids: &[OrderUid],
+    db: &super::Postgres,
+    events: &[(OrderUid, OrderEventLabel)],
     timestamp: DateTime<Utc>,
-    label: OrderEventLabel,
-) {
-    for uid in uids {
-        if let Err(err) = order_events::insert_order_event(
-            ex,
-            &OrderEvent {
-                order_uid: ByteArray(uid.0),
-                timestamp,
-                label,
-            },
-        )
-        .await
-        {
-            tracing::warn!(?label, ?uid, ?err, "failed to store order event");
-        }
+) -> Result<()> {
+    let mut ex = db.0.begin().await.context("begin transaction")?;
+    for (uid, label) in events {
+        let event = OrderEvent {
+            order_uid: ByteArray(uid.0),
+            timestamp,
+            label: *label,
+        };
+
+        order_events::insert_order_event(&mut ex, &event).await?
     }
+    ex.commit().await?;
+    Ok(())
 }
--- a/crates/autopilot/src/database/order_events.rs
+++ b/crates/autopilot/src/database/order_events.rs
@@ -1,0 +1,59 @@
+pub use database::order_events::OrderEventLabel;
+use {
+    anyhow::{Context, Result},
+    chrono::{DateTime, Utc},
+    database::{
+        byte_array::ByteArray,
+        order_events::{self, OrderEvent},
+    },
+    model::order::OrderUid,
+    sqlx::PgConnection,
+    tracing::Instrument,
+};
+
+impl super::Postgres {
+    /// Stores the given order event for all passed [`OrderUid`]s in a
+    /// background thread to not block the process.
+    pub fn store_order_events(&self, uids: Vec<OrderUid>, label: OrderEventLabel) {
+        if uids.is_empty() {
+            return;
+        }
+
+        let now = Utc::now();
+        let db = self.0.clone();
+        let insert = async move {
+            let mut ex = db.acquire().await.context("acquire")?;
+            store_order_events(&mut ex, &uids, now, label)
+                .await
+                .context("store events")
+        };
+        tokio::task::spawn(
+            async move {
+                if let Err(err) = insert.await {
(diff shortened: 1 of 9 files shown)


MartinquaXD:
One thing to consider is that an order can currently be `Considered XOR Executing` but these events don't get stored at the same point in time. That means an order could be inside the winning solution (i.e. not qualify for `Considered`) but fail to make it to the execution phase (no `Executing` event). That means it's theoretically possible that we lose some information here. LMK if you have strong opinions on the issue or if it's fine as it is.

nlordell:
> Looked good enough to merge but should be followed up with a PR implementing and actual e2e test for that.

YES PLEASE! While you're at it, there is a `TODO` in the E2E tests to check DB state for solver competition information after the auction settles which would be nice to fix with your proposed E2E test improvements.

crates/autopilot/src/run_loop.rs
```diff
@@ -379,4 +391,24 @@ impl RunLoop {
...
+            .iter()
+            .flat_map(|(_, response)| response.orders.iter())
+            .collect();
+        let considered_orders: Vec<OrderUid> = considered_orders
+            .difference(&winning_orders)
+            .map(|o| **o)
+            .collect();
+        self.database
+            .store_order_events(considered_orders, OrderEventLabel::Considered);
+    }
```
nlordell, line 413: One way to avoid the issue with `considered XOR executing` that you mentioned would be to label all orders included in solutions as `considered`, and then when we are going to execute them, then add an `executing` event.

This way we don't have such an issue with missing data, and an order for which there was a proposed solution that did not execute (regardless of the reason, whether it was second place or because of an error in between choosing the winner and executing) will get an event.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -168,17 +169,25 @@ impl SolvableOrdersCache {
         let mut counter = OrderFilterCounter::new(self.metrics, &db_solvable_orders.orders);
 
         let orders = filter_banned_user_orders(db_solvable_orders.orders, &self.banned_users);
-        counter.checkpoint("banned_user", &orders);
+        let removed = counter.checkpoint("banned_user", &orders);
+        self.database
+            .store_order_events(removed, OrderEventLabel::Invalid);
```
nlordell, line 174: Seeing this repetition, does it makes sense to add a:

```rust
counter.checkpoint_with_events("banned_user", &orders, OrderEventLabel::Invalid);
```

To prevent accidentally adding new checkpoints and forgetting to store the order events in the DB?

crates/autopilot/src/solvable_orders.rs
```diff
@@ -210,24 +219,34 @@ impl SolvableOrdersCache {
...
+        self.database
+            .store_order_events(removed, OrderEventLabel::Filtered);
 
         let orders = apply_fee_objective_scaling_factor(orders, &self.fee_objective_scaling_factor);
-        counter.checkpoint("fee_scaling_overflow", &orders);
+        let removed = counter.checkpoint("fee_scaling_overflow", &orders);
+        self.database
+            .store_order_events(removed, OrderEventLabel::Filtered);
 
         let auction = Auction {
```
nlordell, line 251: `counter.record()` logs warnings for "uncheckpointed" orders. Should we also have something catch-all like:

```rust
        self.database
            .store_order_events(removed, OrderEventLabel::Filtered);
```

nlordell (approved):
LGTM! Some small suggestions, but the PR looks good to merge as is.

crates/autopilot/src/solvable_orders.rs
MartinquaXD: I also though about that but then you either have to make the DB a member of `OrderFilterCounter` or have a helper function on `SolvableOrdersCache` which takes a lot of arguments. Both seem kind of weird and we already have the problem of being able to forget creating a checkpoint after introducing a new filter so I'm not sure this is worth it.

database/sql/V054__store_order_events.sql
```diff
@@ -0,0 +1,16 @@
...
+  'considered',
+  'traded',
+  'cancelled'
+);
+
+CREATE TABLE order_events (
+    order_uid bytea NOT NULL,
+    timestamp timestamptz NOT NULL,
+    label OrderEventLabel NOT NULL
+);
```
vkgnosis, line 16: In this PR you only define event insertion but we likely already know in what way we want to read them. For this there should be an index to make it efficient. You know this better than I but I would guess we want a sorted index (b-tree) over `(order_uid, timestamp)` to efficiently retrieve all the events for an order ordered by time.

crates/autopilot/src/database/order_events.rs
```diff
@@ -0,0 +1,64 @@
...
+    },
+    model::order::OrderUid,
+    sqlx::PgConnection,
+    tracing::Instrument,
+};
+
+impl super::Postgres {
+    /// Stores the given order event for all passed [`OrderUid`]s in a
+    /// background task to not block the thread.
+    pub fn store_order_events(&self, uids: Vec<OrderUid>, label: OrderEventLabel) {
```
vkgnosis, line 16: I'd like to make this function more general and if you need a specific helper version of it to define that locally. For me the following changes make sense here:
- normal async fn
- accept `Vec<(OrderUid, Label, Timestamp)>`
- do all insertions in one transaction
- add a metric for how long the query takes like for the other DB operations

This version should be easy to convert into the slight variations needed in other places. See my comment in the solvable_orders module.

I don't like forcing the background thread because it isn't necessary. Talking to the DB should be fast and it's easier to understand the code without background thread. Especially fast when we batch all the events into one transaction as I suggest for solvable_orders.

MartinquaXD: > accept Vec<(OrderUid, Label, Timestamp)>

The reason I didn't add that was because it wasn't necessary, yet. It appears to me the function (signature) I have at the moment is general purpose enough for the needs we have at the moment.

> do all insertions in one transaction

I wanted to avoid transactions to prevent a single failed insertion from dropping all other events.
However, I'm not sure what would cause the insertion to fail but I thought better safe than sorry. If this is a non-issue I agree that a single transaction is preferable.

vkgnosis: I don't see how a single event insertion can fail.

vkgnosis: > enough for the needs we have at the moment.

Yes but it also influences how people use the function. For example my suggestion solvable_orders might seem harder to implement because it doesn't fit the current specialized signature well. But yeah it's always a trade off to find the right amount of generalization.

crates/autopilot/src/run_loop.rs
MartinquaXD: My initial inclination was to that as well but in the issue you were pretty specific:
> 'considered': the order was included in a settlement that did not win

But since you are fine with relaxing that constraint a bit I'm happy to simplify the code. 👍

MartinquaXD: [906404e](https://github.com/cowprotocol/services/pull/1641/commits/906404ee1812468c1d02cadc8fdf1239636cc41d)

crates/autopilot/src/solvable_orders.rs
MartinquaXD: [d0d0770](https://github.com/cowprotocol/services/pull/1641/commits/d0d0770fc23ad39fcb6ff2c25eaa2f242aa1adf9)

crates/autopilot/src/database/order_events.rs
MartinquaXD: I decided to go only with `[(Uid, Label)]` because we probably never will waste a meaningful amount of time between determining labels.
[3550229](https://github.com/cowprotocol/services/pull/1641/commits/3550229b1bb49f862538032fb074ecb3ca857940)

vkgnosis (approved):

crates/autopilot/src/run_loop.rs
nlordell: Yeah, sorry for moving the goal posts, just thought it made sense given how the code turned out.

PR #1644, opened 2023-07-05
Separate Hooks from Interactions

nlordell:
This PR separates the concept of a "user-hook" from a "interaction". This distinction is useful because:
- It allows "hooks" to have different fields than interactions (namely, they don't have a `value` and specify need to specify a `gasLimit`; we can potentially add other fields in the future like `canRevert` as well).
- It means that the custom hook trampolining gets encoded at the orderbook level - this means that an order's `interactions.{pre,post}` already encode the required trampolining so:
  - All our simulation logic uses interactions exactly as they will be executed as part of a settlement
  - Simplifies settlement encoding logic (driver assumes that the order's interaction is meant to be encoded as-is)
  
In follow ups I will:
- [x] Extend the quoting logic's fee verification to also include the `gasLimit` for the hooks; allowing the orderbook to enforce fees even when verified quotes are disabled (implemented in #1646)
- [x] Revert many uneccessary changes from #1605 (implemented in #1645)

### Test Plan

E2E tests for pre-interactions (now pre-hooks) continue to pass!


vkgnosis (approved):

crates/shared/src/order_validation.rs
```diff
@@ -513,6 +533,23 @@ impl OrderValidating for OrderValidator {
...
+                return Err(ValidationError::Partial(
+                    PartialValidationError::UnsupportedCustomInteraction,
+                ));
+            }
+            // custom interactions are not allowed for partially fillable orders
+            if order.partially_fillable {
+                return Err(ValidationError::Partial(
+                    PartialValidationError::UnsupportedCustomInteraction,
+                ));
+            }
```
MartinquaXD, line 549: I think with the introduction of hooks this restriction could actually be lifted eventually.
AFAIK the only problem here is that was unclear in the context of partial fills whether a custom interaction should be executed on every fill or only on the first or last fill.
With hooks we could theoretically support arbitrarily complex rules traders could use.
That being said, I think it would be too much hassle for a probably very niche scenario.

MartinquaXD (approved):

crates/shared/src/order_validation.rs
nlordell: We had it in the MVP spec to not support this for partially fillable orders. Agreed that now that we have "hooks" we can add support eventually (by specifying when the hook should be executed - first part, when its done, every time, etc.).

PR #1646, opened 2023-07-05
Include Additional Hook Gas In Quoted Fee

nlordell:
This PR includes the additional hook `gas_limit` added in #1644 into the quoted fee amount.

### Test Plan

Existing tests continue to pass (including E2E test which uses hooks). Also, modified a unit test to consider additional gas costs from signature verification and hooks.

### DIFF
--- a/crates/shared/src/order_quoting.rs
+++ b/crates/shared/src/order_quoting.rs
@@ -896,8 +896,11 @@ mod tests {
                 from: H160([3; 20]),
                 ..Default::default()
             }),
-            signing_scheme: QuoteSigningScheme::Eip712,
-            additional_gas: 0,
+            signing_scheme: QuoteSigningScheme::Eip1271 {
+                onchain_order: false,
+                verification_gas_limit: 1,
+            },
+            additional_gas: 2,
         };
         let gas_price = GasPrice1559 {
             base_fee_per_gas: 1.5,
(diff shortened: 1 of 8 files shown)


crates/shared/src/order_quoting.rs
```diff
@@ -883,7 +896,11 @@ mod tests {
                 from: H160([3; 20]),
                 ..Default::default()
             }),
-            signing_scheme: QuoteSigningScheme::Eip712,
+            signing_scheme: QuoteSigningScheme::Eip1271 {
+                onchain_order: false,
+                verification_gas_limit: 1,
+            },
+            additional_gas: 2,
```
nlordell, line 903: The quote’s `gas` amount was `3`, and now we have an additional `1 + 2` gas from signature verification...

crates/shared/src/order_quoting.rs
```diff
@@ -992,8 +1009,8 @@ mod tests {
                 },
                 sell_amount: 100.into(),
                 buy_amount: 42.into(),
-                fee_amount: 15.into(),
-                full_fee_amount: 30.into(),
+                fee_amount: 30.into(),
+                full_fee_amount: 60.into(),
```
nlordell, line 1013: ... Meaning the fee amounts should double.

vkgnosis (approved):

MartinquaXD (approved):

PR #1648, opened 2023-07-05
Fix solver HTTP timeout bug

cowfee:
We had seen some solver timeouts happening in staging. I think the reason for this was because the HTTP timeout for the solver was being set to exactly the same value as the auction deadline. This could cause the HTTP timeout to happen _before_ the solver interrupts its search algorithm. I added a small 500 millisecond buffer to the HTTP request to account for this.

### Test Plan

See if the error goes away next time we apply this to staging.


crates/driver/src/infra/solver/mod.rs
```diff
@@ -125,7 +125,13 @@ impl Solver {
             .client
             .post(self.config.endpoint.clone())
             .body(body)
-            .timeout(timeout.into());
+            .timeout(
+                // Add a small buffer to the timeout to account for network latency. Otherwise the
+                // HTTP timeout might happen before the solver times out its search algorithm.
+                (timeout.duration() + chrono::Duration::milliseconds(500))
```
cowfee, line 131: Do you think 500 milliseconds is the right value? Should it be lower? Or higher?

MartinquaXD: I think 500ms should probably enough. However, I would say we should rather deduct that from the solver deadline and leave the http timeout as is.
My reasoning is that this would give solvers stronger guarantees. For example in case they have to manage inventory risk for their solutions they don't suddenly have to account for additional 500ms they wouldn't really know about. These timing concerns came up a lot while talking to the MooMaker team.

MartinquaXD (approved):

crates/driver/src/infra/solver/mod.rs
nlordell: Agree with @MartinquaXD

nlordell (approved):

PR #1653, opened 2023-07-06
Send Order Interactions To Colocated Driver

nlordell:
In debugging why custom order interactions were not working on Goerli for my demo, I realized that we weren't serializing order interactions in the auction that we send to co-located drivers 🤦.

This PR fixes this issue, we now correctly marshal the interactions to the co-located drivers for solving.

### Test Plan

Duplicated the hooks E2E test (formerly named `pre_interaction`, and renamed to `hooks` here, since we now call these hooks, and the E2E test included both pre- and post- hooks, so it was badly named to begin with).

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -14,6 +14,7 @@ use {
     itertools::Itertools,
     model::{
         auction::{Auction, AuctionId},
+        interaction::InteractionData,
         order::{LimitOrderClass, OrderClass},
     },
     primitive_types::{H160, H256},
@@ -216,6 +217,17 @@ impl RunLoop {
                             (Class::Limit, surplus_fee)
                         }
                     };
+                    let map_interactions =
+                        |interactions: &[InteractionData]| -> Vec<solve::Interaction> {
+                            interactions
+                                .iter()
+                                .map(|interaction| solve::Interaction {
+                                    target: interaction.target,
+                                    value: interaction.value,
+                                    call_data: interaction.call_data.clone(),
+                                })
+                                .collect()
+                        };
                     solve::Order {
                         uid: order.metadata.uid,
                         sell_token: order.data.sell_token,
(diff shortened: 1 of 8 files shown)


crates/e2e/tests/e2e/colocation_hooks.rs
```diff
@@ -0,0 +1,160 @@
...
+        "--account-balances-optimistic-pre-interaction-handling=true".to_string(),
+        "--enable-colocation=true".to_string(),
+        "--drivers=http://localhost:11088/test_solver".to_string(),
+    ]);
+    services
+        .start_api(vec![
+            "--enable-custom-interactions=true".to_string(),
+            "--account-balances-optimistic-pre-interaction-handling=true".to_string(),
+        ])
+        .await;
```
nlordell, line 63: This E2E test is a copy of the `hooks.rs` (former `pre_interactions.rs` E2E test) with the exception of the different setup here to use co-location instead of the legacy setup.

crates/driver/src/infra/observe/mod.rs
```diff
@@ -262,14 +262,21 @@ pub fn solver_response(solver: &solver::Name, endpoint: &Url, res: Result<&str,
...
     solver: &solver::Name,
     mempool: &Mempool,
+    settlement: &Settlement,
     res: &Result<eth::TxId, boundary::Error>,
 ) {
     match res {
         Ok(txid) => {
-            tracing::info!(%solver, ?txid, ?mempool, "sending transaction via mempool succeeded");
+            tracing::info!(
+                %solver, ?txid, ?mempool, ?settlement,
```
nlordell, line 271: Having the settlement here in the observation was really useful for debugging.

crates/autopilot/src/run_loop.rs
```diff
@@ -230,8 +242,8 @@ impl RunLoop {
                         owner: order.metadata.owner,
                         partially_fillable: order.data.partially_fillable,
                         executed: Default::default(),
-                        pre_interactions: Default::default(),
-                        post_interactions: Default::default(),
+                        pre_interactions: map_interactions(&order.interactions.pre),
+                        post_interactions: map_interactions(&order.interactions.post),
```
nlordell, line 246: This file has the only logic changes, just trivially marshaling auction data to the co-located solvers.

MartinquaXD (approved):

PR #1661, opened 2023-07-10
Pass token info to `driver`

cowfee:
There were problems with the ParaSwap solver indicating that token decimals are missing. This was because the autopilot wasn't passing the token info to the driver correctly.

### Test Plan

CI still passes, will test this again on staging at some point.

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -78,6 +78,8 @@ pub mod solve {
         #[serde_as(as = "Option<DisplayFromStr>")]
         pub price: Option<U256>,
         pub trusted: bool,
+        pub decimals: Option<u8>,
+        pub symbol: Option<String>,
     }
 
     #[serde_as]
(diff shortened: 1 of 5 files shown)


crates/autopilot/src/driver_model.rs
```diff
@@ -78,6 +78,8 @@ pub mod solve {
         #[serde_as(as = "Option<DisplayFromStr>")]
         pub price: Option<U256>,
         pub trusted: bool,
+        pub decimals: Option<u8>,
+        pub symbol: Option<String>,
```
MartinquaXD, line 82: I would move this into the driver. It's not core to the protocol and some solvers don't care about that at all.

nlordell: Agree with @MartinquaXD on this one.

cowfee: I agree as well but this will take more time to do.

nlordell: Can you add a `TODO` and create a GitHub issue?

vkgnosis (approved):

PR #1670, opened 2023-07-11
Port Swapr Liquidity to Colocated Driver

nlordell:
This PR ports Swapr liquidity to the co-located driver.

This is not a complicated effort, since it mostly shares logic with Uniswap V2 (with the exception of the "liquidity collector" setup and variable fees).

### Test Plan

Run the driver with Swapr liquidity enabled and quote:

```
% curl -s http://localhost:11088/mybologna/quote -H 'Content-Type: application/json' --data '@-' <<JSON | jq
{
  "sellToken": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
  "buyToken": "0xa1d65e8fb6e87b60feccbc582f7f97804b725521",
  "kind": "sell",
  "deadline": "2024-03-25T00:00:00.000Z"
}
JSON
{
  "amount": "2382618097769939792",
  "interactions": [
    {
      "value": "0",
      "callData": "0x095ea7b3000000000000000000000000b9960d9bca016e9748be75dd52f02188b9d0829f0000000000000000000000000000000000000000000000000000000000000000"
    },
    {
      "value": "0",
      "callData": "0x095ea7b3000000000000000000000000b9960d9bca016e9748be75dd52f02188b9d0829fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
    },
    {
      "target": "0xb9960d9bca016e9748be75dd52f02188b9d0829f",
      "value": "0",
      "callData": "0x8803dbee0000000000000000000000000000000000000000000000002110c28948fe83500000000000000000000000000000000000000000000000000e043da61725000000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab41ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000a1d65e8fb6e87b60feccbc582f7f97804b725521"
    }
  ],
  "solver": "0x7e5f4552091a69125d5dfcb7b8c2659029395bdf"
}
```

Notice the calldata target on the [`0xb9960d9bca016e9748be75dd52f02188b9d0829f`](https://etherscan.io/address/0xb9960d9bca016e9748be75dd52f02188b9d0829f#code) `DXswapRouter` (i.e. Swapr) contract.

Created #1669 to track work for creating E2E tests for all liquidity types (for quoting, solving and settling).

### DIFF
--- a/crates/driver/src/boundary/liquidity/swapr.rs
+++ b/crates/driver/src/boundary/liquidity/swapr.rs
@@ -1,10 +1,9 @@
 use {
     crate::{
-        boundary,
+        boundary::{self, Result},
         domain::liquidity::{self, swapr},
         infra::{self, blockchain::Ethereum},
     },
-    anyhow::Result,
     shared::{
         current_block::CurrentBlockStream,
         price_estimation,
@@ -16,15 +15,16 @@ use {
 /// The base unit for basis points, i.e. how many basis points in 100%.
 const BPS_BASE: u32 = 10_000;
 
-pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Option<liquidity::Liquidity> {
+pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Result<liquidity::Liquidity> {
     // invalid Swapr fee ratio; does not have exact BPS representation
-    if (pool.fee.numer() * BPS_BASE) % pool.fee.denom() != 0 {
-        return None;
-    };
+    anyhow::ensure!(
+        (pool.fee.numer() * BPS_BASE) % pool.fee.denom() == 0,
+        "invalid Swapr fee ratio; does not have exact BPS representation",
+    );
 
     let bps = (pool.fee.numer() * BPS_BASE) / pool.fee.denom();
     let fee = swapr::Fee::new(bps)?;
-    Some(liquidity::Liquidity {
+    Ok(liquidity::Liquidity {
         id,
         gas: price_estimation::gas::GAS_PER_UNISWAP.into(),
         kind: liquidity::Kind::Swapr(swapr::Pool {
--- a/crates/driver/src/boundary/liquidity/swapr.rs
+++ b/crates/driver/src/boundary/liquidity/swapr.rs
@@ -17,11 +17,10 @@ use {
 const BPS_BASE: u32 = 10_000;
 
 pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Option<liquidity::Liquidity> {
-    assert_eq!(
-        (pool.fee.numer() * BPS_BASE) % pool.fee.denom(),
-        0,
-        "invalid Swapr fee ratio; does not have exact BPS representation",
-    );
+    // invalid Swapr fee ratio; does not have exact BPS representation
+    if (pool.fee.numer() * BPS_BASE) % pool.fee.denom() != 0 {
+        return None;
+    };
 
     let bps = (pool.fee.numer() * BPS_BASE) / pool.fee.denom();
     let fee = swapr::Fee::new(bps)?;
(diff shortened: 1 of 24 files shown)


crates/driver/src/boundary/liquidity/swapr.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// The base unit for basis points, i.e. how many basis points in 100%.
+const BPS_BASE: u32 = 10_000;
+
+pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Option<liquidity::Liquidity> {
+    assert_eq!(
+        (pool.fee.numer() * BPS_BASE) % pool.fee.denom(),
+        0,
+        "invalid Swapr fee ratio; does not have exact BPS representation",
+    );
```
vkgnosis, line 24: This will panic. Is that what you want?

vkgnosis (approved):

crates/driver/src/boundary/liquidity/swapr.rs
nlordell: I guess I can make it s `debug_assert_eq`. My rationale for `panic`-ing here is that it should really only happen if there is a configuration issue.

Now that I think about it more, a misbehaving node could return an invalid result from the `swapFee` and cause us to panic... I will remove it.

nlordell: Removed panic.

crates/driver/src/domain/liquidity/swapr.rs
```diff
@@ -1,4 +1,51 @@
...
+}
+
+impl Fee {
+    /// Creates a new fee from the specified basis points. Returns `None` for
+    /// invalid fee values (i.e. outside the range `[0, 1000]`).
+    pub fn new(bps: u32) -> Option<Self> {
+        if !(0..=1000).contains(&bps) {
+            return None;
+        }
+        Some(Self(bps))
```
MartinquaXD, line 44:
```suggestion
        (0..=1000).contains(&bps).then_some(Self(bps))
```

MartinquaXD (approved):

crates/driver/src/boundary/liquidity/swapr.rs
```diff
@@ -0,0 +1,54 @@
...
+    solver::{liquidity::ConstantProductOrder, liquidity_collector::LiquidityCollecting},
+};
+
+/// The base unit for basis points, i.e. how many basis points in 100%.
+const BPS_BASE: u32 = 10_000;
+
+pub fn to_domain(id: liquidity::Id, pool: ConstantProductOrder) -> Option<liquidity::Liquidity> {
+    // invalid Swapr fee ratio; does not have exact BPS representation
+    if (pool.fee.numer() * BPS_BASE) % pool.fee.denom() != 0 {
+        return None;
```
cowfee, line 22: I don't like the idea of returning `Option` on error instead of `Result`. `Option` means that a value could be missing, `Result` means that something can go wrong.

cowfee: But I know that you're doing this in some other places as well, ideally the approach should be all-or-nothing. Even if you have to write just a single struct to represent the error I still think that's more correct and expresses your intent more clearly.

nlordell: Your logic is undeniable. Agreed on all fronts.

cowfee (approved):

crates/driver/src/boundary/liquidity/swapr.rs
nlordell: I switched to using unit-structs for representing these errors. I went with an "all-or-nothing" approach.

93024239136c195067de1ceee808179536c27af8
93024239136c195067de1ceee808179536c27af8

PR #1671, opened 2023-07-11
Mention non backend app data schema

vkgnosis:
[Slack discussion](https://cowservices.slack.com/archives/C056TDLDE0Y/p1689072332416029)

Also fixed some trailing whitespaces because my editor does it automatically.

### Test Plan

CI

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -53,7 +53,7 @@ paths:
         This is a *best effort* cancellation, and might not prevent solvers from
         settling the orders (if the order is part of an in-flight settlement
         transaction for example). Authentication must be provided by an
-        [EIP-712](https://eips.ethereum.org/EIPS/eip-712) 
+        [EIP-712](https://eips.ethereum.org/EIPS/eip-712)
         signature of an `OrderCancellations(bytes[] orderUids)` message.
       requestBody:
         description: Signed `OrderCancellations`.
@@ -98,7 +98,7 @@ paths:
       summary: Cancel an order by marking it invalid with a timestamp.
       description: |
         The successful deletion might not prevent solvers from settling the order.
-        Authentication must be provided by providing an 
+        Authentication must be provided by providing an
         [EIP-712](https://eips.ethereum.org/EIPS/eip-712) signature of an
         `OrderCancellation(bytes orderUid)` message.
       parameters:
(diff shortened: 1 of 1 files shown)


crates/orderbook/openapi.yml
```diff
@@ -668,14 +668,25 @@ components:
...
+
+                ```json
+                {
+                  "version": "0.7.0",
+                  "appCode": "YOUR_APP_CODE",
+                  "metadata": {}
+                }
+
                 The total byte size of this field's UTF-8 encoded bytes is limited to 1000.
+                ```
```
alfetopito, line 689: This is including the comment into the JSON code block AFAICT.
Should be moved to line `687` instead, no?

vkgnosis: Thanks. Fixed.

vkgnosis: Thanks. Fixed.

alfetopito (approved):

crates/orderbook/openapi.yml
```diff
@@ -668,13 +668,24 @@ components:
...
                 order with the `appData` field set appropriately. If this isn't the case, signature
-                verification will fail. For easier debugging it is recommended to additionally set the 
+                verification will fail. For easier debugging it is recommended to additionally set the
                 `appDataHash` field.
 
                 The field must be the encoding of a valid JSON object. The JSON object can contain
                 arbitrary application specific data (JSON key values). The optional key `backend` is
                 special. It **MUST** conform to the schema documented in `BackendAppData`.
 
+                The intended use of the other keys of the object is follow the standardized format
```
anxolin, line 678: What do you mean by "the other keys in the object"? 

I would love we use this opportunity to advertise how following the standard allows to attach arbitrary information to the order, this is important because it will enable among others to:
- Credit the order to a specific app
- Add additional instructions to solvers about your order preferences (i.e. CoW hooks, wait for cows, and any future protocol extension)
- Attach UTM codes to credit the order to a campaign

anxolin (commented):
Thanks 🙏

crates/orderbook/openapi.yml
vkgnosis: "other keys" refers to the previous paragraph that introduces the `backend` key.

vkgnosis: I guess long term it makes sense to document all of the app data in one place. There is no reason to document BackendAppData here and the other app data somewhere else.

vkgnosis: Then we could even enforce that the non backend relevant app data also matches the schema.

MartinquaXD (approved):

PR #1674, opened 2023-07-12
Implement Liquidity Presets for Driver Configuration

nlordell:
This PR changes the driver configuration format to accept `preset = "$name"` to make running the driver with known liquidity sources easier.

For example. you can now specify:

```toml
[[liquidity.$source]]
preset = "$name"
```

This makes running the driver locally against supported networks much easier (you don't need to look up router contract addresses or pool initialization codes).

I also realized that all of the `${uniswap_clone}Factory` contracts were not being used, so I cleaned them up from the `contracts` crate.

ALSO, we have PancakeSwap configured in Mainnet in our cluster, so I added it as a preset. I mostly did this to verify that the abstractions were good, and that it was trivial to add a new preset (yes, it was super trivial).

### Test Plan

Tests continue to pass, so no regression, and ran locally to try out presets.


vkgnosis (approved):

MartinquaXD (approved):
Looks alright to me.

crates/driver/src/infra/config/mod.rs
```diff
@@ -14,7 +14,7 @@ pub struct Config {
     pub disable_access_list_simulation: bool,
     pub disable_gas_simulation: Option<eth::Gas>,
     pub solvers: Vec<solver::Config>,
-    pub liquidity: liquidity::Config,
+    pub liquidity: LiquidityConfig,
```
cowfee, line 17: This isn't great. This `Config` type isn't supposed to know anything about the config files. It's just a type which collects all infra config objects. So this should be `liquidity::Config` like it was before...

cowfee, line 17: This isn't great. This `Config` type isn't supposed to know anything about the config files. It's just a type which collects all infra config objects. So this should be `liquidity::Config` like it was before...

crates/driver/src/infra/config/file/load.rs
```diff
@@ -47,41 +44,7 @@ pub async fn load(path: &Path) -> infra::Config {
...
-                .liquidity
-                .uniswap_v3
-                .into_iter()
-                .map(|config| liquidity::config::UniswapV3 {
-                    router: config.router.into(),
-                    max_pools_to_initialize: config.max_pools_to_initialize,
-                })
-                .collect(),
-        },
+        liquidity: config.liquidity,
```
cowfee, line 47: ...and the mapping from the config file representation (`LiquidityConfig`) into `liquidity::Config` should happen here, rather than in `run.rs`.

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -203,51 +203,106 @@ struct TenderlyConfig {
     save_if_fails: bool,
 }
 
-#[derive(Debug, Default, Deserialize)]
+#[derive(Clone, Debug, Default, Deserialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
-struct LiquidityConfig {
+pub struct LiquidityConfig {
```
cowfee, line 208: And if you do it that way, none of these types need to be public.

crates/driver/src/infra/config/mod.rs
nlordell: We did this with `ContractAddresses`, so I took the same approach here.

Essentially, it needs to be slightly different because runtime values are needed to do the mapping correctly. Specifically (like for `ContractAddresses`, the `eth::NetworkId` needs to be known.

Alternatively, and I played with this as well, but also didn't feel super clean, was to pass in the `eth::Ethereum` or `eth::NetworkId` to the `config::load` function. This would allow both `ContractAddresses` and `LiquidityConfig` to not leak.

cowfee: Hmm, OK. I agree that passing the `eth` object in `load` is not the best. I think maybe it's time to jettison this `Config` type entirely. I don't expect you to do that though (at least not in this PR).

cowfee (approved):

PR #1675, opened 2023-07-12
Start storing `settlement_observations`

MartinquaXD:
With solver driver colocation we no longer know upfront the `tx_from` and `tx_nonce` to expect for a given competition.
That's why we implemented the driver to append the settled `auction_id` to the calldata.
With that we can again associate which settlement belongs to which auction which allows us to correctly update the `settlement_obersvations` table again.

### Test Plan
Updated unit test for decoding the metadata
TODO: 1 big e2e test that verifies all the CIP-20 data is there when I implemented storing the `solver_competition` data


crates/autopilot/src/decoded_settlement.rs
```diff
@@ -197,16 +175,68 @@ impl TryFrom<database::orders::OrderExecution> for OrderExecution {
...
+        }
+        Self::try_new(without_selector, function, false).map_err(Into::into)
+    }
+
+    fn try_new(data: &[u8], function: &Function, with_metadata: bool) -> Result<Self> {
+        let metadata_len = if with_metadata {
+            anyhow::ensure!(
+                data.len() % 32 == Self::META_DATA_LEN,
+                "calldata does not contain the expected number of bytes to include metadata"
+            );
```
MartinquaXD, line 204: Apparently the calldata format is not guaranteed to have unique layout for a given ABI. However, in practice the ecosystem expects a certain ABI (it's already encoded in the `ethabi` crate so we wouldn't even be able to parse non-conforming calldata with or without appended `auction_id`). That's why I doubled down on this format and expect `calldata.len() % 32 == 12` when we try to extract the `auction_id`. 32 bytes for the word size (`uint256`), 4 bytes for the function selector, 8 bytes for the appended `auction_id`.
This function expects the function select to be removed already, hence `== Self::META_DATA_LEN`.

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -202,6 +209,65 @@ impl OnSettlementEventUpdater {
...
+            .any(|p| Some(H160(p.participant.0)) == tx.from)
+        {
+            tracing::debug!(auction_id, "extracted auction_id from settlement metadata");
+            Ok(Some(auction_id))
+        } else {
+            tracing::warn!(
+                auction_id,
+                ?tx,
+                "settlement was submitted by solver that didn't participate in the competition"
+            );
```
MartinquaXD, line 267: IMO this is a pretty serious issue so I would liked to have made this an `error` log. But because we have to environments (staging, prod) with 2 separate `auction_id` counters we may see `auction_id` collisions in the future which makes this check not bullet proof and might result in noisy false alarms.
Imagine 6 months from now prod reaches the same the current `auction_id` of staging. Then we settle an auction on prod and staging sees it. The staging DB might already have settled the same auction and would notice that the prod winner didn't even compete in the staging auction.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -197,16 +175,68 @@ impl TryFrom<database::orders::OrderExecution> for OrderExecution {
...
+        let without_selector = input
+            .strip_prefix(&function.selector())
+            .ok_or(DecodingError::InvalidSelector)?;
+
+        // Decoding calldata without expecting metadata can succeed even if metadata
+        // was appended. The other way around would not work so we do that first.
+        if let Ok(decoded) = Self::try_new(without_selector, function, true) {
+            return Ok(decoded);
+        }
+        Self::try_new(without_selector, function, false).map_err(Into::into)
```
vkgnosis, line 196: I'm slightly confused that we have to try once with and once without metadata. The way I would expect it it to work is that parsing the calldata with the function's ABI tells you how many bytes were consumed. The leftover bytes must then be the metadata. Maybe we don't have the right APIs for this and maybe this is related to your comment below.

vkgnosis (approved):

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: > The way I would expect it it to work is that parsing the calldata with the function's ABI tells you how many bytes were consumed.

Yeah, this would have been my preferred solution as well (this [comment](https://github.com/cowprotocol/ethcontract-rs/pull/936#issuecomment-1630558317) had exactly that in mind). But unless I'm missing something this functionality can not be patched using `ethcontract-rs` and I didn't want to copy lots of code over from `ethabi` to know exactly how many bytes would be used by that implementation.

crates/autopilot/src/on_settlement_event_updater.rs
nlordell: I would make this an `error` TBH. We can instead encode additional metadata other than the auction ID (like a 4-byte "environment" selector as `b"prod"` or `b"barn"`).

We can even pass a `salt` in the `auction` or in the `/settle` API call to the driver that must be appended.

nlordell (approved):

nlordell (approved):
LGTM

Agree that the E2E test would be nice here.

crates/autopilot/src/decoded_settlement.rs
nlordell: Note that Solidity ABI allows for arbitrary padding - so even then I don't think this would be "correct".

crates/autopilot/src/on_settlement_event_updater.rs
MartinquaXD: With the `salt` we might as well go back to where we started. 😅  (autopilot generating a secret key the solver is supposed to append)
Although this would require another table which associates the `auction_id` to the `salt`.

nlordell: I also don't think they are likely to collide for any period of time.

MartinquaXD: I think last time this came up I checked and estimated that the prod `auction_id` would catch up in staging in 6 months. Granted, this still doesn't guarantee that collisions will happen (not all `auction_id`s will lead to settlements) but the chance is there. Especially if we would really push for 1 block per auction.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: Totally agree, with that assessment. (That's why I suggested the idea of generating a secret key in the first place during the design phase).
Not sure why it we switched over to using the `auction_id`.

crates/autopilot/src/on_settlement_event_updater.rs
nlordell: > estimated that the prod auction_id would catch up in staging in 6 months

At which point, `prod` would surpass staging right? So its unlikely to stay "noisy" for any period of time - it might collide for a settlement or two at most.

nlordell: Agree we should prevent this with code though - at which point, I do think we need some salt and, as you said:

> With the `salt` we might as well go back to where we started 😅

crates/autopilot/src/decoded_settlement.rs
nlordell: It seamed reasonable at the time. I think it was because this was the simplest solution. See https://cowservices.slack.com/archives/C049P351GKF/p1688227366128809.


I think we can release with `settle(...) || auctionId` and can switch to `settle(...) || salt`.

crates/autopilot/src/on_settlement_event_updater.rs
MartinquaXD: Sorry, I meant it would catch up to the current staging ID. Meaning that in 6 months prod auction_ids will always have the potential to collide with a 6 months old staging ids.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: > I think it was because this was the simplest solution.

Strange argument given the fact that half of the logic was already there.

crates/autopilot/src/on_settlement_event_updater.rs
nlordell: > And even if one environment actually overtakes the other one it would only move the potential for collisions into the other environment.

Right, but my point is, if we assume an average `auctions/secs`, that the collision possibility will only happen when the two "auction IDs / time lines" collide.

So, once prod overtakes staging, I imagine auction IDs will continue to grow at a quicker rate in production than in staging and make it so they don't collide anymore.

I agree that this is not ideal and we should solve it though.

MartinquaXD: Collisions (at least in the context of this particular thread) are cases when e.g. staging indexes a settlement from prod but already has data for that auction_id (because it generated and settled that auction_id first). Now when staging looks at the participants for this auction it will not find the `tx_from` because it came from a prod solver which did not partake in the staging environment.
At this point it's unclear whether some rogue staging solver settled this auction out of order or if a law abiding solver from the other environment settled this auction fair and square.
This will never change. Not with both auction ids being close to each other nor with 1 environment being way ahead of the other.
The only thing we could do to decide what happened (rogue solver vs. settlement from other env) is to check if our env already indexed the auction data for this id. If we already know about the auction we could just assume the solution came from the other env.

MartinquaXD: > check if our env already indexed the settlement_observation

Actually we can't even do that at the moment because the only thing associating an `auction_id` with a transaction is the `auction_transactions` table which we will no longer populate because we are missing the `nonce`. 😩

MartinquaXD: [6590139](https://github.com/cowprotocol/services/pull/1675/commits/659013951d89d750b2be660a5f5f39ed9ce72b1b) and more importantly [7ba9e4b](https://github.com/cowprotocol/services/pull/1675/commits/7ba9e4bb0e5ec28c1319ff19f70b2fccb465ec15) try to address this issue as accurately as possible.
Could you give it another careful look to see if we are good @nlordell?
Another big issue I noticed is that theoretically a malicious solver could submit a solution with an auction_id for which we already recorded all the data. This could have resulted in us overwriting existing data.
Now we only insert into the `settlement_observations` table if the solution was provided by the winning solver for the given address and we don't already have data for that. To ensure correctness I had to do the entire update in a single big DB transaction.

crates/autopilot/src/decoded_settlement.rs
```diff
@@ -40,16 +40,20 @@ type DecodedSettlementTokenized = (
...
 pub struct DecodedSettlement {
     // TODO check if `EncodedSettlement` can be reused
     pub tokens: Vec<Address>,
     pub clearing_prices: Vec<U256>,
     pub trades: Vec<DecodedTrade>,
     pub interactions: [Vec<DecodedInteraction>; 3],
+    /// Data that was appended to the regular call data of the `settle()` call
+    /// as a form of on-chain meta data. This gets used to associated a
+    /// settlement with an auction.
+    pub metadata: Option<Bytes<[u8; Self::META_DATA_LEN]>>,
```
cowfee, line 53: I don't like that we're inventing this new concept of "settlement meta data". The last few bytes are just the auction ID. This is known ahead of time.

cowfee (commented):
Wouldn't it be simpler to also encode the auction ID into the settlements published by the autopilot? And then you _always_ know that the last 8 bytes (or whatever) are the auction ID. This only matters during the interim period anyway, and it seems like much of the complexity in this PR comes from this discrepancy.

crates/autopilot/src/decoded_settlement.rs
MartinquaXD: Yeah, this felt more correct when I tried to implement a more generic solution. But because that had subtle issues I went for the bare bones where it indeed makes more sense to simply call that settlement id.

MartinquaXD:
Adding the id in the current driver seems like a good idea. 👍

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -195,13 +212,72 @@ impl OnSettlementEventUpdater {
...
+
+        let score = database::settlement_scores::fetch(ex, auction_id).await?;
+        let data_already_recorded =
+            database::auction_transaction::data_exists(ex, auction_id).await?;
+        match (score, data_already_recorded) {
+            (None, _) => {
+                tracing::debug!(
+                    auction_id,
+                    "calldata claims to settle auction that has no competition"
+                );
```
sunce86, line 259: nit: This case will be entered for each transaction exactly once on different environment (will be entered in prod for staging transaction).
Nice to have for completness but maybe worth skipping since does not tell us much.

sunce86 (approved):
Looked into detail and I think you covered all cases.

PR #1677, opened 2023-07-12
Log settlement submission error

vkgnosis:
We noticed it was confusing that settlement submission didn't work but no message was printed.

### Test Plan

CI

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -504,7 +504,10 @@ impl Driver {
                     Some(receipt.transaction_hash)
                 }
                 Err(SubmissionError::Revert(hash)) => Some(hash),
-                _ => None,
+                Err(err) => {
+                    tracing::warn!(?err, "settlement submission error");
+                    None
+                }
             };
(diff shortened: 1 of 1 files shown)


crates/solver/src/driver.rs
```diff
@@ -504,7 +504,7 @@ impl Driver {
                     Some(receipt.transaction_hash)
                 }
                 Err(SubmissionError::Revert(hash)) => Some(hash),
-                _ => None,
+                Err(other) => tracing::warn!("settlement submission error {err:?}"),
```
cowfee, line 507: Should still return `None` :smile`

cowfee (approved):

crates/solver/src/driver.rs
MartinquaXD: Also:
```rust
Err(err) => tracing::warn!(?err, "settlement submission error")
```

vkgnosis: Fixed. Wrote this without running check.

nlordell (approved):

Issue #1672, opened 2023-07-11
Colocation tasks before production


cowfee:
## Before Staging
<details>
  <summary>✅</summary>

- Store CIP-20 data:
  - [x] Insert data into `settlement_observations` (looks like a trivial bug).
  - [x] Insert data into `solver_competition` (we have to build the JSON blob we store and also store that). https://github.com/cowprotocol/services/issues/1603 and update solver rewards script.
  - [x] Define new shape of solver_competition and save it to db https://github.com/cowprotocol/services/issues/1614
- [x] Respect auction deadline in the `solvers` crate. https://github.com/cowprotocol/services/issues/1664
- Sort orders in the driver before sending the auction to solvers:
  - [x] Sort market orders before limit orders.
  - [x] Sort well-priced orders before mispriced ones, using the reference prices provided in the auction. Make the sorting order randomized, weighted by how likely each order is to be settled. So more likely-to-settle orders will probably come first, and less likely-to-settle orders will come later, but the sorting order shouldn't be deterministic. Here's a [reference PR](https://github.com/cowprotocol/services/pull/762) which implemented this in the old code.
  - [x] Filter orders based on user balance. If the user doesn't have enough balance to settle an order, it should be filtered out. Note that if there are multiple orders from the same user, the later orders should be filtered out if the user's balance only covers the former orders. So the filtering should be context-aware.
- https://github.com/cowprotocol/services/issues/1666:
  - [x] Balancer.
    - [x] https://github.com/cowprotocol/services/pull/1698
    - [x] https://github.com/cowprotocol/services/pull/1700
    - [x] https://github.com/cowprotocol/services/pull/1701
    - [x] https://github.com/cowprotocol/services/pull/1702
  - [x] https://github.com/cowprotocol/services/pull/1670
- Fetch more token metadata in the driver (currently, the autopilot passes all this metadata to the driver, but in the future the driver should do this on its own):
  - [x] Token decimals.
  - [x] Token symbols.
  - [x] Token balance available in the settlement contract.
- [x] [Non blocking] Add `/reveal` to driver https://github.com/cowprotocol/services/issues/1744
- [x] Score calculation https://github.com/cowprotocol/services/issues/1893
</details>

## Before Production
- [ ] Make sure solver reward payout scripts work as expected
- [ ] cowprotocol/k8s-autodeploy#3
- [ ] cowprotocol/services#1949

## After Production (Cleanups)
- [ ] cowprotocol/services#1673
[...]

PR #1678, opened 2023-07-12
`driver`: sort orders by likelihood of fulfillment

cowfee:
Progress on https://github.com/cowprotocol/services/issues/1672.

Sort orders in the driver by likelihood of fulfillment before passing them to the solver.

### Test Plan

Added an automated test.


crates/driver/src/domain/competition/auction.rs
```diff
@@ -11,25 +12,84 @@ use {
...
+            match left_kind.cmp(&right_kind) {
+                std::cmp::Ordering::Equal => {
+                    // If the orders are of the same kind, then sort by likelihood of fulfillment
+                    // based on token prices.
+                    left.likelihood(tokens).cmp(&right.likelihood(tokens))
+                }
+                other => other,
+            }
+        });
+        orders.reverse();
```
MartinquaXD, line 55: I think this should be equivalent.
```suggestion
        orders.sort_by_key(|order| {
            // Market orders are preferred over limit orders, as the expectation is that
            // they should be immediately fulfillable. Liquidity orders come last, as they
            // are the most niche and rarely used.
            let class = match order.kind {
                competition::order::Kind::Market => 2,
                competition::order::Kind::Limit { .. } => 1,
                competition::order::Kind::Liquidity => 0,
            };
            std::cmp::Reverse((class, order.likelihood(tokens)))
        });
```

crates/driver/src/domain/competition/auction.rs
```diff
@@ -11,25 +12,84 @@ use {
...
+
+        // TODO Filter out orders based on user balance
+
+        Self(orders)
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &competition::Order> {
```
MartinquaXD, line 62: Wouldn't it be more rusty to implement `IntoIterator` instead?
Same remark on `Tokens`.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -159,6 +172,23 @@ impl Order {
...
+        match (
+            tokens.get(self.buy.token).price,
+            tokens.get(self.sell.token).price,
+        ) {
+            (Some(buy_price), Some(sell_price)) => {
+                let buy = buy_price.apply(self.buy.amount);
+                let sell = sell_price.apply(self.sell.amount);
+                conv::u256::to_big_rational(buy.0) / conv::u256::to_big_rational(sell.0)
+            }
+            _ => num::BigRational::zero(),
```
MartinquaXD, line 189: nit: Prices for traded tokens should always be in the auction. Would be nice if that could be encoded in the types somehow.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -159,6 +172,23 @@ impl Order {
...
+    /// This is used to prioritize orders when solving.
+    pub fn likelihood(&self, tokens: &auction::Tokens) -> num::BigRational {
+        match (
+            tokens.get(self.buy.token).price,
+            tokens.get(self.sell.token).price,
+        ) {
+            (Some(buy_price), Some(sell_price)) => {
+                let buy = buy_price.apply(self.buy.amount);
+                let sell = sell_price.apply(self.sell.amount);
+                conv::u256::to_big_rational(buy.0) / conv::u256::to_big_rational(sell.0)
```
MartinquaXD, line 187: This could panic if `sell_amount` or `sell_price` is 0. **Should** never happen but it would be easy enough to account for that here.

crates/driver/src/tests/cases/order_sorting.rs
```diff
@@ -0,0 +1,36 @@
...
+                .limit()
+                .reduce_amount(1000000000000000u128.into()),
+        )
+        .order(ab_order().rename("fourth order").limit())
+        .solution(ab_solution())
+        .done()
+        .await;
+
+    // Only check that the solve endpoint can be called successfully, which means
+    // that the solver received the orders sorted.
```
MartinquaXD, line 34: Not sure I get it. Can you point out where the code would throw an error if the orders were to be received in unsorted order?
Could be that I'm just not familiar enough with the test setup code.

crates/driver/src/tests/cases/order_sorting.rs
```diff
@@ -0,0 +1,36 @@
...
+    setup,
+    setup::{ab_order, ab_pool, ab_solution},
+};
+
+/// Test that orders are sorted correctly before being sent to the solver:
+/// market orders come before limit orders, and orders that are more likely to
+/// fulfill come before orders that are less likely (according to token prices
+/// in ETH).
+#[tokio::test]
+#[ignore]
```
MartinquaXD, line 11: Test is ignored. Is this intentional?

crates/driver/src/tests/cases/order_sorting.rs
MartinquaXD: I think I found it. All good. 👌

cowfee: The test setup code is confusing. It could be better...

crates/driver/src/tests/cases/order_sorting.rs
cowfee: All our E2E tests are ignored.

crates/driver/src/domain/competition/auction.rs
vkgnosis: You can implement `IntoIterator` for `&T`. In practice the difference between an `IntoIterator` implementation and an `iter` function is that the former can be used in for loop syntax `for a in b` without writing `b.iter()`. Also allows you to be generic over anything that can turn into an Iterator. In this case it doesn't matter much.

crates/driver/src/domain/competition/order/mod.rs
cowfee: This is difficult to do, because you actually want to ensure two things:
1. That the token information is the same for every order in the auction.
2. That the token information is present for every order in the auction.

I experimented with moving token info into the `Order` type and I wasn't happy with the result.

cowfee: https://github.com/cowprotocol/services/pull/1678/commits/32c5a55cdf0695e5ef9293fee9f4206d14755443

crates/driver/src/domain/competition/order/mod.rs
cowfee: https://github.com/cowprotocol/services/pull/1678/commits/32c5a55cdf0695e5ef9293fee9f4206d14755443

crates/driver/src/domain/competition/auction.rs
cowfee: https://github.com/cowprotocol/services/pull/1678/commits/32c5a55cdf0695e5ef9293fee9f4206d14755443

crates/driver/src/tests/setup/driver.rs
```diff
@@ -64,7 +65,11 @@ impl Driver {
 pub fn solve_req(test: &Test) -> serde_json::Value {
     let mut tokens_json = Vec::new();
     let mut orders_json = Vec::new();
-    for quote in test.quotes.iter() {
+    // The orders are shuffled before being sent to the driver, to ensure that the
+    // driver sorts them correctly before forwarding them to the solver.
+    let mut quotes = test.quoted_orders.clone();
+    quotes.shuffle(&mut rand::thread_rng());
```
vkgnosis, line 71: This makes tests less deterministic. Is there a bad better solution?

vkgnosis (approved):

sunce86:
Why are we sorting the orders?

cowfee:
> Why are we sorting the orders?

Because the in-house solvers (specifically the single order solver) solver orders one by one in the order they receive them. Another reason is because it's useful to give the guarantee to the solvers that those orders which come first will be more likely to be fulfillable than those that come later.

> And why is this relevant to colocation?

It happens in the driver.

PR #1686, opened 2023-07-17
Add logs to http request/response

sunce86:
Added more logs to observe the http communication with external solvers.

This is needed to check if the requests are properly crafted for the Raven colocated driver.

### DIFF
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -182,6 +182,13 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             let body = serde_json::to_vec(&model).unwrap();
             request = request.body(body);
         };
+        // temporary log, not needed once the code is stable for colocation
+        tracing::debug!(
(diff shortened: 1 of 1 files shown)


crates/shared/src/http_solver.rs
```diff
@@ -182,14 +182,21 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             let body = serde_json::to_vec(&model).unwrap();
             request = request.body(body);
         };
+        // temporary log, not needed once the code is stable for colocation
```
cowfee, line 185: I'd still leave it.

cowfee (approved):

sunce86:
Lets merge it with 1 approval so we can check the communication asap.

vkgnosis:
I know it's already merged but it would be better to keep this as a trace log and change the log filter configuration to display it. That's one of the main points of the tracing setup and trace logs, that we can selectively enable and disable logs without code changes. I'd like to revert this PR and do it through the config.

PR #1704, opened 2023-07-26
Add Route for Uploading AppData

nlordell:
This PR adds a route for uploading some `fullAppData` to the orderbook so that orders

This is done in a very simple way, by providing a `PUT api/v1/app_data/${hash}` route that can be used for uploading full app data documents.

Uploads verified that:
- The hashes match
- The app data is valid

Additionally, the `PUT` is (as specified by HTTP) idempotent.

### Test Plan

Adjusted existing database tests for new insertion semantics and added some assertions in app-data E2E test.


crates/orderbook/src/app_data.rs
```diff
@@ -0,0 +1,74 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum SaveError {
+    #[error("app data is invalid: {0}")]
+    Invalid(anyhow::Error),
+    #[error("app data already exists")]
+    Duplicate,
```
vkgnosis, line 54: Should this be an error?

vkgnosis (approved):

crates/orderbook/src/app_data.rs
nlordell: I thought the API was easier to deal with if it was.

The alternative would be to have a:

```rust
enum Saved {
    New,
    AlreadyExists,
}
```

And return a

```rust
Result<Saved, SaveError>
```

I originally went this way but the API was a bit clunkier. Happy to change it if you feel this is more correct.

vkgnosis: I have a weak preference for not erroring if the data is already there. The point of this route is to make this app data available in the database. If it is already there then that isn't an error to the consumer of this API. The consumer doesn't need to know whether the app data was freshly saved or was already there.

nlordell: Actually, it ended up looking "nicer" IMO. Thanks for the suggestion!

275b21082365c6d966dcb30ee80688e9f79acf86

PR #1705, opened 2023-07-26
Record Unused AppData Metrics 

nlordell:
Follow-up to #1704 .

### Test Plan

Run the `autopilot` and `orderbook` locally and:
```
% curl -s 'http://localhost:9589/metrics' | rg unused_app_data | rg -v '^#'
gp_v2_autopilot_unused_app_data 0
% curl -s 'http://localhost:8080/api/v1/app_data/0x586e9b1e1681ba3ebad5ff5e6f673d3e3aa129fcdb76f92083dbc386cdde4312' -X PUT --data-raw '{"hello":"world"}' > /dev/null
% sleep 60
% curl -s 'http://localhost:9589/metrics' | rg unused_app_data | rg -v '^#'
gp_v2_autopilot_unused_app_data 1
```

### DIFF
--- a/crates/autopilot/src/database.rs
+++ b/crates/autopilot/src/database.rs
@@ -24,13 +24,23 @@ impl Postgres {
         Ok(Self(PgPool::connect(url).await?))
     }
 
-    pub async fn update_table_rows_metric(&self) -> sqlx::Result<()> {
+    pub async fn update_database_metrics(&self) -> sqlx::Result<()> {
         let metrics = Metrics::get();
+
+        // update table row metrics
         for &table in database::ALL_TABLES {
             let mut ex = self.0.acquire().await?;
             let count = count_rows_in_table(&mut ex, table).await?;
             metrics.table_rows.with_label_values(&[table]).set(count);
         }
+
+        // update unused app data metric
+        {
+            let mut ex = self.0.acquire().await?;
+            let count = count_unused_app_data(&mut ex).await?;
+            metrics.unused_app_data.set(count);
+        }
+
         Ok(())
     }
 }
@@ -40,11 +50,31 @@ async fn count_rows_in_table(ex: &mut PgConnection, table: &str) -> sqlx::Result
     sqlx::query_scalar(&query).fetch_one(ex).await
 }
 
+async fn count_unused_app_data(ex: &mut PgConnection) -> sqlx::Result<i64> {
+    let query = r#"
+        SELECT
+            COUNT(*)
+        FROM app_data AS a
+        LEFT JOIN orders o
+            ON a.contract_app_data = o.app_data
+        WHERE
+            o.app_data IS NULL
+        ;
+    "#;
+    sqlx::query_scalar(query).fetch_one(ex).await
+}
+
 #[derive(prometheus_metric_storage::MetricStorage)]
 struct Metrics {
     /// Number of rows in db tables.
     #[metric(labels("table"))]
     table_rows: prometheus::IntGaugeVec,
+
+    /// Number of unused app data entries.
+    ///
+    /// These are entries in the `app_data` table that do not have a
+    /// corresponding order in the `orders` table.
+    unused_app_data: prometheus::IntGauge,
 
     /// Timing of db queries.
     #[metric(name = "autopilot_database_queries", labels("type"))]
@@ -59,7 +89,7 @@ impl Metrics {
 
 pub async fn database_metrics(db: Postgres) -> ! {
     loop {
-        if let Err(err) = db.update_table_rows_metric().await {
+        if let Err(err) = db.update_database_metrics().await {
             tracing::error!(?err, "failed to update table rows metric");
         }
         tokio::time::sleep(Duration::from_secs(60)).await;


crates/autopilot/src/database.rs
```diff
@@ -40,12 +50,32 @@ async fn count_rows_in_table(ex: &mut PgConnection, table: &str) -> sqlx::Result
     sqlx::query_scalar(&query).fetch_one(ex).await
 }
 
+async fn count_unused_app_data(ex: &mut PgConnection) -> sqlx::Result<i64> {
+    let query = r#"
+        SELECT
+            COUNT(*)
+        FROM app_data AS a
+        LEFT JOIN orders o
+            ON a.contract_app_data = o.app_data
```
vkgnosis, line 59: Is this going to work with LEFT JOIN or does it need LEFT OUTER JOIN?

vkgnosis (approved):

crates/autopilot/src/database.rs
```diff
@@ -24,13 +24,23 @@ impl Postgres {
...
+        // update table row metrics
         for &table in database::ALL_TABLES {
             let mut ex = self.0.acquire().await?;
             let count = count_rows_in_table(&mut ex, table).await?;
             metrics.table_rows.with_label_values(&[table]).set(count);
         }
+
+        // update unused app data metric
+        {
+            let mut ex = self.0.acquire().await?;
```
MartinquaXD, line 39: This pattern already existed before but wouldn't it make sense to only acquire a single connection and reuse it for all updates here?

crates/e2e/tests/e2e/app_data.rs
```diff
@@ -105,5 +105,60 @@ async fn app_data(web3: Web3) {
...
-    assert_eq!(order3_.metadata.full_app_data, Some(app_data.to_string()));
+    assert_eq!(order3_.metadata.full_app_data.as_deref(), Some(app_data));
+
+    // invalid app data
+    let invalid_app_data = r#"{"backend":"invalid"}"#;
+    let order4 = create_order(OrderCreationAppData::Full {
+        full: invalid_app_data.to_string(),
+    });
+    let err = services.create_order(&order4).await.unwrap_err();
+    dbg!(err);
```
MartinquaXD, line 116: left over `dbg!()` statement? (2 more in this test)

crates/orderbook/openapi.yml
```diff
@@ -402,6 +402,34 @@ paths:
             application/json: {}
         404:
           description: No full `appData` stored for this hash.
+    put:
+      summary: Registers a full `appData` so it can be referenced by `appDataHash`.
+      description: |
+        Uploads a full `appData` to orderbook so that orders created with the
```
MartinquaXD, line 408:
```suggestion
        Uploads a full `appData` to the orderbook so that orders created with the
```

crates/orderbook/src/app_data.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[error("app data is invalid: {0}")]
+    Invalid(anyhow::Error),
+    #[error("app data already exists")]
+    Duplicate,
+    #[error("computed app data hash {computed:?} doesn't match expected {expected:?}")]
+    HashMismatch {
+        expected: AppDataHash,
+        computed: AppDataHash,
+    },
+    #[error("stored app data {existing:?} is different than the specified data")]
```
MartinquaXD, line 60: Should these names use `camelCase` since this is also what gets used in the openapi spec?

crates/autopilot/src/database.rs
MartinquaXD: Since this doesn't start a transaction I think there should be no locks involved.
On the other hand it doesn't matter that much. Just looked weird to me.

vkgnosis: I think that was it.

crates/e2e/tests/e2e/app_data.rs
nlordell: I kept these in on purpose, as the rest of the test uses this pattern (and assumed it was for a reason - specifically to manually check the error contents when running the test with `--nocapture`).

crates/orderbook/src/app_data.rs
nlordell: Sorry, I didn't quite understand the suggestion - which name should be camel case?

crates/autopilot/src/database.rs
nlordell: They are equivalent AFAICT. From the docs (emphasis my own):

> The words `INNER` and `OUTER` are optional in all forms. `INNER` is the default; `LEFT`, `RIGHT`, and `FULL` **imply an outer join**.

Also, as far as I can tell by quickly Googling, `LEFT JOIN` is equivalent to `LEFT OUTER JOIN`.

crates/orderbook/src/app_data.rs
MartinquaXD: In the openapi spec it says `appData` instead of `app data` or `appDataHash` instead of `app data hash`.
Might make sense to unify the spelling across openapi spec and error message.

MartinquaXD (approved):

crates/orderbook/src/app_data.rs
nlordell: This change was part of the parent PR and just showed here because I didn't rebase this PR.

Addressed here:

PR #1706, opened 2023-07-26
Update openapi docs for priceQuality

sunce86:
Updates the `PriceQuality` enum.

Lately we introduced the variant `verified` which means that specific quote needs to be verified/simulated in order to be accepted as valid. 

Related to [slack](https://cowservices.slack.com/archives/C036G0J90BU/p1690299239980499) discussion.

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -508,6 +508,10 @@ components:
       description: |
         How good should the price estimate be?
 
+        Fast: The price estimate is chosen among the fastest N price estimates.
+        Optimal: The price estimate is chosen among all price estimates.
+        Verified: The price estimate is chosen among all verified/simulated price estimates.
+
         **NOTE**: Orders are supposed to be created from `verified` price estimates.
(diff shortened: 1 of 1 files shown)


crates/orderbook/openapi.yml
```diff
@@ -508,9 +508,9 @@ components:
       description: |
         How good should the price estimate be?
 
-        **NOTE**: Orders are supposed to be created from `optimal` price estimates.
+        **NOTE**: Orders are supposed to be created from `verified` price estimates.
       type: string
-      enum: [fast, optimal]
+      enum: [fast, optimal, verified]
```
vkgnosis, line 513: It would be nice if we explained what the values mean. Someone looking at these docs won't understand the difference. Not required for this PR.

vkgnosis (approved):

crates/orderbook/openapi.yml
sunce86: Good point. Added description.

PR #1707, opened 2023-07-26
Cache IPFS app data

vkgnosis:
I observed automated order creation through our API with an order that has no full app data. The order later fails validation. This puts pressure on our IPFS gateway because we have to check whether the app data exists every time. This PR relives some of this pressure by caching IPFS app data.

This PR contains several changes:
- Move IPFS app data fetching from the ipfs module to a dedicated ipfs_app_data module.
- Refactor IPFS code to not treat all reqwest errors as missing app data. Missing app data is only the result when the request and response fetching is successful with a status code that isn't 200.
- Add a cache to the ipfs_app_data module.

### Test Plan

existing tests and observe Pinata request counts later

### DIFF
--- a/crates/orderbook/src/ipfs.rs
+++ b/crates/orderbook/src/ipfs.rs
@@ -1,7 +1,7 @@
 use {
-    anyhow::{anyhow, Context, Result},
-    model::app_id::AppDataHash,
-    reqwest::{Client, StatusCode},
+    anyhow::{Context, Result},
+    reqwest::{Client, ClientBuilder, StatusCode},
+    std::time::Duration,
     url::Url,
 };
 
@@ -12,10 +12,10 @@ pub struct Ipfs {
 }
 
 impl Ipfs {
-    pub fn new(client: Client, base: Url, query: Option<String>) -> Self {
+    pub fn new(client: ClientBuilder, base: Url, query: Option<String>) -> Self {
         assert!(!base.cannot_be_a_base());
         Self {
-            client,
+            client: client.timeout(Duration::from_secs(5)).build().unwrap(),
             base,
             query,
         }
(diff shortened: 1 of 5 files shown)


crates/orderbook/src/ipfs.rs
```diff
@@ -12,10 +12,13 @@ pub struct Ipfs {
...
 impl Ipfs {
-    pub fn new(client: Client, base: Url, query: Option<String>) -> Self {
+    pub fn new(base: Url, query: Option<String>) -> Self {
         assert!(!base.cannot_be_a_base());
         Self {
-            client,
+            client: ClientBuilder::new()
+                .timeout(Duration::from_secs(5))
+                .build()
+                .unwrap(),
```
nlordell, line 21: This makes the HTTP client ignore "global" configuration such as `User-Agent`. The `HttpCLientFactory` has a configure method that you can use to override `ClientBuilder` configurations:

https://github.com/cowprotocol/services/blob/d04c14b87c33335da9ba70e1b7ef947b3dd13d26/crates/shared/src/http_client.rs#L36-L39

And an example of where this is used for overriding HTTP client timeouts:

https://github.com/cowprotocol/services/blob/d04c14b87c33335da9ba70e1b7ef947b3dd13d26/crates/shared/src/bad_token/token_owner_finder.rs#L245

crates/orderbook/src/ipfs_app_data.rs
```diff
@@ -0,0 +1,119 @@
...
+        let new = new_app_data_cid(contract_app_data);
+        let fetch = |cid: String| async move {
+            let result = self.ipfs.fetch(&cid).await;
+            let result = match result {
+                Ok(Some(result)) => {
+                    tracing::debug!(?contract_app_data, %cid, "found full app data");
+                    result
+                }
+                Ok(None) => {
+                    tracing::debug!(?contract_app_data, %cid,"no full app data");
```
nlordell, line 45:
```suggestion
                    tracing::debug!(?contract_app_data, %cid, "no full app data");
```

crates/orderbook/src/ipfs.rs
vkgnosis: This code needs a custom time out which is why I intentionally did not use the global factory. I forgot about other things like user agent though. Will switch it back to the factory.

crates/orderbook/src/ipfs_app_data.rs
```diff
@@ -0,0 +1,119 @@
...
+    ipfs: Ipfs,
+    cache: Mutex<TimedSizedCache<AppDataHash, Option<String>>>,
+}
+
+impl IpfsAppData {
+    pub fn new(ipfs: Ipfs) -> Self {
+        Self {
+            ipfs,
+            cache: Mutex::new(TimedSizedCache::with_size_and_lifespan_and_refresh(
+                1000, 600, false,
```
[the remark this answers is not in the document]
vkgnosis: This is an argument for putting every parameter to every function call into a variable that describes the parameter. That's unreasonable. What the values do is documented by the the function. I don't have a problem with doing it that way in this instance but I'm not sure that this argument makes sense.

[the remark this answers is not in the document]
nlordell: > This is an argument for putting every parameter to every function call into a variable that describes the parameter.

I think in most cases, arguments can be inferred by context. For example, `order.with_sell_amount(10000)` - it is immediately obvious that `10000` is a "sell amount" and there is no need to first create a binding to describe it.

I also agree that as a general rule we shouldn't do it. The reason I mentioned it here is that, in this case, it stood out to me as "what do these values mean".

As I write this, however, I realize that the function name here provides enough context:

```rust
with_size_and_lifespan_and_refresh(
  1000 /* size */, 600 /* lifetime */, false /* refresh */,
)
```

So I retract my suggestion.

PR #1710, opened 2023-07-27
Move hooks to metadata field

nlordell:
As discussed in [Slack](https://cowservices.slack.com/archives/C0375NV72SC/p1690365762570749) this PR changes the backend to use the `metadata` field for specifying hooks and deprecates the `backend`.

For simplicity's sake, we decided to ignore the `backend` field entirely if the `metadata` field is specified. While this is technically breaking behaviour, AFAICT, all uses of "hooks" have been without `metadata` being specified at all (following the [documentation](https://docs.cow.fi/overview/cow-hooks/cow-hooks-example)).

### Test Plan

Added comprehensive unit tests. Adjusted E2E tests to use "metadata" field.

### DIFF
--- a/crates/orderbook/src/orderbook.rs
+++ b/crates/orderbook/src/orderbook.rs
@@ -24,6 +24,7 @@ use {
     },
     primitive_types::H160,
     shared::{
+        app_data,
         metrics::LivenessChecking,
         order_validation::{OrderValidating, ValidationError},
     },
@@ -209,6 +210,11 @@ impl Orderbook {
         &self,
         contract_app_data: &AppDataHash,
     ) -> Result<Option<String>> {
+        // we reserve the 0 app data to indicate empty app data.
+        if contract_app_data.is_zero() {
+            return Ok(Some(app_data::EMPTY.to_string()));
+        }
+
         if let Some(app_data) = self
             .database
             .get_full_app_data(contract_app_data)
--- a/crates/shared/src/app_data.rs
+++ b/crates/shared/src/app_data.rs
@@ -2,18 +2,22 @@ use {
     anyhow::{anyhow, Context, Result},
     model::{app_id::AppDataHash, order::Hooks},
     serde::Deserialize,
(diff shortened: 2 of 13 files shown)


crates/orderbook/src/orderbook.rs
```diff
@@ -206,6 +207,11 @@ impl Orderbook {
     ///
     /// The full app data can be located in the database or on IPFS.
     pub async fn find_full_app_data(&self, contract_app_data: &AppDataHash) -> Option<String> {
+        // we reserve the 0 app data to indicate empty app data.
+        if contract_app_data.is_zero() {
+            return Some(app_data::EMPTY.to_string());
+        }
```
vkgnosis, line 213: Is not needed because I added database entries for all zero app data mapping to `{}`. On the other hand if you start with a fresh database like in an e2e test that would stop working.

vkgnosis (approved):

crates/shared/src/app_data.rs
```diff
@@ -49,33 +53,259 @@ impl Validator {
...
+        );
+
+        // Note that if `metadata` is specified, then the `backend` field is
+        // ignored.
+        assert_app_data!(
+            r#"
+                {
+                    "metadata": {},
+                    "backend": {
+                        "hooks": {
```
alfetopito, line 290: Could this become a problem for the short term while this is live in the UI?
Would it be easy to do a db query to check for existing orders with `metadata` and `backend` keys?

alfetopito (approved):
Looks good AFAICT, just 2 comments

crates/orderbook/src/orderbook.rs
nlordell: Ah ok - makes sense. This is is why I ran into issues locally.

I would prefer to keep this as explicit behaviour instead of relying on the DB being set up (unless it became part of the migration process).

anxolin (commented):
Thanks a lot for making the change 🙏

PR #1712, opened 2023-07-28
Add Raven as legacy http price estimator

harisang:
This PR adds Raven as a legacy http price estimator, so that it can be enabled in production. This should be viewed as a temporary solution until we move to the colocated driver solution.

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -219,7 +219,7 @@ pub struct Arguments {
     /// The API endpoint to call the Raven solver for price estimation
     #[clap(long, env)]
     pub raven_solver_url: Option<Url>,
-   
+
     /// The API path to use for solving.
     #[clap(long, env, default_value = "solve")]
     pub raven_solver_path: String,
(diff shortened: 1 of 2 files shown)


crates/shared/src/price_estimation.rs
```diff
@@ -214,6 +216,14 @@ pub struct Arguments {
     #[clap(long, env, default_value = "solve")]
     pub yearn_solver_path: String,
 
+    /// The API endpoint to call the Raven solver for price estimation
+    #[clap(long, env)]
+    pub raven_solver_url: Option<Url>,
+   
+    /// The API path to use for solving.
+    #[clap(long, env, default_value = "solve")]
+    pub raven_solver_path: String,
```
harisang, line 225: Tbh, I am not sure if this is needed,  and whether additional changes are needed because of that, as the solving part is already taken care of somewhere else (as Raven already participates in price estimation).

sunce86: In general yes, let's keep it

sunce86 (approved):

vkgnosis (approved):

PR #1713, opened 2023-07-31
Decrease log level to trace

MartinquaXD:
Reduces noisy logs to `trace` so they can easily be turned off when they are no longer need (suggested [here](https://github.com/cowprotocol/services/pull/1686#issuecomment-1645794201)).

### Test Plan
CI


vkgnosis (approved):

sunce86:
Was this the reason for increased log sizes Anuj mentioned?

MartinquaXD:
> Was this the reason for increased log sizes Anuj mentioned?

These logs made up ~8% of the total number of logs and they tended to be way longer than the average log message so I think it's safe to assume that these were the culprit.

PR #1715, opened 2023-07-31
Add Support For Arbitrary Legacy Solver Estimators

nlordell:
This PR adds support for adding arbitrary solvers as legacy HTTP solvers (a follow up to #1712 which allows more solvers to be added via configuration instead of requiring code changes).

This PR introduces a new `--price-estimation-legacy-solver` flag for configuring these solvers. These are meant to be used as an "intermediate" step while we get more solvers onboarded to CIP-27. It is different than setting up a co-located driver + legacy solver adapter in that:
- It respects the reported `feeEstimate` 
- Does not do simulation verification

This is meant for on-boarding solvers to be able to start participating without generating calldata for an interim period.

Note that this change slightly changes the `PriceEstimatorFactory` semantics. Specifically, price etimator entries are unique by `name: String` and not longer split between `PriceEstimatorKind` and `name: String` as before (depending on the "source" of the price estimator).

### Test Plan

Ran locally connected to the Raven solver:

```
% cargo run -p orderbook -- --baseline-sources UniswapV2 --price-estimators None --native-price-estimators Baseline --price-estimation-legacy-solvers 'raven|${RAVEN_URL}/solve'
...
2023-07-31T13:34:23.486Z DEBUG request{id=0}: shared::price_estimation::competition: new price estimate query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 100000000000000000000, kind: Sell, verification: None } result=Ok(Estimate { out_amount: 1, gas: 447759, solver: 0x0000000000000000000000000000000000000000 }) estimator="raven"
...
```

Also added some parsing unit tests.

### Release notes

Once this PR merges, Quasimodo, Yearn and Raven solvers should be configured with this new method instead of having special "built-in" solvers for them.

### DIFF
--- a/crates/autopilot/src/lib.rs
+++ b/crates/autopilot/src/lib.rs
@@ -1,14 +1,3 @@
-pub mod arguments;
-pub mod database;
-pub mod decoded_settlement;
-pub mod driver_api;
-pub mod driver_model;
-pub mod event_updater;
-pub mod fok_limit_orders;
-pub mod on_settlement_event_updater;
-pub mod run_loop;
-pub mod solvable_orders;
-
 use {
     crate::{
         database::{
@@ -57,7 +46,7 @@ use {
         metrics::LivenessChecking,
         oneinch_api::OneInchClientImpl,
         order_quoting::OrderQuoter,
-        price_estimation::factory::{self, PriceEstimatorFactory},
+        price_estimation::factory::{self, PriceEstimatorFactory, PriceEstimatorSource},
         recent_block_cache::CacheConfig,
         signature_validator::Web3SignatureValidator,
         sources::{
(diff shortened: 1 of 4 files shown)


crates/autopilot/src/lib.rs
```diff
@@ -1,14 +1,3 @@
...
-pub mod decoded_settlement;
-pub mod driver_api;
-pub mod driver_model;
-pub mod event_updater;
-pub mod fok_limit_orders;
-pub mod on_settlement_event_updater;
-pub mod run_loop;
-pub mod solvable_orders;
-
 use {
```
vkgnosis, line 1: To me it makes more sense to keep module declarations above use imports.

vkgnosis (approved):

crates/autopilot/src/lib.rs
nlordell: Me too... but `rust-analyzer` inserts `use statements; mod statements; code`, so I just fixed the order to LSP adding of use imports works as expected.

Otherwise, you end up with:

```rust
use new_path;

mod existing_mod;

use existing_path;
```

nlordell: I'm OK with either or, but it is inconsistent in the repo. I decided to just pick what my LSP server prefers 🤷.

vkgnosis: I wonder if this a rust analyzer setting. It would be better to change that if it exists.

sunce86 (approved):
Beautiful.

PR #1722, opened 2023-08-01
Notify about tx hash instead of block hash

MartinquaXD:
For some reason the notification about a successfully mined solution contained a block hash which is fairly useless.
This PR notifies about the tx hash instead which can be easily found on etherscan.

### Test Plan
compiler


sunce86 (approved):
O.o

PR #1723, opened 2023-08-02
Fix soft cancellations support

MartinquaXD:
Follow up fix to https://github.com/cowprotocol/services/pull/1544 (please read that one for more context)

That PR only cleared `transactions` after a soft cancellations because I thought that `submitted_transactions` would always be overwritten with `transactions` afterwards.
However, this actually only happens when `transactions` is not empty which means the previous PR didn't have an effect.

Now we clear `transactions` to avoid waiting for any tx that will never be mined and we forcefully update `submitted_transactions` to actually ensure that we will not wait for the gas price to increase during the next submission of the same solver.

This PR also adds a log that states whenever we are waiting for the gas price to increase since this would have been useful debugging information a couple of times.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -497,6 +497,12 @@ impl<'a> Submitter<'a> {
                 if gas_price.max_priority_fee_per_gas < replacement_price.max_priority_fee_per_gas
                     || gas_price.max_fee_per_gas < replacement_price.max_fee_per_gas
                 {
+                    tracing::debug!(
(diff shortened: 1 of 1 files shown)


MartinquaXD:
Not sure how to test this correctly so @sunce86 could you give this and the previous PR a critical look since your are the most familiar with the submission logic and I would like to avoid merging another half baked PR.

vkgnosis (approved):

sunce86 (approved):
It's been a while since I looked into this code but I _think_ this code change does what is intended with this PR.

PR #1727, opened 2023-08-03
Log orders in auction

vkgnosis:
While debugging slow order execution times I noticed that there is no way to tell when an order is in the current auction. We log orders that get filtered out from the auction but not when an order shows up. In the colocated code this is solved through order observation events but we don't have these for non colocated code. This PR fixes that by logging how the auction that the driver handles changes. This is smarter than logging the full auction because it only logs the difference.

I made this a debug log because we usually want to know about this in hindsight and it isn't too spammy compared to other driver logs.

### Test Plan

Check logs after deploying

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -1,13 +1,4 @@
 use {
-    crate::settlement_rater::SettlementRating,
-    shared::http_solver::model::{AuctionResult, SubmissionResult},
-    tracing::info_span,
-};
-
-pub mod gas;
-pub mod solver_settlements;
-
-use {
     crate::{
         auction_preprocessing,
         driver_logger::DriverLogger,
@@ -18,6 +9,7 @@ use {
         orderbook::OrderBookApi,
         settlement::{PriceCheckTokens, Settlement},
         settlement_ranker::SettlementRanker,
+        settlement_rater::SettlementRating,
         settlement_simulation,
         settlement_submission::{SolutionSubmitter, SubmissionError},
         solver::{Auction, Solver, Solvers},
@@ -29,7 +21,7 @@ use {
     gas_estimation::GasPriceEstimating,
     model::{
         auction::{AuctionId, AuctionWithId},
-        order::OrderUid,
+        order::{Order, OrderUid},
         solver_competition::{
             self,
             CompetitionAuction,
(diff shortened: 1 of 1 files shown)


crates/solver/src/driver.rs
```diff
@@ -233,6 +235,17 @@ impl Driver {
         Ok(())
     }
 
+    fn observe_auction_orders(&mut self, orders: &[Order]) {
+        let orders: HashSet<OrderUid> = orders.iter().map(|order| order.metadata.uid).collect();
+        for order in orders.difference(&self.previous_auction_orders) {
+            tracing::debug!(%order, "new order in auction");
```
MartinquaXD, line 241: nit: The same order may drop in and out of auctions so `new order` might be slightly misleading.
```suggestion
            tracing::debug!(%order, "order now in auction");
```

MartinquaXD: As a side note: did you consider logging only a single message containing all ids? It seems like having 1 log per order might be a bit more noisy than needed (especially during start ups where all orders in the auction will be printed).

MartinquaXD (commented):
Seems like a nice and simple way to only log the relevant information. 👍 
Overall I'm happy with the PR. Just not approving because I see you already enabled auto-merge and I wanted to bring up these 2 things before merging.

crates/solver/src/driver.rs
vkgnosis: Disabled auto merge while discussing.

I'll change the log message to be more clear about "new".

>  logging only a single message

Either way the amount of information is the same but the number of log messages changes. Doing it this way felt more natural to me because the log message is about individual orders and not about all of them as a set. I care about seeing that some specific order is in the auction. I don't care about seeing that these orders showed up together.

MartinquaXD: > But if you're looking at all the logs of an auction then maybe logging once is nicer.

Yeah, I was thinking about that. However, this is probably not so big of an issue so I'm also fine with having 1 log per order.

MartinquaXD (approved):

crates/solver/src/driver.rs
vkgnosis: I decided I like merged logs more.

PR #1729, opened 2023-08-03
Proper default log level

MartinquaXD:
Currently pods running the new `solvers` are logging lots of debug information of dependencies.
The new default log level will only log warnings and higher of dependencies and debug and higher of any code inside our repo.

### Test Plan
Manually ran solver to see that logs are still being printed


nlordell (approved):

nlordell:
Thanks!

PR #1730, opened 2023-08-03
Adapt Gnosis solvers score reporting to capped rewards

harisang:
This PR optimizes the score reporting for Gnosis solvers according to the capped rewards scheme that is currently implemented. So far, Gnosis solvers have reported their score according to the pure, uncapped version of CIP-20. This has resulted in the solvers often discounting their solutions more than needed, since the rewards/penalties end up being capped.

To understand the logic behind this change, we briefly sketch how we reached the implemented formulas. We note that all quantities used here are denominated in ETH, and we also note that in the code we have replaced the currently hardcoded cap of 0.01 ETH of CIP-20 with a parameter `cap` (in the analysis below, we still use this 0.01 in order to make the analysis slightly easier to read).

According to CIP-20, a solver that won a batch auction gets paid:

- in case of a successful settlement: `payment = min{surplus + fees - reference score, gas + 0.01}`
- in case of revert: 'payment = - min{reference score, 0.01}

Since the goal of a solver is to maximize their (expected) profit, if p denotes an estimated probability of success, we have

`E[profit] = p * (min{surplus + fees - reference score, gas + 0.01} - gas) - (1-p) * min{reference score, 0.01}`

Let `obj := surplus + fees - gas`. Then, we can write

`E[profit] = p * min{obj - reference score, 0.01} - (1-p) * min{reference score, 0.01}`.

Note that since in order to be the winner one needs to have the highest score, we have that `score >= reference score`, which implies that

`E[profit] >=  p * min{obj - score, 0.01} - (1-p) * min{score, 0.01}`.

We now want to ensure that `E[profit] >=  0`, which is equivalent to

`min{obj - score, 0.01} >= min{score, 0.01} * (1-p)/p`.     (1)

For that, we do some case analysis for the best choice of `score`, depending on the value of `obj` and `p`.

- Case where `p * obj <= 0.01` AND `(1-p) * obj <= 0.01`. In this case, we must have `score <= p * obj`. We will show this by a contradiction. Suppose otherwise; i.e., suppose that we report a score that satisfies`score > p * obj`. Then, we have `obj - score < (1-p) * obj <= 0.01` and so the left-hand side of (1) is smaller than `(1-p) * obj`. On the other hand, the right-hand side of (1) is at least `min{p * obj, 0.01} * (1-p)/p = min{obj,  0.01 /p} * (1-p) = (1-p) * obj`, where the last equality holds since we have assumed that `p * obj <= 0.01` which is equivalent to `obj <= 0.01/p`. Putting everything together we get that the left-hand side of (1) is strictly smaller than `(1-p) * obj` while the right-hand side of (1) is equal to exactly `(1-p) * obj`, and so (1) is violated. We conclude that we must have `score <= p * obj` in order not to end up with a negative profit (in the worst case), and since the actual payment only depends on the reference score and not the score reported by the winner, we should report a score as large as possible. Thus, in this case, we set `score = p * obj`. We can also verify that this is indeed a valid choice, as the left-hand side becomes equal to `(1-p) * obj` and the right-hand side becomes equal to `p * obj * (1-p)/p =(1-p) * obj `, i.e., under the assumptions made, (1) is satisfied with equality.

- The remaining cases to consider are the following:

    1. `p * obj > 0.01` AND `(1-p) * obj <= 0.01`
    2. `p * obj > 0.01` AND `(1-p) * obj > 0.01`
    3. `p * obj <= 0.01` AND `(1-p) * obj > 0.01`

    We will decompose these cases in a slightly different way, as follows:  
    
    1a. `p * obj > 0.01` AND `(1-p) * obj <= 0.01` AND `p >= 0.5` 
    1b. `p * obj > 0.01` AND `(1-p) * obj <= 0.01` AND `p < 0.5`
    2a. `p * obj > 0.01` AND `(1-p) * obj > 0.01` AND `p >= 0.5`.
    2b. `p * obj > 0.01` AND `(1-p) * obj > 0.01` AND `p< 0.5`.
    3a. `p * obj <= 0.01` AND `(1-p) * obj > 0.01` AND `p >= 0.5` .
    3b. `p * obj <= 0.01` AND `(1-p) * obj > 0.01` AND `p < 0.5`  
    
    Note that case (1b) cannot happen as `(1-p) * obj > p * obj` if `p < 0.5`, and similarly case (3a) cannot happen. By merging case (1a) with (2a), and, separately, case (2b) with (3b), we conclude that we only need to consider the following two cases.

- Case where `p * obj > 0.01` AND `p >= 0.5`. In this case, we must have `score <= obj - 0.01 * (1-p)/p`. Again, we will use contradiction. Suppose otherwise, i.e., that we report a score that satisfies `score > obj - 0.01 * (1-p)/p`. Then, the left hand-side is strictly smaller than  `0.01 * (1-p)/p` while the right-hand side is equal to `0.01 * (1-p)/p`. To see that the right-hand side is equal to that, note that we have assumed that `obj > 0.01/p` which means that `score > 0.01/p - 0.01 * (1-p)/p = 0.01`. Thus, we again get a violation of (1), which means that we must have `score <= obj - 0.01 * (1-p)/p`. With a similar reasoning as before, we set `score = obj - 0.01 * (1-p)/p`. We now verify that (1) is indeed satisfied. The left-hand side in this case is `0.01 * (1-p)/p` and the right-hand side is (as computed before) `(0.01/p - 0.01 * (1-p)/p) * (1-p)/p  = 0.01 * (1-p)/p`, and so, again (1) is satisfied with equality.

- Case where `(1-p) * obj > 0.01` AND `p < 0.5`. In this case, we must have `score <= 0.01 * p/(1-p)`. To see this, we note that if the right-hand is equal to `0.01 * (1-p)/p`, then (1) cannot be satisfied, since `(1-p)/p > 1` and the left-hand side is always <= 0.01. Thus we must have score <= 0.01, and we can actually strengthen this and ask that `score * (1-p)/p <= 1`, which is equivalent to `score <= 0.01 * p/(1-p)`. Thus, we set `score = 0.01 * p/(1-p)`, and we now verify that this is indeed a valid choice. The left-hand side of (1) is now equal to 0.01, since we have `obj - score > 0.01/(1-p) - 0.01 * p/(1-p) = 0.01`, while the right-hand side is equal to 0.01 by construction, which means that (1) is satisfied with equality.


fhenneke (approved):
I checked the logic behind the new reporting and it looks correct. Since it is a bit involved we should create a document explaining it somewhere.
The increase of the score in the test is expected.

I cannot really comment on quality of the code itself.

crates/solver/src/solver/score_computation.rs
```diff
@@ -53,12 +53,12 @@ impl ScoreCalculator {
         let success_probability = 1. / (1. + exponent.exp());
         let obj = surplus + fees - gas_amount * gas_price;
         let score: f64;
-        if success_probability * obj <= 0.01 && (1.0 - success_probability) * obj <= 0.01 {
+        if success_probability * obj <= 10000000000000000 && (1.0 - success_probability) * obj <= 10000000000000000 {
```
fhenneke, line 56: potentially pull the cap out as a variable.

crates/solver/src/solver/score_computation.rs
```diff
@@ -51,7 +51,16 @@ impl ScoreCalculator {
             - self.gas_price_factor * gas_price / 10_000_000_000.
             - self.nmb_orders_factor * nmb_orders as f64;
         let success_probability = 1. / (1. + exponent.exp());
-        let score = success_probability * (surplus + fees - gas_amount * gas_price);
+        let obj = surplus + fees - gas_amount * gas_price;
+        let score: f64;
+        let cap = 0.01;
```
fhenneke, line 56: Are you sure the scaling is correct? `gas_amount` and `gas_price` seem to be large numbers which suggests that wei are being used as unit. The cap should use the same scaling then.

harisang: Pushed a fix for that as I think you are right.

crates/solver/src/solver/score_computation.rs
```diff
@@ -51,7 +51,16 @@ impl ScoreCalculator {
...
+        let obj = surplus + fees - gas_amount * gas_price;
+        let score: f64;
+        let cap = 10000000000000000.0;
+        if success_probability * obj <= cap && (1.0 - success_probability) * obj <= cap {
+            score = success_probability * obj;
+        } else if success_probability * obj > cap && success_probability >= 0.5 {
+            score = obj - cap * (1.0 - success_probability) / success_probability;
+        } else {
+            score = cap * success_probability / (1.0 - success_probability);
+        }
```
nlordell, line 63: A slightly more "Rust-y" way to write this is:
```suggestion
        let cap = 10000000000000000.0;
        let score = if success_probability * obj <= cap && (1.0 - success_probability) * obj <= cap {
            success_probability * obj
        } else if success_probability * obj > cap && success_probability >= 0.5 {
            obj - cap * (1.0 - success_probability) / success_probability
        } else {
            cap * success_probability / (1.0 - success_probability)
        };
```

crates/solver/src/solver/score_computation.rs
```diff
@@ -51,7 +51,16 @@ impl ScoreCalculator {
             - self.gas_price_factor * gas_price / 10_000_000_000.
             - self.nmb_orders_factor * nmb_orders as f64;
         let success_probability = 1. / (1. + exponent.exp());
-        let score = success_probability * (surplus + fees - gas_amount * gas_price);
+        let obj = surplus + fees - gas_amount * gas_price;
+        let cap = 10000000000000000.0;
+        let score = if success_probability * obj <= cap && (1.0 - success_probability) * obj <= cap
```
MartinquaXD, line 56: I'm not able to verify that this logic makes sense without an explanation. 😅 
A minor improvement to readability could be to add
```rust
let revert_probability = 1.0 - success_probability;
```
and use that inside the conditions and calculations.

MartinquaXD (commented):
Would like to delay approval until you post the explanation behind the logic.

harisang:
> Would like to delay approval until you post the explanation behind the logic.

I just updated the original description of the PR with an explanation on how to derive the formulas. Will bother @fhenneke once again to briefly check whether the description makes sense or not.

MartinquaXD (approved):

crates/solver/src/solver/score_computation.rs
```diff
@@ -51,7 +51,16 @@ impl ScoreCalculator {
             - self.gas_price_factor * gas_price / 10_000_000_000.
             - self.nmb_orders_factor * nmb_orders as f64;
         let success_probability = 1. / (1. + exponent.exp());
-        let score = success_probability * (surplus + fees - gas_amount * gas_price);
+        let obj = surplus + fees - gas_amount * gas_price;
+        let cap = 10000000000000000.0;
```
sunce86, line 55: Do you see `cap` changing it's value any time soon? If so, it might be worth reading it from configuration.

harisang: It's not clear, as we are currently considering different options. Tbh, and given that this fix was long overdue (as Gnosis solvers underbidding has affected certain auctions), I would keep it as is for now.

sunce86 (approved):

fhenneke:
I checked the derivation and it looks correct.
I agree that we should communicate our thoughts on optimal bidding strategies with solvers. I myself would not have been able to come up with the clean three cases used in the code. (I arrived at the same strategy, with a different line of argument and 5 instead of 3 cases. After seeing Haris' solution, I was able to see that those 5 cases can be reduced to 3.)

harisang:
As a final comment, it would be good to zoom out a bit from the formulas and note the following:

- The last case `((1-p) * obj > 0.01 AND p < 0.5` is actually very uncommon since it asks for `p < 0.5`. I note this as the final score reporting might look strange in that last case, as it does not contain the obj at all and is only determined by `p`. Of course, the `obj` must satisfy `obj > 0.01 / (1-p)`, but still the final score reporting might look surprising. This happens because the cap seriously affects this case.
- The most common cases are the first two. In the first case, the score reporting aligns with the optimal strategy for the uncapped version of the mechanism, while in the second case, solvers are still incentivized to maximize the obj but also incorporate some risk considerations, which both are desirable properties.

Issue #1603, opened 2023-06-19
Required Solver Competion Fields


nlordell:
Currently, the `solver_competitions` table is intended for debugging purposes and not meant to provide a stable API for data consumption. However, our analytics makes use of the following keys for the **winning solution**:

- solver address
- calldata (while this can be retrieved from the blockchain, it would require an archive node, so we should store it in our DB - we already fetch it for the CIP-20 solver rewards payout).
- uninternalized calldata
- simulation block

As such, we need to provide a stable API for these fields (potentially create a new table for storing this).

We should also add a check for this to the colocation e2e test in `crates/e2e/tests/e2e/colocation_univ2.rs`.

cowfee:
[Related slack thread](https://cowservices.slack.com/archives/C0375NV72SC/p1687192847302879).

PR #1731, opened 2023-08-03
Save settlement call_data to db

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1603

Simulation block goes into `settlement_scores` table where also winner is saved.
Calldata is saved into separate table `settlement_call_data`.

The main concern is that the colocated driver now responds with calldata as part of the `\solve` response. We already discussed this few times, and the last discussion is [here](https://cowservices.slack.com/archives/C049P351GKF/p1689757539644619) (spoiler, no decision was made). To unblock colocation, it seems the easiest is to just add it to `\solve`.

Long term, my personal preference is to introduce additional `\reveal` request between `\solve` and `\settle` as it is more flexible and future proof. However, practically, we don't have to decide this until we push solvers into running their own drivers when we should have a stable driver api defined.

TODO RELEASE: update solver rewards script after colocation is enabled (to use these fields instead of solver competition table)


crates/database/src/settlement_call_data.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+pub struct SettlementCallData {
+    pub auction_id: AuctionId,
+    pub call_data: Vec<u8>,
+    pub uninternalized_call_data: Vec<u8>,
+}
+
+pub async fn insert(
+    ex: &mut PgTransaction<'_>,
```
MartinquaXD, line 15: This should allow you to call this function with an connection returned from `db.acquire()` or a transaction returned from `db.begin()`.

```suggestion
    ex: &mut PgConnection,
```

database/sql/V055__create_settlement_call_data.sql
```diff
@@ -0,0 +1,10 @@
+-- Add non-optional `simulation_block` field to settlement_scores table. 
+ALTER TABLE settlement_scores ADD COLUMN simulation_block bigint NOT NULL DEFAULT 0;
+ALTER TABLE settlement_scores ALTER COLUMN simulation_block DROP DEFAULT;
+
+-- Winning settlement call data for each auction.
+CREATE TABLE settlement_call_data (
```
MartinquaXD, line 6: Please don't forget to update the [`database/README.md`](https://github.com/cowprotocol/services/blob/main/database/README.md) when altering the DB schema.

crates/autopilot/src/run_loop.rs
```diff
@@ -85,6 +85,7 @@ impl RunLoop {
         // Shuffle so that sorting randomly splits ties.
         solutions.shuffle(&mut rand::thread_rng());
         solutions.sort_unstable_by_key(|solution| solution.1.score);
+        let competition_simulation_block = self.current_block.borrow().number;
```
MartinquaXD, line 88: This can be inaccurate.
I think at the moment our logic simply simulates using the `latest` block which may be different across these simulations. If we want to get an accurate reading we probably need to use some helper solidity code to return the used block number alongside the simulation result using state overrides.
That way we would know exactly which block number was used for each solution.

sunce86: At the moment we don't do any simulations in the colocated runloop. Maybe we could implement that when we add simulations (if we add them ever).
For now, this line just mimics what the non-colocated [driver is doing right now](https://github.com/cowprotocol/services/blob/main/crates/solver/src/driver.rs#L359-L361).

crates/database/src/settlement_call_data.rs
sunce86: Fixed

database/sql/V055__create_settlement_call_data.sql
sunce86: Ah yees. Added.

crates/autopilot/src/run_loop.rs
MartinquaXD: Ahh, true. Do you know what script or so needs the simulation block number and whether this worse than before approximation is still good enough?
If we can get by without this block number maybe it's worth it. This might require changes in other software so I think it's fine to leave this in this PR.

sunce86: [Here](https://github.com/cowprotocol/solver-rewards/blob/b3783694e36f66b266ec34f7c05b0f3e1db0d7f7/internal_transfers/actions/src/accounting.ts#L66) it is.  To me, it seems like it will do the job.

sunce86: Ofc, will need to update the solver rewards script before colocation is enabled.

crates/autopilot/src/database/competition.rs
```diff
@@ -42,6 +43,11 @@ pub struct Competition {
     /// chain before this block height.
     pub block_deadline: u64,
     pub order_executions: Vec<OrderExecution>,
+    pub competition_simulation_block: u64,
+    /// Winner settlement call data
+    pub call_data: Vec<u8>,
+    /// Uninternalized winner settlement call data
+    pub uninternalized_call_data: Vec<u8>,
```
MartinquaXD, line 50: Could probably use [ByteArray](https://github.com/cowprotocol/services/pull/1051/files) to get nicer Debug printing for the calldata bytes.

crates/database/src/settlement_call_data.rs
```diff
@@ -0,0 +1,53 @@
+use {crate::auction::AuctionId, sqlx::PgConnection};
+
+#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+pub struct SettlementCallData {
+    pub auction_id: AuctionId,
+    pub call_data: Vec<u8>,
+    pub uninternalized_call_data: Vec<u8>,
```
MartinquaXD, line 7: I think you should be able to use our [ByteArray](https://github.com/cowprotocol/services/pull/1051/files) here to get nicer debug printing of the calldata bytes.

crates/autopilot/src/database/competition.rs
```diff
@@ -117,6 +127,17 @@ impl super::Postgres {
...
+        database::settlement_call_data::insert(
+            &mut ex,
+            SettlementCallData {
+                auction_id: competition.auction_id,
+                call_data: competition.call_data.clone(),
+                uninternalized_call_data: competition.uninternalized_call_data.clone(),
+            },
+        )
+        .await
+        .context("settlement_call_data::insert")?;
```
MartinquaXD, line 139: Could you add another verification step [here](https://github.com/cowprotocol/services/blob/main/crates/e2e/tests/e2e/colocation_univ2.rs#L101-L122) to e2e test this?

crates/autopilot/src/database/competition.rs
sunce86: But ByteArray requires size to be known at compile time?

MartinquaXD: Ahh, right. Alternatively you should be able to use [this](https://github.com/cowprotocol/services/pull/836/files).

sunce86: Nice, added.

crates/database/src/settlement_call_data.rs
sunce86: Probably no need to add it here also (also `database` crate doesn't have `derivative` dependency).

MartinquaXD (approved):

PR #1733, opened 2023-08-03
Fix Co-located Native Token Price Errors

nlordell:
In debugging native token price requests in the co-located driver, I noticed two small issues:

1. We are overflowing because of too high sell amounts for the synthetic auction we create for buy quotes. 
2. We are very picky with rounding errors (good! means no unexpected behaviour), but don't account for their possibility in the Baseline solver implementation.

This PR fixes these two issues by:
1. Reducing the synthetic buy amount to be `2 ** 144`. The rationale being that `uint112` is the reserve size for Uniswap pools, `256 - 112 == 144` so it is would allow trading a full `uint112::MAX` without overflowing, and trading more `uint112` is unlikely and would not work with Uniswap anyway.
2. Fixing rounding errors on output buy amounts for buy orders with the Baseline solver.

### Test Plan

Added a `solvers` crate E2E test that demonstrates the rounding error.


vkgnosis (approved):

crates/driver/src/domain/quote.rs
```diff
@@ -179,7 +179,7 @@ impl Order {
             // contract, so buy orders requiring excessively large sell amounts
             // would not work anyway.
             order::Side::Buy => eth::Asset {
-                amount: (eth::U256::one() << 192).into(),
+                amount: (eth::U256::one() << 144).into(),
```
MartinquaXD, line 182: Is there a reason to use `144` specifically like there was to use `192`. If so it would be nice to link that here.

MartinquaXD (approved):

crates/driver/src/domain/quote.rs
MartinquaXD: Ah, sorry. This was the second time I read this PR but this time only focused on the code. 🤦‍♂️

crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -72,10 +72,21 @@ impl Baseline {
...
+                        // The baseline solver generates a path with swapping
+                        // for exact output token amounts. This leads to
+                        // potential rounding errors for buy orders, where we
+                        // can buy slightly more than intended. Fix this by
+                        // capping the output amount to the order's buy amount
+                        // for buy orders.
+                        let mut output = route.output();
+                        if let order::Side::Buy = order.side {
+                            output.amount = cmp::min(output.amount, order.buy.amount);
+                        }
```
MartinquaXD, line 84: It looks like these rounding issues can also cause us to buy slightly less than expected which also results in bad native prices. This can easily be reproduced on xdai.
Maybe we have to fix the rounding issue in the baseline algorithm itself?

MartinquaXD: I think if we are confident that the algorithm is sound and these are indeed only rounding issues we could consider replacing `cmp::min` with something like this:
```rust
if matches!(order.side, order::Side::Buy) && is_roughly_equal(output.amount, order.buy.amount) {
    output.amount = order.buy.amount
}
```
This should be a last resort, though.

PR #1735, opened 2023-08-03
Make simulation reentrancy guard less fragile

MartinquaXD:
[Context](https://cowservices.slack.com/archives/C0375NV72SC/p1690980795844549) on slack

Sasha noticed that the quote verification on`goerli` discards all quotes with the reason `prepareSwap can only be called once`. This message comes from a reentrancy guard that is supposed to prevent solvers tricking our simulation logic with specifically crafted calldata. I manually tested this reentrancy guard on `mainnet` without any issues which is why I was surprised that this causes issues on `goerli`.

I think what happens is this:
Default initializing a **MUTABLE** member variable with some value in solidity should in principle just be syntax sugar for writing a constructor that sets that value explicitly (immutable variables don't require a constructor). Since we are only using state overrides for the simulation helper contracts we are not actually running any constructor. That means which value the variable is initialized with depends on the node implementation executing our simulation. Since our reentrancy guard used a `bool` before it was basically a 50-50 chance whether the flag gets initialized to `true` or `false`.
When I manually tested the logic on `mainnet` it defaulted to `false` which resulted in correct behavior of the reentrancy guard. However, on `goerli` (using `nodereal`) the default behavior seems to be to initialize this bool to `true` which would always cause reverts.

The new approach is to store the flag in the contracts storage. The EVM spec requires uninitialized storage to return `0` which is equal to `false`. That means the new reentrancy guard implements the same idea but without the uncertainty how nodes handle uninitialized member variables.

### Test Plan
I manually tested this approach on both my setups without any issues.


nlordell (approved):
Wow! Incredible debugging!

crates/contracts/solidity/Trader.sol
```diff
@@ -21,8 +21,8 @@ contract Trader {
     address constant private TRADER_IMPL = address(0x10000);
 
     /// @dev Flag that ensures that `prepareSwap` gets called exactly once to
-    /// prevent custom interaction from calling it.
-    bool private _alreadyCalled = false;
+    /// prevent custom interactions from calling it.
+    address private _prepareSwapMarker;
```
nlordell, line 25: A more common implementation pattern to deal with this kind of thing is as follows:

```solidity
bytes32 constant private ALREADY_CALLED_SLOT = keccak256("Trader.alreadyCalled") - 1;
```

And then:

```solidity
function alreadyCalled() private returns (bool value) {
    bytes32 slot = ALREADY_CALLED_SLOT;
    assembly {
        value := sload(slot)
        sstore(slot, 1)
    }
}
```

And finally:

```solidity
require(!alreadyCalled(), "prepareSwap can only be called once");
```

nlordell: The reason for this suggestion is that it *may* be that `ReentrencyGuard` relies on specific values in that storage slot.

MartinquaXD: Thanks! Will try it out. What purpose does `-1` serve when computing the slot here?

MartinquaXD: [52b3ff7](https://github.com/cowprotocol/services/pull/1735/commits/52b3ff78b24b6b81c7b4d695c74abefa137a42b6)

PR #1737, opened 2023-08-04
Remove cow based fee subsidy

MartinquaXD:
The plan to use additional fee subsidies for cow token holders as a means of tokenomics didn't work and the DAO did not want to extend the cow based subsidies.
Since we are exploring other ideas for tokenomics there is no reason to believe this would come back so we might as well get rid of the code.

### Test Plan
compiler


sunce86 (approved):
Nice. We can easily revert if needed.

Issue #1614, opened 2023-06-23
Missing CIP20 data in settlement observations table


vkgnosis:
From [Slack](https://cowservices.slack.com/archives/C0375NV72SC/p1687266975324709?thread_ts=1687192847.302879&cid=C0375NV72SC)  reported by Harris.

Maybe I missed it in the discussion, but if not, we just realized that the
settlement_observations  table is also missing entries. Eg this one: https://etherscan.io/tx/0x51643b1a53c0791ae7943817241d6c3989aebcdb35858b89b824aec09af3b869 . Just raising it here since I feel the discussion was mostly about the solver_competitions table. Note that the  settlement_observations  table is crucial for CIP-20

I think this happened while we were testing staging mainnet colocation.

PR #1741, opened 2023-08-07
Save solver competition from autopilot

sunce86:
Progress on https://github.com/cowprotocol/services/issues/1672

Fixes https://github.com/cowprotocol/services/issues/1614 (solver competition part, since observations part is done).


crates/autopilot/src/run_loop.rs
```diff
@@ -181,15 +221,16 @@ impl RunLoop {
...
             tracing::info!(?competition, "saving competition");
             if let Err(err) = self.save_competition(&competition).await {
                 tracing::error!(?err, "failed to save competition");
                 return;
             }
 
-            tracing::info!(url = %self.drivers[index].url, "settling with solver");
-            match self.settle(id, &self.drivers[index], &solution).await {
+            tracing::info!(url = %self.drivers[*index].url, "settling with solver");
+            match self.settle(id, &self.drivers[*index], solution).await {
```
MartinquaXD, line 233: This looks like a (pre-existing) bug. Since the indices refer to solutions sorted by score and not by their driver I believe this does will usually tell the wrong driver to execute its solution.

crates/autopilot/src/run_loop.rs
```diff
@@ -169,6 +178,37 @@ impl RunLoop {
...
+                    .map(|(index, response)| SolverSettlement {
+                        solver_address: response.submission_address,
+                        score: Some(Score::Solver(response.score)),
+                        ranking: Some(solutions.len() - index),
+                        orders: response
+                            .orders
+                            .iter()
+                            .map(|o| Order {
+                                id: *o,
+                                ..Default::default()
```
MartinquaXD, line 202: It probably makes sense to leave a TODO or open an issue about later removing the fields which can not be populated with colocated drivers.

Also I'm sorry but I my brain must have been AFK when I reviewed https://github.com/cowprotocol/services/pull/1731 because I now realize that we should not expect call data from the `/solve` request. Otherwise solvers will not be able to properly utilize RFQ based liquidity.
I see that you implemented the calldata collection via `/solve` to unblock the colocation efforts but we definitely have to revisit that implementation before releasing the driver.

crates/autopilot/src/run_loop.rs
sunce86: AFAIS indexes are paired with solutions making a tuple `(index, solution)` therefore indexes are following it's original solution even after sorting by score.

Drivers: [0,1,2]
Solutions: [(0,A),(1,B),(2,C)]
Sorted solutions example: [(1,B), (2,C), (0,A)]

crates/autopilot/src/run_loop.rs
sunce86: I agree. However, it seems that we don't have to decide until we push external solvers to run their own driver. By that time, we should have `/reveal` endpoint added where the `calldata` will be revealed by the winning driver. 
Do you think I should do it asap? (IMO it does not block us from enabling internal colocation).

crates/autopilot/src/run_loop.rs
```diff
@@ -103,15 +110,17 @@ impl RunLoop {
...
-            let reference_score = solutions
-                .last()
-                .map(|(_, response)| response.score)
-                .unwrap_or_default();
-            let mut participants = solutions
+            let reference_score = match solutions.len() > 1 {
+                true => solutions
+                    .get(solutions.len() - 2)
+                    .map(|(_, response)| response.score)
+                    .unwrap_or_default(),
```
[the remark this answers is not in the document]
sunce86: Fixed using iterators.

crates/autopilot/src/run_loop.rs
MartinquaXD: I agree that it only has to happen before external parties are expected to run their own driver.
It's fine if that only happens after we iron out all the remaining kinks of the new driver.

MartinquaXD (approved):

crates/autopilot/src/run_loop.rs
sunce86: Cool. Added https://github.com/cowprotocol/services/issues/1744 to capture the work around this.

sunce86: Experimented quickly with adding /reveal and ended with https://github.com/cowprotocol/services/pull/1745

PR #1742, opened 2023-08-07
Make e2e crate public

devanoneth:
This makes most of the e2e crate public. This is so that solvers can use the crate in order to craft their own e2e tests and not need to copy code or write their own test harness.

Leaving this in draft for now as:
1. I want to gain feedback on this general idea.
2. I might need to make more changes as I'm still developing our own tests against my local fork.


MartinquaXD:
> I might need to make more changes as I'm still developing our own tests against my local fork.

Could you roughly outline how you set up your tests? I'm trying to understand how your tests might look like and how this PR would improve your situation.

Do you have a fork of this repo as part of your test repository and want to reference the `e2e` test code from there somehow?

devanoneth:
>Could you roughly outline how you set up your tests? I'm trying to understand how your tests might look like and how this PR would improve your situation.

Sure, I run the tests for our solver very similarly to how you do in your tests, for example here: https://github.com/cowprotocol/services/blob/main/crates/e2e/tests/e2e/limit_orders.rs

Except that it has a different `OrderCreation` variable, it runs our solver's HTTP server and gets the `old_driver` to hit our solver instead of using one of the built-in solvers. Thus, with minimal code changes on our side, we can have rust orchestrated e2e tests against our solver.

> Do you have a fork of this repo as part of your test repository and want to reference the e2e test code from there somehow?

Exactly. Just like how the `local_node_single_limit_order` relies on code in the `e2e::setup` module, so do our tests. But, if the `setup` module stays inside the `test` dir, it cannot be exported / relied on by other projects. With this refactor, all of your tests still run and other codebases can use the helpful functions that you have created in the `setup` module. e.g. `run_test` and all of the function in the `services.rs` file.

docker/hardhat/hardhat.config.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
                 accountsBalance: "1000000000000000000000000"
             },
             gas: 1e7,
-            gasPrice: 1
+            gasPrice: 1,
+            chainId: 1
```
devanoneth, line 11: I put this in so that when running against a forked node the correct contract addresses are loaded in e.g. by running `BalancerV2Vault::deployed(web3).await.unwrap()`.

I've tested locally and it doesn't break the existing tests AFAIK, but looking forward to seeing CI run against this.

crates/e2e/src/nodes/forked_node.rs
```diff
@@ -0,0 +1,81 @@
...
+        &self.transport
+    }
+}
+
+/// Implements functions that are only available in a forked node.
+///
+/// Relevant RPC calls for the Hardhat forked network can be found at:
+/// https://hardhat.org/hardhat-network/docs/reference#hardhat-network-methods
+impl<T: Transport> ForkedNodeApi<T> {
+    pub fn fork(&self, fork_url: &String) -> CallFuture<bool, T::Out> {
```
MartinquaXD, line 67: `fork_url` should probably be a `URL` and only transform to a string inside the function.
Same goes for all the other URLs that are currently strings in this PR.

docker/hardhat/hardhat.config.js
MartinquaXD: Hmm, this would be an issue if somebody's solver only works on some other chain but still wants to use a forked node. But I guess we can figure that out when the first person is interested in that.

MartinquaXD (commented):
The idea makes sense to me overall. 👌

docker/hardhat/hardhat.config.js
devanoneth: Yeah, I hear you on that. Unfortunately hardhat doesn't support changing the `chainId` once the process is running, I already checked if I could do that to avoid changing this file at all. We could add different networks though and invoke the correct network via docker compose. Or, maybe there is an even better solution we could come to once needed.

devanoneth:
Thanks for the review @MartinquaXD! I think I've addressed all your comments so far. Also appreciate that you are happy to have this upstream. Will help all solvers run e2e tests against their own infra. :)

crates/e2e/src/setup/mod.rs
```diff
@@ -82,69 +94,52 @@ where
...
-
-    run(f, filters, Some(solver_address), Some(fork_url)).await
+    run(f, Some(extra_filters), Some(solver_address), Some(fork_url)).await
 }
 
-async fn run<F, Fut>(
+async fn run<F, Fut, T>(
     f: F,
-    filters: Vec<String>,
+    filters: Option<impl IntoIterator<Item = T>>,
```
MartinquaXD, line 134: This does not need the `Option`. If you don't want to pass additional filter you can pass `std::iter::empty()` or probably just `[]`.

crates/e2e/src/setup/mod.rs
```diff
@@ -69,6 +69,18 @@ const DEFAULT_FILTERS: [&str; 9] = [
     "orderbook::api::request_summary=off",
 ];
 
+fn take_filters<T>(custom_filters: Option<impl IntoIterator<Item = T>>) -> Vec<String>
```
MartinquaXD, line 72: Sorry, I just needed some function name for my example. 😅 
In this case sth like `with_default_filters` or so would make sense to me.

MartinquaXD (commented):
Last few nits and then it's good to go from my perspective.
Maybe @sunce86 could also give it a look.

ape-dev-cs:

devanoneth:
My colleague @ape-dev-cs has stepped in to close this off as I'm AFK for the next few days. :)

MartinquaXD (approved):

Issue #1744, opened 2023-08-08
Add /reveal to driver


sunce86:
Right now, there is a hard constraint that protocol needs to know the both the final calldata and uninternalized calldata of the winning solution (because of solver rewards script using them).

Driver currently returns this data as part of the response from `/solve`, however, this requires even non winning solutions to provide calldata. Instead, we want to:

1. Call `/solve` on all drivers and get scores
2. Rank by score and choose winner. Call `/reveal` on winner and get calldata.
3. Call `/settle` on winner.

PR #1745, opened 2023-08-08
Add /reveal to driver

sunce86:
Fixes https://github.com/cowprotocol/services/issues/1744

Calldata is no longer received from `/solve` but from `/reveal`. Also, took the chance and moved the `orders` to `/reveal` also. 

The consequence of this is that we save the `calldata` and `orders` for winning solution only.

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -5,10 +5,13 @@ use {
             Postgres,
         },
         driver_api::Driver,
-        driver_model::solve::{self, Class},
+        driver_model::{
+            reveal,
+            solve::{self, Class},
+        },
         solvable_orders::SolvableOrdersCache,
     },
-    anyhow::{anyhow, Context, Result},
+    anyhow::{anyhow, ensure, Context, Result},
     chrono::Utc,
     database::order_events::OrderEventLabel,
     itertools::Itertools,
@@ -94,11 +97,6 @@ impl RunLoop {
         solutions.sort_unstable_by_key(|solution| solution.1.score);
         let competition_simulation_block = self.current_block.borrow().number;
 
-        let events: Vec<_> = solutions
-            .iter()
-            .flat_map(|(_, s)| s.orders.iter().map(|o| (*o, OrderEventLabel::Considered)))
-            .collect();
-
         // TODO: Keep going with other solutions until some deadline.
         if let Some((index, solution)) = solutions.last() {
             // The winner has score 0 so all solutions are empty.
@@ -106,7 +104,22 @@ impl RunLoop {
                 return;
             }
 
+            tracing::info!(url = %self.drivers[*index].url, "revealing with solver");
+            let revealed = match self.reveal(id, &self.drivers[*index]).await {
+                Ok(result) => result,
+                Err(err) => {
+                    tracing::error!(?err, "solver {index} failed to settle");
+                    return;
+                }
+            };
+
+            let events = revealed
+                .orders
+                .iter()
+                .map(|o| (*o, OrderEventLabel::Considered))
+                .collect::<Vec<_>>();
             self.database.store_order_events(&events).await;
+
             let auction_id = id;
             let winner = solution.submission_address;
             let winning_score = solution.score;
@@ -123,13 +136,13 @@ impl RunLoop {
             let block_deadline = competition_simulation_block
                 + self.submission_deadline
                 + self.additional_deadline_for_rewards;
-            let call_data = solution.calldata.internalized.clone();
-            let uninternalized_call_data = solution.calldata.uninternalized.clone();
+            let call_data = revealed.calldata.internalized.clone();
+            let uninternalized_call_data = revealed.calldata.uninternalized.clone();
             // Save order executions for all orders in the solution. Surplus fees for
             // partial limit orders will be saved after settling the order
             // onchain.
             let mut order_executions = vec![];
-            for order_id in &solution.orders {
+            for order_id in &revealed.orders {
                 let auction_order = auction
                     .orders
                     .iter()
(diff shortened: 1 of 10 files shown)


crates/autopilot/src/run_loop.rs
```diff
@@ -187,26 +199,41 @@ impl RunLoop {
...
-                                ..Default::default()
-                            })
-                            .collect(),
-                        call_data: response.calldata.internalized.clone(),
-                        uninternalized_call_data: Some(response.calldata.uninternalized.clone()),
-                        // TODO: revisit once colocation is enabled (remove not populated fields)
-                        // Not all fields can be populated in the colocated world
-                        ..Default::default()
+                    .map(|(index, response)| {
+                        let (orders, call_data, uninternalized_call_data) =
```
MartinquaXD, line 203: Might be slightly nicer to read if it would look like this:
```
let mut settlement = SolverSettlement{..};
if is_winner {
    settlement.orders = ...;
    ...
}
return settlement;
```

crates/autopilot/src/run_loop.rs
```diff
@@ -234,7 +261,10 @@ impl RunLoop {
             }
 
             tracing::info!(url = %self.drivers[*index].url, "settling with solver");
-            match self.settle(id, &self.drivers[*index], solution).await {
+            match self
+                .settle(id, &self.drivers[*index], solution, &revealed)
+                .await
```
MartinquaXD, line 266: The reason why you want to introduce `/reveal` instead of taking the calldata from `/settle` is because the request to `/settle` might fail or in some other way not return the calldata, right?

crates/driver/src/domain/competition/mod.rs
```diff
@@ -231,15 +237,15 @@ impl Competition {
...
 
 /// Solution information revealed to the protocol by the driver before the
-/// settlement happens.
-/// Initially the calldata was planned to be revealed to the protocol after
-/// the \settle is called, but this would require protocol to wait for the
-/// settlement to be settled onchain before saving the solver competition data.
-/// This is unacceptable because saving the solver competition data could fail
-/// and the protocol would have inconsistent db state.
+/// settlement happens. Note that the calldata is only revealed once the
+/// protocol instructs the driver to reveal, and not before.
```
MartinquaXD, line 241: Looks like this doc comment could be split to have the relevant information on `Solved` and `Revealed`.

crates/autopilot/src/run_loop.rs
```diff
@@ -94,18 +97,27 @@ impl RunLoop {
...
+                    tracing::error!(?err, "solver {index} failed to reveal");
+                    return;
+                }
+            };
+
+            let events = revealed
+                .orders
+                .iter()
+                .map(|o| (*o, OrderEventLabel::Considered))
+                .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
sunce86: Do you know why we need this information? At least I don't see a purpose to know if some order was included in some solution that did not win?
Is it worth forcing the `/solve` to return this data just for debugging purposes? 
Reveal should rarely fail since it is simple -> doesn't even communicate to underlying solver, it just reads already saved settlement and returns simple data -> should be quite fast.

crates/autopilot/src/run_loop.rs
```diff
@@ -94,18 +97,27 @@ impl RunLoop {
...
             // The winner has score 0 so all solutions are empty.
             if solution.score == 0.into() {
                 return;
             }
 
+            tracing::info!(url = %self.drivers[*index].url, "revealing with solver");
+            let revealed = match self.reveal(id, &self.drivers[*index]).await {
+                Ok(result) => result,
+                Err(err) => {
+                    tracing::error!(?err, "solver {index} failed to reveal");
```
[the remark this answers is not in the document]
sunce86: Change the log to `warn`?

crates/autopilot/src/run_loop.rs
sunce86: Exactly. Autopilot could be killed while `settle` is executing and we also can't force driver to return early (before starting the submission process).
The only alternative way I can think of (if we want to avoid introducing `/reveal`) is to do it in some background autopilot worker where we would periodically check all auction ids with missing calldata and fetch them from drivers but it would require drivers to save that data in their own db which I don't like.

crates/autopilot/src/run_loop.rs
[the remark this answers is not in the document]
MartinquaXD: > Do you know why we need this information?

It's part of our SLI initiative which aims to give us a detailed overview where in our systems orders spend a lot of time. If an order is not part of the winning solution but some runner up we could estimate with this information how much we could gain from letting multiple solver submit disjoint settlements for example.

> Is it worth forcing the /solve to return this data just for debugging purposes?

Hmm, you have a point. This was not part of the initial API design and we probably shouldn't change that because we want to have some debugging information.
On the other hand this doesn't seem like a big ask from a solver's perspective. They already have to estimate a score for their solution so surely they must have an idea which orders this solution would settle.
I think it's fine to not have that for now.

crates/autopilot/src/run_loop.rs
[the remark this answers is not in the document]
MartinquaXD: Yeah, I would go with that. We could also have some alert based on metrics if this problem keeps happening but it seems very far fetched that only this endpoint would fail.

crates/autopilot/src/run_loop.rs
MartinquaXD: Makes sense. 👍 
I'm very much looking forward to new smart contracts that make it so we no longer have to worry about internal buffer trading. 😩

crates/autopilot/src/run_loop.rs
sunce86: Indeed, refactored.

crates/driver/src/domain/competition/mod.rs
sunce86: You are right. I've updated the docs to reflect the changes.

Issue #1664, opened 2023-07-11
Make DEX-Based Solvers Respect Auction Deadline


nlordell:
Currently, DEX-based solvers in the `solvers` crate will run auctions for all orders. Change the code to only solve orders until it gets close to the deadline.

PR #1751, opened 2023-08-10
Respect deadline in solvers crate

sunce86:
Progress on https://github.com/cowprotocol/services/issues/1672

Fixes https://github.com/cowprotocol/services/issues/1664


crates/solvers/src/domain/solver/dex/mod.rs
```diff
@@ -36,20 +36,42 @@ impl Dex {
...
+                    }
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let solutions = futures::future::join_all(futures)
+            .await
+            .into_iter()
+            .flatten()
+            .collect();
```
MartinquaXD, line 74: Historically we preferred solving orders sequentially instead of all of them in parallel because it would overwhelm the dex API.
Especially with auctions now containing limit orders this might result in hundreds of requests at once.

sunce86: Indeed. I was carried away with the //concurrency comment above the code.
Fixed.

fleupold (approved):

crates/solvers/src/domain/solver/dex/mod.rs
```diff
@@ -37,16 +37,34 @@ impl Dex {
...
         let mut solutions = Vec::new();
         for order in auction.orders.iter().cloned() {
             let span = tracing::info_span!("solve", order = %order.uid);
-            if let Some(solution) = self
-                .solve_order(order, &auction.tokens, auction.gas_price)
-                .instrument(span)
-                .await
-            {
+            let order_uid = order.uid;
+            let deadline = auction
```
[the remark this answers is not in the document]
sunce86: Should be already reduced here:
https://github.com/cowprotocol/services/blob/main/crates/driver/src/infra/solver/mod.rs#L128-L130

crates/solvers/src/domain/solver/dex/mod.rs
```diff
@@ -37,16 +37,34 @@ impl Dex {
...
+                .deadline
+                .signed_duration_since(chrono::offset::Utc::now())
+                .to_std();
+
+            if deadline.is_err() || deadline.unwrap().is_zero() {
```
MartinquaXD, line 50: Setting up a timeout for every order individually results in a lot of awkward error handling and lots of logs about missed timeouts (per order).
Alternatively you could wrap the whole solving process in a single future and only apply a timeout to that:
```rust
        let mut solutions = Vec::new();
        let solve_orders = async {
            for order in auction.orders.iter().cloned() {
                let span = tracing::info_span!("solve", order = %order.uid);
                if let Some(solution) = self
                    .solve_order(order, &auction.tokens, auction.gas_price)
                    .instrument(span)
                    .await
                {
                    solutions.push(solution);
                }
            }
        };
        let time_to_deadline = auction
            .deadline
            .signed_duration_since(chrono::offset::Utc::now())
            .to_std()
            .unwrap_or_default();
        if tokio::time::timeout(time_to_deadline, solve_orders).await.is_err() {
            tracing::debug!("reached deadline; stopping to solve");
        }
```

sunce86: After all, I guess we don't need to know the exact orders that were skipped. If needed, we can always go back to timeout per order which indeed is less readable.
Fixed.

MartinquaXD (approved):

crates/solvers/src/domain/solver/dex/mod.rs
MartinquaXD: One could easily get the list of skipped orders with this approach by initializing the `orders` iterator outside the future. Then after the deadline has been reached you could collect the remaining orders from that iterator and put that into a single log.
But I would still be cautious about that since this will print a very big log for every solver implementation on every auction.

MartinquaXD: To not issue these big logs every time we could have `reached deadline; stopping to solve` be a `debug` log and introduce a second log that contains the skipped orders which is `tracing`.
And one slight issue that occurred to me about the hoisting the iterator out is that we'll always miss logging the one order that is currently being solved when the deadline expires (one could argue that that one should also be considered skipped).

sunce86: Sorry I did not see these comments before setting the PR to merge automatically.

PR #1756, opened 2023-08-14
Delete a few lines

MartinquaXD:
7vpw42

Theoretically we can use any price estimator via the colocated driver setup. That means we can get rid of all the price estimation implementations in `shared` and all the things only needed to configure them.

Once we also move all the solvers to the colocated driver world we can probably get rid of a few more things (mostly the API wrappers of 1Inch, 0x and so on).

This should probably only be merged after we fixed the rounding issue in the `Baseline` solver and I think `Paraswap` still needs to fetch the token meta data to function properly.

### Test Plan
Compiler, CI

Some e2e tests still need to be adjusted because they use removed CLI arguments (e.g. `--(native-)price-estimators`).


nlordell:
So... rewarding...

MartinquaXD:
Closing since this depends on a few things I didn't have on my radar or take longer than anticipated.
But we'll come back to this. 🧹

PR #1760, opened 2023-08-15
Use new balancer API

MartinquaXD:
@nlordell suggested to try out the new balancer API because that should be faster, offers more optimal solutions and already takes care of crafting the tx which means we can get rid of some code.

The new API is still in alpha and has a few bugs and a strange response format which might change in the future but overall this PR should be all we need to switch over to the new API.

### Test Plan
Did a few manual tests which showed this blocking bug: https://github.com/balancer/balancer-api/issues/234
Will finish up by adapting the existing tests when the API is in better shape.


MartinquaXD:
Not a high priority right now and would have to be ported into the new solvers repo in the gnosis org.

PR #1761, opened 2023-08-15
Baseline Solver: Fallback To Other Solutions When Traversal Fails

nlordell:
@MartinquaXD identified that we are _still_ running into rounding issues with buy orders for the baseline solver.

It turns out that this is because Balancer weighted pools are "unstable". What this means is if we compute the input amount `X` required for an exact output amount `Y`, then using the exact same pool state compute the output amount `Y'` for an exact input amount `X`, it can happen that `Y' < Y`. This is a problem for buy orders, as the baseline solver computes output amounts backwards from `Y` to `X`, but needs to build solutions trading from `X` to `Y`. However, the computed amounts will only buy `Y'` which is **not sufficient for executing the buy order**.

This PR changes the baseline solver logic to filter solutions after attempting the sell token -> buy token traversal. This allows paths with Balancer weighted pools that observe the above behaviour to be skipped so that the baseline solver can at least produce a solution.

In a follow up I intend to:
- [x] Converge towards an executed sell amount with a path that uses Balancer weighted pools (experimentally, we can converge in very few iterations): https://github.com/cowprotocol/services/pull/1766
- [ ] Fix an unrelated issue for Balancer weighted pool computation that was discovered during investigations (#1763).

### Example

Consider the following Balancer weighted pool:

```
{
    "kind": "weightedproduct",
    "tokens": {
        "0x177127622c4a00f3d409b75571e12cb3c8973d3c": {
            "balance": "1963528800698237927834721",
            "scalingFactor": "1000000000000000000",
            "weight": "0.5",
        },
        "0x9c58bacc331c9aa871afd802db6379a98e80cedb": {
            "balance": "1152796145430714835825",
            "scalingFactor": "1000000000000000000",
            "weight": "0.5",
        }
    },
    "fee": "0.005",
    "id": "1",
    "address": "0x21d4c792ea7e38e0d0819c2011a2b1cb7252bd99",
    "gasEstimate": "88892"
}
```

Note that if we compute the required input amount in order to buy exactly `9056454904357528` tokens, it will return `15503270361045187239` (both in the Rust code and [on-chain](https://dashboard.tenderly.co/shared/simulation/4291a5cb-64f5-478f-9ad4-2aa7a5bf248d)). However, unexpectedly, if we were to sell the same sell amount `15503270361045187239` to the pool, we would only receive `9056454904357125` (also both in Rust code and [on-chain](https://dashboard.tenderly.co/shared/simulation/148e3aa0-7fd4-42b0-9559-7d0fee071d47)), `403` less than what we would need to satisfy such a buy order. This is the root of the rounding problem we are seeing in the Baseline solver.

### Test Plan

Added a new E2E test for the Baseline solver that verifies the fallback logic.


nlordell:
Annnnnd, it turns out that different weighted pool versions use different math:

From V1 to V3/4:

This:
```sol
    /**
     * @dev Returns x^y, assuming both are fixed point numbers, rounding up. The result is guaranteed to not be below
     * the true value (that is, the error function expected - actual is always negative).
     */
    function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
        uint256 raw = LogExpMath.pow(x, y);
        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);

        return add(raw, maxError);
    }
```

Became this:

```sol
    /**
     * @dev Returns x^y, assuming both are fixed point numbers, rounding up. The result is guaranteed to not be below
     * the true value (that is, the error function expected - actual is always negative).
     */
    function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
        // Optimize for when y equals 1.0, 2.0 or 4.0, as those are very simple to implement and occur often in 50/50
        // and 80/20 Weighted Pools
        if (y == ONE) {
            return x;
        } else if (y == TWO) {
            return mulUp(x, x);
        } else if (y == FOUR) {
            uint256 square = mulUp(x, x);
            return mulUp(square, square);
        } else {
            uint256 raw = LogExpMath.pow(x, y);
            uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);

            return add(raw, maxError);
        }
    }
```

nlordell:
Created https://github.com/cowprotocol/services/issues/1763 to track the issue above.

crates/solvers/src/boundary/baseline.rs
```diff
@@ -44,34 +44,50 @@ impl<'a> Solver<'a> {
...
+                            estimate.value,
+                        )?;
+
+                        let sell = estimate.value;
+                        let buy = segments
+                            .last()
+                            .map(|segment| segment.output.amount)
+                            .unwrap_or(sell);
+
+                        if sell <= request.sell.amount && buy >= request.buy.amount {
```
MartinquaXD, line 69: Pointing out that `bool::then_some()` is a thing (although it might end up with bad formatting due to the long expression.

Also would be nice to add a small comment explaining why that is needed here so people don't have to track down the reason using `git blame`.

MartinquaXD (approved):
Thanks for tracking down that bug! 🕵️ 
That would have taken me forever.

crates/solvers/src/boundary/baseline.rs
nlordell: > Pointing out that `bool::then_some()` is a thing

Thanks... I **always** forget about it (and agree that it generally makes nicer code).

If it is OK with you, I would not implement these suggestions (although I agree with them) just because the code changes again in https://github.com/cowprotocol/services/pull/1766.

MartinquaXD (approved):

nlordell:
Merging this without https://github.com/cowprotocol/services/issues/1763 as this should unblock Gnosis Chain co-location.

fleupold (commented):
Change looks good. Not needed assuming we are enabling the pools soon again, but if I understand correctly this filters out whole paths where Balancer pools are involved (vs. just the unstable pools). Would this be something that could be pushed down into the per-hop calculation?

nlordell:
Sorry @fleupold! Saw your message a bit late.

> Would this be something that could be pushed down into the per-hop calculation?

Yes, it could, but it was a bit messy to put into the baseline solver code.

However, I have a, IMO, better solution to converge to sufficient sell amount without sacrificing the "best" price in https://github.com/cowprotocol/services/pull/1766 (that being said, there would be veeeery rare cases where scaling up the sell amount would result in a slightly worse settlement, but this would be fairly minor as both settlements would be very close in this case and likely not make a meaningful price difference)

PR #1766, opened 2023-08-16
Baseline: Converge To Sufficient Sell Amount

nlordell:
Follow up to #1761

This PR adjusts the Baseline solver algorithm to iterate on progressively larger sell amounts in an attempt to traverse from an order's sell token to its buy token ensuring that the executed buy amount is enough to cover the buy order.

This is needed because some baseline liquidity (notable Balancer weighed pools) are "unstable", meaning that if we compute the input amount `X` required for an exact output amount `Y`, then using the exact same pool state compute the output amount `Y'` for an exact input amount `X`, it can happen that `Y' < Y`, meaning that the executed interactions would not generate enough tokens to cover the buy order.

### Test Plan

Adjusted E2E test to demonstrate that the path with the better price (over an unstable Balancer pool) is used.

### DIFF
--- a/crates/shared/src/baseline_solver.rs
+++ b/crates/shared/src/baseline_solver.rs
@@ -2,9 +2,13 @@
 //! onchain liquidity.
 
 use {
+    crate::conversions::U256Ext,
     ethcontract::{H160, U256},
     model::TokenPair,
-    std::collections::{HashMap, HashSet},
+    std::{
+        cmp,
+        collections::{HashMap, HashSet},
+    },
 };
 
 /// The maximum number of hops to use when trading with AMMs along a path.
@@ -68,15 +72,15 @@ pub fn estimate_buy_amount<'a, L: BaselineSolvable>(
                 let (best_liquidity, amount) = liquidity
                     .get(&TokenPair::new(*current, previous)?)?
                     .iter()
-                    .map(|liquidity| {
-                        (
+                    .filter_map(|liquidity| {
+                        Some((
                             liquidity,
-                            liquidity.get_amount_out(*current, (amount, previous)),
-                        )
+                            liquidity.get_amount_out(*current, (amount, previous))?,
+                        ))
                     })
-                    .max_by(|(_, amount_a), (_, amount_b)| amount_a.cmp(amount_b))?;
+                    .max_by_key(|(_, amount)| *amount)?;
                 path.push(best_liquidity);
-                Some((amount?, *current, path))
+                Some((amount, *current, path))
             },
         )
         .map(|(amount, _, liquidity)| Estimate {
@@ -93,6 +97,39 @@ pub fn estimate_sell_amount<'a, L: BaselineSolvable>(
     path: &[H160],
     liquidity: &'a HashMap<TokenPair, Vec<L>>,
 ) -> Option<Estimate<'a, U256, L>> {
+    // Some baseline liquidity is "unstable", where if you compute an input
+    // amount large enough to buy X tokens, selling the computed amount over the
+    // same pool in the exact same state will yield X-𝛿 tokens. To work around
+    // this, for each hop,  to converge to some `sell` amount >= the required
+    // buy amount.
+    let get_amount_in =
+        |liquidity: &L, in_token: H160, (exact_out_amount, out_token): (U256, H160)| {
+            /// Upper bound on number of iterations to find a sufficiently high
+            /// sell amount.
+            const MAX_ITERATIONS: usize = 3;
+
+            let mut in_amount = liquidity.get_amount_in(in_token, (exact_out_amount, out_token))?;
+            for _ in 0..MAX_ITERATIONS {
+                let out_amount = liquidity.get_amount_out(out_token, (in_amount, in_token))?;
+                dbg!(in_amount, out_amount);
+                if out_amount >= exact_out_amount {
+                    return Some(in_amount);
+                }
+
+                // The computed output amount is not enough; so scale the sell
+                // amount up a bit.
+                let bump = cmp::max(
+                    (exact_out_amount - out_amount)
+                        .checked_mul(in_amount)?
+                        .checked_ceil_div(&out_amount)?,
+                    U256::from(1),
+                );
+                in_amount = in_amount.checked_add(bump)?;
+            }
+
+            None
+        };
+
     let buy_token = path.last()?;
     path.iter()
         .rev()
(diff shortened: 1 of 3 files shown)


crates/solvers/src/tests/baseline/buy_order_rounding.rs
```diff
@@ -220,19 +220,16 @@ async fn balancer() {
...
+    // amount. This is inevitable because of rounding.
     assert_eq!(
         solution,
         json!({
             "solutions": [{
                 "id": 0,
                 "prices": {
                     "0x177127622c4a00f3d409b75571e12cb3c8973d3c": "1000000000000000000",
-                    "0xe91d153e0b41518a2ce8dd3d7944fa863463a97d": "1004013040121365096289871"
+                    "0xe91d153e0b41518a2ce8dd3d7944fa863463a97d": "15503270361046562907"
```
nlordell, line 232: Price improvement 🎉!

nlordell:
@fleupold - I will try to push the "sell amount convergence" to each hop in the `baseline_solver` code here, as it would have slightly better results.

crates/solvers/src/boundary/baseline.rs
```diff
@@ -54,23 +60,47 @@ impl<'a> Solver<'a> {
...
+                                return Some((segments, sell));
+                            }
+
+                            // The computed output amount is not enough for the
+                            // order, so scale the sell amount up a bit.
+                            let bump = cmp::max(
+                                (request.buy.amount - buy)
+                                    .checked_mul(sell)?
+                                    .checked_ceil_div(&buy)?
+                                    .checked_mul(2.into())?,
```
MartinquaXD, line 97: I initially thought this doing a binary search but it's actually doing this, right?
```
bump = convert_to_sell(missing_buy_amount) * 2
```
Am I getting this right? Seems surprising to me.

MartinquaXD (commented):
Nice idea! Just one thought about the math of it all.

nlordell:
As @fleupold suggested, I modified this to apply the logic per-hop instead of doing it over the whole path. This has a major advantage that if we can't find an appropriate executed sell amount for a single baseline liquidity, we don't disregard the entire path, but just that liquidity.

crates/solvers/src/boundary/baseline.rs
nlordell: I changed it to `bump = convert_to_sell(missing_buy_amount)` - the *2 is confusing because it wasn't actually needed 🫣.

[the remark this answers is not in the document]
nlordell: It was a logic mistake - removed the *2.

Yes, it could be used to overshoot, but it shouldn't be needed.

nlordell: > doing something like a binary search

We would need to know a reasonable upper bound, which we don't. Instead I'm linearly interpolating what the sell amount should be, so something like:

```
bump = convert_to_sell(missing_buy_amount)
```

crates/solvers/src/boundary/baseline.rs
```diff
@@ -44,47 +44,43 @@ impl<'a> Solver<'a> {
...
+                        self.traverse_path(&estimate.path, request.sell.token.0, estimate.value)?;
+                    let buy = segments
+                        .last()
+                        .map(|segment| segment.output.amount)
+                        .unwrap_or(estimate.value);
+
+                    (estimate.value <= request.sell.amount && buy >= request.buy.amount)
+                        .then_some((segments, estimate))
+                })
+                .min_by_key(|(_, estimate)| estimate.value)?,
```
MartinquaXD, line 66: nit: I liked that you called the estimate either `sell` or `buy` in the previous version.
That made it trivial to understand why `min_by_key()` was correct in one case and `max_by_key()` was correct in the other case.

MartinquaXD (approved):
I'm not super familiar with the internals of the baseline solver but the change looks reasonable to me. 👍

crates/solvers/src/tests/baseline/buy_order_rounding.rs
```diff
@@ -247,12 +244,166 @@ async fn balancer() {
...
+                        "order": "0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a\
+                                    2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a\
+                                    2a2a2a2a",
+                        "executedAmount": "9056454904357528"
                     }
+                ],
+                "interactions": [
+                    {
+                        "kind": "liquidity",
+                        "internalize": false,
```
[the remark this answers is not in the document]
nlordell: I would rather do that in the `driver` if the solver has "auto-internalization" enabled. We would need to have `trusted && balance > outputAmount`

crates/shared/src/baseline_solver.rs
```diff
@@ -104,19 +141,16 @@ pub fn estimate_sell_amount<'a, L: BaselineSolvable>(
...
                     .iter()
-                    .map(|liquidity| {
-                        (
+                    .filter_map(|liquidity| {
+                        Some((
                             liquidity,
-                            liquidity.get_amount_in(*current, (amount, previous)),
-                        )
+                            get_amount_in(liquidity, *current, (amount, previous))?,
+                            //liquidity.get_amount_in(*current, (amount, previous))?,
```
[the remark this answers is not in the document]
nlordell: 218adf851d76f18c6a420c0f2cbd1ba30a00c70e

crates/solvers/src/boundary/baseline.rs
nlordell: 218adf851d76f18c6a420c0f2cbd1ba30a00c70e

crates/solvers/src/boundary/baseline.rs
```diff
@@ -44,47 +44,40 @@ impl<'a> Solver<'a> {
...
+                        &self.amms,
+                    )?;
+                    let segments =
+                        self.traverse_path(&sell.path, request.sell.token.0, sell.value)?;
+                    let buy = segments
+                        .last()
+                        .map(|segment| segment.output.amount)
+                        .unwrap_or(sell.value);
+
+                    (sell.value <= request.sell.amount && buy >= request.buy.amount)
```
fleupold, line 63: We are doing the two sided check for buy order but not for sell orders. Should we add it for sell orders as well to be safe? Also, I'm wondering if we should log in the unexpected case that `buy < request.buy.amount`

fleupold (approved):
Looks good to me. Very nice fix!

crates/solvers/src/boundary/baseline.rs
nlordell: > We are doing the two sided check for buy order but not for sell orders.

Mostly because we don't have the same round-tripping logic for sell orders, so it can't ever happen. But I agree that we should likely do a check for both (as there is an implicit assumption on the behavior of the `baseline_solver` module code).

> Also, I'm wondering if we should log in the unexpected case that `buy < request.buy.amount`

Will add for both sell and buy orders.

PR #1769, opened 2023-08-16
Only count winner if we actually got a quote

MartinquaXD:
Currently we sort price estimate by some condition and pick the estimator that provided the first entry as the winner.
However, it can happen that the first entry is an error. That happens if all estimators returned an error. In that case it doesn't make sense to say some solver provided a "winning error".

This PR makes it so that metrics related to winning are only incremented if the best solution is not an error.

I decided to leave the arguably somewhat misleading log of "winning price estimate" unchanged because it's only used for debugging and introducing a new log specifically when all solvers produce errors seems to do more harm than good.

### Test Plan
CI

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -215,30 +215,33 @@ impl PriceEstimating for RacingCompetitionPriceEstimator {
             let estimator = self.inner[estimator_index].0.as_str();
             tracing::debug!(?query, ?result, estimator, "winning price estimate");
 
-            // Collect stats for winner predictions.
             let requests = self.inner.len() as u64;
-            if let (Some(competition), Ok(_)) = (&self.competition, &result) {
-                let trade = Trade::from(query);
-                let estimator = EstimatorIndex(estimator_index);
-                if let Some(predictions) = predictions.get(&trade) {
-                    let was_correct = predictions.iter().any(|p| p.winner == estimator);
-                    metrics().record_prediction(&trade, was_correct);
-                }
-                competition.record_winner(trade, estimator);
-                metrics()
-                    .requests
-                    .with_label_values(&["saved"])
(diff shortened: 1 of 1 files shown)


sunce86 (approved):

sunce86:
> I decided to leave the arguably somewhat misleading log of "winning price estimate" unchanged because it's only used for debugging and introducing a new log specifically when all solvers produce errors seems to do more harm than good.

Why not putting in under `result.is_ok()`?

nlordell (approved):

MartinquaXD:
> Why not putting in under result.is_ok()?

I think I sparked confusion unwillingly. The `winning price estimate` already contains the whole result. So we already have the information whether or not we have an error in the log.
[error log](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/c0e240e0-d9b3-11ed-b0e6-e361adffce0b/cowlogs-prod-2023.08.16?id=jJxY_okB9jDEhm2yzsKc)
[ok log](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/c0e240e0-d9b3-11ed-b0e6-e361adffce0b/cowlogs-prod-2023.08.16?id=8mNa_okBgmOFBpCY1Vn0)

sunce86:
Yes but we also have `new price estimate` log that logs every estimate so I was wondering do we need ever to know the best error solution.

MartinquaXD:
> Yes but we also have new price estimate log that logs every estimate so I was wondering do we need ever to know the best error solution.

It's not so much that we want to know which error was the best but rather having a single convenient log to query in order to know exactly who provided the quote for an order. Either we know the actual winner (if `result.is_ok()`) or we know that none of them provided a quote if (`result.is_err()`). Since we already have all the info in a single log I didn't see the need to add another log for that.

PR #1770, opened 2023-08-16
Make paraswap work as a price estimator

MartinquaXD:
Paraswap so far could not be used as a price estimator. For some reason the paraswap API requires us to send the decimals of the tokens we want to swap.
This information was currently only available in `/solve` requests but not in `/quote` requests. I took the `token_info` components from the `shared` crate an integrated it into the `driver's` `infra` and `boundary` modules (as it seems to be the practice for shared code).

Note that we currently fetch the token info for `/solve` requests in the `autopilot` but IMO this should also be moved to the `driver`. Will probably do that in follow up PR.

Additionally this test also showed that the `partner` id is actually no longer optional so I adjusted the examples and `config` parsing accordingly.
```
2023-08-16T13:27:36.111Z  WARN request{id="1"}:/solve:solve{order=0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000}: solvers::domain::solver::dex: failed to get swap err=Api("Validation failed: \"partner\" must be a string")
```

And finally the `paraswap` API also complained about us using `0x00...001` address so I replaced it with the prod solver address in the example file.
```
2023-08-16T13:24:41.682Z  WARN request{id="0"}:/solve:solve{order=0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000}: solvers::domain::solver::dex: failed to get swap err=Api("Validation failed: \"userAddress\" must be a valid address")
```

### Test Plan
Manual set up the system locally verified that I was able to get a response for a `/native_price` request.


nlordell:
> Note that we currently fetch the token info for `/solve` requests in the autopilot but IMO this should also be moved to the `driver`. Will probably do that in follow up PR.

Please do! This is an annoying piece of tech-debt that would be nice to nip in the bud.

crates/driver/src/boundary/mod.rs
```diff
@@ -26,6 +26,7 @@ pub mod liquidity;
 pub mod mempool;
 pub mod quote;
 pub mod settlement;
+pub mod token_info;
```
nlordell, line 29:
```suggestion
pub mod tokens;
```

I don't think `info` is particularly useful in the name.

crates/driver/src/infra/token_info/mod.rs
```diff
@@ -0,0 +1 @@
+pub use crate::boundary::token_info::Fetcher;
```
nlordell, line 1: To be consistent with the rest of the design, the `infra` code should operate on `domain` types and not `boundary` types (i.e. `TokenInfo`).

As such, the `infra::Fetcher` should probably return a `domain::competition::Tokens` and not a `boundary::token_info::TokenInfo`. Also, I think the `boundary` model should serve as an adapter between legacy code, and the rest of the driver code.

Personally, I would port the token fetching to the new driver architecture instead of using a `boundary` module since it is small enough (and then we can delete the "token info fetching" from `shared`, `orderbook` and `autopilot` once we fully co-locate).

MartinquaXD: I ended up implementing the `Fetcher` in `infra` directly so we don't use more `shared` code.
Things to note:
* did not use a trait
* used `RwLock` instead of `tokio::Mutex` to not block while making requests to the node
* dropped the whole `BatchCall` setup since those calls should be batched under the hood by your fancy `Web3` extension (although I believe that this still needs to be used in the driver)

crates/driver/src/infra/tokens.rs
```diff
@@ -0,0 +1,88 @@
...
+            .await?;
+            Ok((*address, Metadata { decimals, symbol }))
+        });
+        futures::future::join_all(futures).await
+    }
+
+    /// Returns the `Metadata` for the given tokens. Note that the result will
+    /// not contain data for tokens that encountered errors while fetching
+    /// the data.
+    pub async fn get_token_infos(
```
nlordell, line 55: naming nit: I think that `token_info.get_token_infos()` is quite redundant in terms of naming. I would just call this method `fetch` or something, as then `token_info.fetch()` or `tokens.fetch` is, IMO, more descriptive.

crates/driver/src/infra/tokens.rs
```diff
@@ -0,0 +1,88 @@
...
+        collections::HashMap,
+        sync::{Arc, RwLock},
+    },
+};
+
+#[derive(Clone, Debug, Default)]
+pub struct Metadata {
+    pub decimals: u8,
+    pub symbol: String,
+}
```
nlordell, line 16: Might make sense to move this into the `domain` module next to the `auction::Tokens` type, as this is core to the domain logic (i.e. it gets included with the tokens in the augmented auction).

crates/driver/src/domain/quote.rs
```diff
@@ -97,11 +95,19 @@ impl Order {
         )
     }
 
-    fn fake_auction(
+    async fn fake_auction(
         &self,
         gas_price: eth::GasPrice,
         weth: eth::WethAddress,
+        token_info: &infra::tokens::Fetcher,
```
nlordell, line 102: naming nit: I personally slightly prefer `tokens` (which was used in `Api::tokens` for example). I don't feel too strongly about the naming, but would more strongly suggest that the naming should consistent throughout the codebase.

crates/driver/src/infra/tokens.rs
```diff
@@ -0,0 +1,88 @@
...
+        addresses: &[eth::TokenAddress],
+    ) -> Vec<Result<(eth::TokenAddress, Metadata)>> {
+        let futures = addresses.iter().map(|address| async {
+            let erc20 = ERC20::at(&self.web3, address.0 .0);
+            // Use `try_join` because these calls get batched under the hood
+            // so if one of them fails the other will as well.
+            // Also this way we won't get incomplete data for a token.
+            let (decimals, symbol) = futures::future::try_join(
+                erc20.methods().decimals().call(),
+                erc20.methods().symbol().call(),
```
nlordell, line 44: We didn't do this before - but I think that we should consider fetching balances as well here. Specifically, the existing token buffers allow solvers to make additional optimizations or shortcuts that might be useful to know at quoting time.

I **don't think it should be part of this PR** - just wanted to point it out as food for thought.

nlordell (approved):
Code looks good to me - mostly smaller nits.

One thing that I don't like and feels like a code-smell is that the tokens are only fetched for quoting and not solving. It would be really nice to clean that up in a follow up PR so that we don't have alternative code paths for solving and quoting - which led to this discrepancy in the first place (noting that you already mentioned you planned on doing that - just reiterating that I think it is important).

One additional comment out that - I don't think that the current `Auction` abstraction fits super nicely in the driver anymore - specifically, it is kind of two concepts at the moment - it is partially the auction input that we get from the protocol and partially the "augumented" auction that we send to the solvers. Maybe it makes sense to have two separate `auction` types - one to model the input to `/solve` and another to model the data that will eventually reach the solver (so `auction` + liquidity + token data + other things we may add in the future).

crates/driver/src/infra/token_info/mod.rs
nlordell: > did not use a trait

Good - I don't think we should have mocking. That being said - this should be reflected in an E2E test IMO.

> used `RwLock` instead of `tokio::Mutex` to not block while making requests to the node

Sounds reasonable. You may double-fetch some tokens this way, but I think that is worth not delaying the fetch too long (which makes more sense in the context of using a `driver` for quoting as well).

> dropped the whole `BatchCall` setup

Nice.

> exposed `web3` type from Blockchain

Not a fan of this, left a comment on the approach I would use (spoiler - I think it should use `eth.contract_at::<ERC20>(token_address)`).

crates/driver/src/infra/tokens.rs
MartinquaXD: > Specifically, the existing token buffers allow solvers to make additional optimizations or shortcuts that might be useful to know at quoting time.

I thought we explicitly wanted to forbid internalisation while quoting. Or do you have other optimisations in mind?

crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -51,6 +51,10 @@ impl Ethereum {
         })
     }
 
+    pub fn web3(&self) -> &DynWeb3 {
+        &self.web3
+    }
+
```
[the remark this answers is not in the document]
MartinquaXD: [e130639](https://github.com/cowprotocol/services/pull/1770/commits/e13063914e98dec16c6d3d14c7e92f8b6b0bbf0b)

crates/driver/src/domain/quote.rs
MartinquaXD: [3f5bda9](https://github.com/cowprotocol/services/pull/1770/commits/3f5bda9081801437c2320c2d89c0b41e3636c81b)

crates/driver/src/infra/tokens.rs
MartinquaXD: [a473160](https://github.com/cowprotocol/services/pull/1770/commits/a473160f0116036981fd94dbfe37532e46fc6cd8)

PR #1772, opened 2023-08-16
Move token info fetching entirely into driver

MartinquaXD:
Follow up to https://github.com/cowprotocol/services/pull/1770

As discussed [here](https://github.com/cowprotocol/services/pull/1770#pullrequestreview-1581196291) it's more correct to move all the fetching of all token metadata that only enables certain optimisations but are not part of the core protocol into the driver.

### Test Plan
e2e tests

### DIFF
--- a/crates/driver/src/infra/blockchain/mod.rs
+++ b/crates/driver/src/infra/blockchain/mod.rs
@@ -171,8 +171,8 @@ impl Ethereum {
         erc20.methods().symbol().call().await.map_err(Into::into)
     }
 
-    /// Returns the current [`eth::TokenAmount`] balance of the specified account for
-    /// a given token.
+    /// Returns the current [`eth::TokenAmount`] balance of the specified
+    /// account for a given token.
     pub async fn erc20_balance(
         &self,
         holder: eth::Address,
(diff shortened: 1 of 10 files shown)


crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -161,14 +161,51 @@ impl Ethereum {
             .map_err(Into::into)
     }
 
-    pub async fn decimals(&self, address: eth::TokenAddress) -> Result<u8, Error> {
-        let erc20 = self.contract_at::<contracts::ERC20>(address.0);
-        erc20.methods().decimals().call().await.map_err(Into::into)
+    /// Returns the token's decimals. Returns `None` if the token does not
+    /// implement this optional method.
+    pub async fn decimals(&self, token: eth::TokenAddress) -> Result<Option<u8>, Error> {
+        let erc20 = self.contract_at::<contracts::ERC20>(token.0);
```
nlordell, line 167: What I meant in my other comment was to use `contract_at` in the `tokens::Fetcher` code, not necessarily to add new ERC-20 specific methods to the `Ethereum` type (which feels a bit weird, as you pointed out before).

nlordell (approved):

crates/driver/src/infra/blockchain/mod.rs
MartinquaXD: Ahh, okay. Misunderstood you.
Fixed in: [d2bbf31](https://github.com/cowprotocol/services/pull/1772/commits/d2bbf312faeab8a43cb562607bbf219f5f9c8152)

PR #1775, opened 2023-08-18
Remove number-conversion dependency

MartinquaXD:
The driver should rely less and less on shared crates.
This PR removes the dependency on the `number-conversion` crate.
Also instead of having free functions for everything I moved the conversion logic into the `U256Ext` trait to make things a bit nicer to read (hopefully).

`conv::u256` and the remaining code relied on 3 functions that were still missing in the driver crate. These have been copied from `number-conversions`.

### Test Plan
CI, e2e tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1645,7 +1645,6 @@ dependencies = [
  "mockall",
  "model",
  "num",
- "number-conversions",
  "primitive-types",
  "prometheus",
  "prometheus-metric-storage",
--- a/crates/driver/Cargo.toml
+++ b/crates/driver/Cargo.toml
@@ -59,7 +59,6 @@ contracts = { path = "../contracts" }
 ethcontract = { workspace = true }
 gas-estimation = { workspace = true }
 model = { path = "../model" }
-number-conversions = { path = "../number-conversions" }
 shared = { path = "../shared" }
 solver = { path = "../solver" }
 tracing = { workspace = true }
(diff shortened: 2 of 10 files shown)


MartinquaXD:
Alternatively one could consider moving all the logic into `number-conversion` and not rely on `driver::util::conv::u256`. Conversions are not unique to each binary and `eth::U256` is simply a re-export of `primitive-types::U256` so that would also not be an issue.
Overall I think we might be at a point where it makes sense to start thinking about how the final dependency graph is supposed to look like. Where does it make sense to share code and where should it actually be ported or copied over?

nlordell (approved):

nlordell:
I think the ultimate vision is to have this in a separate repo - that being said I don't see this happening soon so it may make sense to setup a `number-conversions` crate for this specifically. That being said, my "C background" is just telling me - it is not worth setting up an additional crate just to share these small functions, but I may just be plain wrong about this.

> final dependency graph is supposed to look like

I _really_ think we need to get rid of `shared` - it became a "dumping ground" for all the code that we wanted to share between our different binaries, and, in retrospect, I think this lead to some bad design decisions in the past (specifically, I think the way price estimation works is a bit wonky - we have a `baseline_solver` crate that is generic on AMM types to deal with different AMM types between the `solver` and `orderbook` binaries, instead of the "baseline" solver service that we have today, which is nicer IMO).

PR #1776, opened 2023-08-18
Per solver liquidity

MartinquaXD:
External solvers already complained that they would rather not receive any liquidity in their `/solve` requests.
This PR makes it so that we can control that behavior on a per-solver bases in the driver's `config.toml` file.

I preferred this approach because this distinction only makes sense if you have a single driver serving multiple solvers. We will most likely be the only ones doing that and our solvers don't support only a subset of liquidity so this simple flag is sufficient.

### Test Plan
Manual local test

### DIFF
--- a/crates/driver/example.toml
+++ b/crates/driver/example.toml
@@ -41,13 +41,6 @@ base-tokens = [
 # router = "0xb9960d9bca016e9748be75dd52f02188b9d0829f"
 # pool-code = "0xd306a548755b9295ee49cc729e13ca4a45e00199bbd890fa146da43a50571776"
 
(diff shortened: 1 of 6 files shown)


MartinquaXD:
Alternatively we could set up 2 drivers in our infra (1 with liquidity and one without) but that seems suboptimal.

nlordell:
> Alternatively we could set up 2 drivers in our infra (1 with liquidity and one without) but that seems suboptimal.

This is how we had imagined it previously. I would personally slightly prefer this approach as we would have a clear separation of "drivers for external solvers" and "drivers for internal solvers".

PR #1777, opened 2023-08-18
Pretty format gas price

sunce86:
Pretty format gas price.
Right now it is very hard to visually compare gas prices in logs. It's hard to notice a difference between 30000000000 (30gwei) and 300000000000 (300 gwei)

Dependent on https://github.com/cowprotocol/gas-estimation/pull/11.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -492,7 +492,7 @@ impl<'a> Submitter<'a> {
             }
 
             tracing::debug!(
-                "creating transaction with gas price (gas_price {}, gas estimate {}",
+                "creating transaction with gas price {}, gas estimate {}",
                 gas_price,
                 params.gas_estimate,
             );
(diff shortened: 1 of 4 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -481,22 +481,19 @@ impl<'a> Submitter<'a> {
                     || gas_price.max_fee_per_gas < replacement_price.max_fee_per_gas
                 {
                     tracing::debug!(
-                        ?gas_price,
-                        ?replacement_price,
                         sleep = ?params.retry_interval,
-                        "keep waiting for gas price to increase enough"
+                        "keep waiting for gas price to increase enough, gas price={}, replacement price={}",
+                        gas_price,
+                        replacement_price,
```
nlordell, line 487: nit, I would still use `tracing` module here. You can use `%` instead of `?` to format with `Display` instead of `Debug`.

```rust
                    tracing::debug!(
                        %gas_price,
                        %replacement_price,
                        sleep = ?params.retry_interval,
                        "keep waiting for gas price to increase enough"
                        "keep waiting for gas price to increase enough",
```

sunce86: Fixed.

nlordell (approved):

PR #1778, opened 2023-08-18
Refactor ERC-20 Interactions in Driver

nlordell:
This PR just proposes a small refactor in the `driver` to make code that uses ERC-20s easier. I also flattened the `Spender` type, since having the token tied to the spender felt a bit awkward.

It introduces a `blockchain::token::Erc20` type for fetching ERC-20 token information.

### Test Plan

Rust CI - tests continue to pass.

### DIFF
--- a/crates/driver/src/infra/blockchain/token.rs
+++ b/crates/driver/src/infra/blockchain/token.rs
@@ -1,0 +1,85 @@
+use {super::Error, crate::domain::eth};
+
+/// An ERC-20 token.
+///
+/// https://eips.ethereum.org/EIPS/eip-20
+pub struct Erc20 {
+    contract: contracts::ERC20,
+}
+
+impl Erc20 {
+    pub(super) fn new(contract: contracts::ERC20) -> Self {
+        Self { contract }
+    }
+
+    /// Returns the [`eth::TokenAddress`] of the ERC20.
+    pub fn address(&self) -> eth::TokenAddress {
+        self.contract.address().into()
+    }
+
+    /// Fetch the ERC20 allowance for the spender. See the allowance method in
+    /// EIP-20.
+    ///
+    /// https://eips.ethereum.org/EIPS/eip-20#methods
+    pub async fn allowance(
+        &self,
+        owner: eth::Address,
+        spender: eth::Address,
+    ) -> Result<eth::allowance::Existing, Error> {
+        let amount = self.contract.allowance(owner.0, spender.0).call().await?;
+        Ok(eth::Allowance {
+            token: self.contract.address().into(),
+            spender,
+            amount,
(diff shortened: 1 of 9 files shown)


crates/driver/src/infra/blockchain/token.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// Returns the [`eth::TokenAddress`] of the ERC20.
+    pub fn address(&self) -> eth::TokenAddress {
+        self.contract.address().into()
+    }
+
+    /// Fetch the ERC20 allowance for the spender. See the allowance method in
+    /// EIP-20.
+    ///
+    /// https://eips.ethereum.org/EIPS/eip-20#methods
```
MartinquaXD, line 23: I guess the links above the function implementations could either be dropped or modified to link to the actual function on the website.

crates/driver/src/infra/blockchain/token.rs
```diff
@@ -0,0 +1,85 @@
+use {super::Error, crate::domain::eth};
+
+/// An ERC-20 token.
+///
+/// https://eips.ethereum.org/EIPS/eip-20
+pub struct Erc20 {
+    contract: contracts::ERC20,
```
MartinquaXD, line 7: Nice component. 👍

MartinquaXD (approved):

crates/driver/src/infra/blockchain/token.rs
nlordell: They point to the `methods` section (Which includes all ERC-20 contract method documentation). It was mostly inherited from the `allowance` function documentation.

MartinquaXD: Yes. But it's also possible to link to the exact method that is getting implemented.

nlordell: 🤯

fleupold (approved):
Looks good, but I don't see where you limit the allowance to the requested one in the code (was expecting that given the PR title)

nlordell:
> but I don't see where you limit the allowance to the requested one in the code (was expecting that given the PR title)

Yes, sorry - I forgot to change the title. I was looking through the `driver` code and noticed we do `approve(spender, type(uint256).max)`. For some reason, I (mis)remembered that we had decided to change this behaviour in the `solver` crate - but it turns out that we didn't actually change it - so I ended up leaving the refactor I had and removed the change to set the exact approval that the solvers requested.

I updated the PR title.

crates/driver/src/infra/blockchain/token.rs
nlordell: Fixed.

PR #1781, opened 2023-08-18
Refactor Driver Configuration Initialization

nlordell:
This PR does a refactor for a code smell that was irking me for a while.

The `config::file` module contains _private_ types used just for parsing TOML configurations for the driver and turning them into their `infra::${component}::Config` counterparts. This separation follows a pattern we have in the `driver` code where we separate DTOs and data models from the actual types used within the codebase (it adds boilerplate, but I am in general fond of this).

There were 2 exceptions to the above rule: contract addresses and liquidity configuration. This is because the default configuration values for these are **network dependent**. This PR changes configuration file loading to be network aware, and makes all of the `config::file` module types private (as they should be IMO). For this, w

As a bonus on top, we check that the Ethereum RPC connection's network matches the configuration file (this helps prevent misconfiguration, our other processes make use of this technique as well).

### Test Plan

CI continues to pass 🎉.

### DIFF
--- a/crates/driver/src/infra/blockchain/mod.rs
+++ b/crates/driver/src/infra/blockchain/mod.rs
@@ -27,9 +27,9 @@ pub struct Network {
 }
 
 impl Rpc {
-    /// Instantiate an RPC connection to an Ethereum (or Ethereum-compatible)
-    /// node at the specifed URL.
-    pub async fn connect(url: &url::Url) -> Result<Self, Error> {
+    /// Instantiate an RPC client to an Ethereum (or Ethereum-compatible) node
+    /// at the specifed URL.
+    pub async fn new(url: &url::Url) -> Result<Self, Error> {
         let web3 = boundary::buffered_web3_client(url);
         let id = web3.net().version().await?.into();
         let chain = web3.eth().chain_id().await?.into();
--- a/crates/driver/src/infra/blockchain/mod.rs
+++ b/crates/driver/src/infra/blockchain/mod.rs
@@ -13,24 +13,53 @@ pub mod token;
 
 pub use self::contracts::Contracts;
 
+/// An Ethereum RPC connection.
+pub struct Rpc {
+    web3: DynWeb3,
+    network: Network,
+}
+
+/// Network information for an Ethereum blockchain connection.
+#[derive(Clone, Debug)]
+pub struct Network {
+    pub id: eth::NetworkId,
+    pub chain: eth::ChainId,
+}
+
+impl Rpc {
+    /// Instantiate an RPC connection to an Ethereum (or Ethereum-compatible)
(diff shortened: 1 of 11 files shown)


crates/driver/src/infra/config/file/load.rs
```diff
@@ -44,7 +55,129 @@ pub async fn load(path: &Path) -> infra::Config {
             }
         }))
         .await,
-        liquidity: config.liquidity,
+        liquidity: liquidity::Config {
```
nlordell, line 58: this code just moved here from `run.rs` with `s/eth.network_id()/&network.id/g`.

crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -12,24 +12,53 @@ pub mod contracts;
...
+#[derive(Clone, Debug)]
+pub struct Network {
+    pub id: eth::NetworkId,
+    pub chain: eth::ChainId,
+}
+
+impl Rpc {
+    /// Instantiate an RPC connection to an Ethereum (or Ethereum-compatible)
+    /// node at the specifed URL.
+    pub async fn connect(url: &url::Url) -> Result<Self, Error> {
```
MartinquaXD, line 31: Why not call this `new()`? It's not like calling this function will open a connection so the name seems misleading.

crates/driver/src/infra/blockchain/mod.rs
```diff
@@ -12,24 +12,53 @@ pub mod contracts;
...
 impl Ethereum {
-    /// Access the Ethereum blockchain through an RPC API hosted at the given
-    /// URL.
-    pub async fn ethrpc(url: &url::Url, addresses: contracts::Addresses) -> Result<Self, Error> {
-        let web3 = boundary::buffered_web3_client(url);
-        let chain_id = web3.eth().chain_id().await?.into();
-        let network_id = web3.net().version().await?.into();
-        let contracts = Contracts::new(&web3, &network_id, addresses);
+    /// Access the Ethereum blockchain through an RPC API.
+    pub async fn new(rpc: Rpc, addresses: contracts::Addresses) -> Result<Self, Error> {
```
MartinquaXD, line 59: side note: Since this type is essential for running the driver we might as well panic on an error here and make error handling `Etherum::new()` obsolete.

MartinquaXD (approved):
Nice, always a good sign when you can suddenly make so many things private. 🧹

fleupold (approved):

crates/driver/src/infra/blockchain/mod.rs
nlordell: It does though - specifically it uses the connection to fetch and cache the chain ID and network ID.

PR #1783, opened 2023-08-19
Multiple Balancer Weighted Pool Versions

### LINKED ISSUE #1763: Incorrect Input/Output Computation for Balancer Weighted Pools
Balancer weighted pools V3+ have slightly different math functions which causes the input and output computations to not match what actually happens on-chain. The deviation is small but consistent.

Specifically the implementation of the `powUp` math function changed from this:

```solidity
    /**
     * @dev Returns x^y, assuming both are fixed point numbers, rounding up. The result is guaranteed to not be below
     * the true value (that is, the error function expected - actual is always negative).
     */
    function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
        uint256 raw = LogExpMath.pow(x, y);
        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);

        return add(raw, maxError);
    }
```

To this in V3+:

```solidity
    /**
     * @dev Returns x^y, assuming both are fixed point numbers, rounding up. The result is guaranteed to not be below
     * the true value (that is, the error function expected - actual is always negative).
     */
    function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
        // Optimize for when y equals 1.0, 2.0 or 4.0, as those are very simple to implement and occur often in 50/50
        // and 80/20 Weighted Pools
        if (y == ONE) {
            return x;
        } else if (y == TWO) {
            return mulUp(x, x);
        } else if (y == FOUR) {
            uint256 square = mulUp(x, x);
            return mulUp(square, square);
        } else {
            uint256 raw = LogExpMath.pow(x, y);
            uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);

            return add(raw, maxError);
        }
    }
```

This means that the Rust implementation of `get_amount_in` and `get_amount_out` for 50/50 and 80/20 pools (which are very common) will yield different input/output amounts than what is observed on-chain.

The fix is to support different Math functions for different weighted product pools.

nlordell:
Fixes #1763 

Currently, there are 4 versions of weighted pools:
* **original** (called `v0` in code): The original weighted pool implementation with the original weighted pool math. This is what was already implemented in Rust.
* **v2**: This version uses the same math as the original, but was deprecated because of a security issue (in fact, we do not index these pools at all).
* **v3**: This version introduces a few new features. Notably, they modify the `FixedPoint::pow{Up,Down}` functions to early exit on specific inputs (for example, `FixedPoint::powUp(a, 1.0)` will just return `a`, where previously it would do math to compute the exponentiation which resulted in a small error).
* **v4**: Uses `CREATE2` for pool creation, the math is identical to V3.

As you can see, between V0 and V3+, the math for computing input and output amounts changed. This PR adds a "version" field to Balancer weighted pools that models the version of the weighted pool implementations above. This version is used to decide which weighted math calculation version should be used for the pool.

### Test Plan

Added new E2E tests that verify input and output amount computations for Balancer weighted pools V3.

#### Verifying Test Case Values

<details><summary><code>tests::baseline::bal_liquidity::weighted</code></summary>

Command:
```sh
python3 -c "print($(curl -s https://mainnet.infura.io/v3/$INFURA_PROJECT_ID -X POST -H 'content-type: application/json' --data '@-' <<JSON | jq -r '.result'
{
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [{
    "from": "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
    "to": "0x92762b42a06dcdddc5b7362cfb01e631c4d44b40",
    "data": "0x9d2c110c000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000026272545808bbac15450000000000000000000000000000000000000000000f85779b802c3ee5b6afc700000000000000000000000000000000000000000000000000000000000000000000000000000000000000006810e776880c02933d47db1b9fc05908e5386b96000000000000000000000000def1ca1fb7fbcdc777520aa7f396b4e015f497ab0000000000000000000000000000000000000000000000000de0b6b3a764000092762b42a06dcdddc5b7362cfb01e631c4d44b40000200000000000000000182000000000000000000000000000000000000000000000000000000000111d8310000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab410000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab4100000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000000"
  }, "latest"],
  "id": 0
}
JSON
))"
```

Output:
```
1657855325872947866705
```

</details>

<details><summary><code>tests::baseline::bal_liquidity::weighted_v3plus</code></summary>

Command:
```sh
python3 -c "print($(curl -s https://rpc.gnosischain.com/ -X POST -H 'content-type: application/json' --data '@-' <<JSON | jq -r '.result'
{
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [{
    "from": "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
    "to": "0x21d4c792ea7e38e0d0819c2011a2b1cb7252bd99",
    "data": "0x9d2c110c00000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000a4fd79613c2f4bd8b50000000000000000000000000000000000000000000435c584802d92015299af00000000000000000000000000000000000000000000000000000000000000000000000000000000000000009c58bacc331c9aa871afd802db6379a98e80cedb000000000000000000000000177127622c4a00f3d409b75571e12cb3c8973d3c0000000000000000000000000000000000000000000000000de0b6b3a764000021d4c792ea7e38e0d0819c2011a2b1cb7252bd9900020000000000000000001e0000000000000000000000000000000000000000000000000000000001c2a9e10000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab410000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab4100000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000000"
  }, "latest"],
  "id": 0
}
JSON
))"
```

Output:
```
1663373703594405548696
```

</details>

<details><summary><code>tests::baseline::buy_order_rounding::balancer_weighted_v3plus</code></summary>

Command:
```sh
python3 -c "print($(curl -s https://rpc.gnosischain.com/ -X POST -H 'content-type: application/json' --data '@-' <<JSON | jq -r '.result'
{
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [{
    "from": "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
    "to": "0x21d4c792ea7e38e0d0819c2011a2b1cb7252bd99",
    "data": "0x9d2c110c000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000026272545808bbac15450000000000000000000000000000000000000000000f85779b802c3ee5b6afc700000000000000000000000000000000000000000000000000000000000000010000000000000000000000009c58bacc331c9aa871afd802db6379a98e80cedb000000000000000000000000177127622c4a00f3d409b75571e12cb3c8973d3c00000000000000000000000000000000000000000000003635c9adc5dea0000021d4c792ea7e38e0d0819c2011a2b1cb7252bd9900020000000000000000001e0000000000000000000000000000000000000000000000000000000001c2a9e10000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab410000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab4100000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000000"
  }, "latest"],
  "id": 0
}
JSON
))"
```

Output:
```
603167793526702182
```

</details>

### DIFF
--- a/crates/shared/src/sources/balancer_v2/swap/weighted_math.rs
+++ b/crates/shared/src/sources/balancer_v2/swap/weighted_math.rs
@@ -77,6 +77,50 @@ pub fn calc_in_given_out(
     balance_in.mul_up(ratio)
 }
 
+pub fn calc_out_given_in_v3(
+    balance_in: Bfp,
+    weight_in: Bfp,
+    balance_out: Bfp,
+    weight_out: Bfp,
+    amount_in: Bfp,
+) -> Result<Bfp, Error> {
+    if amount_in > balance_in.mul_down(*MAX_IN_RATIO)? {
+        return Err(Error::MaxInRatio);
+    }
+
+    let denominator = balance_in.add(amount_in).expect("See proof above.");
+    let base = balance_in.div_up(denominator)?;
+    let exponent = weight_in.div_down(weight_out)?;
+    let power = base.pow_up_v3(exponent)?;
+
+    balance_out.mul_down(power.complement())
+}
+
+pub fn calc_in_given_out_v3(
+    balance_in: Bfp,
+    weight_in: Bfp,
+    balance_out: Bfp,
+    weight_out: Bfp,
+    amount_out: Bfp,
+) -> Result<Bfp, Error> {
+    if amount_out > balance_out.mul_down(*MAX_OUT_RATIO)? {
+        return Err(Error::MaxOutRatio);
+    }
+
+    let denominator = balance_out.sub(amount_out).expect(
+        "if amount_out > balance_out >= balance_out.mul_down(*MAX_OUT_RATIO) contradicting above \
+         inequality",
+    );
+    let base = balance_out.div_up(denominator)?;
+    let exponent = weight_out.div_up(weight_in)?;
+    let power = base.pow_up_v3(exponent)?;
+    let ratio = power
+        .sub(Bfp::one())
+        .expect("power=0 => base=0 => balance_out=0 => amount_out=0 => denominator div by zero");
+
+    balance_in.mul_up(ratio)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/driver/src/boundary/liquidity/balancer/v2/mod.rs
+++ b/crates/driver/src/boundary/liquidity/balancer/v2/mod.rs
@@ -13,6 +13,7 @@ use {
         BalancerV2StablePoolFactory,
         BalancerV2Vault,
         BalancerV2WeightedPoolFactory,
+        BalancerV2WeightedPoolFactoryV3,
         GPv2Settlement,
     },
     shared::{
@@ -112,6 +113,18 @@ async fn init_liquidity(
                     (
                         BalancerFactoryKind::Weighted,
                         BalancerV2WeightedPoolFactory::at(&web3, factory.into())
+                            .raw_instance()
+                            .clone(),
+                    )
+                })
+                .collect::<Vec<_>>(),
+            config
+                .weighted_v3plus
+                .iter()
+                .map(|&factory| {
+                    (
+                        BalancerFactoryKind::WeightedV3,
+                        BalancerV2WeightedPoolFactoryV3::at(&web3, factory.into())
                             .raw_instance()
                             .clone(),
                     )
--- a/crates/driver/src/boundary/liquidity/balancer/v2/weighted.rs
+++ b/crates/driver/src/boundary/liquidity/balancer/v2/weighted.rs
@@ -6,6 +6,7 @@ use {
             liquidity::{self, balancer},
         },
     },
+    shared::sources::balancer_v2::pool_fetching::WeightedPoolVersion,
     solver::liquidity::{balancer_v2, WeightedProductOrder},
 };
 
@@ -38,6 +39,10 @@ pub fn to_domain(id: liquidity::Id, pool: WeightedProductOrder) -> Result<liquid
                     .collect::<Result<_>>()?,
             )?,
             fee: pool.fee.as_uint256().into(),
+            version: match pool.version {
+                WeightedPoolVersion::V0 => balancer::v2::weighted::Version::V0,
+                WeightedPoolVersion::V3Plus => balancer::v2::weighted::Version::V3Plus,
+            },
         }),
     })
 }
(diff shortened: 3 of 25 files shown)


crates/shared/src/sources/balancer_v2/swap/weighted_math.rs
```diff
@@ -77,6 +77,50 @@ pub fn calc_in_given_out(
     balance_in.mul_up(ratio)
 }
 
+pub fn calc_out_given_in_v3(
```
MartinquaXD, line 80: Instead of duplicating these functions I think you could simply have 1 base version that takes the specific `pow_up` function pointer. And then `calc_out_given_in_v3()` would just be a wrapper around the base version with the new `pow_up` function. [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7848d04b7fe2113f105306393bdb1f99)

fleupold (approved):
Wow, really nice 🕵️ work. I guess having each pool be it's own type and implement its math directly would be too much of an architectural change, so leaving the free methods and branching based on the version field seems good.

nlordell:
> implement its math directly would be too much of an architectural change

I played around with this. The reason I didn't go this route is that it felt like all these weighted pools are just weighed pools (so the data is identical) with different versions (or different flavours). If they became fundamentally different, I would have preferred to setup separate liquidity types to represent each.

crates/shared/src/sources/balancer_v2/swap/weighted_math.rs
nlordell: I purposefully kept the functions separate and preferred to copy the small bit of code instead of passing in a function pointer. My rationale is that, from a conceptual level, the difference in how input/output amounts get computed is version-based and it is easy to imagine that it is not just the `pow_up` function that is different but other things (different rounding directions for other operations, etc.). The code that was kept common in `WeighedPool::get_amount_{in,out}` is what is implemented in the `BaseMinimalSwapInfoPool` implementation (so the base class that didn't change).

nlordell: We tried out the suggestion and spoke in person, and agree it is OK as is.

MartinquaXD (approved):

PR #1784, opened 2023-08-19
Print raw response in case of decoding error

sunce86:
Occasionally we get decoding error:

```
Caused by:
   0: method 'balanceOf(address):(uint256)' failure: web3 error: Decoder error: Error("EOF while parsing a list", line: 1, column: 555)
   1: web3 error: Decoder error: Error("EOF while parsing a list", line: 1, column: 555)
```

 probably due to not receiving complete data from the client node. This happens quite rarely.

This PR adds printing the raw body in case of decoding error, so we can observe the response and reproduce the error.

Related to alert: https://cowservices.slack.com/archives/C0371SB243E/p1692411771471219

### DIFF
--- a/crates/shared/src/ethrpc/http.rs
+++ b/crates/shared/src/ethrpc/http.rs
@@ -93,14 +93,21 @@ async fn execute_rpc<T: DeserializeOwned>(
     // Log the raw text before decoding to get more information on responses that
     // aren't valid json. Debug encoding so we don't get control characters like
     // newlines in the output.
-    tracing::trace!(name = %inner.name, %id, body = %text.trim(), "received response");
     if !status.is_success() {
         return Err(Web3Error::Transport(TransportError::Message(format!(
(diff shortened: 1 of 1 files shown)


crates/shared/src/ethrpc/http.rs
```diff
@@ -93,14 +93,21 @@ async fn execute_rpc<T: DeserializeOwned>(
     // Log the raw text before decoding to get more information on responses that
     // aren't valid json. Debug encoding so we don't get control characters like
     // newlines in the output.
-    tracing::trace!(name = %inner.name, %id, body = %text.trim(), "received response");
```
nlordell, line 96: Can we leave the `trace` log here? I think it is useful (I still make use of it when debugging locally for example). It lets you see all RPC requests/responses that the services are making.

sunce86: Ofc. reverted.

PR #1785, opened 2023-08-19
Revert UniV3

sunce86:
Reverts https://github.com/cowprotocol/services/pull/1736 but keep few nice [things](https://github.com/cowprotocol/services/pull/1785/commits/fd25e6b66e153501599665ac91f4243aed9b3920).

The reason being that we need list of ALL existing indexed pools and not only a subset (more context [here](https://cowservices.slack.com/archives/C0361CDD1FZ/p1692451971491739?thread_ts=1692439510.428959&cid=C0361CDD1FZ))

I hoped we could keep the change related to `totalValueLockedETH`, but when I tested it, the results were:
Before: number of pools `6474`
After: number of pools `2100` ordered by `totalValueLockedETH` where the first pool was [0x277667eb3e34f134adf870be9550e9f323d0dc24](https://etherscan.io/address/0x277667eb3e34f134adf870be9550e9f323d0dc24) so I don't trust graph to sort this properly.

### DIFF
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -129,7 +129,7 @@ impl PoolsCheckpointHandler {
     pub async fn new(
         chain_id: u64,
         client: Client,
-        max_pools_to_initialize_cache: u64,
+        max_pools_to_initialize_cache: usize,
     ) -> Result<Self> {
         let graph_api = UniV3SubgraphClient::for_chain(chain_id, client)?;
         let mut registered_pools = graph_api.get_registered_pools().await?;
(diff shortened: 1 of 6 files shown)


crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -147,10 +145,21 @@ impl PoolsCheckpointHandler {
...
+        // liquidity
+        registered_pools.pools.sort_unstable_by(|a, b| {
+            a.total_value_locked_eth
+                .partial_cmp(&b.total_value_locked_eth)
+                .unwrap()
+        });
         let pool_ids = registered_pools
             .pools
-            .iter()
+            .clone()
```
MartinquaXD, line 158: You don't have to clone the whole thing. You can do..iter().rev().map(value.clone())

MartinquaXD (approved):

fleupold (approved):
Checked that this indeed reverts #1736 (modulo the type and default size changes)

PR #1789, opened 2023-08-21
Handle deadline error for colocated driver

sunce86:
Related to the alert https://cowservices.slack.com/archives/C037PB929ME/p1692593972494599

When orderbook sends quote request to colocated driver (that we now use for quoting on staging) and the driver does not respond in a timely manner, `DeadlineExceeded` error is returned back.

For this particular error I think we should have a separate error type in orderbook api since it does not belong to any other type and it shouldn't be traced with `tracing::error` log level since drivers theoretically can spend as much time as they want.

### DIFF
--- a/crates/shared/src/api.rs
+++ b/crates/shared/src/api.rs
@@ -241,6 +241,10 @@ impl IntoWarpReply for PriceEstimationError {
                 error("NoLiquidity", "not enough liquidity"),
                 StatusCode::NOT_FOUND,
             ),
+            Self::DeadlineExceeded => with_status(
+                error("DeadlineExceeded", "quoting deadline exceeded"),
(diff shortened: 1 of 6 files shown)


MartinquaXD (approved):
This looks fine for now but I think we should refactor price estimation error reporting eventually. It shouldn't be the case that adding a new low level price estimation error will result in a new error type in the public API.
I think we should have a API level price estimation error which just consists of a few broad error categories (no liquidity, internal error, maybe bad request) and then convert these low level categories to one of those error variants to not leak too many details to the public API.

This is especially true when you consider that these error case will probably never be returned from the price competition. (a single `Ok` will result in all `Err` being ignored and even if you have only `Err` results only the most severe will be reported).

PR #1688, opened 2023-07-19
Get surplus_fee from solvers for FOK limit orders

sunce86:
Implements https://github.com/cowprotocol/services/issues/1657

We no longer use the protocol calculated `surplus_fee` for FOK limit orders, but we ask the solvers to provide the fee and use that fee to build the settlement.

This PR contains only the SWITCH to a new logic, while the complete cleanup of `surplus_fee` calculation and it's dependencies will be done in a follow up PR.

PR #1797, opened 2023-08-23
[DRAFT] Surplus_fee Calculation Cleanup

sunce86:
Follow-up PR for https://github.com/cowprotocol/services/pull/1688

This PR does a cleanup of `surplus_fee` calculation in `autopilot` and all dependent code.


crates/driver/src/tests/cases/mod.rs
```diff
@@ -39,8 +39,5 @@ pub const ETH_ORDER_AMOUNT: u128 = 40000000000000000000u128;
 pub const DEFAULT_SCORE_MIN: u128 = 2000000000000000000u128;
 pub const DEFAULT_SCORE_MAX: u128 = 500000000000000000000000000000u128;
 
-/// The default surplus fee for limit orders.
-pub const DEFAULT_SURPLUS_FEE: u128 = 0u128;
```
MartinquaXD, line 43: Do we already have a test that verifies that the `driver` is computing `surplus_fee`s dynamically?

database/sql/V056__remove_surplus_fee.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE orders 
+  DROP COLUMN surplus_fee,
+  DROP COLUMN surplus_fee_timestamp;
```
MartinquaXD, line 3: Please remember to update the `database/README.md`
Also I think having a small comment for each DB migration explaining why it's necessary is kind of nice.

MartinquaXD (approved):
Love it! 🧹

database/sql/V056__remove_surplus_fee.sql
sunce86: Added, thanks. I always forget to update :)

nlordell (approved):
YAY

database/sql/V056__remove_surplus_fee.sql
```diff
@@ -0,0 +1,8 @@
+-- These field are not needed anymore since they were used to calculate the 
+-- surplus fee for limit orders beforehand. Now, since the surplus fee 
+-- is provided by the solvers at the time of solving, we can remove these feilds 
+-- and rely on the order_execution table to get the executed surplus fee.
+
+ALTER TABLE orders 
+  DROP COLUMN surplus_fee,
+  DROP COLUMN surplus_fee_timestamp;
```
nlordell, line 8: ubernit: Missing newline at EOF.

crates/model/src/order.rs
```diff
@@ -915,9 +906,6 @@ impl OrderClass {
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Hash, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct LimitOrderClass {
-    #[serde_as(as = "Option<DecimalU256>")]
-    pub surplus_fee: Option<U256>,
-    pub surplus_fee_timestamp: Option<DateTime<Utc>>,
     #[serde_as(as = "Option<DecimalU256>")]
     pub executed_surplus_fee: Option<U256>,
```
nlordell, line 910: Not related to this PR, but would LOVE to see this property go as well (I think it requires different `Order` types for different situations - for example this is only needed by the `orderbook` API on `GET api/v1/orders/${uid}` to be able to display executed fee amounts, but not in the order used for the `GET api/v1/auction`).

database/sql/V056__remove_surplus_fee.sql
sunce86: Added. Btw, why do we need it?

nlordell: It's not needed per se, but it makes future diffs nicer. Adding code to the end of the file becomes a one line change instead of a two line change (one extra line change to add the `\n` to the end of the last line).

PR #1801, opened 2023-08-23
Baseline: Add Support for Stable Pools

nlordell:
This PR introduces stable pool support in the Baseline solver. The math already existed, and just minor tweaking was needed:

- Use `BTreeMap` for token states: Balancer V2 requires [sorted tokens](https://github.com/balancer/balancer-v2-monorepo/blob/227683919a7031615c0bc7f144666cdf3883d212/pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol#L37-L42) for the pools we implement - this ensures we have consistent rounding characteristics when computing invariants/converging (as there are divisions that happen based on token Balances in the order of the tokens).
- The Stable math uses a constant `AMP_PRECISION` (amplification parameter precision) despite our liquidity allowing for variable precision, so we fixed the `BaselineSolvable` implementation to normalize the `amplification_parameter` to this constant base.

Additionally, we **removed `StablePoolFactory` and only kept `StablePoolFactoryV2`**. Why? The original `StablePoolFactory` has been [deprecated](https://forum.balancer.fi/t/vulnerability-disclosure/3179) because of a security vulnerability, and liquidity has been migrated to `StablePoolV2`s or `ComposableStablePool`s. Additionally, the original `StablePool`s had [slightly different rounding](https://github.com/balancer/balancer-v2-monorepo/pull/1330) which led to small errors in our input/output amount computation. While they are technically still usable, they are mostly void of liquidity and (IMO) not worth the additional complexity to support the additional stable pool flavour (in a similar manner that we support multiple weighted pool flavours) for a deprecated pool type with a [vulnerability warning](https://app.balancer.fi/#/ethereum/pool/0x06df3b2bbb68adc8b0e302443692037ed9f91b42000000000000000000000063) in the Balancer UI.

In a follow up, I will add support for `ComposableStablePool`s (which additionally require some additional logic to filter out the BPT token from its token list and fetching amplification parameters).

### Test Plan

I verified the logic from the Balancer `StableMath.sol` contract and it lined up. Also added new baseline solver E2E tests that line up with amounts computed on-chain:

- [Simulation](https://dashboard.tenderly.co/shared/simulation/1f5ff18d-7388-4388-815e-830c1a6229e7) for the sell order in the `baseline::bal_liquidity::stable` test case.
- [Simulation](https://dashboard.tenderly.co/shared/simulation/9054058c-1622-4448-a44e-047fc6cef3ee) for the buy order in the `baseline::bal_liquidity::stable` test case.


crates/shared/src/sources/balancer_v2/pools/stable.rs
```diff
@@ -55,8 +55,8 @@ impl AmplificationParameter {
     }
 
     /// This is the format used to pass into smart contracts.
-    pub fn as_u256(&self) -> U256 {
-        self.factor * self.precision
+    pub fn with_base(&self, base: U256) -> Option<U256> {
+        Some(self.factor.checked_mul(base)?.checked_mul(base)? / self.precision)
```
fleupold, line 59: This math looks quite different to what we had before. Was this causing issues?

crates/solvers/src/api/mod.rs
```diff
@@ -59,16 +59,11 @@ async fn solve(
...
         tracing::trace!(?auction);
 
-        let solutions = state
-            .solve(auction)
-            .await
-            .into_iter()
-            .next()
-            .map(|solution| dto::Solutions::from_domain(&[solution]))
-            .unwrap_or_default();
+        let solutions = state.solve(auction).await;
```
fleupold, line 62: unrelated

crates/solvers/src/boundary/liquidity/stable.rs
```diff
@@ -0,0 +1,65 @@
...
+        *pool.amplification_parameter.denom(),
+    )
+    .ok()?;
+
+    Some(Pool {
+        common: CommonPoolState {
+            id,
+            address,
+            swap_fee,
+            paused: false,
```
fleupold, line 50: Is this guaranteed because we only fetch unpaused pools?

fleupold (approved):
I don't really grasp the concept of the amplification parameter, but given the unit test I'm somewhat optimistic this will work.

crates/solver/src/liquidity.rs
```diff
@@ -352,19 +352,16 @@ impl std::fmt::Debug for StablePoolOrder {
...
+pub fn token_pairs<T>(reserves: &BTreeMap<H160, T>) -> Vec<TokenPair> {
+    reserves
+        .keys()
+        .flat_map(|&token_a| {
+            reserves
+                .keys()
+                .filter_map(move |&token_b| match TokenPair::new(token_a, token_b) {
+                    Some(pair) if pair.get().0 == token_a => Some(pair),
+                    _ => None,
+                })
```
MartinquaXD, line 364: The pattern matching looks quite confusing.
```suggestion
            reserves.keys().filter_map(move |&token_b| {
                let pair = TokenPair::new(token_a, token_b)?;
                (pair.get().0 == token_a).then_some(pair)
            })

```

crates/solvers/src/boundary/liquidity/stable.rs
```diff
@@ -0,0 +1,65 @@
...
+        pool_fetching::{AmplificationParameter, CommonPoolState, TokenState},
+        swap::fixed_point::Bfp,
+    },
+    std::collections::BTreeMap,
+};
+
+/// Converts a domain pool into a [`shared`] Balancer V2 stable pool. Returns
+/// `None` if the domain pool cannot be represented as a boundary pool.
+pub fn to_boundary_pool(address: H160, pool: &liquidity::stable::Pool) -> Option<Pool> {
+    // NOTE: this is only used for encoding and not for solving, so it OK to
```
MartinquaXD, line 15:
```suggestion
    // NOTE: this is only used for encoding and not for solving, so it's OK to
```

crates/solvers/src/boundary/liquidity/stable.rs
```diff
@@ -0,0 +1,65 @@
...
+        .map(|reserve| {
+            Some((
+                reserve.asset.token.0,
+                TokenState {
+                    balance: reserve.asset.amount,
+                    scaling_exponent: reserve.scale.exponent(),
+                },
+            ))
+        })
+        .collect::<Option<BTreeMap<_, _>>>()?;
```
MartinquaXD, line 38:
```suggestion
        .map(|reserve| {
            (
                reserve.asset.token.0,
                TokenState {
                    balance: reserve.asset.amount,
                    scaling_exponent: reserve.scale.exponent(),
                },
            )
        })
        .collect();
```

crates/solvers/src/boundary/liquidity/stable.rs
```diff
@@ -0,0 +1,65 @@
...
+        amplification_parameter,
+    })
+}
+
+/// Converts a rational to a Balancer fixed point number.
+fn to_fixed_point(ratio: &eth::Rational) -> Option<Bfp> {
+    // Balancer "fixed point numbers" are in a weird decimal FP format (instead
+    // of a base 2 FP format you typically see). Just convert our ratio into
+    // this format.
+    let base = U256::from(1_000_000_000_000_000_000_u128);
```
MartinquaXD, line 62:
```suggestion
    let base = U256::exp10(18);
```

crates/solvers/src/boundary/liquidity/weighted_product.rs
```diff
@@ -38,7 +38,7 @@ pub fn to_boundary_pool(address: H160, pool: &liquidity::weighted_product::Pool)
                 },
             ))
         })
-        .collect::<Option<HashMap<_, _>>>()?;
+        .collect::<Option<BTreeMap<_, _>>>()?;
```
MartinquaXD, line 41: The compiler can deduce the type which means you don't have to import `BTreeMap`.
```suggestion
        .collect::<Option<_>>()?;
```

crates/solvers/src/api/mod.rs
nlordell: I already made a separate PR for this: https://github.com/cowprotocol/services/pull/1799

It should go away once that PR merges and I rebase (but needed to make the test pass).

crates/solvers/src/boundary/liquidity/stable.rs
nlordell: Yes - these are the pools that have already been fetched sent to the `baseline` solver. This "boundary" module is just mapping the pools that we get for solving into the `shared` crate types to make use of the shared code (even if we don't use all fields like this one). Additionally, the `shared::baseline_solver` code never actually checks this value.

crates/shared/src/sources/balancer_v2/pools/stable.rs
nlordell: > Was this causing issues?

Exactly, this was the main bug I found (see PR description). The "amplification parameter" is a rational value, the "actual value" of `1.23` could be represented as `{ factor: 123, precision: 100 }` or `{ factor: 1230000, precision: 1000000 }`. Note that, with the previous implementation, they would yield **different** `as_u256` representation, despite representing the same "actual value". This was a bug since the stable math uses a [constant precision value](https://github.com/balancer/balancer-v2-monorepo/blob/227683919a7031615c0bc7f144666cdf3883d212/pkg/pool-stable/contracts/StableMath.sol#L30) of `1000`, so we want to convert the amplification parameter from  
`{ factor: 123, precision: 100 }` to `{ factor: 1230, precision: 1000 }` and yield a U256 representation of `1230`.

> This math looks quite different to what we had before

Note that if `base == precision`, then the results would be the same 😄 - except for overflow semantics.

crates/solvers/src/boundary/liquidity/weighted_product.rs
nlordell: 7d31e635d98917da4d35aeea68dabc42df11e20a

crates/solvers/src/boundary/liquidity/stable.rs
nlordell: 7d31e635d98917da4d35aeea68dabc42df11e20a

crates/solvers/src/boundary/liquidity/stable.rs
nlordell: 7d31e635d98917da4d35aeea68dabc42df11e20a

crates/solver/src/liquidity.rs
nlordell: 7d31e635d98917da4d35aeea68dabc42df11e20a

MartinquaXD (approved):

crates/solvers/src/boundary/liquidity/stable.rs
nlordell: 7d31e635d98917da4d35aeea68dabc42df11e20a

crates/shared/src/sources/balancer_v2/swap.rs
```diff
@@ -573,7 +577,8 @@ mod tests {
         let tusd = H160::from_low_u64_be(3);
         let tokens = vec![dai, usdc, tusd];
         let scaling_exps = vec![0, 12, 12];
-        let amplification_parameter = AmplificationParameter::new(570.into(), 1000.into()).unwrap();
+        let amplification_parameter =
+            AmplificationParameter::new(570000.into(), 1000.into()).unwrap();
```
nlordell, line 581: THis value was just wrong... 
image

nlordell: The test passed because of the incorrect `as_uint256` implementation on `AmplificationParameter`

crates/shared/src/sources/balancer_v2/pools/stable.rs
nlordell: There was actually a booboo here - it has since been fixed.

PR #1802, opened 2023-08-24
Scaling Exponent To Factor

nlordell:
Prerequisite for composable stable pools.

Previously, our Balancer V2 indexing logic _assumed_ that all token's scaling factors were computed as `18 - token.decimals()`. While this is generally the case (for weighted and stable pools), composable stable pools use "rate converters" to adjust token scaling factors:

image

This PR changes the codebase to instead use `scaling_factors` directly instead of `scaling_exponents` everywhere for Balancer pools. While the PR is mostly just converting scaling exponents to their factor representation - there are a couple of logic changes that resulted from no longer having exponents available:

- The in amount convergence code now uses the approximate out amount deficit converted to in tokens instead of a constant 1e-12 tokens. We still multiply the bump by 10 on each iteration, so we converge quickly. I want to try @fleupold's idea of iterating `for i in 1..=18` and bumping by `U256::exp10(i)` (which would have more easily predictable behaviour, but first want to see how this affects the Baseline solver solving times before adding more iterations to the algorithm).
- We **fixed** an issue in the legacy solver adapter where we were sending scaling factors, when the legacy solver HTTP API expected what it calls "scaling rates", which is the inverse of a scaling factor.

### Test Plan

CI continues to pass (required adjustments to the exact Balancer amounts for the rounding tests).


crates/shared/src/sources/balancer_v2/swap.rs
```diff
@@ -267,18 +248,20 @@ fn converge_in_amount(
...
-    // bit. We start with some epsilon and multiply the amount to bump by 10 for
-    // each iteration. When swapping tokens with 18 decimals, this corresponds
-    // to bumping from [1e-12, 1e-6).
-    let mut bump = in_token_epsilon(in_exponent, out_exponent);
+    // bit. We start by approximating the out amount deficit to in tokens at the
+    // trading price and multiply the amount to bump by 10 for each iteration.
+    let mut bump = (exact_out_amount - out_amount)
+        .checked_mul(in_amount)?
+        .ceil_div(&out_amount.max(U256::one()))
+        .max(U256::one());
```
nlordell, line 263: This is the first logic change mentioned in the PR description.

crates/solver/src/settlement_simulation.rs
```diff
@@ -504,17 +504,17 @@ mod tests {
...
-                    scaling_exponent: 6
+                    balance: 50_716_887_827_666_u128.into(),
+                    scaling_factor: 1_000_000_000_000_u128.into(),
                 },
                 "0xdac17f958d2ee523a2206206994597c13d831ec7".parse().unwrap() => TokenState{
-                    balance: U256::from(38436050628181u128),
-                                        scaling_exponent: 6
-                                    },
+                    balance: 38_436_050_628_181_u128.into(),
+                    scaling_factor: 1_000_000_000_000_u128.into(),
```
nlordell, line 516: The scaling factors in this test were wrong (they were `decimals` instead of `18 - decimals`). Since they don't matter for the test, it didn't make any difference changing them, but I just wanted to fix up the values to be consistent with the actual pool.

Note the tokens are DAI, USDC, and USDT

crates/solvers/src/boundary/legacy.rs
```diff
@@ -231,7 +231,7 @@ fn to_boundary_auction(
                     scaling_rates: state
                         .reserves
                         .iter()
-                        .map(|reserve| (reserve.asset.token.0, reserve.scale.get()))
+                        .map(|reserve| (reserve.asset.token.0, to_scaling_rate(&reserve.scale)))
```
nlordell, line 234: Bug fix mentioned in the PR description.

crates/solver/src/settlement_simulation.rs
```diff
@@ -504,17 +507,17 @@ mod tests {
...
+                    balance: 50_716_887_827_666_u128.into(),
+                    scaling_factor: Bfp::exp10(12),
                 },
                 "0xdac17f958d2ee523a2206206994597c13d831ec7".parse().unwrap() => TokenState{
-                    balance: U256::from(38436050628181u128),
-                                        scaling_exponent: 6
-                                    },
+                    balance: 38_436_050_628_181_u128.into(),
+                    scaling_factor: Bfp::exp10(12),
+                },
```
nlordell, line 520: The scaling factors in this test were wrong (they were `decimals` instead of `18 - decimals`). Since they don't matter for the test, it didn't make any difference changing them, but I just wanted to fix up the values to be consistent with the actual pool.

Note the tokens are DAI, USDC, and USDT

crates/shared/src/sources/balancer_v2/swap/fixed_point.rs
```diff
@@ -114,6 +114,15 @@ impl Bfp {
         *ONE
     }
 
+    pub fn exp10(exp: i32) -> Self {
+        let exp = exp.saturating_add(18);
+        if exp < 0 {
+            return Self::zero();
```
MartinquaXD, line 120: Seems surprising to return `0` here. Is this intended?
If so I think a doc comment about this behavior would be nice so you get informed about it by your IDE when typing this function.

crates/solver/src/solver/baseline_solver.rs
```diff
@@ -829,7 +829,7 @@ mod tests {
         );
         assert_eq!(
             pool_1.get_amount_in(tokens[0], (1.into(), tokens[1])),
-            Some(8999613.into())
+            Some(15999226.into())
```
MartinquaXD, line 832: Is this change (and later ones) due to the change in the convergence function?

MartinquaXD (approved):
Looks good over all. Just a few nits.

crates/shared/src/sources/balancer_v2/pool_fetching/pool_storage.rs
```diff
@@ -269,7 +269,7 @@ mod tests {
                         id: H256([1; 32]),
                         address: H160([1; 20]),
                         tokens: vec![H160([0x11; 20]), H160([0x22; 20])],
-                        scaling_exponents: vec![0, 0],
+                        scaling_factors: vec![Bfp::exp10(0), Bfp::exp10(0)],
```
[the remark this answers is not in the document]
nlordell: I used `Bfp::exp10` on purpose here - specifically it is so the `0` numerical constants stay the same in the unit tests (proving that we didn't change anything).

For example, a `scaling_exponent` of 0 would turn into a scaling factor `10**0` which is `1`, while a `scaling_exponent` of 12 would turn into a scaling factor of `10**12` which is `1000000000000`. I find the former (i.e. writing `Bfp::exp10(0)` or `Bfp::exp10(12)`) to more clearly convey that their is "no real diff" here, besides the representation changing. In other words, I did it this way to try and make reviewing easier.

crates/solver/src/solver/baseline_solver.rs
nlordell: Yes.

crates/shared/src/sources/balancer_v2/pools/liquidity_bootstrapping.rs
```diff
@@ -123,21 +123,21 @@ mod tests {
             H160([1; 20]) => TokenState {
                 common: common::TokenState {
                     balance: bfp!("1000.0").as_uint256(),
-                    scaling_exponent: 0,
+                    scaling_factor: bfp!("1"),
```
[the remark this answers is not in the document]
nlordell: See 👇

crates/shared/src/sources/balancer_v2/pools/liquidity_bootstrapping.rs
```diff
@@ -123,21 +123,21 @@ mod tests {
...
                 },
                 weight: bfp!("0.3"),
             },
             H160([3; 20]) => TokenState {
                 common: common::TokenState {
                     balance: 15_000_000.into(),
-                    scaling_exponent: 12,
+                    scaling_factor: bfp!("1000000000000"),
```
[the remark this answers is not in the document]
nlordell: For sure. Actually, I meant to change all `scaling_exponent: x` to `scaling_factor: Bfp::exp10(x)` but left the ones in this file out by accident (I had started writing it this way but changed to use `Bfp::exp10` partway through the PR for the reasons I explained [above](https://github.com/cowprotocol/services/pull/1802#discussion_r1305529060)).

In fact I noticed this and [corrected it](https://github.com/cowprotocol/services/pull/1810/files#diff-58f36a21d0735bada1f70dbf97ab3caf8b35f544ae6d02be349f24b9e36461f1) already in one of the other PRs in the stack, so if it OK with you, I would leave it as here and expect it get fixed when #1810 merges).

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -130,7 +130,10 @@ impl Auction {
                                     r.asset.token.into(),
                                     StableReserve {
                                         balance: r.asset.amount.into(),
-                                        scaling_factor: r.scale.factor(),
+                                        scaling_factor: bigdecimal::BigDecimal::new(
+                                            eth::U256::from(r.scale).to_big_int(),
+                                            18,
+                                        ),
```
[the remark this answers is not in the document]
nlordell: Hmm, I don't agree.

Specifically, I think that conversion of `ScalingFactor` to a `BigDecimal` is not really a domain concern (not core to the "domain logic", but a detail of how we (de)serialize `ScalingFactor`s in JSON). With that in mind I do think that this is the correct place to have this logic (i.e. in the `dto` module).

**That being said**, I agree you need some implicit assumptions to know that dividing by 1e18 is required, which is not clear from the provided API. In particular, does `U256::from(ScalingFactor(2.1))` equal `U256(2)` or `U256(2e18)`? Right now it is `U256(2e18)` but the former is also a reasonable assumption and leaves ambiguity in the `From` implementation. I would propose removing the `From<ScalingFactor>` (as well as `From<Weight>` as it has a similar problem) implementation and have a method that more clearly conveys what is happening.

@MartinquaXD does this make sense to you?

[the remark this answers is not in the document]
MartinquaXD: Yes. A free function for this is fine as well. As long as you are not forced to know that you have to divide by `1e18` every time you want to convert the type it seems like an improvement.

crates/shared/src/sources/balancer_v2/swap/fixed_point.rs
nlordell: I implemented `exp10` to truncate. In particular, we have:

```
Bfp::exp10(-1) == Bfp(1e17)
Bfp::exp10(-2) == Bfp(1e16)
...
Bfp::exp10(-18) == Bfp(1e0) == Bfp(1)
Bfp::exp10(-19) == Bfp(1e-1) == Bfp(0) // truncating
...
```

I will add a doc-comment and some unit tests that demonstrate this.

nlordell: 472e804ff0b56c53eca785fe36f67be74362193e

crates/driver/src/infra/solver/dto/auction.rs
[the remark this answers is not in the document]
nlordell: 472e804ff0b56c53eca785fe36f67be74362193e

PR #1804, opened 2023-08-24
Move ethrpc into separate crate

MartinquaXD:
Moves `ethrpc` into its own crate because it's self contained and this way we can easily reuse it in our different crates without pulling it from `shared`.
Over all the idea is to trim down `shared`.

This PR tries to move the code with minimal changes. Whether or not code inside the new crate should be refactored or removed should be evaluated in a separate PR.

### Test Plan
No behavior changes so existing CI and e2e tests should be enough.


crates/ethrpc/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    },
+};
+
+pub const MAX_BATCH_SIZE: usize = 100;
+
+pub type Web3 = DynWeb3;
+pub type Web3Transport = DynTransport;
+pub type Web3CallBatch = CallBatch<Web3Transport>;
+
+pub struct Arguments {
```
nlordell, line 25: nit: I think we should call this `Config` - it is no longer arguments.

crates/shared/src/ethrpc.rs
```diff
@@ -69,44 +44,29 @@ impl Display for Arguments {
...
             self.ethrpc_max_concurrent_requests
         )?;
+        writeln!(f, "ethrpc_batch_delay: {:?}", self.ethrpc_batch_delay)?;
 
         Ok(())
     }
 }
 
+impl From<&Arguments> for ethrpc::Arguments {
+    fn from(args: &Arguments) -> ethrpc::Arguments {
```
nlordell, line 54: nit: Instead of a `From` conversion, it might make sense to just have an accessor function for this:

```rust
impl Arguments {
    pub fn ethrpc(&self) -> ethrpc::Config {
        // ...
    }
}
```

nlordell (approved):
Looks good, just some small nits.

PR #1805, opened 2023-08-24
Create observe crate

MartinquaXD:
Another PR to split `shared` into smaller crates.
Specifically this PR puts `tracing`, `global-metrics`, `panic_hook` inititialization and code related to `request_id`s in task local storage into the `observe` crate.
Slightly unfortunate is that the `driver` already has an `observe` module which resulted in slight name shadowing troubles. But I'm strongly in favor of ripping logs out of the `driver observe` module which will leave only metrics in there so at that point we might as well rename the `driver`'s `observe` module to `metrics`.


nlordell (approved):
LGTM. Didn't carefully review all 79 files, but I am assuming it is just renaming module paths and such.

crates/observe/Cargo.toml
```diff
@@ -1,11 +1,16 @@
 [package]
```
nlordell, line 1: Can you create the `LICENCE-*` symlinks in this directory as well?

PR #1809, opened 2023-08-25
Remove obsolete BlockRetriever strategies

MartinquaXD:
The newest `BlockRetrieverStrategy` (`GetBlockAndCall`) is the most accurate and reliable one so there is really no need anymore for the other ones.
This PR drops the obsolete strategies and the related code.


crates/shared/src/baseline_solver.rs
```diff
@@ -61,10 +61,10 @@ pub fn estimate_buy_amount<'a, L: BaselineSolvable>(
     let sell_token = path.first()?;
     path.iter()
         .skip(1)
-        .fold(
-            Some((sell_amount, *sell_token, Vec::new())),
+        .try_fold(
```
nlordell, line 64: oops! 🙈

crates/shared/src/current_block/get_block_and_call.rs
```diff
@@ -1,8 +1,9 @@
 use {
```
nlordell, line 1: I would rename the module as well `get_block_and_call` is not helpful IMO here (since we no longer have different strategies).

Maybe just `retriever.rs` or `fetcher.rs` to indicate its the logic for the `current_block::{retriever,fetcher}`?

nlordell (approved):

crates/shared/src/baseline_solver.rs
MartinquaXD: Still don't understand why git is not able to update a PR diff correctly when the PR is based off of a different branch that then got merged to `main`.
Merged `main` into this branch and now things are correct again.

PR #1818, opened 2023-08-28
Add None solver and allow to be set if external solver is set

### LINKED ISSUE #1791: Make solvers vector optional in solver crate
Currently it is not possible to run the solver / old driver against only an external solver. Trying to pass an empty vector of some kind to the `--solvers` flag in the `solver` crate causes an error. Previously, it was possible to instruct the old driver to run against only the `CowDexAg` solver and have an external solver accept those requests. This was removed in #1743.

If you agree, I'm happy to open a PR to make the `solvers` vector in the `Arguments` struct an `Option` and allow it to be `None` iff at least one external solver is set.

devanoneth:
Fixes #1791, #901

Adds a `None` SolverType which allows for the `solver` binary to run only against an external solver.

### DIFF
--- a/crates/solver/src/solver.rs
+++ b/crates/solver/src/solver.rs
@@ -400,8 +400,8 @@ pub async fn create(
 ) -> Result<Solvers> {
     // Tiny helper function to help out with type inference. Otherwise, all
     // `Box::new(...)` expressions would have to be cast `as Box<dyn Solver>`.
-    fn shared(solver: impl Solver + 'static) -> Option<Arc<dyn Solver>> {
-        Some(Arc::new(solver))
+    fn shared(solver: impl Solver + 'static) -> Arc<dyn Solver> {
+        Arc::new(solver)
     }
 
     let buffer_retriever = Arc::new(BufferRetriever::new(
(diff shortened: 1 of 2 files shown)


crates/solver/src/solver.rs
```diff
@@ -483,6 +484,7 @@ pub async fn create(
             let score_calculator = score_configuration.get_calculator(solver_type);
 
             let solver = match solver_type {
+                SolverType::None => None,
```
MartinquaXD, line 487:
```suggestion
                SolverType::None => return None,
```
This enables a few improvements:
* makes pattern matching block further down obsolete
* don't have to return `Option` from `shared`
* no `.expect()` on `shared()` further down

crates/solver/src/solver.rs
```diff
@@ -567,12 +569,21 @@ pub async fn create(
...
+                }
+            };
+
             shared(OptimizingSolver {
                 inner: solver,
                 post_processing_pipeline: post_processing_pipeline.clone(),
                 score_calculator,
             })
         })
+        .flatten()
```
MartinquaXD, line 586: Instead of calling `.map().flatten()` `clippy` suggests to use `.filter_map()` instead.

crates/solver/src/solver.rs
devanoneth: Ah of course, thank you!

MartinquaXD (approved):

MartinquaXD:
Thank you for your contribution! 🙏

PR #1820, opened 2023-08-29
Move current block stream into `ethrpc`

MartinquaXD:
The overall idea is to move block streams and event streams into the `ethrpc` crate. This PR takes care of the `current_block` stream.

Note that the files `shared::ethrpc` and `shared::current_block` still exist but they only contain the CLI logic.
In a follow up refactoring it would make sense to move that last remaining logic into `shared::arguments`.

### Test Plan
Only moved code around so the e2e tests should be enough.


crates/shared/src/current_block.rs
```diff
@@ -1,325 +1,57 @@
...
-pub async fn block_number_to_block_number_hash(
-    web3: &Web3,
-    block_number: BlockNumber,
-) -> Option<BlockNumberHash> {
-    web3.eth()
-        .block(BlockId::Number(block_number))
+/// Command line arguments for creating global block stream.
+#[derive(Debug, Parser)]
+#[group(skip)]
+pub struct Arguments {
```
nlordell, line 21: Maybe move this to `shared::arguments::Arguments`?

crates/ethrpc/src/current_block/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+/// latest block number.
+#[async_trait::async_trait]
+pub trait BlockRetrieving: Send + Sync + 'static {
+    async fn current_block(&self) -> Result<BlockInfo>;
+    async fn block(&self, number: u64) -> Result<BlockNumberHash>;
+    async fn blocks(&self, range: RangeInclusive<u64>) -> Result<Vec<BlockNumberHash>>;
+}
+
+#[async_trait::async_trait]
+impl BlockRetrieving for Web3 {
```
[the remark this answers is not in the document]
MartinquaXD: Yep. Had a few other refactoring ideas as well but wanted to do them in a separate PR as well. 👌

crates/shared/src/current_block.rs
MartinquaXD: Initially I wanted to do that together with moving `ethrpc` arguments because I forgot that last time. But I can do that now as well.

nlordell: Follow up is fine.

nlordell (approved):

fleupold (approved):
Assuming no logic changes

🧹

PR #1828, opened 2023-08-30
Enable Simulation-Based ERC-1271 Signature Validation

nlordell:
Follow up to #1827 

This PR adds the required simulation options to enable simulation-based ERC-1271 signature verification in order to support smart contract orders that do required setup as pre-hooks.

An example of this would be to have a Composable CoW order be created as a pre-hook and traded in the same transaction. With account abstraction, this would allow for gasless smart orders!

### Test Plan

Will share a manual test once completed. Tx hash or it didn't happen!


fleupold (approved):
So tidy

nlordell:
Merging with one approval as this is not contain any interesting logic changes.

PR #1831, opened 2023-08-31
Magic: StorageAccessbile Based Simulations

nlordell:
Follow up to https://github.com/cowprotocol/services/pull/1827#discussion_r1310425237 - thanks @fleupold for the wonderful idea!

This PR changes instances of `eth_call` with state override simulations **where we are only overriding the Settlement contract code** to make use of `StorageAccessible`. It does this with an additional support `Reader` contract in order to avoid needing to manage deployments on-chain for the `StorageAccessible` target (i.e. the reader logic). See the comment in the `Reader.sol` file for more details on how this works.

Specifically, we changed the following two simulations to no longer require state overrides (i.e. we do not need a special node to handle those requests, and it will Just Work™ on Gnosis Chain):
* Account balance simulations with pre-hooks
* ERC-1271 signature validation with pre-hooks

In a follow up PRs I will:
1. https://github.com/cowprotocol/services/pull/1832
2. Remove the redundant `Web3` strategy for both account balance and signature verification simulation

### Release notes

Hooks E2E test continues to pass **without the nasty `----account-balances-optimistic-pre-interaction-handling=true` hack** which used to disable balance checks for orders with pre-hooks. This means that not only are we E2E testing that executing orders with pre-hooks are working, we are also checking that balance simulations work 🎉.

You can double check this is actually doing something by applying a patch like this:

```diff
diff --git a/crates/e2e/tests/e2e/hooks.rs b/crates/e2e/tests/e2e/hooks.rs
index 42d8d072..84d978b5 100644
--- a/crates/e2e/tests/e2e/hooks.rs
+++ b/crates/e2e/tests/e2e/hooks.rs
@@ -66,7 +66,7 @@ async fn test(web3: Web3) {
             full: json!({
                 "metadata": {
                     "hooks": {
-                        "pre": [permit, steal_cow],
+                        "pre": [/*permit, */steal_cow],
                         "post": [steal_weth],
                     },
                 },
```

And see the E2E test fail as expected with:

```
thread 'hooks::local_node_test' panicked at 'called `Result::unwrap()` on an `Err` value: (400, "{\"errorType\":\"InsufficientAllowance\",\"description\":\"order owner must give allowance to VaultRelayer\"}")', crates/e2e/tests/e2e/hooks.rs:83:41
```

I also ran the `cargo test -p shared -- --nocapture --ignored account_balances::simulation` manual test which continues to work.


crates/contracts/solidity/Reader.sol
```diff
@@ -0,0 +1,45 @@
...
+    /// another application of this trick.
+    ///
+    /// The reader contract does this to:
+    /// 1. Deploy some arbitrary contract code
+    /// 2. Use the `StorageAccessible` pattern to execute the contract code
+    ///    deployed in step 1. within the another contract context (usually the
+    ///    settlement contract - which implements this pattern)
+    ///
+    /// This allows us to make use of `StorageAccessible` without actually
+    /// deploying a contract :).
```
fleupold, line 23: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FUm3ljJl8jrnHy%2Fgiphy.gif&token=mindblown" data-gitmeme-token="mindblown"></a>

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FMUeQeEQaDCjE4%2Fgiphy.gif&token=bowing" data-gitmeme-token="bowing"></a>

crates/contracts/solidity/Reader.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+import { IStorageAccessible } from "./interfaces/IStorageAccessible.sol";
+
+/// @title A contract for simulating delegate calls on the Settlement contract.
+contract Reader {
```
fleupold, line 7: Nit, while the name is very concise, I find it hard to. by itself grasp what this contract is doing. Maybe something like DeployExtensionAndSimulateDelegatecall ?

fleupold (approved):
mind = blown

crates/contracts/solidity/Reader.sol
nlordell: Maybe `SimulateCode` or something? We don't actually deploy (well we do, but that is an implementation detail - if it were able to simulate code in some other way, then we could use that without it affecting the interface).

nlordell: Renamed things a bit (also `s/reader/storage_accessible/g` since the module is really for building calls to `StorageAccessible` contracts) - LMK if the new naming is better or if I should change it again.

crates/contracts/build.rs
```diff
@@ -523,11 +523,10 @@ fn main() {
     generate_contract("Trader");
     generate_contract("Solver");
 
-    // Support contract used for balance simulation.
+    // Support contract used for various order simulations.
```
MartinquaXD, line 526: Otherwise it sounds like the comment should rather be above `SimulateCode` (instead of above a block of contracts).
```suggestion
    // Support contracts used for various order simulations.
```

crates/shared/src/signature_validator/simulation.rs
```diff
@@ -131,7 +121,7 @@ impl Simulation {
     fn decode(output: &[u8]) -> Result<Self> {
         let function = contracts::support::Signatures::raw_contract()
             .abi
-            .function("balance")
+            .function("validate")
```
MartinquaXD, line 124: Did this never cause problems?

crates/contracts/solidity/SimulateCode.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma solidity ^0.8.17;
+
+import { IStorageAccessible } from "./interfaces/IStorageAccessible.sol";
+
+/// @title A contract for simulating arbitrary code in the context of a
+/// `StorageAccessible` contract.
+contract SimulateCode {
+    /// @dev This looks like a constructor but it is not... In fact, nodes
+    /// support `eth_call`s for contract creation and **return the code of the
+    /// contract that would be created**. This means we can use contructors to
```
MartinquaXD, line 11:
```suggestion
    /// contract that would be created**. This means we can use constructors to
```

MartinquaXD (approved):
Awesome stuff!

crates/shared/src/signature_validator/simulation.rs
nlordell: It did - the PR this was based off of was untested. I noticed this while implementing #1832

crates/contracts/solidity/SimulateCode.sol
nlordell: Interesting - my spell-check nvim plugin isn't working for Solidity files 🤔.

PR #1833, opened 2023-09-01
1Inch Price Oracle

nlordell:
This PR introduces a 1Inch price oracle (using their [spot price aggregator](https://docs.1inch.io/docs/spot-price-aggregator/introduction) onchain infrastructure).

The nice thing about this approach is that it relies on on-chain infrastructure, so we don't have to worry about paying for API credits (and with an archive node can even replay historic data).

### Test Plan

No tests yet. Just proposing an idea. Will follow up with an actual `solvers` implementation for native token prices if the team thinks I should.


fleupold:
Interesting. I tried to see how they get the prices, and it looks like a very naive implementation from our baseline solver onchain (I'm not really sure what's offchain about it). From what I can tell, they only support Uniswap, Mooniswap and Synthetix oracles (not even Balancer).
I'm a bit worried that if we have many almost identical implementations of a poor router, we need a high threshold before we can consider a quote reliable enough.

nlordell:
So, the choise of the 1Inch on-chain oracle was mostly because I was looking into their `token-prices` API and it looks like:
1. It is not an official public API (we would have to ask permission to use it)
2. We need to spoof some headers in order to get around their access control

I started with this type as a "base" for a price oracle in order to implement a price oracle solver (which would be independent on the price source).

fleupold:
> It is not an official public API (we would have to ask permission to use it)

I was thinking about using https://portal.1inch.dev/documentation/spot-price/swagger (but it only has 357 tokens)

MartinquaXD (commented):
The code makes sense to me but if that oracle is really just a worse version of `Baseline` than I think we should not implement it.

nlordell:
One argument for keeping this around is that they:
- They notably implement UniV3
- Does not require us to setup and index "liquidity" if/when we go to new chains

They actually support quite a bit more than our Baseline solver:

image

> I'm not really sure what's offchain about it

It is definitely "on-chain" in that you read the price from a contract on-chain. I think they called the contract "off-chain" to make it extra clear API consumers that it is unsafe for as an on-chain price oracle. From their docs:

> To avoid price manipulations within a transaction, the spot price aggregator should ONLY be used off-chain. DO NOT use it on-chain.

crates/solvers/src/infra/oracle/oneinch.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub ethrpc: reqwest::Url,
+
+    /// The address of the `OffchainOracle` contract. Specify `None` to use the
+    /// default contract address for the connected chain.
+    pub oracle: Option<eth::ContractAddress>,
+}
+
+impl OneInch {
+    pub async fn new(config: Config) -> Result<Self, Error> {
+        let web3 = ethrpc::web3(module_path!(), &config.ethrpc);
```
MartinquaXD, line 19: Enabling the `trace` logs of `ethrpc` already puts a lot of pressure on elastic search. I think having a short name here would actually make a difference.
```suggestion
        let web3 = ethrpc::web3("OneInchOracle", &config.ethrpc);
```

fleupold (approved):
Given the list of sources, I'm more inclined to approve this estimator. Can you link where you found this though, when I checked the onchain oracles they were all of type Uni v2, Mooniswap and Synthetix

nlordell:
I found it here: https://docs.1inch.io/docs/spot-price-aggregator/introduction/#ethereum-mainnet

Note that the contract address changed very recently (6 days ago), so I updated it in the PR.

nlordell:
I have not merged this yet, as I'm waiting to have a concrete `solvers::Solver` implementation for this before introducing code that we don't end up using.

nlordell:
Closing

PR #1838, opened 2023-09-04
Configure 0x For Tests

nlordell:
In trying to run the orderbook locally to reproduce an issue, I discovered that since 0x changed their API to require an access key, we can no longer run some of the tests.

This PR fixes that by changing the 0x API `Default` implementation to read the URL and API key from environment variables.

### Test Plan

Running the 0x tests with a configured API key now work.

### DIFF
--- a/crates/shared/src/zeroex_api.rs
+++ b/crates/shared/src/zeroex_api.rs
@@ -390,9 +390,18 @@ impl DefaultZeroExApi {
         })
     }
 
-    /// Create a 0x HTTP API client using the default URL and HTTP client.
+    /// Create a 0x HTTP API client for testing using the default HTTP client.
+    ///
+    /// This method will attempt to read the `ZEROEX_URL` (falling back to the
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Maybe we can include a hint on how to get the correct URL for a future developer that comes across this test.

crates/shared/src/zeroex_api.rs
```diff
@@ -390,17 +390,18 @@ impl DefaultZeroExApi {
...
-            base_url: Self::DEFAULT_URL.parse().unwrap(),
-        }
-    }
-
-    /// Create a 0x HTTP API client using the default URL and HTTP client.
+    /// Create a 0x HTTP API client for testing using the default HTTP client.
+    ///
+    /// This method will attempt to read the `ZEROEX_URL` (falling back to the
+    /// default URL) and `ZEROEX_API_KEY` (falling back to no API key) from the
+    /// local environment when creating the API client.
```
nlordell, line 397: @fleupold - added doc-comment here.

PR #1843, opened 2023-09-06
Remove SubmissionPreference

sunce86:
I came across this field when checking what else needs to be ported to colocated driver.

This field is not used in colocated world and I would rather remove it since it is not used anymore (no solver is sending it) (and there is not much sense for it to exist). 

Colocated driver is supposed to determine the submission strategies based on revert risk. And we also landed on a pretty decent public mempool + Mevblocker setup.


nlordell (approved):
Makes sense to me!

MartinquaXD (approved):

PR #1846, opened 2023-09-07
Settlement score mandatory

sunce86:
After https://github.com/cowprotocol/services/pull/1843, this is another initiative to make `Settlement` cleaner for the needs of porting the risk/score calculation to colocated driver.

Now the `score` is non-optional, with the default value `Score::RiskAdjusted` which is equivalent to `score: None` (from the POV how the protocol handles `score: None`) but more explicit.

Note: ~~Initially I wanted to make the score non optional in the `SettledBatchAuctionModel` during deserialization, but I couldn't make piece between #[flatten] and #[default] on untagged enum. If anyone knows about any workaround let me know.~~ Added custom deserializator


nlordell (approved):
Makes sense to me!

MartinquaXD (approved):

PR #1847, opened 2023-09-07
Add Timing Metrics for IPFS

nlordell:
This PR adds timing metrics for fetching AppData from IPFS. I noticed that we are getting the occasional alert around order creation taking longer than expected, this metric should help us see if it lines up with IPFS app data queries taking too long (instead of manually looking at log timestamps).

I also considered adding a label on "outcome", but didn't think it would be super valuable here.

### Test Plan

No logic changes, so Rust CI.

### Release notes

Note that the existing IPFS metric names got renamed (to remove the duplicate `api_api` from the name). Will update Grafana once PR merges.

### DIFF
--- a/crates/orderbook/src/ipfs_app_data.rs
+++ b/crates/orderbook/src/ipfs_app_data.rs
@@ -13,11 +13,14 @@ pub struct IpfsAppData {
 }
 
 #[derive(prometheus_metric_storage::MetricStorage, Clone, Debug)]
-#[metric(subsystem = "api")]
+#[metric(subsystem = "ipfs")]
 struct Metrics {
     /// Number of completed IPFS app data fetches.
     #[metric(labels("outcome", "source"))]
-    ipfs_app_data: prometheus::IntCounterVec,
+    app_data: prometheus::IntCounterVec,
+
+    /// Timing of IPFS app data fetches.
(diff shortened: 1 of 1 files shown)


crates/orderbook/src/ipfs_app_data.rs
```diff
@@ -13,11 +13,14 @@ pub struct IpfsAppData {
...
-#[metric(subsystem = "api")]
+#[metric(subsystem = "ipfs")]
 struct Metrics {
     /// Number of completed IPFS app data fetches.
     #[metric(labels("outcome", "source"))]
-    ipfs_app_data: prometheus::IntCounterVec,
+    app_data: prometheus::IntCounterVec,
+
+    /// Timing of IPFS app data fetches.
+    fetches: prometheus::Histogram,
```
MartinquaXD, line 23: Should this new metric also be initialized on start up to avoid broken graphs when no data has been generated?

MartinquaXD (approved):

crates/orderbook/src/ipfs_app_data.rs
nlordell: Its a histogram, so I don't think its needed. AFAIK, we don't do it for our other histograms.

PR #1849, opened 2023-09-08
Respect Deadline in Baseline Solver

nlordell:
This PR changes the baseline solver to respect the auction deadline.

### Test Plan

CI - very minor logic change I didn't think it was worth introducing a test for.

### DIFF
--- a/crates/solvers/src/domain/solver/baseline.rs
+++ b/crates/solvers/src/domain/solver/baseline.rs
@@ -50,11 +50,12 @@ impl Baseline {
         auction
             .orders
             .iter()
+            .take_while(|_| auction.deadline.remaining().is_some())
             .filter_map(|order| {
                 let sell_token = auction.tokens.reference_price(&order.sell.token);
                 self.requests_for_order(UserOrder::new(order)?)
                     .find_map(|request| {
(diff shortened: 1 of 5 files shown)


crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -50,11 +50,12 @@ impl Baseline {
         auction
             .orders
             .iter()
+            .take_while(|_| auction.deadline.remaining().is_some())
             .filter_map(|order| {
                 let sell_token = auction.tokens.reference_price(&order.sell.token);
                 self.requests_for_order(UserOrder::new(order)?)
                     .find_map(|request| {
-                        tracing::trace!(?request, "finding route");
+                        tracing::trace!(order =% order.uid, ?request, "finding route");
```
MartinquaXD, line 58: This will result in responses being emitted **after** the deadline has passed. Should we maybe deduct a small buffer to ensure the response arrives at the driver **before** the deadline?

nlordell: The `driver` already does this.

MartinquaXD (approved):

PR #1851, opened 2023-09-08
Concurrent solvers

nlordell:
This PR add concurrency to the single order solvers in the `solvers` binary. Fairly simple, nothing complicated.

### Test Plan

Tests continue to pass with new concurrency parameter.


crates/solvers/src/domain/solver/dex/mod.rs
```diff
@@ -28,27 +33,35 @@ impl Dex {
...
-            for order in auction.orders.iter().filter_map(order::UserOrder::new) {
-                let span = tracing::info_span!("solve", order = %order.get().uid);
-                if let Some(solution) = self
-                    .solve_order(order, &auction.tokens, auction.gas_price)
-                    .instrument(span)
-                    .await
-                {
-                    solutions.push(solution);
-                }
+            // Collect user orders into a vector for chunking. Note that we
```
MartinquaXD, line 44: I think you are reinventing [`futures::stream::buffered()`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered) here.
`buffered()` would be a bit nicer because we are always saturating the max number of active parallel requests instead of having the whole next batch being blocked by a single slow request.

crates/solvers/src/infra/config/dex/file.rs
```diff
@@ -63,6 +71,8 @@ pub async fn load<T: DeserializeOwned>(path: &Path) -> (super::Config, T) {
             config.absolute_slippage.map(eth::Ether),
         )
         .expect("invalid slippage limits"),
+        concurrent_requests: NonZeroUsize::new(config.concurrent_requests)
+            .expect("invalid concurrent requests values"),
```
MartinquaXD, line 75: Would be more helpful to state that that value has to be greater than 0.

crates/solvers/src/infra/config/dex/file.rs
```diff
@@ -25,6 +25,10 @@ struct Config {
     #[serde_as(as = "Option<serde_with::DisplayFromStr>")]
     absolute_slippage: Option<eth::U256>,
 
+    /// The number of concurrent requests to make to the DEX aggregator API.
+    #[serde(default = "default_concurrent_requests")]
+    concurrent_requests: usize,
```
MartinquaXD, line 30: Would it be possible to directly deserialize `NonZeroUsize` here?

crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -49,8 +50,8 @@ impl Baseline {
 
         auction
             .orders
-            .iter()
-            .take_while(|_| auction.deadline.remaining().is_some())
+            .par_iter()
+            .take_any_while(|_| auction.deadline.remaining().is_some())
```
MartinquaXD, line 54: Bit unfortunate that we don't really take advantage of prioritized orders anymore but I guess `Baseline` is fast enough to go through all orders so it wouldn't really matter anyway.

nlordell: Hmm... Good point - maybe it makes sense for the baseline solver to keep it as it was. Also it would allow for better concurrency.

nlordell: Maybe we should use `tokio::task::spawn_blocking` for this.

crates/solvers/src/infra/config/dex/file.rs
nlordell: It is!

crates/solvers/src/infra/config/dex/file.rs
nlordell: Serialized directly to the non-zero type.

crates/solvers/src/domain/solver/dex/mod.rs
nlordell: I was, and this was a great suggestion.

I wrestled with the Rust compiler for a long time, but got it to work.

crates/solvers/src/domain/solver/baseline.rs
nlordell: Will leave this for another time, its a larger change.

MartinquaXD (approved):

crates/solvers/src/domain/solver/baseline.rs
nlordell: https://github.com/cowprotocol/services/pull/1852

PR #1852, opened 2023-09-08
Push Solving Logic to Background Thread

nlordell:
The Tokio [documentation](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) points out that (emphasis my own):

> In general, issuing a blocking call or **performing a lot of compute** in a future without yielding is problematic, as it may prevent the executor from driving other futures forward.

In particular this is an issue with the Baseline solver which can take 100s of ms to go through an auction (see below). The solution is simple - do the solving in a background thread so that the main Tokio runtime doesn't get affected. Note that this function requires a `Send` future, so we need to wrap the Baseline solver state into an `Arc` so that we can clone and send it over cheaply.

We apply the same to he Naive solver (although, it takes significantly less time, in the 30s of ms, but still enough IMO to warrant the `spawn_blocking` call).

### Test Plan

Existing tests continue to pass.

### Baseline Solver Metrics:

image


crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -15,35 +15,64 @@ use {
...
     /// The maximum number of attempts to solve a partially fillable order.
     /// Basically we continuously halve the amount to execute until we find a
     /// valid solution or exceed this count.
-    pub max_partial_attempts: usize,
+    max_partial_attempts: usize,
 }
 
 impl Baseline {
+    /// Creates a new baseline solver for the specified configuration.
+    pub fn new(config: config::baseline::Config) -> Self {
```
MartinquaXD, line 51: Strictly speaking the solver in `domain` should not use a config type from `infra`, right?
Seems like `Baseline` should define what kind of configuration it expects and the `infra` module should return that.
I think this doesn't have to be addressed in this PR. Just noticed it.

MartinquaXD (approved):

crates/solvers/src/domain/solver/baseline.rs
nlordell: Good point.

I think this is an issue with all `domain::solver::*::new` methods that take `infra::config` types. At least it is consistent (but different to what we do in the `driver` module.

nlordell: https://github.com/cowprotocol/services/issues/1853

tracked.

PR #1855, opened 2023-09-08
Fix score deserialization

sunce86:
Fix after https://github.com/cowprotocol/services/pull/1846

One of the solvers is sending `score: null` and it fails deserialization. This PR lands on a much simpler logic to support this edge case.


MartinquaXD (approved):

sunce86:
Merging with 1 approval to be included in next week release.

PR #1866, opened 2023-09-12
Sum up executed fees

MartinquaXD:
Our executed fee values in the order metadata were not accurate for partially fillable limit orders with multiple fills.
Now we are summing the executed fees for all trade executions so it should be good now.


sunce86 (approved):
Ah yes, I just realized this as well.

fleupold (approved):
Nice!

PR #1871, opened 2023-09-13
Fix error handling for signature validator

sunce86:
Context: https://cowservices.slack.com/archives/C0361CDD1FZ/p1694595411953299

We need to convert web3 error first to `ExecutionError` so the reverts are properly classified so that our error conversion method properly classify revert as `SignatureValidationError::Invalid`

### DIFF
--- a/crates/shared/src/signature_validator.rs
+++ b/crates/shared/src/signature_validator.rs
@@ -1,5 +1,6 @@
 use {
-    ethcontract::Bytes,
+    crate::ethcontract_error::EthcontractErrorType,
+    ethcontract::{errors::ExecutionError, Bytes},
     ethrpc::Web3,
     hex_literal::hex,
     model::interaction::InteractionData,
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):

fleupold (approved):
Change is good. I'm generally not a friend of automatically deriving any error into the "Other" error type (cf. https://github.com/cowprotocol/services/blob/main/crates/shared/src/signature_validator.rs#L30) as this doesn't force the producer of errors to think about what higher level error they want to produce. I'd be happy if we got rid of all `#[from] anyhow::Error`

nlordell:
I don't think the fix is correct - the `Signatures.sol` is designed to never revert, if it does something unexpected is happening. 

> I'm generally not a friend of automatically deriving any error into the "Other" error type

Note that this isn't what happened here - we had a manual implementation of `From<web3::Error>` that was explicitly converting all `web3` errors to the `Other` type. I think we drop the `From` implementation as you suggested should add comment explaining the rationale (call should never revert since the `Signature.sol` handles reverts already).

nlordell:
> I don't think the fix is correct - the `Signatures.sol` is designed to never revert, if it does something unexpected is happening.

Turns out the unexpected thing that was happening was unexpected semantics of Solidity `try ... catch`. I think that a nicer fix would be to consolidate the revert handling into one place instead of having the contract handle reverts, and the Rust code also handling reverts.

PR #1873, opened 2023-09-14
Refactor Signature Validation

nlordell:
This PR is a follow-up to #1871

Specifically, that PR added error handling for the `eth_call` simulation such that reverts are properly classified as `SignatureValidationError::Invalid`. The `Signatures.sol` contract was _supposed to_ handle all reverts from signature verifiers, so that error classification shouldn't have been needed.

**However**, it turns out I made some assumptions about the Solidity `try ... catch` semantics that turned out to be invalid and cause the `Signatures.sol` to still revert in some cases. In particular, `Signatures::validate` will revert if:
- `signer` is not a contract
- `signer.isValidSignature` returns less than 32 bytes of data

With this in mind, it doesn't really make sense to try and handle revert errors in both the Solidity code and the Rust code, so just handle all revert errors in Rust in the `signature_validation::simulation::Validator` code and get rid of the `try ... catch` in the Solidity code.

I also refactored the `From<ethcontract::ExecutionError>` implementation into the existing `From<web3::Error>` (which was no longer being used, and IMO made more sense to be there).

Furthermore, https://github.com/cowprotocol/services/pull/1842 removed some `signature_validator::{arguments, web3}` submodules whose files were accidentally left behind, so I removed them here as well.

### Test Plan

CI - the existing E2E tests for signature validation continue to pass, and added a new one to check that errors are correctly detected.


crates/shared/src/signature_validator/arguments.rs
```diff
@@ -1,66 +0,0 @@
-use {
```
nlordell, line 1: This module had already been removed in #1842, but the code was accidentally left behind.

crates/shared/src/signature_validator/web3.rs
```diff
@@ -1,108 +0,0 @@
-use {
```
nlordell, line 1: This module had already been removed in #1842, but the code was accidentally left behind.

crates/shared/src/signature_validator/simulation.rs
```diff
@@ -106,51 +96,39 @@ impl SignatureValidating for Validator {
...
-impl SimulationResult {
-    fn decode(value: u8) -> Result<Self> {
-        match value {
-            0 => Ok(Self::Ok),
-            1 => Ok(Self::Invalid),
-            2 => Ok(Self::Revert),
-            _ => anyhow::bail!("invalid simulation result variant {value}"),
+impl From<web3::Error> for SignatureValidationError {
+    fn from(err: web3::Error) -> Self {
+        // TODO: This is needed to parse Hardhat revert errors, which
```
nlordell, line 117: Fixed in https://github.com/cowprotocol/ethcontract-rs/pull/962

Will remove this extra branch when the new `ethcontract` version hits.

crates/contracts/solidity/Signatures.sol
```diff
@@ -44,16 +39,10 @@ contract Signatures {
...
-                : Result.INVALID;
-        }
-        catch {
-            gasUsed = gasUsed - gasleft();
-            result = Result.REVERT;
-        }
+        bytes4 magicValue = signer.isValidSignature(order, signature);
+        gasUsed = gasUsed - gasleft();
+
+        require(magicValue == ERC1271_MAGICVALUE, "didn't say the magic word");
```
sunce86, line 45: 😄

crates/shared/src/signature_validator/simulation.rs
sunce86: Probably worth creating issue so we don't forget.

sunce86 (approved):
Nice.

crates/shared/src/signature_validator/simulation.rs
nlordell: I already made the PR and plan on doing it today. Can create an issue if you think its worth it.

MartinquaXD (approved):

crates/shared/src/signature_validator/simulation.rs
nlordell: https://github.com/cowprotocol/services/issues/1875

PR #1878, opened 2023-09-19
Optional bad token detection

MartinquaXD:
This PR explores the viability of replacing bad token detection with verified quotes.
The idea is that in order to create a quote you have to use a verified quote. If any of the solvers returns a quote that can actually be simulated we consider the token supported even if it uses some weird logic under the hood.

Note that this code makes 1 important assumption: When verified quotes are enabled all quotes in our DB are assumed to be verified. This is obviously wrong when we switch from unverified quotes to verified because the existing quotes don't get deleted right away. I kept it like this to keep the scope of this PR smaller + I think adding the extra complexity for this will end up as dead weight because we want to move to 100% verified quotes quickly. LMK if you have thoughts on this.

In order to completely rely on verified quotes to determine whether or not we support a token we now simply have to configure a `--trade-simulator` and no `--tracing-node-url` in our deployment.

### Test Plan
To make sure switching over to verified quotes doesn't break anything when deployed I migrated all our tests to use verified quotes. Which resulted in a few changes:
1. we now pass the settlement contract address into the verification helper contract because its address in tests is different from the deployed address
2. for 2 of the partial fill tests the trader has to start out with the entire `sell_token` balance and then we temporarily decrease their balance to force a partial fill
3. the quote for an ethflow order now gets requested by the actual trader instead of the ethflow contract

I also added a new test that verifies that an order must not be created when we can't find a verified quote for it.


nlordell:
> Can we make this a counter/warning or something softer for now to see in prod how many orders, and from what integrations, would get into trouble?

Can this be deduced by logs? I think if we encounter this log whenever an order is created without a quote ID:

```
tracing::debug!(quote_id =? quote.id, "computed fresh quote for order creation");
```

**Edit**: Oops! This is only true for market orders, and I assume that @fleupold is probably referring to not being so strict for limit orders.

nlordell:
Fat fingers

MartinquaXD:
Closing since this PR is getting replaced by a series of smaller PRs.

PR #1901, opened 2023-09-27
nlordell suggestions: Refactor scores

sunce86:
Applied on top of https://github.com/cowprotocol/services/pull/1856

Tries to implement suggestions from https://github.com/cowprotocol/services/pull/1856#pullrequestreview-1646356353


crates/driver/src/domain/mempools.rs
```diff
@@ -34,8 +34,31 @@ impl Mempools {
...
 }
 
 #[derive(Debug, Error)]
 #[error("no mempools configured, cannot execute settlements")]
 pub struct NoMempools;
+
+/// Defines if the mempools are configured in a way that could lead to
+/// significant costs in case a settlement fails onchain submission.
+#[derive(Debug, Clone, Copy)]
+pub enum HighRisk {
```
nlordell, line 61: I'm not a fan of this name. Specifically, from what I understand, the "risk" here encodes whether or not a settlement can revert? In that case, I would not use the same nomenclature as we used in the past but something more specific to what we are encoding here (whether or not the mempool executes reverted transactions or just ignores them).

nlordell (commented):
Personally, I like this suggestion alot. My only gripe is with the "high risk" naming which is unclear to me (it seems like a term carried over from legacy code and a good candidate for renaming to something more concrete).

nlordell:
> however not super happy with what I came up with

What in particular are you not happy about?

sunce86:
> > however not super happy with what I came up with
> 
> What in particular are you not happy about?

I spent quite some time refactoring this and hoped I could extract more functionality from `boundary::settlement::score()` to the `driver::domain`, but essentially settlement and score calculation are currently tightly coupled.

This can probably be done in the future, but it seems the way forward would be to define `objective_value::Inputs` in the driver domain and get it from `settlement` and break the dependency.. but will probably be able to do this only once we drop `solver` crate.

crates/driver/src/tests/cases/mod.rs
```diff
@@ -41,3 +41,6 @@ pub const DEFAULT_SCORE_MAX: u128 = 500000000000000000000000000000u128;
 
 /// The default solver fee for limit orders.
 pub const DEFAULT_SOLVER_FEE: u128 = 100u128;
+
+/// The default maximum value to be payout out to solver per solution
+pub const DEFAULT_SCORE_CAP: u128 = 10000000000000000u128;
```
sunce86, line 46: @fhenneke This comment I had to add make it obvious that `score_cap` is a bad name for this. Btw, `score_cap` is used in [this function](https://github.com/cowprotocol/services/blob/main/crates/solver/src/settlement_rater.rs#L405).

Do we have a more descriptive alternatives to name this variable? `reward_cap`, `payout_cap` or something else?

Right now, when I look into autopilots `score_cap`, based on it's name it looks like it is some kind of maximum score `driver` is allowed to submit to autopilot...

fhenneke: Indeed `reward_cap` is a more descriptive name. The value puts a limit on how much we pay on top of gas reimbursements, i.e. the reward.
The number does not on its own give a bound on the total payment. The payment is bounded by `reward_cap + gas_costs`.

PR #1902, opened 2023-09-28
Scale Orders By Remaining Amount And Available Balance

nlordell:
# Description

This PR essentially combines #1861 and #1900 into a single PR. The reason for providing an alternative PR is that some of the decisions taking in #1900 (on which #1861 is based on) are directly related to the requirements of #1861. I hope this makes reviewing easier 🤞.

This PR fixes an issue where partially fillable order amounts were not being scaled before passing to the solver engines as expected, leading to invalid settlements.

# Changes

1. The order `Partial` now carries an available amount and not how much was executed. This allows us to encode things like the remaining order balance in the same field (so "available" here means how much of a partially fillable order is left and covered by user balance).
2. Use the scaled available amounts to pass to the solvers instead of the full amounts.

## How to test

Rust CI - added unit tests to verify available amount scaling logic. An E2E test that fully checks the balance allocation is under development and will be proposed in a follow up PR.


fleupold (approved):
LGTM (still think the logic around updating order.available based on allocated_balance could be better documented for future maintainers of the code), but it makes sense to me now.

nlordell:
> still think the logic around updating order.available based on allocated_balance could be better documented for future maintainers of the code

In that case I will add a comment explaining it. I was hoping that with the types the code was clearer, but if it is still not the case, then I will do as you suggested.

nlordell:
For future selves, there is a lot of interesting discussion in #1900 and #1861 that you should consult when trying to fully understand this PR.

Issue #1898, opened 2023-09-27
Remove Score::Discount variant


sunce86:

sunce86:
Implemented.

PR #1906, opened 2023-09-29
Remove score discount

sunce86:
# Description
Implements https://github.com/cowprotocol/services/issues/1898
Implemented in a separate PR so it doesn't gets merged too soon (next week release is too soon for this change _I think_)

# Changes

As per [discussion](https://cowservices.slack.com/archives/C036JAGRQ04/p1695637642640269), score discount is deprecated.

# Release Notes
To be merged next week, so it becomes a part of weekly release ~10th of October @harisang

### DIFF
--- a/crates/driver/src/boundary/settlement.rs
+++ b/crates/driver/src/boundary/settlement.rs
@@ -211,9 +211,6 @@ impl Settlement {
     ) -> Result<competition::Score> {
         let score = match self.inner.score {
             http_solver::model::Score::Solver { score } => score,
-            http_solver::model::Score::Discount { .. } => {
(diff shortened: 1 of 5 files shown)


harisang:
Just pinged solvers and told them about this.

fleupold (approved):
@harisang can you make sure this is also reflected in the official documentation for the instance.json file we have at docs.cow.fi ?

harisang:
> @harisang can you make sure this is also reflected in the official documentation for the instance.json file we have at docs.cow.fi ?

Good catch. Actually, only `score` is mentioned in our official documentation, so all other alternatives are not there. I will update the documentation to contain the `success_probability` style of score reporting.

PR #1918, opened 2023-10-04
Additional fixes for driver multiple solutions

sunce86:
# Description
Follow up PR addressing the comments: https://github.com/cowprotocol/services/pull/1909#discussion_r1345260527

Solution ID can be unique:
1. Across multiple drivers (hard to implement)
2. Across multiple competitions in a single driver
3. In single competition in a single driver

Currently, solution ID is unique in a single competition (3), which seems enough for now, but if we need it to be unique across multiple competitions in a single driver it can be implemented without changing the `autopilot/driver` interface, since it is a `driver`s implementation detail.

# Changes

- [x] `Request` no longer optional.
- [x] Add comments for solution ID.

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -137,8 +137,8 @@ pub mod solve {
     #[derive(Clone, Debug, Default, Deserialize)]
     #[serde(rename_all = "camelCase", deny_unknown_fields)]
     pub struct Solution {
-        /// Unique ID of the solution, used to identify it in subsequent
-        /// requests (reveal, settle).
+        /// Unique ID of the solution (per driver competition), used to identify
+        /// it in subsequent requests (reveal, settle).
         #[serde_as(as = "serde_with::DisplayFromStr")]
         pub solution_id: u64,
         pub score: U256,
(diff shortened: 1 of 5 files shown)


fleupold (approved):

crates/autopilot/src/driver_model.rs
```diff
@@ -137,8 +137,8 @@ pub mod solve {
     #[derive(Clone, Debug, Default, Deserialize)]
     #[serde(rename_all = "camelCase", deny_unknown_fields)]
     pub struct Solution {
-        /// Unique ID of the solution, used to identify it in subsequent
-        /// requests (reveal, settle).
+        /// Unique ID of the solution (per driver competition), used to identify
+        /// it in subsequent requests (reveal, settle).
```
nlordell, line 141: Should we maybe make this unique per driver to prevent some unintended bugs?

nlordell: (fine as is, just food for thought).

sunce86: I think we can do this when we implement sending multiple solutions within `driver`. 
I captured this concern in the issue: https://github.com/cowprotocol/services/issues/1920

PR #1926, opened 2023-10-08
Fetch Recent Liquidity for Quotes

nlordell:
# Description

This PR was my take on recent liquidity fetching for the driver. Unfortunately, I did not see #1924 when I started working on it, but I believe some of my proposed changes are worthwhile including, so I rebased it onto the aforementioned PR.

In particular, this PR (just like the one its based on) uses recent instead of latest liquidity for quoting. This improves latency of quotes.

# Changes

- [x] Introduce a `liquidity::When` enum to specify when to fetch liquidity for (latest vs recent) instead of using a `bool` flag.
- [x] Add an E2E test to verify this. Note that this kind of E2E test is imperfect (since it involves timings, which is never great), but it is capturing the expected behavior of quotes (and because the cache update loop is long enough, it doesn't fail the many times I ran it).

## How to test

E2E test that verifies liquidity cache is not being updated for the `/quote` request.

### DIFF
--- a/crates/driver/src/boundary/liquidity/mod.rs
+++ b/crates/driver/src/boundary/liquidity/mod.rs
@@ -129,7 +129,7 @@ impl Fetcher {
     pub async fn fetch(
         &self,
         pairs: &HashSet<liquidity::TokenPair>,
-        allow_stale: bool,
+        when: infra::liquidity::When,
     ) -> Result<Vec<liquidity::Liquidity>> {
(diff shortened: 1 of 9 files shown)


fleupold (approved):
Nice test!

Not sure about the order of merging with #1925, but we should probably merge so that it makes it into Tuesday's release cut (to keep testing in xDAI prod).

PR #1938, opened 2023-10-10
Isolate Safe creation to dedicated function

fedgiac:
# Description
A small PR that isolates some boilerplate needed for generating Safes in e2e tests for future use.

# Changes

These changes are exclusively a refactor, no logic changes or new code involved.

## How to test

E2e tests still work.

### DIFF
--- a/crates/e2e/src/setup/onchain_components.rs
+++ b/crates/e2e/src/setup/onchain_components.rs
@@ -5,6 +5,8 @@ use {
         ERC20Mintable,
         GnosisSafe,
         GnosisSafeCompatibilityFallbackHandler,
+        GnosisSafeProxy,
+        GnosisSafeProxyFactory,
     },
     ethcontract::{transaction::TransactionBuilder, Account, Bytes, PrivateKey, H160, H256, U256},
     hex_literal::hex,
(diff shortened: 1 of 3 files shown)


crates/e2e/src/setup/onchain_components.rs
```diff
@@ -66,6 +68,55 @@ macro_rules! tx_safe {
...
+pub struct GnosisSafeInfrastructure {
+    pub factory: GnosisSafeProxyFactory,
+    pub fallback: GnosisSafeCompatibilityFallbackHandler,
+    pub singleton: GnosisSafe,
+    web3: Web3,
+}
+
+impl GnosisSafeInfrastructure {
+    pub async fn new(web3: &Web3) -> Self {
+        let singleton = GnosisSafe::builder(web3).deploy().await.unwrap();
```
MartinquaXD, line 80: Why is this called singleton?
Is it because it gets created with `create2` and always ends up with the same address? What happens if it gets deployed multiple times?

MartinquaXD (approved):

fleupold (commented):
Can't we just add an async deploy method on the existing Safe struct in that file? 
It already has a deployed method for loading existing Safes.

Currently all test cases instantiate the infra just to deploy a single Safe right after (I'm also not sure caching the infra is important for performance unless we were to create a lot safes) and it would avoid developers having to use two different types in their test.

Not related to this PR, but I think we should also consider breaking up onchain_components into smaller submodules for readability (with onchain_components/safe.rs being one)

crates/e2e/src/setup/onchain_components.rs
nlordell: > Why is this called singleton?

It is worth mentioning that this is the terminology used by the Safe contracts/docs/etc., so it makes sense to keep it in our code as well.

nlordell (approved):
🧹

fedgiac:
> Can't we just add an async deploy method on the existing Safe struct in that file? [...]

The `colocation_hooks` test deploys a safe through a hook, having the infrastructure available is important for the test since it's only an indirect deployment.
But I agree it's confusing that you can choose to use both `Safe` or `GnosisSafe` in a test. I plan to merge this PR as it is since it isolates some shared components, but later I'll update the `Safe` struct to replace standalone uses of `GnosisSafe`.

PR #1939, opened 2023-10-11
Add metadata.signer field to app data and reject mismatching orders

fedgiac:
# Description

Adds a new field to the app data, `metadata.signer`, that stores the address for which the order is signed.
If `metadata.signer` doesn't match the owner of the order, the order is rejected at creation time.
Note that it's still possible to omit the signer field or not provide some explicit metadata, in which case the order is still accepted.

Note that the changes also apply to order replacement, since it follows the same code path as order creation; this is however untested.

The new field follows the exact same logic as the `from` field in the order creation. We could consider deprecating the `from` field once we make providing app data mandatory.

[Related document](https://docs.google.com/document/d/1OpWXRZyqaQjEQqzX_d4AI5iphRV0Z4k5fCHnEs5jWwo/edit#heading=h.8aw19fkz5xgi).

# Next steps

Right now, it's possible to create an order without explicit app data that is later found to be from an invalid owner. A future PR will add an extra state to the order that marks it as invalid if an order becomes invalid when new information on the app data becomes available.

## How to test

New e2e test.


crates/model/src/order.rs
```diff
@@ -413,18 +413,44 @@ impl OrderCreation {
...
+                    Some(from)
+                } else {
+                    return Err(VerificationError::AppdataFromMismatch(
+                        AppdataFromMismatch {
+                            from,
+                            app_data_signer,
+                        },
+                    ));
+                }
+            }
```
nlordell, line 450: Rust supports stuff like this with `if` guards (which reduces the right-ward shift of the code already a bit):

```suggestion
            (Some(from), Some(app_data_signer)) if from == app_data_signer => Some(from),
            (Some(from), Some(app_data_signer)) => return Err(VerificationError::AppdataFromMismatch(
                AppdataFromMismatch {
                    from,
                    app_data_signer,
                },
            )),
```

nlordell (approved):
LGTM with an E2E test to boot!

fleupold (approved):
Looks good!

crates/model/src/order.rs
```diff
@@ -413,18 +413,41 @@ impl OrderCreation {
     }
 
     /// Recovers the owner address for the specified domain, and then verifies
-    /// it matches the expected address.
+    /// it matches the expected addresses.
+    ///
+    /// Expected addresses can come from two sources: directly from the order
+    /// creation and as part of the calldata.
```
MartinquaXD, line 419: Do you mean `appdata` here?
```suggestion
    /// creation and as part of the appdata.
```

crates/model/src/order.rs
```diff
@@ -489,11 +512,18 @@ impl OrderCreationAppData {
...
+    pub from: H160,
+    pub app_data_signer: H160,
+}
+
 #[derive(Debug)]
 pub enum VerificationError {
     UnableToRecoverSigner(anyhow::Error),
     UnexpectedSigner(signature::Recovered),
     MissingFrom,
+    AppdataFromMismatch(AppdataFromMismatch),
```
MartinquaXD, line 526: I think the new error needs to be added to the openapi spec.

MartinquaXD (approved):
Looks really good!

PR #1943, opened 2023-10-11
Auction ID to `solvers` Tracing Span

nlordell:
# Description

It turns out that we forgot to add `auction=$id` tracing spans to the `solvers` API handler. This PR does that to make debugging easier.

Note that we need to manually get the current span and move it into the new thread (and can't use `tracing::Instrumented::in_current_span`) because the blocking task gets executed on a separate thread outside of the `tokio` thread-pool which doesn't properly propagate spans.

# Changes

- [x] Modified the `auction::Id` abstraction to no longer use `Option` but have a more explicit enum type to better document the semantics of auction IDs (either it is a quote, or an auction with a specific numeric ID), instead of implying that `None` was used for quoting. This also helps in the logs filter out by spans when things are run as part of a quote and as part of an actual solver competition.
- [x] Added tracing span to the `solvers` API endpoint

## How to test

Run locally to see the span:

```
2023-10-11T15:14:17.261Z TRACE request{id="9"}:/solve:auction{id=quote}: solvers::domain::solver::baseline: finding route order=0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 request=Request { sell: Asset { amount: 1000000000000000000, token: TokenAddress(0xe91d153e0b41518a2ce8dd3d7944fa863463a97d) }, buy: Asset { amount: 1, token: TokenAddress(0xddafbb505ad214d7b80b1f830fccc89b60fb7a83) }, side: Sell }
```

### DIFF
--- a/crates/solvers/src/api/dto/auction.rs
+++ b/crates/solvers/src/api/dto/auction.rs
@@ -16,7 +16,10 @@ impl Auction {
     /// Converts a data transfer object into its domain object representation.
     pub fn to_domain(&self) -> Result<auction::Auction, Error> {
         Ok(auction::Auction {
-            id: self.id.map(auction::Id),
+            id: match self.id {
+                Some(id) => auction::Id::Solve(id),
+                None => auction::Id::Quote,
+            },
             tokens: auction::Tokens(
                 self.tokens
                     .iter()
--- a/crates/solvers/src/api/mod.rs
+++ b/crates/solvers/src/api/mod.rs
@@ -62,7 +62,7 @@ async fn solve(
         let auction_id = auction.id;
         let solutions = state
             .solve(auction)
-            .instrument(tracing::info_span!("auction", %auction_id))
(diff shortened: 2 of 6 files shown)


fleupold (approved):

MartinquaXD (approved):

crates/solvers/src/api/dto/auction.rs
```diff
@@ -16,7 +16,10 @@ impl Auction {
     /// Converts a data transfer object into its domain object representation.
     pub fn to_domain(&self) -> Result<auction::Auction, Error> {
         Ok(auction::Auction {
-            id: self.id.map(auction::Id),
+            id: match self.id {
+                Some(id) => auction::Id::Solve(id),
+                None => auction::Id::Quote,
+            },
```
sunce86, line 22: Why not having separate /quote api with it's own dto?

PR #1944, opened 2023-10-11
Refactor safe use in e2e tests

fedgiac:
# Description
Implements some [suggestion from Felix](https://github.com/cowprotocol/services/pull/1938#pullrequestreview-1670074705) on how we abstract a Safe multisig.

Note that some of the new abstractions are already used in #1939.

# Changes

In the first commit, some new abstractions are created under the `Safe` struct to reduce some of the boilerplate in the tests.
In the second commit, all code related to the safe is moved to its own submodule `onchain_components::safe`; the only changes are imports and `GnosisSafeInfrastructure` → `safe::Infrastructure`.

## How to test

See that all e2e tests continue to work.


crates/e2e/src/setup/onchain_components/safe.rs
```diff
@@ -0,0 +1,297 @@
...
+        DomainSeparator,
+    },
+    shared::ethrpc::Web3,
+    web3::{
+        signing,
+        signing::{Key, SecretKeyRef},
+    },
+};
+
+//#[macro_export]
```
MartinquaXD, line 23: I guess the `macro_export` can be dropped then?

MartinquaXD (approved):

crates/e2e/src/setup/onchain_components/safe.rs
nlordell: If you don't want to `macro_export`, I _think_ you can add (after the macro definition) a:

```rust
pub(crate) use self::tx_safe;
```

nlordell (approved):
Nice clean up.

Approving assuming the commented out `#[macro_export]` will be cleaned up somehow (either un-commented or removed).

PR #1953, opened 2023-10-12
Fix score deserialization

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/1951

The issue was that the decimal was interpreted as hex, so decimal `16442840735913131` becomes `25671225992463593777 `


MartinquaXD (approved):

MartinquaXD:
Will reenable colocation after this is merged.

PR #1954, opened 2023-10-12
Unify HTTP handling in the solvers crate

nlordell:
# Description

This PR unifies HTTP handling in the `solvers` crate in way that:
- All HTTP requests automatically trace relevant information (method, URL, body, status code, etc.) while preserving the callsite's module path in the log event (useful for filtering, so you can enable `trace` JUST for a specific DEX aggregator API for example).
- Improved error handling:
  1. We first try to parse with an API specific error type - this allows us to extract well-formed errors
  2. In the case that we got an HTTP success but failed to parse the response data, emit a JSON error
  3. In the case we got an HTTP error status code, emit an HTTP status error

# Changes

- [x] Added an `util::http::roundtrip` helper that can be used for HTTP requests.
- [x] Refactored existing HTTP requests to use this new utility.

## How to test

CI tests DEX aggregator with mocked services, so we can be confident that it continues to work with this refactor.

## Related Issues

Supersedes #1941

### DIFF
--- a/crates/solvers/src/util/http.rs
+++ b/crates/solvers/src/util/http.rs
@@ -12,6 +12,12 @@ use {
     std::str,
 };
 
+/// Roundtrip an HTTP request. This will `TRACE` log the request and responses.
+///
+/// This is a thin macro wrapper around [`roundtrip_internal`] that ensures that
+/// logs are attributed from the callsite and not from this module. This allows
+/// log filtering to be done in a more fine-grained manner and based on where
+/// the HTTP roundtripping is happening.
 macro_rules! roundtrip {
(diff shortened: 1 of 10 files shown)


crates/solvers/src/util/http.rs
```diff
@@ -0,0 +1,111 @@
...
+                } else {
+                    Error::Status(status, body)
+                })
+            })),
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
```
MartinquaXD, line 88: doc comment on each variant would be nice as their difference can be subtle.

MartinquaXD (approved):

PR #1964, opened 2023-10-13
[TRIVIAL] Add driver restart log

sunce86:
# Description
Adds a simple log to easily find when the driver restarted and with config.


fleupold (approved):
Are we logging sensitive information in the config? We may want to implement a custom write method to redact some secrets.

sunce86:
> Are we logging sensitive information in the config? We may want to implement a custom write method to redact some secrets.

Let's implement this in follow up since I need the log for debugging.

PR #1965, opened 2023-10-13
[TRIVIAL] Rename Flashbots in Colocated Driver

sunce86:
# Description
Renames `Flashbots` submission to `MEVBlocker` in `driver` crate. Now the `Flashbots` is only mentioned in `boundary` part of the `driver` which is allowed.


fleupold (approved):
🧹

PR #1969, opened 2023-10-13
Add instrument for submission kind

sunce86:
# Description
Adds mempool kind to every submission log, similar to what we have in non-colocated driver.

Didn't want to implement `Debug` nor `Display` just so to avoid any future misusage of ? and % in printing.

### DIFF
--- a/crates/driver/src/boundary/mempool.rs
+++ b/crates/driver/src/boundary/mempool.rs
@@ -52,10 +52,10 @@ pub enum Kind {
 
 impl Kind {
     /// for instrumentization purposes
-    pub fn format_variant(&self) -> String {
+    pub fn format_variant(&self) -> &'static str {
         match self {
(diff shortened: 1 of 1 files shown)


crates/driver/src/boundary/mempool.rs
```diff
@@ -49,6 +50,16 @@ pub enum Kind {
     },
 }
 
+impl Kind {
+    /// for instrumentization purposes
+    pub fn format_variant(&self) -> String {
```
MartinquaXD, line 55: Probably makes sense to return a &'static str to avoid allocating over and over.

fleupold (approved):
You could alternatively implement `Display` for `Kind`

crates/driver/src/boundary/mempool.rs
sunce86: Indeed. Fixed.

PR #1991, opened 2023-10-19
Use simulated gas for score computation

MartinquaXD:
# Description
@sunce86 [suggested](https://github.com/cowprotocol/services/pull/1960/files#r1363834421) to use the more accurate simulated gas amount for DEX swaps for the gas computation.

# Changes
Instead of computing the score before turning the swap into a solution (which is where we get the accurate gas) we now also compute the score when turning the swap into a solution.

### DIFF
--- a/crates/solvers/src/domain/dex/mod.rs
+++ b/crates/solvers/src/domain/dex/mod.rs
@@ -4,7 +4,7 @@
 
 use {
     crate::{
-        domain::{auction, eth, order, solution},
+        domain::{self, auction, eth, order, solution},
         infra,
         util,
     },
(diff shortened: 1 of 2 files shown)


fleupold (approved):

MartinquaXD:
>  Is it possible to do similar for Baseline? do we have simulator available there?

AFAICS this is not easily doable because the baseline solver does not build the calldata for the auction. It simply takes the liquidity provided by the driver and returns a list of ids (referring to those LPs) and the input arguments for each. Then the driver will build the final calldata (which would be needed for the simulation).

PR #1999, opened 2023-10-20
Make /settle call blocking and report tx_hash

MartinquaXD:
# Description
Currently the autopilot wastes a lot of time waiting for transactions that will never appear. This happens because the driver's `/settle` endpoint operates in a fire-and-forget (get request, kick off submission in background, return immediately). That way the driver does not have any way to communicate that it will not be able to submit the solution to the autopilot and it will have to monitor the blockchain until the deadline is reached.

# Changes
This PR makes it so that `/settle` actually blocks in the driver until the solution gets submitted and returns the tx_hash. If the driver is not able to submit the solution (e.g. the simulations for the solution start reverting) it will return an error to the autopilot which will immediately move to the next auction.

Note that this is not really how it's supposed to work but this solution is fine as long as we are running all the drivers (and can therefore assume a reasonable behavior). We should revisit this decision before we can encourage external parties to run their own driver.

## How to test
e2e tests

### DIFF
--- a/crates/autopilot/src/driver_model.rs
+++ b/crates/autopilot/src/driver_model.rs
@@ -194,6 +194,7 @@ pub mod reveal {
 pub mod settle {
     use {
         model::bytes_hex,
+        primitive_types::H256,
         serde::{Deserialize, Serialize},
         serde_with::serde_as,
     };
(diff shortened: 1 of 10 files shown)


fleupold (approved):
I believe you also have to set a custom timeout on the http request as otherwise `/settle` will likely return after 10s like all our other http requests.

PR #2003, opened 2023-10-20
Forward solution ID through notifications

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2002

Now, with each notification, we are sending both auction_id and solution_id which should be enough for all debugging purposes.

Solution ID is represented as `Vec<u64>` since it can contain multiple solutions, as in case of merged settlements.

### DIFF
--- a/crates/solvers/src/domain/solution.rs
+++ b/crates/solvers/src/domain/solution.rs
@@ -23,8 +23,14 @@ impl From<u64> for Id {
 }
 
 impl Id {
+    /// Generates a new solution id that is unique per /solve request.
     pub fn generate() -> Self {
         Id(COUNTER.fetch_add(1, Ordering::Relaxed))
+    }
+
+    /// Resets the solution id counter to 0.
+    pub fn reset() {
+        COUNTER.store(0, Ordering::Relaxed);
     }
 }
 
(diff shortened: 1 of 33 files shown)


crates/solvers/src/domain/solution.rs
```diff
@@ -4,12 +4,31 @@ use {
...
-    std::{collections::HashMap, slice},
+    std::{
+        collections::HashMap,
+        slice,
+        sync::atomic::{AtomicU64, Ordering},
+    },
 };
 
+static COUNTER: AtomicU64 = AtomicU64::new(1);
+pub fn solution_id() -> Id {
```
MartinquaXD, line 15: nit: Now you'd have to use this as `solution::solution_id()`. Maybe it's slightly nicer to read to have it as a member on `Id` like `Id::generate()` or `Id::next()`.

MartinquaXD (approved):

sunce86:
I've updated the PR according to comment from: https://github.com/cowprotocol/services/issues/2002#issuecomment-1773722203

Since we don't want to send notifications for merged settlements, then `solution_id` can be a single value.

crates/solvers/src/domain/solution.rs
sunce86: Yes, added.

sunce86:
I had to drop COUNTER because of issues with parallel calls of `solve` endpoint.

PR #2006, opened 2023-10-20
Drop access list from `eth::Tx` `Debug` impl

MartinquaXD:
# Description
Drop the `access_list` field from the `Debug` impl because it's pretty noisy and not very helpful for debugging purposes.

### DIFF
--- a/crates/driver/src/domain/eth/mod.rs
+++ b/crates/driver/src/domain/eth/mod.rs
@@ -328,8 +328,7 @@ impl std::fmt::Debug for Tx {
             .field("from", &self.from)
             .field("to", &self.to)
             .field("value", &self.value)
-            .field("input", &hex::encode(&self.input.0))
-            .field("access_list", &self.access_list)
+            .field("input", &self.input)
             .finish()
     }
 }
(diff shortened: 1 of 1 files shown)


sunce86 (approved):
> note that Bytes already implements hex string printing

Ah, didn't know. My bad.

sunce86:
The only reason I think it might be smart to have custom debug impl is to drop access list printing since it's not important for the simulation result (except for out of gas error which don't happen)

MartinquaXD:
> The only reason I think it might be smart to have custom debug impl is to drop access list printing since it's not important for the simulation result (except for out of gas error which don't happen)

But this is not the case in the current code either, right?
Happy to drop the access list printing, just checking if I'm missing something here.

sunce86:
> > The only reason I think it might be smart to have custom debug impl is to drop access list printing since it's not important for the simulation result (except for out of gas error which don't happen)
> 
> But this is not the case in the current code either, right? Happy to drop the access list printing, just checking if I'm missing something here.

Yup, not the case, I just noticed it.

fedgiac (approved):

PR #2011, opened 2023-10-24
Colocation notify - bad score

sunce86:
# Description
Implements #4 from https://github.com/cowprotocol/services/issues/1891

# Changes
- [ ] All specific errors that occur during score calculation are propagated to solver engine.
- [ ] A bit of type safety changes to make the `driver` domain logic cleaner.

### DIFF
--- a/crates/driver/src/domain/competition/score.rs
+++ b/crates/driver/src/domain/competition/score.rs
@@ -40,9 +40,11 @@ pub struct SuccessProbability(pub f64);
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("objective value is non-positive")]
-    ObjectiveValueNonPositive,
+    ObjectiveValueNonPositive(#[from] boundary::settlement::Error),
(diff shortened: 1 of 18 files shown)


crates/driver/src/domain/competition/score.rs
```diff
@@ -33,16 +36,68 @@ impl From<eth::U256> for Score {
...
+#[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
+pub struct ObjectiveValue(pub eth::U256);
+
+impl From<eth::U256> for ObjectiveValue {
+    fn from(value: eth::U256) -> Self {
+        Self(value)
+    }
+}
+
+impl std::cmp::PartialEq<ObjectiveValue> for Score {
```
MartinquaXD, line 79: Maybe a small comment why comparing `ObjectiveValue` with `Score` is reasonable would be cool.

MartinquaXD (approved):
Looks alright to me.

PR #2012, opened 2023-10-24
Dont reset node during test

MartinquaXD:
# Description
Ever since we replaced `hardhat` with `anvil` we got flaky tests.
The issue seems to be related to deadlocks inside `anvil`. It's possible that these deadlocks could be related to running the same node instance for all tests and resetting it over and over again.

# Changes
Drop `anvil` from the `docker-compose` file and instead spawn a fresh `anvil` instance for every `e2e` test.

## How to test
Rerun CI a couple of times and :pray: that we don't see flaky errors?

### DIFF
--- a/crates/e2e/src/nodes/mod.rs
+++ b/crates/e2e/src/nodes/mod.rs
@@ -1,9 +1,113 @@
 pub mod forked_node;
 pub mod local_node;
 
+/// The default node URL that should be used for e2e tests.
 pub const NODE_HOST: &str = "http://127.0.0.1:8545";
 
-#[async_trait::async_trait(?Send)]
-pub trait TestNode {
-    async fn reset(&self);
+/// A blockchain node for development purposes. Dropping this type will
+/// terminate the node.
+pub struct Node {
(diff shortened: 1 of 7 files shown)


crates/e2e/src/nodes/mod.rs
```diff
@@ -1,9 +1,113 @@
...
+
+    /// Spawn a new node instance using the list of given arguments.
+    async fn spawn_process(args: &[&str]) -> Self {
+        use tokio::io::AsyncBufReadExt as _;
+
+        // Allow using some custom logic to spawn `anvil` by setting `ANVIL_COMMAND`.
+        // For example if you set up a command that spins up a docker container.
+        let command = std::env::var("ANVIL_COMMAND").unwrap_or("anvil".to_string());
+
+        let mut process = tokio::process::Command::new(command)
```
MartinquaXD, line 49: Mostly taken from the node implementation in the `driver` test crate. I think sharing this implementation is probably not super useful so I just copied it here.

sunce86 (approved):
Let's try it.

PR #2013, opened 2023-10-24
Colocation Notify - failed simulation

sunce86:
# Description
Implements (2) from https://github.com/cowprotocol/services/issues/1891

# Changes
- [ ] Simulation failure error (together with tx data) propagated from the `driver` to the `solvers`.

Unfortunately, not all data can (and should) be populated to support the legacy version. `block_number` is one of those that is needed but it is not provided by simulators. This is the never ending issue we generally have with doing simulations on a specific block number.

The easiest fix would be to add get_block_number() call to each web3 simulation (enso and tenderly seem to give block_number in their responses), but it wouldn't be accurate.

At the moment I decided to skip it, since, after all, solvers can pretty much figure out the block number on their own, there are usually 1 or 2 candidates...


crates/driver/src/infra/simulator/mod.rs
```diff
@@ -142,17 +142,17 @@ pub enum SimulatorError {
...
-    err: SimulatorError,
-    tx: eth::Tx,
+    pub err: SimulatorError,
+    pub tx: eth::Tx,
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("basic: {0:?}")]
+    #[error("{0:?}")]
```
MartinquaXD, line 151: I **believe** this is what `#[error(transparent)]` is for.

crates/driver/src/infra/solver/dto/notification.rs
```diff
@@ -64,12 +70,23 @@ pub struct Notification {
...
     DuplicatedSolutionId,
 }
 
+#[serde_as]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Tx {
+    pub from: H160,
+    pub to: H160,
+    pub input: Vec<u8>,
```
MartinquaXD, line 86: Should probably be adjusted to send a hex string instead of vector of numbers.

crates/driver/src/infra/solver/dto/notification.rs
```diff
@@ -64,12 +70,23 @@ pub struct Notification {
...
+    pub value: U256,
```
MartinquaXD, line 87: `#[serde_as(as = "HexOrDecimalU256")]`

crates/solvers/src/api/routes/notify/dto/notification.rs
```diff
@@ -74,12 +80,23 @@ pub struct Notification {
...
 }
 
+#[serde_as]
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct Tx {
+    from: H160,
+    to: H160,
+    input: Vec<u8>,
+    value: U256,
```
MartinquaXD, line 97: Same serialization comment here.

crates/solvers/src/boundary/legacy.rs
```diff
@@ -578,6 +581,22 @@ fn to_boundary_auction_result(notification: &notification::Notification) -> (i64
...
         Kind::EmptySolution => AuctionResult::Rejected(SolverRejectionReason::NoUserOrders),
+        Kind::SimulationFailed(tx) => AuctionResult::Rejected(
+            SolverRejectionReason::SimulationFailure(TransactionWithError {
+                error: "".to_string(),
+                transaction: SimulatedTransaction {
+                    from: tx.from.into(),
+                    to: tx.to.into(),
+                    data: tx.input.clone().into(),
+                    internalization: InternalizationStrategy::SkipInternalizableInteraction,
+                    block_number: Default::default(),
```
MartinquaXD, line 592: Still needs to be populated.

crates/solvers/src/boundary/legacy.rs
```diff
@@ -578,6 +581,22 @@ fn to_boundary_auction_result(notification: &notification::Notification) -> (i64
...
     let auction_result = match &notification.kind {
         Kind::EmptySolution => AuctionResult::Rejected(SolverRejectionReason::NoUserOrders),
+        Kind::SimulationFailed(tx) => AuctionResult::Rejected(
+            SolverRejectionReason::SimulationFailure(TransactionWithError {
+                error: "".to_string(),
+                transaction: SimulatedTransaction {
+                    from: tx.from.into(),
+                    to: tx.to.into(),
+                    data: tx.input.clone().into(),
+                    internalization: InternalizationStrategy::SkipInternalizableInteraction,
```
MartinquaXD, line 591: I think the internalization flag should probably not be hardcoded. The simulation could also only fail if internalized AFAIK.

crates/solvers/src/boundary/legacy.rs
```diff
@@ -578,6 +581,22 @@ fn to_boundary_auction_result(notification: &notification::Notification) -> (i64
...
+                    block_number: Default::default(),
+                    tx_index: Default::default(),
```
MartinquaXD, line 593: Does it make sense to have the `tx_index` here?
If we are simulating on the latest block the `tx_index` shouldn't make a difference because the simulator can't know what other transactions would come before our tx_index, right?

crates/solvers/src/domain/eth/mod.rs
```diff
@@ -37,9 +39,51 @@ pub struct Asset {
...
+/// An onchain transaction.
+#[derive(Clone)]
+pub struct Tx {
+    pub from: Address,
+    pub to: Address,
+    pub value: Ether,
+    pub input: Bytes<Vec<u8>>,
+}
+
+impl std::fmt::Debug for Tx {
```
MartinquaXD, line 80: Appears to be identical to simply using `#[derive(Debug)]`

crates/solvers/src/util/bytes.rs
```diff
@@ -0,0 +1,30 @@
+/// A thin wrapper around a collection of bytes. Provides hex debug formatting.
```
MartinquaXD, line 1: We have by now 3 different ways how to handle this.
1. `Bytes` wrapper
2. [`hex.rs`](https://github.com/cowprotocol/services/blob/main/crates/solvers/src/util/fmt/hex.rs)
3. using `Derivative` + `shared::debug_bytes`

I think we should pick one and use it everywhere but IMO using the wrapper type should not be the way. This forces you to go through another layer of indirection even when you don't use the `Debug` implementation which doesn't seem great.

crates/solvers/src/boundary/legacy.rs
fedgiac: My understanding from the description is that this is intentionally left unpopulated, same as the one below.
Still, I'd like to see something like `block_number: PLACEHOLDER_UNAVAILABLE_BLOCK` where `PLACEHOLDER_UNAVAILABLE_BLOCK = 0` that clarifies this is intentionally left blank.

fedgiac (commented):
I find it a bit strange that we introduce a few new identical `Tx` types rather than reusing `eth::Tx`.
I also don't understand the impact of populating the legacy notification with fake data for `block_number` and `tx_index`.
Except for this, I've nothing more to add.

crates/driver/src/infra/solver/dto/notification.rs
```diff
@@ -64,12 +70,23 @@ pub struct Notification {
...
     ScoringFailed(ScoreKind),
     NonBufferableTokensUsed(BTreeSet<H160>),
     SolverAccountInsufficientBalance(U256),
     DuplicatedSolutionId,
 }
 
+#[serde_as]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Tx {
```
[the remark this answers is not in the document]
sunce86: The reason we have this struct is to adhere the DDD that wants to separate the domain object and dto object (where the serialization is encapsulated).
The reason I dropped access list is that access lists don't influence simulation status (success/revert) therefore are not needed to be part of this struct that is mainly used to reproduce the simulation.

crates/solvers/src/boundary/legacy.rs
sunce86: Not really useful, at least we did not use it, I am not sure why it was originally added.

crates/solvers/src/domain/eth/mod.rs
sunce86: Nice catch. Copy paste.

crates/solvers/src/api/routes/notify/dto/notification.rs
sunce86: added

crates/driver/src/infra/solver/dto/notification.rs
sunce86: used `#[serde_as(as = "serialize::U256")]` for consistency. 

replacing all occurences with `#[serde_as(as = "HexOrDecimalU256")]` should be probably done in a separate PR.

crates/driver/src/infra/solver/dto/notification.rs
sunce86: true

crates/driver/src/infra/simulator/mod.rs
sunce86: Not sure if transparent calls for Display but should be the irrelevant in this case.

crates/solvers/src/boundary/legacy.rs
sunce86: I aimed here to actually drop the `internalization` flag as it is not needed IMO, since all solver engines are sending raw interactions to the `driver` and, in case of simulation failure, they get the failed transaction with calldata, so that would be enough to see if it was internalized or not (if the calldata in notification differs from what they sent to us, and they have `solution id` to match on).

So, I just want to be careful and add the bare minimum of data needed (to reduce complexity).

Here, I would rather add additional variant `InternalizationStrategy::Undefined` or similar, to make the code accurate and not misleading, instead of providing actual value.

WDYT?

crates/solvers/src/util/bytes.rs
sunce86: Would you like to create a cleanup ticket for this to make it consistent across repo? might be a good first issue.

crates/solvers/src/boundary/legacy.rs
sunce86: @fedgiac makes sense, added also the reference to the tracked issue https://github.com/cowprotocol/services/issues/2018

crates/driver/src/infra/solver/dto/notification.rs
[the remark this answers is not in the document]
fedgiac: > The reason we have this struct is to adhere the DDD that wants to separate the domain object and dto object (where the serialization is encapsulated).

I see, it's about separating core logic and interface, if I understood DDD design principle. All good then.

> access lists don't influence simulation status (success/revert)

This isn't the case as far as I know, for example sending ETH to a safe could fail if the access list doesn't account for it.

crates/solvers/src/boundary/legacy.rs
sunce86: Added `InternalizationStrategy::Unknown`

crates/solvers/src/util/bytes.rs
sunce86: Tracked https://github.com/cowprotocol/services/issues/2030

crates/driver/src/infra/solver/dto/notification.rs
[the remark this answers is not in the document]
sunce86: I tried to find if this is still valid, since we already talked about dropping access lists completely. 
However, in the meantime, I added access list as you suggested. I used the `web3::types::AccessList` as I assume this one is the most convenient since all the data is already envisioned to be used to recreate tx for simulation.

crates/solvers/src/boundary/legacy.rs
MartinquaXD: Sorry, saw the follow up comments too late.
Wouldn't it then make more sense to drop the `internalization` here altogether? Seems like in all the existing use cases having the `Unknown` variant doesn't really make sense and therefore compromising the use of an existing type for this doesn't seem worth it.

sunce86: Yes we can definitely drop it. Maybe I was too cautious to not cause any backward compatibility issues since this is after all a debugging feature.

sunce86: Tried removing but we are using this field to recreate tenderly link in the non-colocated driver also, so we will have to remove it only once the colocation is activated. 
I added the comment in the code to remember this.

MartinquaXD (approved):

sunce86:
Merging so I can observe it in staging before tomorrow colocation run in prod. @fedgiac feel free to add more comments

PR #2019, opened 2023-10-26
Move duplicated ids check to competition::solve

sunce86:
# Description
Pretty fix: moves the check to main competition::solve where all the checks are done.

# Changes
- [ ] Solutions are filtered instead of failing early for all solutions.
- [ ] Added metric and fixed log for a check to be searchable with `discarded solution:` prefix


fedgiac (approved):
Looks reasonable to me.
I checked and I _think_ the only other place where we use `Solver::solve` is to get a quote in `domain::Quote::quote`, where we anyway don't do any other validation of the incoming solutions.

PR #2022, opened 2023-10-26
Colocation notify - bad scores refactor and fixes

sunce86:
# Description
Contains various fixes around scores validation. Currently, score validation is messy and very complicated. I tried to summarize all checks to only 4 of them:

1. Score > 0
2. Score <= Quality, where Quality = surplus + fees.
3. SuccessProbability in range [0,1]
4. ObjectiveValue > 0, where ObjectiveValue = Quality - GasCost

(1) and (2) are checked for all scores as per CIP20. 
(3) and (4) are checked only for `RiskAdjusted` scores (scores based on success probability used by `ScoreCalculator`), and they represent validation of the input to the `ScoreCalculator` which demands these checks in order for computation to be meaningful.

# Changes
- [ ] Moved out `ObjectiveValueNonPositive` and `SuccessProbabilityOutOfRange` so that all errors from score calculator are now boundary errors as they should be (I don't want to import `ScoringError` into `boundary`)
- [ ] Fixed a bug: `score > quality` instead of `score > objective_value`
- [ ] `ScoreHigherThanObjective` removed completely since this is an internal error in score calculator and it should never happen unless there is a bug in the code. Anyway, this is definitely not a special case I want to handle.
- [ ] `TooHighScore` replaced with `ScoreHigherThanQuality`


fhenneke (commented):
I was not able to actually review the code.

From my understanding of our auction mechanism:
- (1) and (2) are required from the CIP. (1) (and (2)?) could also be checked by the autopilot when ranking. But having those checks in the driver makes sense too.
- (3) is about submitting correct success probabilities. It needs to be checked, otherwise the computation of the score might do unexpected things. E.g. it can lead to scores larger than quality.
- (4) is a reasonable check. It should not strictly be necessary to check this since it will also automatically result in scores smaller than zero. But it is useful debugging info. Note that [this line](https://github.com/cowprotocol/services/blob/bb1b428cae4922b7de96c305d5b35e33df59746b/crates/solver/src/settlement_rater.rs#L468) is what computes the score in case of negative objective. The computed value can be negative even without the objective being negative though (due to additional costs in case of failure). So some form of check for score > 0 is required anyways.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -268,20 +268,26 @@ impl Settlement {
...
-                let failure_cost =
-                    matches!(revert_protection, mempools::RevertProtection::Disabled)
-                        .then(|| self.gas.estimate * auction.gas_price())
-                        .unwrap_or(zero());
+                let failure_cost = match revert_protection {
+                    mempools::RevertProtection::Enabled => zero(),
+                    mempools::RevertProtection::Disabled => gas_cost,
+                };
+                let objective_value = quality - gas_cost;
+                if objective_value.is_zero() {
```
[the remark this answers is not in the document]
sunce86: Yes, since negative scores are converted to 0 since U256 is unsigned...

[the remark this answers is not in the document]
fhenneke: If that applies to other parts of the code, then it might be useful to filter out nonpositive  objectives (as in `quality <= gas_costs`). The formulas probably do work as intended in case of objective  0 and result in scores of 0. But there might be edge cases I have not thought about.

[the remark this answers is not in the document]
sunce86: Do you think I should apply this check in both cases? (solver provided direct score vs solver provided success probability)?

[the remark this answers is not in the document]
fhenneke: With direct score submission, we should probably not check `quality > gas_cost`. At least this is a restriction not mentioned in the CIP and not strictly necessary. For example if our cost estimates are slightly off this could lead to exclusion of good solutions.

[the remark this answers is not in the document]
sunce86: > With direct score submission, we should probably not check quality > gas_cost.

Ok, this is the current state of the code. For direct score submission, (1) and (2) are checked. 

Another question, if (3) and (4) passed in score calculator, can it ever be true `score > objective_value` ?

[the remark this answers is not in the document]
sunce86: > Another question, if (3) and (4) passed in score calculator, can it ever be true score > objective_value

Based on the UTs and my understanding, score should be at max equal to objective value, but just want to confirm.

crates/driver/src/domain/competition/score.rs
```diff
@@ -76,31 +89,101 @@ impl From<eth::U256> for ObjectiveValue {
...
     ZeroScore,
-    #[error("objective value {0:?} is higher than the objective {1:?}")]
-    ScoreHigherThanObjective(Score, ObjectiveValue),
-    #[error("success probability is out of range {0:?}")]
-    SuccessProbabilityOutOfRange(SuccessProbability),
-    #[error("invalid objective value")]
+    /// Protocol does not allow solutions that are claimed to be "better" than
+    /// the actual value they bring (quality). It is expected that score
+    /// is always lower than quality, because there is always some
+    /// revert risk that needs to be incorporated into the score and lower it.
```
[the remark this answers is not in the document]
sunce86: thanks

crates/driver/src/domain/competition/solution/settlement.rs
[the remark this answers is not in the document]
fhenneke: If (3) (and (4)) holds then we should always have `score <= objective_value`. I.e. `score > objective_value` is _not_ possible then.

crates/solver/src/settlement_rater.rs
```diff
@@ -362,10 +358,15 @@ impl ScoreCalculator {
             self.score_cap.clone(),
         )?;
         if optimal_score > *objective_value {
-            return Err(ScoringError::ScoreHigherThanObjective(
-                optimal_score,
-                objective_value.clone(),
-            ));
+            tracing::error!(
+                "Sanity check failed, score higher than objective, should never happen unless a \
+                 bug in a computation"
```
fedgiac, line 363: Nit: "unless _there's_ a bug". (Same in line below.)
Maybe it makes sense to also print the values? If we see that they differ by a rounding error we could start looking in different places compared to seeing two completely different values.

crates/driver/src/domain/competition/score.rs
```diff
@@ -76,31 +89,102 @@ impl From<eth::U256> for ObjectiveValue {
...
+            Self(value)
+        }
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Error {
+        /// Solution has success probability that is outside of the allowed
+        /// range [0, 1]
+        #[error("success probability is out of range {0:?}")]
+        SuccessProbabilityOutOfRange(SuccessProbability),
```
fedgiac, line 179: The existence of this error is very weird to me, as it shouldn't be called a probability in the first place if it isn't in this range. How is it possible that we get a nonsense probability?

crates/driver/src/infra/notify/mod.rs
```diff
@@ -24,23 +24,21 @@ pub fn scoring_failed(
...
+        score::Error::RiskAdjusted(score::risk::Error::SuccessProbabilityOutOfRange(
+            success_probability,
+        )) => notification::Kind::ScoringFailed(
+            notification::ScoreKind::SuccessProbabilityOutOfRange(*success_probability),
+        ),
+        score::Error::RiskAdjusted(score::risk::Error::ObjectiveValueNonPositive) => {
+            notification::Kind::ScoringFailed(notification::ScoreKind::ObjectiveValueNonPositive)
         }
+        score::Error::RiskAdjusted(score::risk::Error::Boundary(_)) => return,
         score::Error::Boundary(_) => return,
```
fedgiac, line 42: Is it expected that we don't notify about anything if any of these errors occur?

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -282,12 +282,6 @@ impl Settlement {
                     mempools::RevertProtection::Disabled => gas_cost,
                 };
                 let objective_value = quality - gas_cost;
-                if objective_value.is_zero() {
-                    return Err(score::Error::RiskAdjusted(
-                        score::risk::Error::ObjectiveValueNonPositive,
-                    ));
-                }
-
```
fedgiac: I was looking at this commit and I don't know why this was removed or where it's handled now. Is it redundant now?

crates/driver/src/domain/competition/score.rs
```diff
@@ -54,19 +41,45 @@ impl num::Zero for Score {
...
 }
 
+/// ObjectiveValue = Quality - GasCost
+impl std::ops::Sub<GasCost> for Quality {
+    type Output = ObjectiveValue;
+
+    fn sub(self, other: GasCost) -> Self::Output {
+        ObjectiveValue(self.0.saturating_sub(other.0 .0))
+    }
+}
```
fedgiac, line 63: I don't really like to silently overload `ObjectiveValue(0)` to mean `ObjectiveValue(≤0)`.
It means that you always need to know if the objective value you are using has been "cleaned" by the function you are using to ignore zero or not.

What could be nice is that if we have an `ObjectiveValue` then it's also valid. We can say that this implementation of `Add` return an `Option<ObjectiveValue>` instead , which is `None` if `objective_value ≤ 0`.

crates/driver/src/domain/competition/score.rs
```diff
@@ -76,31 +89,102 @@ impl From<eth::U256> for ObjectiveValue {
...
-impl std::cmp::PartialEq<ObjectiveValue> for Score {
-    fn eq(&self, other: &ObjectiveValue) -> bool {
-        self.0.eq(&other.0)
+impl std::ops::Add for ObjectiveValue {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        Self(self.0 + other.0)
     }
 }
```
fedgiac, line 98: We implement `Add` for `ObjectiveValue + ObjectiveValue`. But if `self.0 == 0` or `other.0 == 0` then we can't really return an output that makes sense because of the saturating sub. I believe this can't happen right now, however it looks like it could be a source of issues eventually.

fedgiac (commented):
To my understanding, this PR does everything it says in the description and the code looks correct.

crates/driver/src/domain/competition/score.rs
```diff
@@ -76,31 +89,102 @@ impl From<eth::U256> for ObjectiveValue {
...
+    /// Errors only applicable to scores that use success probability.
+    #[error(transparent)]
+    RiskAdjusted(#[from] risk::Error),
+    #[error(transparent)]
     Boundary(#[from] boundary::Error),
 }
+
+pub mod risk {
+    /// Contains functionality and error types for scores that are based on
+    /// success probability.
```
MartinquaXD, line 134: nit: Should probably be a doc comment.
```suggestion
    //! Contains functionality and error types for scores that are based on
    //! success probability.
```

crates/driver/src/domain/competition/score.rs
```diff
@@ -76,31 +89,102 @@ impl From<eth::U256> for ObjectiveValue {
...
+    use {
+        super::{ObjectiveValue, Score},
+        crate::{boundary, domain::eth},
+        bigdecimal::Zero,
+    };
+
+    /// Constructs a score based on the success probability of a solution.
+    impl Score {
+        pub fn new(
```
MartinquaXD, line 143: Doc comment needs to be on the function itself.
```suggestion
    impl Score {
        /// Constructs a score based on the success probability of a solution.
        pub fn new(
```

crates/driver/src/domain/competition/score.rs
MartinquaXD: This could be handled by implementing `TryFrom` instead of `From`. Not sure if this would make the error handling more annoying because the transformation from `f64` to `SuccessProbability` might happen where we don't have the required types to report this error back.

crates/solver/src/settlement_rater.rs
MartinquaXD: There are few enough arguments going into this function that I think we could simply log all of them in the error. That way we wouldn't have to figure out how to get the relevant data in case we ever encounter this error.

MartinquaXD (approved):
Looks alright to me.

crates/driver/src/domain/competition/solution/settlement.rs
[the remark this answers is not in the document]
sunce86: We can resolve this as we agreed on mutual understanding;

crates/driver/src/domain/competition/score.rs
```diff
@@ -54,19 +41,45 @@ impl num::Zero for Score {
     }
 }
 
-/// Represents the probability that a solution will be successfully settled.
-#[derive(Debug, Copy, Clone)]
-pub struct SuccessProbability(pub f64);
+/// Represents the observed quality of a solution. This is not an artifical
+/// value like score. This is a real value that solution provides and it's
+/// defined as surplus + fees.
```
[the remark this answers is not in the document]
sunce86: With `artificial value` I wanted to say that score is a number that can be different for each solver even for the same solution. So, it's not something that is (like objective value) 100% deterministic once you have a solution. But it's rather `a number` that each solver attaches to a solution to bid to win the competition.
But I understand I failed to explain this properly. Your suggestion makes more sense, to keep it simple.

crates/solver/src/settlement_ranker.rs
```diff
@@ -284,11 +281,9 @@ impl SettlementRanker {
...
                 solver.notify_auction_result(
                     auction_id,
-                    AuctionResult::Rejected(SolverRejectionReason::TooHighScore {
-                        surplus,
-                        fees,
-                        max_score,
-                        submitted_score: settlement.score.score(),
+                    AuctionResult::Rejected(SolverRejectionReason::ScoreHigherThanQuality {
+                        score: settlement.score.score(),
+                        quality: max_score,
```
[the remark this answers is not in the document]
sunce86: Makes a lot of sense.

crates/solver/src/settlement_rater.rs
sunce86: Fixed comment and added more parameters to log.

crates/driver/src/domain/competition/score.rs
sunce86: https://github.com/cowprotocol/services/pull/2022/commits/39858e03f8c9b0bcdec8ead269ae5f094584db78

crates/driver/src/domain/competition/solution/settlement.rs
sunce86: This was moved to `score()` function, and with https://github.com/cowprotocol/services/pull/2022/commits/39858e03f8c9b0bcdec8ead269ae5f094584db78 is moved to `Sub` implementation (Quality - GasCost)

crates/driver/src/domain/competition/score.rs
sunce86: Ah, It slipped my mind completely that I can define Output as Option :)
Thanks, updated https://github.com/cowprotocol/services/pull/2022/commits/39858e03f8c9b0bcdec8ead269ae5f094584db78

crates/driver/src/domain/competition/score.rs
sunce86: Resolved with https://github.com/cowprotocol/services/pull/2022/commits/a7ecb3813946118799b1a932d636b042f15cb773

fedgiac (approved):

sunce86:
Aaand I also ended up making `Score` nonzero U256.  https://github.com/cowprotocol/services/pull/2022/commits/d24d64971edf53ee71765d78a35469e8da06a835

crates/driver/src/infra/notify/mod.rs
sunce86: For now yes, it is expected. We still need to figure out the boundary line - which kind of internal errors we want to propagate to solvers.

crates/driver/src/domain/competition/score.rs
sunce86: Fixed.

crates/driver/src/domain/competition/solution/settlement.rs
```diff
@@ -268,20 +268,21 @@ impl Settlement {
...
         revert_protection: &mempools::RevertProtection,
     ) -> Result<competition::Score, score::Error> {
-        let objective_value = self
-            .boundary
-            .objective_value(eth, auction, self.gas.estimate)?;
+        let quality = self.boundary.quality(eth, auction)?;
 
         let score = match self.boundary.score() {
-            competition::SolverScore::Solver(score) => competition::Score(score),
+            competition::SolverScore::Solver(score) => competition::Score(score.try_into()?),
```
MartinquaXD, line 274: I think one of these `.try_into()?` lead to a `/notify` request not being sent.
For example in this [run loop](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2023-10-31T19:37:22.545Z',to:'2023-10-31T19:49:56.704Z'))&_a=(columns:!(log),filters:!(),grid:(columns:('@timestamp':(width:164))),index:ea511870-d9b3-11ed-a9d0-a17451f01cc1,interval:auto,query:(language:kuery,query:'log:%20%2211712235%22%20and%20log:%20%22seasolver%22'),sort:!(!('@timestamp',desc))))

sunce86: Let me check...

sunce86: fixed error handling https://github.com/cowprotocol/services/pull/2031

How we ended up with [score::Solver(0)](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/ea511870-d9b3-11ed-a9d0-a17451f01cc1/cowlogs-staging-2023.10.31?id=rLRBh4sBQgsUfyaRQinM) is still an enigma for me.

sunce86: Apparently seasolver used `scoreDiscount` variant which is deprecated and the deserialization should fallback to `RiskAdjusted` variant with `1.0` success probability.

PR #2027, opened 2023-10-31
build /notify URL respecting multiple path segments

MartinquaXD:
# Description
The current logic to build the `/notify` URL doesn't handle URLs that contain multiple path segments correctly.
Take this http solver URL for example:
`https://miep.cow.solver/1/solve?api_key=1234#row=3`
which consists of the base `https://miep.cow.solver` and the `solve_endpoint` `/1/solve?api_key=1234#row=3`.
The old logic would have computed the notify endpoint to be `https://miep.cow.solver/notify` whereas `https://miep.cow.solver/1/notify?api_key=1234#row=3` actually makes more sense. (`/notify` as a sibling to `/solve`).

# Changes
Always make `/notify` a sibling of `/solve` (including all query parameters and fragments).

## How to test
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d4168862014d3876c28c19153b4267c

### DIFF
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -230,7 +230,9 @@ impl HttpSolverApi for DefaultHttpSolverApi {
     }
 
     fn notify_auction_result(&self, auction_id: AuctionId, result: model::AuctionResult) {
-        let mut url = crate::url::join(&self.base, "notify");
+        let mut url = crate::url::join(&self.base, &self.solve_path);
+        // `/notify` should be a sibling of the `/solve` endpoint
+        url.path_segments_mut().unwrap().pop().push("notify");
 
         let client = self.client.clone();
         let config_api_key = self.config.api_key.clone();


sunce86 (approved):
We should probably check with external solvers if this change brakes something on their side, since AFAIU we go from `https://miep.cow.solver/notify` to `https://miep.cow.solver/1/notify`

sunce86:
I like the Rust playground example 👍

MartinquaXD:
> We should probably check with external solvers if this change brakes something on their side, since AFAIU we go from https://miep.cow.solver/notify to https://miep.cow.solver/1/notify

I noticed that because the `SeaSolver` team said they don't receive any `/notify` request on their staging solver but it works on their prod solver. This suggests a different treatment between colocation vs. current setup.
In the current scheme we for example only configure `https://solvername.xyz/1/` as the base url and let the default logic add `/solve` automatically.
In the colocated world we configure `https://solvername.xyz/1/solve` as the endpoint and split the URL into `https://solvername.xyz` and `1/solve` (to make it compatible with the legacy http component). This then adds `/notify` to the shortened base url.
Colocation actually broke something but I think most people don't notice or care because it's currently just on staging.

PR #2031, opened 2023-10-31
Properly propagate errors for zero scores

sunce86:
# Description
Implements proper error handling when building `Score` from `ZeroU256`.
Accidentally, this was done through `eth::NonZeroU256::from` and then the error was converted to `anyhow::Error`, but what we actually want is to convert to `score::Error::ZeroScore`.

It was noticed in [this log](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/ea511870-d9b3-11ed-a9d0-a17451f01cc1/cowlogs-staging-2023.10.31?id=_bRBh4sBQgsUfyaRQinM).

### DIFF
--- a/crates/driver/src/domain/competition/score.rs
+++ b/crates/driver/src/domain/competition/score.rs
@@ -1,9 +1,5 @@
 use {
-    self::risk::ObjectiveValue,
-    crate::{
-        boundary,
-        domain::{eth, eth::GasCost},
-    },
+    crate::{boundary, domain::eth},
     std::cmp::Ordering,
 };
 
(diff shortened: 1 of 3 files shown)


crates/driver/src/domain/competition/score.rs
```diff
@@ -37,21 +33,6 @@ impl From<eth::U256> for Quality {
...
-        if self.0 > other.0 .0 {
-            Ok(ObjectiveValue(
-                eth::NonZeroU256::new(self.0 - other.0 .0).unwrap(),
-            ))
-        } else {
-            Err(risk::Error::ObjectiveValueNonPositive)
-        }
-    }
-}
-
```
sunce86, line 54: just moved 👇 since it's correlated strictly to `ObjectiveValue`

MartinquaXD (approved):
Looks alright to me but I usually have very low confidence when reviewing anything related to error conversions.

sunce86:
> Looks alright to me but I usually have very low confidence when reviewing anything related to error conversions.

We should be good with this one since commenting out the `impl TryFrom<eth::U256> for Score` leads to compiler errors at two critical `try_into()?` calls that now return proper `score::Error` type.

PR #2033, opened 2023-11-01
Fix parsing score_cap

sunce86:
# Description
@fhenneke noticed that our `score_cap` is [parsed wrongly](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/c0e240e0-d9b3-11ed-b0e6-e361adffce0b/cowlogs-prod-2023.11.01?id=icFRiosBGP6wkTcjJhYX).

Value `1e16` is parsed as `18446744073709551616` due to decimal/hexadecimal misconfiguration.


MartinquaXD (approved):
Just how much pain can the fact actually cause that `U256::from_str()` and `U256::to_string()` use different number system?! 😭

PR #2038, opened 2023-11-01
Add chain_id to /notify query string

devanoneth:
# Description
Add `chain_id` to the query variables on the `/notify` request. We would appreciate this for the Enso solver as all of our infra is multi chain by default. With this we can monitor which chain the notification relates to.

Originally I wanted to add a `metadata` field to the `AuctionResult`, but this accomplishes the same thing and requires no model changes.

# Changes
Add `chain_id` to the query variables on the `/notify` request.

## How to test
Couldn't find an existing `/notify` tests, if they do exist please let me know and I can add some assertions. It is also just a really small change...

## Related Issues

https://github.com/cowprotocol/services/issues/1891#issuecomment-1779295895

### DIFF
--- a/crates/shared/src/http_solver.rs
+++ b/crates/shared/src/http_solver.rs
@@ -234,12 +234,16 @@ impl HttpSolverApi for DefaultHttpSolverApi {
         // `/notify` should be a sibling of the `/solve` endpoint
         url.path_segments_mut().unwrap().pop().push("notify");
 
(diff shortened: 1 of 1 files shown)


sunce86 (approved):
This should cover both legacy and colocated `notify`. Thanks.

MartinquaXD (approved):

PR #2040, opened 2023-11-03
Respect deadline in Baseline solver

sunce86:
# Description
Baseline solver iterates through orders from `auction` and executes some logic for each order in a synchronous way. It checks the deadline before each order. It could happen that the new order logic begins to execute 1ms before the deadline and lasts arbitrary amount of time (500ms?) and significantly break the deadline.

This code adds a checker if the deadline is reached in the main thread, and if it is, ends the background work prematurely and returns whatever was executed up until that point.

## How to test
Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0e23fa016153883faad8b0e7148b4dfa

### DIFF
--- a/crates/solvers/src/domain/solver/baseline.rs
+++ b/crates/solvers/src/domain/solver/baseline.rs
@@ -19,7 +19,7 @@ use {
         infra::config,
     },
     ethereum_types::U256,
-    std::{cmp, collections::HashSet, sync::Arc, time::Duration},
+    std::{cmp, collections::HashSet, sync::Arc},
 };
 
 pub struct Baseline(Arc<Inner>);
@@ -71,31 +71,26 @@ impl Baseline {
         // 100s of ms.
         tracing::trace!("solving {} orders", auction.orders.len());
 
-        let (sender, mut receiver) = tokio::sync::mpsc::channel(auction.orders.len());
+        let (sender, mut receiver) = tokio::sync::mpsc::unbounded_channel();
+        let deadline = auction.deadline.remaining().unwrap_or_default();
 
         let inner = self.0.clone();
         let span = tracing::Span::current();
-        let background_work = tokio::spawn(async move {
+        let background_work = async move {
             let _entered = span.enter();
             inner.solve(auction, sender).await;
-        });
+        };
 
-        // make sure the deadline is respected
-        let deadline = Duration::from_millis(150);
-        let begin = std::time::Instant::now();
+        if tokio::time::timeout(deadline, tokio::spawn(background_work))
+            .await
+            .is_err()
+        {
+            tracing::debug!("reached timeout while solving orders");
+        }
 
         let mut solutions = vec![];
-        while begin.elapsed() < deadline {
-            match receiver.try_recv() {
-                Ok(solution) => solutions.push(solution),
-                Err(_) => {
-                    if background_work.is_finished() {
-                        break;
-                    }
-                    // unblock sync code and wait for channel to populate
-                    tokio::time::sleep(Duration::from_millis(10)).await;
-                }
-            }
+        while let Ok(solution) = receiver.try_recv() {
+            solutions.push(solution);
         }
         tracing::trace!("solved {} orders", solutions.len());
         solutions
@@ -106,7 +101,7 @@ impl Inner {
     async fn solve(
         &self,
         auction: auction::Auction,
-        sender: tokio::sync::mpsc::Sender<solution::Solution>,
+        sender: tokio::sync::mpsc::UnboundedSender<solution::Solution>,
     ) {
         let boundary_solver =
             boundary::baseline::Solver::new(&self.weth, &self.base_tokens, &auction.liquidity);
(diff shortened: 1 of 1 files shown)


crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -69,80 +69,109 @@ impl Baseline {
         // not lock up the [`tokio`] runtime and cause it to slow down handling
         // the real async things. For larger settlements, this can block in the
         // 100s of ms.
+        tracing::trace!("solving {} orders", auction.orders.len());
+
+        let (sender, mut receiver) = tokio::sync::mpsc::channel(auction.orders.len());
```
MartinquaXD, line 74: The over all strategy makes sense to me but I think the code can probably be collapsed a lot.
Most of the time we don't get a solution for a lot of orders. Therefore I would go for `mpsc::unbounded_channel()`. (BTW since you continuously queried the `receiver` a `oneshot` channel would probably have been fine as well)
When we have that we can probably make things more concise by doing something like:
```rust
        let (sender, mut receiver) = tokio::sync::mpsc::unbounded_channel();
        let time_left = tokio::time::Duration::from_millis(150); // (actually compute from auction deadline)
        if tokio::time::timeout(time_left, self.0.solve(auction, sender)).await.is_err() {
            tracing::debug!("reached timeout while solving orders");
        }
        let mut solutions = vec![];
        while let Ok(solution) = receiver.try_recv() {
            solutions.push(solution);
        }
        tracing::trace!("solved {} orders", solutions.len());
```
Technically this will probably run into issues if we encounter an endless loop in the worker thread. If we are worried about that I think we should continue to use `tokio::task::spawn_blocking` and if not doing everything in the same task seems fine to me.

MartinquaXD (commented):
Looks good overall. Just thinking the code can probably be reduced a bit.

crates/solvers/src/domain/solver/baseline.rs
sunce86: i think this doesn't fix the `500ms` issue I mentioned in the description ☝️ 
`tokio::time::timeout` will not return at exactly `time_left` but just after each order is finished.
That's why I introduced the outer loop in main thread to control the maximum time I want to spend in background thread.

MartinquaXD: Yeah, this is why I added the caveat about the endless loop (or your 500ms example). But the main point was: Either we are worried about it then we should continue to use `spawn_blocking` or we are not worried about it then we can probably get away with not spawning anything and just have a few ms time buffer.
But the the compressed code would still work with `task::spawn_blocking()`:
```rust
        let inner = self.0.clone();
        let task = || async move { inner.solve(auction, sender).await };
        let timeout_future = tokio::time::timeout(time_left, tokio::task::spawn_blocking(task));
        if timeout_future.await.is_err() {
        // ...
```
This should be equivalent to your handrolled loop since `poll`ing the spawned task does not block if something inside the spawned task is blocking so the `timeout` should be spot on.

sunce86: Can't make it work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=992e9102decd4d87eab797dc00800b31

MartinquaXD: Curious that `spawn_blocking` indeed does not seem to work but using `spawn` does. ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=06d555a08901a4ef128adfe2a0a7edba))
AFAIK these function should behave the same with the difference that `spawn_blocking` takes a thread from a thread pool where long blocking is not supposed to be a big issue.

crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -69,80 +69,99 @@ impl Baseline {
...
+                        output,
+                        interactions,
+                        gas: route.gas(),
+                    }
+                    .into_solution(auction.gas_price, sell_token, score)?
+                    .with_id(solution::Id(i as u64))
+                    .with_buffers_internalizations(&auction.tokens),
+                )
+            });
+            if let Some(solution) = solution {
```
sunce86, line 158: New code is only from here to the bottom of the function. The rest is copy paste.

crates/solvers/src/domain/solver/baseline.rs
```diff
@@ -69,80 +69,109 @@ impl Baseline {
...
-                                    input: segment.input,
-                                    output: segment.output,
-                                    // TODO does the baseline solver know about this
-                                    // optimization?
-                                    internalize: false,
-                                })
+        for (i, order) in auction.orders.into_iter().enumerate() {
+            let sell_token = auction.tokens.reference_price(&order.sell.token);
+            let user_order = UserOrder::new(&order);
+            if let Some(user_order) = user_order {
```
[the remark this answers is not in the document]
sunce86: Interesting 👍 added.

crates/solvers/src/domain/solver/baseline.rs
sunce86: Yeah. I refactored the solution to be aligned with the example. Also, `unbounded_channel` makes sense.

MartinquaXD (approved):

PR #2042, opened 2023-11-06
Cache balances in driver

MartinquaXD:
# Description
When the driver receives a `/solve` request it fetches the balances for all traders in the auction. The outcome of this exclusively depends on the current block so duplicating this work for every configured solver does not make sense.
This will result in fewer RPC requests and probably also in a faster auction pre-processing time.
The solvers are not really contending on a lock but they are using the same underlying buffered `Web3` type. This is configured to allow up to 10 parallel batch requests with 20 individual requests each. These 200 requests can probably be easily saturated when 10 solvers fetch balances for ~1000 orders which means there will be some solvers who have to wait for the backpressure to resolve.

# Changes
Introduce shared cache for balances which gets updated when the solver queries the balances for an uncached block.
The cache locks until the update is complete so it will avoid all duplicated requests.
Since the balances might get quite big I decided to return a reference counted balances map instead of cloning it in the critical section. Otherwise these big `.clone()`s might add up to a significant delay as well.

## How to test
should be covered by e2e test

### DIFF
--- a/crates/driver/src/domain/eth/balances.rs
+++ b/crates/driver/src/domain/eth/balances.rs
@@ -1,0 +1,95 @@
+use {
+    crate::{
+        domain::{competition::order, eth},
+        infra,
+    },
+    futures::future::join_all,
+    itertools::Itertools,
+    std::{collections::HashMap, sync::Arc},
+    tokio::sync::Mutex,
+};
+
+type BalanceGroup = (order::Trader, eth::TokenAddress, order::SellTokenBalance);
+type Balances = Arc<HashMap<BalanceGroup, order::SellAmount>>;
+
+#[derive(Default, Clone)]
+pub struct Cache(Arc<Mutex<Inner>>);
+
+impl Cache {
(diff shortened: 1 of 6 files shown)


crates/driver/src/domain/eth/balances.rs
```diff
@@ -0,0 +1,95 @@
...
+    itertools::Itertools,
+    std::{collections::HashMap, sync::Arc},
+    tokio::sync::Mutex,
+};
+
+type BalanceGroup = (order::Trader, eth::TokenAddress, order::SellTokenBalance);
+type Balances = Arc<HashMap<BalanceGroup, order::SellAmount>>;
+
+#[derive(Default, Clone)]
+pub struct Cache(Arc<Mutex<Inner>>);
```
sunce86, line 16: nit: would it make sense to use RwLock here since once the first solver populates the balances there is no need to require the reading lock for the rest of the solvers?

sunce86 (approved):
Nice find!

PR #2043, opened 2023-11-06
Configurable http time buffer

sunce86:
# Description
Makes the solver http time buffer configurable. 

This indeed seems like a value we want to setup once and never touch again, which why I don't feel strong to add parameter for this (instead of just updating the hard coded value). 

OTOH, we need this configurable at least until further notice.

### DIFF
--- a/crates/driver/src/infra/config/file/mod.rs
+++ b/crates/driver/src/infra/config/file/mod.rs
@@ -148,6 +148,10 @@ fn default_soft_cancellations_flag() -> bool {
     false
 }
 
+fn default_http_time_buffer_miliseconds() -> u64 {
+    1500
+}
+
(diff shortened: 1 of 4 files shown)


crates/driver/src/infra/config/file/mod.rs
```diff
@@ -148,6 +148,10 @@ fn default_soft_cancellations_flag() -> bool {
     false
 }
 
+fn default_http_time_buffer_miliseconds() -> u64 {
```
MartinquaXD, line 151:
```suggestion
fn default_http_time_buffer_milliseconds() -> u64 {
```

MartinquaXD (approved):
Very useful for the next test run 👍

PR #2044, opened 2023-11-06
Notify timeout early

sunce86:
# Description
Timeouts in legacy solvers are handled inside `solvers` crate and in those cases empty solutions vector is sent to `driver`, leading to these errors being classified in the `driver` as `empty` solutions instead of `timeout`. The consequence of this is that legacy solvers are not properly notified.

This PR sends the `timeout` notification directly from within `legacy` solver so there is no ambiguity. Note that there is no way for `driver` to know this is a `timeout`, so additional `empty` notification will be sent, but afterwards, so it can be ignored by `legacy` solvers.

### DIFF
--- a/crates/solvers/src/domain/solver/legacy.rs
+++ b/crates/solvers/src/domain/solver/legacy.rs
@@ -27,10 +27,13 @@ impl Legacy {
     }
 
     pub async fn solve(&self, auction: auction::Auction) -> Vec<solution::Solution> {
-        match self.0.solve(auction).await {
+        match self.0.solve(&auction).await {
             Ok(solution) => vec![solution.with_id(solution::Id(0))],
             Err(err) => {
                 tracing::warn!(?err, "failed to solve auction");
+                if err.is_timeout() {
+                    self.notify_timeout(auction.id)
+                }
                 vec![]
             }
         }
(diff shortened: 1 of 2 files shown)


crates/solvers/src/domain/solver/legacy.rs
```diff
@@ -27,10 +27,13 @@ impl Legacy {
     }
 
     pub async fn solve(&self, auction: auction::Auction) -> Vec<solution::Solution> {
-        match self.0.solve(auction).await {
+        match self.0.solve(&auction).await {
             Ok(solution) => vec![solution.with_id(solution::Id(0))],
             Err(err) => {
                 tracing::warn!(?err, "failed to solve auction");
+                if err.is_timeout() {
```
MartinquaXD, line 34: nit: If `.is_timout()` is only used in one place it might be nicer to not have the specialized function at all.
Feel free to ignore, though.
```suggestion
                if matches!(err, Error::Timeout) {
```

MartinquaXD (approved):

crates/solvers/src/domain/solver/legacy.rs
sunce86: Yes, I thought one would need less cognitive effort to read `is_timeout` then the matches! version 😄 . But don't feel strongly about it.

Issue #2041, opened 2023-11-06
bug: Colocated solvers timing out on prod mainnet


MartinquaXD:
### Problem
When testing colocated drivers on prod mainnet we saw basically all of them timing out.
This is concerning for 2 reasons:
* it looks like something is making the whole process surprisingly slow (maybe contention on some lock 🤷‍♂️)
* our timeout management in the `driver` is not working. Ideally we should never see timeouts on the autopilot side. Instead the `driver` should abort the computation right before the deadline and return 0 solutions.

### Impact
Colocation is basically not usable like this on prod.

### To reproduce
Only way is to apply colocation on prod mainnet because that is where we have the required traffic to trigger the issues.

### Expected behaviour
The driver should never time out and we should not see a reduced number of proposed solutions compared to the current setup.

### Screenshots/logs
[logs](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2023-11-02T10:21:31.579Z',to:'2023-11-02T11:00:20.182Z'))&_a=(columns:!(log),filters:!(),grid:(columns:('@timestamp':(width:164))),index:c0e240e0-d9b3-11ed-b0e6-e361adffce0b,interval:auto,query:(language:kuery,query:'mainnet%20and%20(log:%20%22timeout%22%20or%20log:%20%22timed%20out%22)%20and%20driver'),sort:!(!('@timestamp',desc))))

sunce86:
A bit more information:

`Autopilot` gives a `driver` a deadline to return the solution. Currently, 15s are given to `driver`.
`Driver` then gives 14s to `solvers` for calculation. Approximately ~1s is left for `driver` to simulate solutions, do the merging, simulate merged settlements, do the scoring...

There are several issues here:
1. `Driver` is left with 1sec, but this deadline is not enforced anywhere, leading to timeout responses in `autopilot`.
2. 1sec is hardcoded and cant be changed via config.
3. 1sec is same for all solvers, which is not appropriate. Some solvers return a single solution while others return a lot of solutions. Some solvers return early, others spend full time given to them (14s).
4. Network delay for sending the response back to `autopilot` is not taken into consideration.

I would suggest the following:

- [ ] Add time buffer for network delay (500ms) and make it configurable. The value is the same for all solvers. Subtract this time from the auction::Deadline on the `driver` side.
[...]

PR #2045, opened 2023-11-07
Speed up balance fetching

MartinquaXD:
# Description
The driver has a routine to prioritize orders and filter orders based on the balance the trader currently has.
With the introduction of pre-interactions this check became much trickier because a user might not have any sell_tokens at the moment but unstakes the needed tokens in a pre-interaction. Because of that we introduced a way to get the actual tradable balance by simulating a balance check using a helper contract.
This is very accurate but unfortunately also very slow. (~10s just for fetching the balances of all traders in an auction)
Luckily most of our orders don't have pre-interactions so we can fall back to a simpler and faster way to query the current balance.

# Changes
Reimplement parts of the balance query of [balances.sol](https://github.com/cowprotocol/services/blob/main/crates/contracts/solidity/Balances.sol#L59-L76) using simple `eth_call`s and use that for any order group that does not contain any pre-interactions.
While testing locally this reduced the time to fetch all balances to ~ 1.2 - 2 seconds.

## How to test
covered by e2e tests

## Related Issues
https://github.com/cowprotocol/services/issues/2041

### DIFF
--- a/crates/driver/src/infra/blockchain/token.rs
+++ b/crates/driver/src/infra/blockchain/token.rs
@@ -1,6 +1,8 @@
 use {
     super::{Error, Ethereum},
     crate::domain::{competition::order, eth},
+    contracts::{dummy_contract, BalancerV2Vault},
+    futures::TryFutureExt,
 };
 
 /// An ERC-20 token.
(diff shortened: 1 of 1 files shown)


crates/driver/src/domain/competition/auction.rs
```diff
@@ -83,7 +83,7 @@ impl Auction {
     /// competition.
     pub async fn prioritize(mut self, eth: &Ethereum, balances: &eth::balances::Cache) -> Self {
         // Sort orders so that most likely to be fulfilled come first.
-        self.orders.sort_by_key(|order| {
+        self.orders.sort_by_cached_key(|order| {
```
MartinquaXD, line 86: I threw this in on top.
Not really related to balance fetching but speeds up sorting from ~2.5s to ~0.2s and doesn't need a separate PR.

sunce86 (approved):
Lg

PR #2046, opened 2023-11-07
Speed up sorting

MartinquaXD:
# Description
Sorting orders based on their price achievability takes a surprisingly long amount of time.

# Changes
Used `sort_by_cached_key()` to speed up sorting from ~1.5s to ~0.2s (prod mainnet auction).
I also used `spawn_blocking()` to run the sorting task since regular tokio tasks are not supposed to be blocked for more than 100ms according to this [blog article](https://ryhl.io/blog/async-what-is-blocking/) (although I'm not sure whether these numbers are still accurate since the tokio runtime started to keep track of every task's compute budget).

## How to test
e2e


sunce86 (approved):
👏

PR #2048, opened 2023-11-07
Increase gas limit

sunce86:
# Description
The point of gas limit factor is to take into consideration possible small changes in gas estimation from block to block during submission.
However, recently seasolver submitted solutions on gnosis chain that have a significant amount of gas refunds. Gas refunds are refunded at the very end of the execution in the EVM, so if the gas limit is set too low, peak gas usage during EVM execution can exceed it and return `OutOfGas` error.

Setting the factor to `2.0` seems safe, since I don't think nothing catastrophic can happen even if we spend double gas in case of an error.

# Changes
Adjusted factor for both submission but also for colocated driver settlement encoding where factor is used to calculate the required eth balance of solver. Although the latter one doesn't seem necessary, I decided to do it anyway for consistency.


MartinquaXD (approved):
Nice! I dreaded debugging this issue. 😬

PR #2049, opened 2023-11-08
Shared auction pre process step

MartinquaXD:
# Description
Follow up to https://github.com/cowprotocol/services/pull/2046.

# Changes
Now we only spawn a single prioritization task that is shared across all solvers because they are doing all the same anyway.

I also moved it entirely to a thread that is okay to block with synchronous code. This means no more duplicate work.
AFAIU This should also improve the throughput when more drivers are configured than the CPU has available cores. Because in those scenarios each driver would have receive the auction and would eventually block it's runtime thread for hundreds of milliseconds by doing CPU bound computations. And since we have more drivers than CPU cores the driver `CPU_CORES + 1` could only start working on their pre-processing when one of the earlier drivers finished.

## How to test
e2e


crates/driver/src/domain/competition/auction.rs
```diff
@@ -75,39 +80,124 @@ impl Auction {
...
+        self.deadline
+    }
+
+    pub fn score_cap(&self) -> Score {
+        self.score_cap
+    }
+}
+
+#[derive(Clone)]
+pub struct PreProcessor(Arc<Mutex<Inner>>);
```
sunce86, line 116: nit: Maybe `AuctionProcessor`

sunce86 (approved):
LG.

One thing I dislike is that we mix the domain logic of sorting/filtering orders with the logic  of how this task is executed (which feels like a non-domain logic), but I also don't have a better suggestion how to solve it.

crates/driver/src/domain/competition/auction.rs
MartinquaXD: I thought this is fine since we tend to use namespaces a lot more in the new code and therefore also reference types as such.
In this case that would result in using the type externally as `auction::AuctionProcessor`.

MartinquaXD:
> One thing I dislike is that we mix the domain logic of sorting/filtering orders with the logic of how this task is executed (which feels like a non-domain logic), but I also don't have a better suggestion how to solve it.

Yeah, something about this solution felt wrong to me as well. I opened the PR regardless because I think this is something we should definitely have for the next test on prod. 
I still have 1 more idea in mind how we could speed up the auction pre-processing. I will try to come up with a nicer separation of concerns when implementing that.

PR #2050, opened 2023-11-08
Fix Quasilabs URL configuration

sunce86:
# Description
During colocation prod test, quasilabs endpoint was misconfigured. In the meantime, I also noticed that it is misconfigured in the staging also.

Quasilabs is the only solver that has endpoint authorization defined as `https://user:pass@host.com/solve`. Based on the current splitting of paths, user and pass get lost. This PR keeps the user and pass as part of the `base`.

# Changes

- [ ] Changed the way we do url split.

## How to test
Updated unit test.

### DIFF
--- a/crates/shared/src/url.rs
+++ b/crates/shared/src/url.rs
@@ -1,7 +1,4 @@
-use {
-    anyhow::{Context, Result},
-    url::Url,
-};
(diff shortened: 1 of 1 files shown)


MartinquaXD (approved):
Nice! 👍

PR #2053, opened 2023-11-10
Colocation - Notify asset flow

sunce86:
# Description
Noticed that some of the solvers are not populating `inputs` and `outputs` fields in `interactions` data, leading to solutions from these solvers being discarded in colocated driver. 
This PR adds a new notification type to inform solvers about their bad expectation, but is also useful for debugging in case a solver does not populate these fields (since the missmatches will be full amounts of tokens in that case).

Missmatches take the range `[-U256, U256]` so `String` seemed like the easiest way to serialize it.


MartinquaXD (approved):
LGTM

Issue #2018, opened 2023-10-26
feat: Simulator to return block number if errored


sunce86:
### Problem
`Driver` simulators (web3, enso, tenderly) should return `block_number` in case they errored on access list estimation or gas estimation.

### Suggested solution
Enso and Tenderly should have this out of the box, while for web3 we can add simple `eth().block(Latest)` call before simulation (we do the same in the non-colocated driver).

### Additional context
Needed to populate the field https://github.com/cowprotocol/services/pull/2013#discussion_r1371229380

PR #2054, opened 2023-11-13
Add block number to simulation error

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2018

# Changes

- [ ] Added block number to each simulation failure. The block number is approximately fetched from the `web3`. 
- [ ] Used this block number for `notify` calls
- [ ] Extracted `notify::simulation_failure` as it is important enough to have it's own function

### DIFF
--- a/crates/driver/src/domain/competition/mod.rs
+++ b/crates/driver/src/domain/competition/mod.rs
@@ -224,6 +224,14 @@ impl Competition {
                         tracing::warn!(block = block.number, ?err, "solution reverts on new block");
                         *score_ref = None;
                         *self.settlement.lock().unwrap() = None;
+                        if let Some(id) = settlement.notify_id() {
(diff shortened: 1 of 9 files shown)


crates/driver/src/domain/competition/mod.rs
```diff
@@ -224,6 +224,9 @@ impl Competition {
                         tracing::warn!(block = block.number, ?err, "solution reverts on new block");
                         *score_ref = None;
                         *self.settlement.lock().unwrap() = None;
+                        if let Some(id) = settlement.notify_id() {
```
sunce86, line 227: Here we have IF because we decided to not send the notification for merged solutions. While this is okkish now since external solvers are sending a single solution anyway, it may become problematic in the future with combinatorial auctions...

MartinquaXD (approved):

PR #2055, opened 2023-11-14
Use OnceCell to manage liquidity source initialization

MartinquaXD:
# Description
We currently have an issue that some solvers receive only uni v2 based liquidity and usually at most 1 solvers receives balancer or uni v3 based liquidity. This is caused by the too simplistic initialization logic inside `BackgroundInitLiquiditySource`. It was written at a time when it would not be queried in parallel (for which the implementation would be fine) but for the new driver (which can query it multiple times in parallel) it results in wrong liquidity availability.
The issue is basically that the old code used `try_lock` because it assumed only the background task running a (potentially) slow initialization logic could use it in parallel. In that case we wanted to return an empty result early.
However with the new code an initialized liquidity source might be locked because it is currently used by another task.

# Changes
The new logic uses `OnceCell` to manage the access to the underlying data source. It enforces that it can only be initialized once and therefore doesn't has to take a lock on `.get()` request. Now multiple solvers can use the liquidity source in parallel and therefore receive the juicy uni v3 and balancer v2 liquidity.

### DIFF
--- a/crates/solver/src/liquidity_collector.rs
+++ b/crates/solver/src/liquidity_collector.rs
@@ -2,9 +2,9 @@ use {
     crate::liquidity::Liquidity,
     anyhow::Result,
     model::TokenPair,
+    once_cell::sync::OnceCell,
     shared::{baseline_solver::BaseTokens, recent_block_cache::Block},
     std::{collections::HashSet, future::Future, sync::Arc, time::Duration},
-    tokio::sync::Mutex,
     tracing::Instrument,
 };
 
@@ -51,7 +51,7 @@ impl LiquidityCollecting for LiquidityCollector {
 /// succeeds. Until the liquidity source has been initialised no liquidity will
 /// be provided.
 pub struct BackgroundInitLiquiditySource<L> {
-    liquidity_source: Arc<Mutex<Option<L>>>,
+    liquidity_source: Arc<OnceCell<L>>,
 }
 
 impl<L> BackgroundInitLiquiditySource<L> {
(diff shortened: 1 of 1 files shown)


crates/solver/src/liquidity_collector.rs
```diff
@@ -82,7 +82,13 @@ impl<L> BackgroundInitLiquiditySource<L> {
...
                         }
                         Ok(source) => {
-                            let _ = inner.lock().await.insert(source);
+                            if inner.set(source).is_err() {
+                                // should never happen but if it does we want to know about it ASAP
+                                tracing::error!(
+                                    source = inner_label,
+                                    "liquidity source already initialized"
+                                );
+                            }
```
squadgazzz, line 91: nit: I understand that should never happen, but wouldn't it be slightly clearer that the metrics are incremented only when the `OnceCell::set` succeeds?
```suggestion
                             if inner.set(source).is_err() {
                                // should never happen but if it does we want to know about it ASAP
                                tracing::error!(
                                    source = inner_label,
                                    "liquidity source already initialized"
                                );
                            } else {
                                tracing::debug!("successfully initialised liquidity source");
                                Metrics::get()
                                    .liquidity_enabled
                                    .with_label_values(&[&inner_label])
                                    .inc();
                            }
```

MartinquaXD: Yeah, that's fair. 👍

squadgazzz (approved):

crates/solver/src/liquidity_collector.rs
```diff
@@ -82,12 +82,20 @@ impl<L> BackgroundInitLiquiditySource<L> {
...
-                            let _ = inner.lock().await.insert(source);
-                            tracing::debug!("successfully initialised liquidity source");
-                            Metrics::get()
-                                .liquidity_enabled
-                                .with_label_values(&[&inner_label])
-                                .inc();
+                            if inner.set(source).is_err() {
+                                // should never happen but if it does we want to know about it ASAP
+                                tracing::error!(
+                                    source = inner_label,
```
sunce86, line 88: nit: probably not needed since we have label instrumented, feel free to ignore

sunce86 (approved):
lg

PR #2065, opened 2023-11-17
Share liquidity fetch requests

MartinquaXD:
# Description
Because all the traits and APIs around liquidity fetching are very weird it's pretty hard to keep track of which liquidity pools are not worth fetching.

# Changes
1. This PR introduces a request sharing logic in the `RecentBlockCache`. With that it shouldn't matter how many solvers have to fetch liquidity in parallel as they should all share the underlying RPC requests.
2. Additionally I modified the logic that keeps track of recently used keys to only contain ids of pools we were able to fetch. This is important because those keys are used to update the cache in a background task and if we have too many useless keys in there we'd waste a lot of work during those background updates.
3. Moved `RequestSharing` garbage collection to background task instead of at every insert. This is mostly relevant because we now use this component for liquidity fetching which issues **MANY** requests. This improvement is actually pretty huge.

Those changes (together with the previous 2 liquidity changes) brought down the liquidity fetching time from ~8s to 0.7-1.5s which should be fast enough for the gnosis solvers.

Fixes: https://github.com/cowprotocol/services/issues/2041


crates/shared/src/recent_block_cache.rs
```diff
@@ -397,18 +446,40 @@ mod tests {
...
     }
 
     fn test_keys(keys: impl IntoIterator<Item = usize>) -> impl Iterator<Item = TestKey> {
         keys.into_iter().map(TestKey)
     }
 
-    #[test]
-    fn marks_recently_used() {
-        let fetcher = FakeCacheFetcher::default();
+    #[tokio::test]
```
MartinquaXD, line 476: These tests had to be marked as async tests because `RequestSharing` that gets now used under the hood spawns a task which would panic on a sync test.

crates/shared/src/recent_block_cache.rs
```diff
@@ -217,7 +255,9 @@ where
 
         let mut cache_hit_count = 0usize;
         let mut cache_hits = Vec::new();
-        let mut cache_misses = HashSet::new();
+        // Use BTreeSet to ensure deterministic iteration order later which enables
+        // request sharing.
+        let mut cache_misses = BTreeSet::new();
```
MartinquaXD, line 260: I think this is no longer needed because I ended up request sharing every individual request instead of a whole chunk.
Will research on next week.

crates/driver/src/boundary/liquidity/mod.rs
```diff
@@ -38,7 +38,7 @@ const BLOCK_POLL_INTERVAL: Duration = Duration::from_secs(1);
 fn cache_config() -> CacheConfig {
     CacheConfig {
         number_of_blocks_to_cache: NonZeroU64::new(10).unwrap(),
-        number_of_entries_to_auto_update: NonZeroUsize::new(200).unwrap(),
+        number_of_entries_to_auto_update: NonZeroUsize::new(1000).unwrap(),
```
MartinquaXD, line 41: Increased the threshold because we'd like to update **ALL** liquidity sources in the background so we ideally only have to read from the cache when a new auction appears.
Eventually we should make these values configurable.

crates/shared/src/recent_block_cache.rs
```diff
@@ -180,33 +184,67 @@ where
...
-        let fetch = || self.fetcher.fetch_values(keys.clone(), block);
-        for _ in 0..self.maximum_retries {
-            match fetch().await {
-                Ok(values) => return Ok(values),
-                Err(err) => tracing::warn!("retrying fetch because error: {:?}", err),
+    async fn fetch_inner(&self, key: K, block: Block) -> Result<Vec<V>> {
+        let retries = self.maximum_retries;
+        let delay = self.delay_between_retries;
+        let fetcher = self.fetcher.clone();
+        let mut created = false;
```
fleupold, line 230: doesn't seem to be read anywhere.

crates/shared/src/recent_block_cache.rs
```diff
@@ -180,33 +184,67 @@ where
...
+        let found_keys: Vec<_> = fetched
+            .iter()
+            .zip(keys.iter())
+            .filter_map(|(results, key)| {
+                results
+                    .as_ref()
+                    .is_ok_and(|res| !res.is_empty())
+                    .then_some(key.clone())
+            })
+            .collect();
```
fleupold, line 213: I think any easier way of getting the found keys is to use `fetched.iter().map(K::for_value).collect();` and maybe not even make it part of this helper method as it's only needed when fetching misses.

crates/shared/src/recent_block_cache.rs
```diff
@@ -249,15 +289,21 @@ where
...
+            cache_hits.extend_from_slice(&fetched);
 
-        cache_hits.extend_from_slice(&uncached_values);
-
-        {
             let mut mutexed = self.mutexed.lock().unwrap();
-            mutexed.insert(cache_miss_block, cache_misses.into_iter(), uncached_values);
+            mutexed.insert(cache_miss_block, chunk.iter().cloned(), fetched);
+            for key in found_keys {
+                mutexed.recently_used.cache_set(key, ());
```
fleupold, line 305: Instead of setting the recently_used cache in two places (`get` and `fetch`) can we just iterate over all results at the end and set recently used if we have an Ok result?

crates/shared/src/recent_block_cache.rs
```diff
@@ -249,15 +289,21 @@ where
         }
 
         let cache_miss_block = block.unwrap_or(last_update_block);
-        let uncached_values = self
-            .fetch_inner(cache_misses.clone(), Block::Number(cache_miss_block))
-            .await?;
+        let cache_misses: Vec<_> = cache_misses.into_iter().collect();
+        // Splits fetches into chunks because we can get over 1400 requests when the
+        // cache is empty which tend to time out if we don't chunk them.
+        for chunk in cache_misses.chunks(200) {
```
fleupold, line 295: nit: if not configurable, can we at least make it a constant?

crates/driver/src/boundary/liquidity/mod.rs
fleupold: Do these need to be chunked as well (given you decided to use 200 entry chunks to cope with 1400 total requests)

crates/shared/src/request_sharing.rs
```diff
@@ -30,13 +36,32 @@ pub type BoxRequestSharing<Request, Response> =
...
+
+    fn collect_garbage(cache: &Cache<Request, Fut>) {
+        let mut cache = cache.lock().unwrap();
+        cache.retain(|_request, weak| weak.upgrade().is_some());
+    }
+
+    fn spawn_gc(cache: Cache<Request, Fut>) {
+        tokio::task::spawn(async move {
+            Self::collect_garbage(&cache);
+            tokio::time::sleep(Duration::from_millis(500)).await;
```
fleupold, line 62: Shouldn't this run in some form of loop? 

If we don't want to create separate task for this, we could probably also collect garbage during background updates in the recent block cache (but I guess other use cases would also like gc).

crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -59,9 +67,9 @@ impl<Factory> PoolInfoFetcher<Factory> {
     }
 
     /// Returns a Balancer base pool contract instance at the specified address.
-    fn base_pool_at(&self, pool_address: H160) -> BalancerV2BasePool {
+    fn base_pool_at(&self, pool_address: &H160) -> Arc<BalancerV2BasePool> {
         let web3 = self.vault.raw_instance().web3();
-        BalancerV2BasePool::at(&web3, pool_address)
+        crate::get_or_init!(BalancerV2BasePool, POOLS, pool_address, &web3)
```
fleupold, line 72: I find this quite ugly (it adds global state + a somewhat hard to understand macro call). How much time is this saving? Would it be sufficient to keep POOLS cache on self or does self get re-created? In any case maybe this should move into a separate PR?

fleupold (commented):
Not a huge fan of the get_or_init macro, otherwise the optimisations look good (with some comments inline)

crates/driver/src/boundary/liquidity/mod.rs
MartinquaXD: The main reason we need chunking for the regular `.get()` function is that it's possible to issue so many liquidity fetch requests when starting the backend that it doesn't fit in the 15s deadline. If we don't chunk in that case we'll never end up caching any data.
When testing this change against the prod mainnet load this did not seem necessary for the background task updating the liquidity.
I would leave it like this for the next test but in the future we can certainly consider chunking that as well.

crates/shared/src/recent_block_cache.rs
MartinquaXD: Right. That's left over debugging logic.

crates/shared/src/recent_block_cache.rs
MartinquaXD: Nice, didn't realize `K::for_value` is a thing. 👍

crates/shared/src/recent_block_cache.rs
MartinquaXD: This would not work in cases where we have to fetch so much liquidity that we can't do it in the 15s solve deadline. In that case we would simply time out and never consider the maybe 80% of the keys we were able to fetch as requested. This would then mean the background task will not fetch them and that could result in us not being able to fetch all the liquidity for a long time if we are unlucky.

crates/shared/src/request_sharing.rs
MartinquaXD: > Shouldn't this run in some form of loop?

Absolutely, good catch!

> If we don't want to create separate task for this, we could probably also collect garbage during background updates in the recent block cache (but I guess other use cases would also like gc).

Yeah, other components that use `RequestSharing` definitely need a way to reclaim memory as well.

crates/shared/src/sources/balancer_v2/pools/common.rs
MartinquaXD: > How much time is this saving?

I captured a flamegraph before and after the optimization which indicates that it makes sense to have it as the `CONTRACT::at()` function calls are not cheap because we have to hash all the function signatures and events emitted by the contract for every single liquidity fetch request.
I'll check if we can move the cached contracts into `Self`. But I think we can't use a generic function for this so we either have to duplicate the function a bunch of times or live with the macro.

flamegraph_before
flamegraph_after

fleupold: From the flam graph it's not obvious to me what the savings are. Would you mind highlighting it?

fleupold (approved):
Still not super happy with the global state for caching pools, but otherwise this looks good to go to me.

crates/shared/src/sources/balancer_v2/pools/common.rs
MartinquaXD: We spend less time in `CONTRACT::with_deployment_info()` which is the function called by `CONTRACT::at()` (pretty high on the big stack in the middle, you can see it by hovering over segments when opening the file in a separate tab).
But since this caching in particular is a point of contention in this PR and could be done in a separate PR I will drop it for now so we can merge the over all more important aspects of this PR.

PR #2070, opened 2023-11-22
Fix storing settlement observations

sunce86:
# Description
Fixes current issue we have in gnosis chain staging, where deep reindexing happened and on_settlement_event_handler doesn't handle it properly.

### DIFF
--- a/crates/database/src/settlement_observations.rs
+++ b/crates/database/src/settlement_observations.rs
@@ -10,10 +10,12 @@ pub struct Observation {
     pub log_index: i64,
 }
 
-pub async fn insert(ex: &mut PgConnection, observation: Observation) -> Result<(), sqlx::Error> {
+pub async fn upsert(ex: &mut PgConnection, observation: Observation) -> Result<(), sqlx::Error> {
     const QUERY: &str = r#"
 INSERT INTO settlement_observations (gas_used, effective_gas_price, surplus, fee, block_number, log_index)
 VALUES ($1, $2, $3, $4, $5, $6)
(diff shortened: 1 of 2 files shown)


crates/database/src/settlement_observations.rs
```diff
@@ -90,5 +92,26 @@ mod tests {
...
         .unwrap();
         assert_eq!(input, output);
+
+        let new_input = Observation {
+            gas_used: 5.into(),
+            effective_gas_price: 6.into(),
+            surplus: 7.into(),
+            fee: 8.into(),
+            block_number: 1,
+            log_index: 1,
```
fedgiac, line 102: Nit: I'd like to make it clear that these values come from `input`: maybe `block_number: input.block_number(), log_index: input.log_index()`. If they were different, the test would pass anyway and we might not notice a change.

fedgiac (approved):
Looks reasonable to me.
The only question is if we want to store the fact that there was such a reorg for accounting purposes.

sunce86:
> reorg

Note that this is not related to reorgs, because reorgs are expected to happen on block [latest-64, latest] while on_settlement_event_handler works with [0..latest-64] block numbers. So, on_settlement_event_handler should never have issues with reorgs.

This actually fixes when deeper reindexing happens (when we manually erase last 1000 blocks from `settlements` database table to intentionally reindex events or when the system is down for 30 minutes and then boots up and need to index older blocks then latest-64).

fleupold (approved):

sunce86:
Merging asap to verify the fix.

PR #2080, opened 2023-11-24
Disable AssetFlow rule.

sunce86:
# Description
During colocation tests, we noticed that some of the solutions were discarded because of this rule that does not exist for legacy driver.

We want first to make sure that all existing legacy driver features are working, before introducing new rules.


fleupold (commented):
Can we fully remove all asset flow related code (I quickly checked and it doesn't look like much)? 
We have in the past too many times created these temporary half-assed deletions which end up staying forever and our codebase became confusing and messy because of this.

The code is stored in version control, so putting the logic back wen we want it should be trivial.

sunce86:
I had a feeling you will suggest that 😄 
Can do that, although I think a lot of code will be removed.

fleupold:
image

fleupold (approved):
I can still see some relicts when grepping `asset flow`

image

Could you remove these as well please?

PR #2082, opened 2023-11-24
Eliminate clippy warnings

squadgazzz:
# Description
No more `clippy` warnings for now

# Changes
Also uses the latest `ethcontract`


fleupold (approved):
What are your clippy setting? They seem overly pedantic imho. You can see the github workflow file for the clippy configuration we enforce as part of CI.

squadgazzz:
> What are your clippy setting? They seem overly pedantic imho. You can see the github workflow file for the clippy configuration we enforce as part of CI.

Probably because I use `nightly` by default.

Issue #2083, opened 2023-11-24
feat: Define protocol fee in autopilot configuration


sunce86:
### Problem
Autopilot should have a set of configuration parameters to define the protocol fee. There are two ways we are considering to take the fee:

1. A cut from the surplus above best quote
2. Fixed fee

This task captures the work needed to implement (1).

It should be possible to define parameters
1. How much percent of surplus above best quote should be taken as a protocol fee
2. Cap in percent of volume
3. Cap in absolute value ($).

The default values should be:
1. 50%
2. 6%
3. 1000$

### Acceptance criteria
- [ ] Autopilot can be configured with different fee parameters.

sunce86:
Closed with https://github.com/cowprotocol/services/pull/2098

Issue #2092, opened 2023-11-28
feat: Don't take protocol fee for in-market limit orders


sunce86:
### Problem
We don't want to take protocol fees for limit orders with in-market price.

### Suggested solution
Don't define protocol fee policies for limit orders with in-market price. To determine if the price is in-market or out-of-market, already existing functionality can be used: https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/solvable_orders.rs#L558

### Additional context
Dependent on https://github.com/cowprotocol/services/issues/2084 which needs to be tackled first.

### Acceptance criteria
- [ ] Protocol fee is not taken for limit orders with in-market price (where price is determined at the moment the order was created).

sunce86:
Done https://github.com/cowprotocol/services/pull/2250

PR #2098, opened 2023-11-28
Define protocol fee params and forward to driver

sunce86:
# Description
Tackles https://github.com/cowprotocol/services/issues/2083 and https://github.com/cowprotocol/services/issues/2084.

Defines fee policy configuration parameters. Based on them, we can take the protocol fee as a cut from price improvement or as a percent of the order volume (fixed fee in absolute value will be implemented later).
On top of that, we also need two flags, for skipping in-market limit orders and twap orders, since default initial configuration will be to run the fee experiment only for out-of-market limit orders.

Note that the quote is not forwarded to the `driver` at this moment, since it is not needed for out-of-market limit orders. 

Also note that I am missing the cap for absolute value of protocol fee (default value 1000$ or 0.5ETH) but this cap is still discussed (will be added later if needed).

Skipping forwarding protocol fee parameters for in-market limit orders will be done in a separate PR.

# Changes

- [ ] Defines basic parameters for protocol fee calculation.
- [ ] Forwards protocol fee parameters to the driver.


crates/autopilot/src/arguments.rs
```diff
@@ -203,6 +203,24 @@ pub struct Arguments {
...
+    pub protocol_fee_factor: f64,
+
+    /// Cap protocol fee with a percentage of the order's volume.
+    #[clap(
+        long,
+        env,
+        default_value = "0",
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub protofol_fee_volume_cap_factor: f64,
```
fleupold, line 223: Given that there will be likely a number of more parameters related to the fee, can we create a sub argument struct for fee related fields (instead of adding a lot more value to the top level list). Eventually we should move towards a `.toml` based config here as well, but I don't see this happening very soon.

crates/autopilot/src/driver_model.rs
```diff
@@ -115,6 +115,7 @@ pub mod solve {
         pub app_data: AppDataHash,
         #[serde(flatten)]
         pub signature: Signature,
+        pub protocol_fee: ProtocolFee,
```
fleupold, line 118: I feel like this should be a list of ProtocolFee objects, so that we can eventually have multiple fees per order (e.g. a LimitOrder that is coming via a widget integration partner may have a volume based fee for the integration partner and the limit order fee from the protocol).

Also I would call the fee_policy or the like (since unlike user_fee it's not specifying a concrete amount but instead an abstract policy we want the driver to enforce).

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +151,16 @@ pub mod solve {
         pub submission_address: H160,
     }
 
+    #[serde_as]
+    #[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct ProtocolFee {
```
fleupold, line 157: This should already be designed for the future (to avoid breaking API changes) as an enum and PriceImprovement being the first subtype. Also please document the what the type of fee means.

fleupold (commented):
How will the fee information be propagated for the frontend to consume? Are we planning to put it on the trade records we write into the database?

Alternatively we could have the API set the fee policy whenever orders are created. Curious if you have thought about the trade-offs?

sunce86:
> How will the fee information be propagated for the frontend to consume? Are we planning to put it on the trade records we write into the database?

It will follow the same flow as solver fees, meaning, fees that are unknown before sending the auction to solvers for solving and also not communicated back to the autopilot before settling, are only possible to evaluate once the transaction is mined and `on_settlement_event_updater` should calculate it and store it into `order_execution` table.

sunce86:
> Alternatively we could have the API set the fee policy whenever orders are created. Curious if you have thought about the trade-offs?

Once I restart the autopilot with updated fee policies I want all orders to follow them from that point in time. It's easier to debug and evaluate the change.

sunce86:
Oh, you meant the fee policy (I was talking about the executed protocol fee).
Yes, we need to store the fee policy somewhere for each order execution.

crates/autopilot/src/arguments.rs
sunce86: Fixed.

crates/autopilot/src/driver_model.rs
sunce86: Renamed and converted to enum.

crates/autopilot/src/driver_model.rs
sunce86: It should be a list if multiple fees are potentially be applied. Fixed.

sunce86:
> > Alternatively we could have the API set the fee policy whenever orders are created. Curious if you have thought about the trade-offs?
> 
> Once I restart the autopilot with updated fee policies I want all orders to follow them from that point in time. It's easier to debug and evaluate the change.

The other alternative would be to save the fee policies to `Order` whenever order is created, but to update this field on each open order on each new auction (so we have this trait I mentioned ☝️ ). Personally I really dislike this approach and would prefer to have a different table for fee policies. 
The reason being that these kind of solutions lead to unnecessary complexities in the system, where we update the same struct from different points in the code and in different points in time. This makes this struct scary to change and hard to reason about, while the only benefit is less DRY code.

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +302,39 @@ impl std::fmt::Display for Arguments {
...
+    /// How much of the order's surplus should be taken as a protocol fee.
+    #[clap(
+        long,
+        env,
+        default_value = "0",
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub protocol_fee_factor: f64,
+
+    /// Cap protocol fee with a percentage of the order's volume.
```
MartinquaXD, line 316: nit: the `--help` text refers to a `percentage` when it's actually a factor that is to be understood as a percentage. Slight difference that bit us in the past.

MartinquaXD (approved):
Configuration and propagation of current fee policies look alright to me.
I'm worried that we might end up with a lot of additional CLI args in the autopilot due to different fee policies but for the start where we only have a single fee policy this seems sufficient.

crates/autopilot/src/arguments.rs
sunce86: I think we are covered with the parsing function that makes sure it's a factor. We also have a factor in the variable name. I wanted to keep the comments focused on the meaning.

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +154,30 @@ pub mod solve {
...
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub enum FeePolicy {
+        /// Applies to limit orders only.
+        /// This fee should be taken if the solver provided good enough solution
+        /// that even after the surplus fee is taken, there is still more
+        /// surplus left above whatever the user expects [order limit price
+        /// or best quote, whichever is better for the user].
```
fleupold, line 164: The fact that this applies to limit orders only is not guaranteed (and a detail of the first policy that is being tested). We can and likely will charge positive quote deviation also for market orders.
I also wouldn't mention the surplus fee at all here since it's confusing. If the user received more than expected, solvers need to pay the protocol.

```suggestion
        /// If the order receives more than expected (positive deviation from quoted amounts)
        /// pay the protocol a factor of the achieved improvement.
```

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +154,30 @@ pub mod solve {
...
+        QuoteDeviation {
+            /// Percentage of the order's `available surplus` should be taken as
+            /// a protocol fee.
+            ///
+            /// `Available surplus` is the difference between the executed_price
+            /// (adjusted by surplus_fee) and the closer of the two: order
+            /// limit_price or best_quote. For out-of-market limit orders,
+            /// order limit price is closer to the executed price. For
+            /// in-market limit orders, best quote is closer to the executed
+            /// price.
```
fleupold, line 174: Let's not define a new keyword here ("available surplus"). We have called this price improvement in the past.

```suggestion
            /// Factor of price improvement the protocol charges as a fee.
            /// Price improvement is the difference between executed price and
            /// limit price or quoted price (whichever is better)
            /// 
            /// E.g. if a user received 2000USDC for 1ETH while having been quoted 1990USDC,
            /// their price improvement is 10USDC. A factor of 0.5 requires the solver to pay
            /// 5USDC to the protocol for settling this order.
```

crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -287,3 +301,9 @@ enum Class {
     Limit,
     Liquidity,
 }
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+enum FeePolicy {
+    QuoteDeviation { factor: f64, volume_cap_factor: f64 },
+}
```
fleupold, line 309: Shouldn't the dto also have documentation (so that it can be generated for the swagger docs)?

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -39,6 +40,9 @@ pub struct Order {
     pub sell_token_balance: SellTokenBalance,
     pub buy_token_balance: BuyTokenBalance,
     pub signature: Signature,
+    /// The types of fees that should be collected by the protocol.
+    /// The driver is expected to apply the fees in the order they are listed.
```
fleupold, line 44: Here we can document this driver's implementation (which is to apply the fee to the solutions it received)

```suggestion
  /// The types of fees the protocol collects from the winning solver.
  /// Unless otherwise configured, the driver modifies solutions to take sufficient fee in the form of positive slippage.
```

fleupold (approved):
Only nits, but I think it's important we pay attention to the way we document things and where. E.g. the autopilot is agnostic to the fact that the fee "is taken from surplus". It just warns participants in the protocol that if they match an order they may have to pay the protocol something at the end of the accounting period.

Our implementation of the driver then is the one that decides to hide fees from solver engines and instead "take" those fees in the form of positive slippage.

We should also make sure we use a common language when talking about the different types of fees (that will be consistent throughout or developer docs, marketing, communication with solvers etc).

Lastly, we should make it easy for maintainers to configure these parameters in a way that are intuitive and contextually clear.

crates/autopilot/src/arguments.rs
```diff
@@ -208,6 +209,11 @@ pub struct Arguments {
     )]
     pub solve_deadline: Duration,
 
+    /// A list of fee policies in the following format: `[LimitOrderOutOfMarket:0.1:0.2,LimitOrderInMarket::2.5]`
+    /// 
+    #[clap(long, env, use_value_delimiter = true)]
+    pub fee_policies: Vec<FeePolicy>,
```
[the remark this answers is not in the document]
sunce86: > What would it semantically mean to define multiple entries with the same policy but different parameters?

It enables us to take a quote deviation fee, and after that another fee as percent of volume (meaning we can chain multiple different fees for a single order). IIUC, this is something we have to support for integrators.

One could argue that we should summarize all of them to a single value and send that single value to the `driver`, but can't do that since not all types of fees are known upfront (volume percent and absolute fee are, quote deviation not).

Then, since db migrations are painful, I decided to model the db table for fee polices in a way that supports all foreseeable types of policies and not just for price improvement type. [That design ](https://github.com/cowprotocol/services/blob/d2ac47e1ad411be5d89a019e831e24f194e1af27/database/sql/V057__add_fee_policy.sql) was created to be memory efficient and decent performance wise (I could have created a separate table for each fee type but then the fetching would be slower).

Why

> pub enum FeePolicy {
>     LimitOrderOutOfMarket(FeePolicyParams),
>     LimitOrderInMarket(FeePolicyParams),
> }

Because I expect to add a third type - TWAP orders.

I hope my reasoning are clearer now. Note that this PR is not yet ready for merging, I just need to get the approval for the last [commit](https://github.com/cowprotocol/services/commit/f783d68a4dfdd528d618bc6c724fde021348065b) that my new logic is ok.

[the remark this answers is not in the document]
sunce86: Also, I hoped that I could support these two types of fees in the first iteration, since it will be trivial with this design:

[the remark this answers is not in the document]
fleupold: > IIUC, this is something we have to support for integrators.

My vision for integrators is that they specify their fee policy as part of the app data and it doesn't become part of the arguments we start the core protocol with.

The core protocol will experiment with different fees (price improvements, volume based, etc) but I foresee its fee config to be relatively concise and not a chain of many different fee compositions. Also, the protocol can only charge fees on order types that are a first class citizen in the core protocol (limit orders).

> Because I expect to add a third type - TWAP orders.

TWAP orders are not part of the core CoW Protocol. They are a set of orders that semantically belong together and are orchestrated and placed from a watch tower, but as far as CoW protocol is concerned users are simply placing limit orders in awkwardly consistent periods.

So my goal here is to split what the protocol boots up with as a fee policy from the different types of protocol fees we can attach to each order that is sent to the solvers.

E.g. the protocol may define that we want to take 0.1% volume fee on all orders. This is what the autopilot should be started with.

If an integrator (e.g. CoW Swap's TWAP feature or Metamask) wants to charge 0.85% volume fee on top, then they can create an order specifying this in the app data and what we end up sending to the driver is two fees (0.1% for the protocol, 0.85% for the integrator), both of which will be charged from solvers if they settle this order at the end. Potentially the protocol will demand a cut (cf. Apple charging 30% of all app store purchases) from the integrator fees.

To sum up, I think we should have a `FeePolicy` struct on the autopilot arguments and the a `protocol_fee: Vec<Fee>` on the order itself.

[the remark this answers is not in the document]
sunce86: Ok, not using configuration type for autopilot domain makes sense, will split it.

Do you think we should keep the possibility that 2 different policies (for example quote deviation and volume based) can be sent for a single order to the driver?

If you don't like having a third enum variant for TWAP orders, how would you model a situation where we set quote deviation policy for limit orders and volume based policy for TWAP orders using:

> price improvement factor
> volume % cap
> absolute cap
> whether or not to include in-market priced limit orders

crates/autopilot/src/arguments.rs
```diff
@@ -208,6 +208,9 @@ pub struct Arguments {
     )]
     pub solve_deadline: Duration,
 
+    #[clap(flatten)]
+    pub quote_deviation_policy: QuoteDeviationPolicy,
```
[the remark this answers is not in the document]
sunce86: outdated

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +302,39 @@ impl std::fmt::Display for Arguments {
...
+#[derive(clap::Parser, Clone)]
+pub struct QuoteDeviationPolicy {
+    /// How much of the order's surplus should be taken as a protocol fee.
+    #[clap(
+        long,
+        env,
+        default_value = "0",
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub protocol_fee_factor: f64,
```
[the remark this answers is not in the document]
sunce86: renamed

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +302,39 @@ impl std::fmt::Display for Arguments {
...
+
+    /// Cap protocol fee with a percentage of the order's volume.
+    #[clap(
+        long,
+        env,
+        default_value = "0",
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub protocol_fee_volume_cap_factor: f64,
```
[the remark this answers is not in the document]
sunce86: outdated

crates/autopilot/src/driver_model.rs
```diff
@@ -115,6 +115,10 @@ pub mod solve {
         pub app_data: AppDataHash,
         #[serde(flatten)]
         pub signature: Signature,
+        /// The types of fees that should be collected by the protocol.
+        /// The driver is expected to apply the fees in the order they are
+        /// listed.
```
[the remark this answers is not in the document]
sunce86: added

crates/autopilot/src/driver_model.rs
sunce86: added

crates/autopilot/src/driver_model.rs
sunce86: added

crates/driver/src/domain/competition/order/mod.rs
sunce86: added

sunce86:
@fleupold @MartinquaXD re-asking for review, third time's a charm :)

This setup now fits nicely with the db table: https://github.com/cowprotocol/services/issues/2111
I've updated the PR to also support volume based fees, since they are anyway needed for quote deviation fee calculation (volume fee cap) so why not adding it right away.

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +303,51 @@ impl std::fmt::Display for Arguments {
...
+pub struct FeePolicy {
+    /// How much of the order's price improvement over max(limit price,
+    /// best_bid) should be taken as a protocol fee.
+    #[clap(
+        long,
+        env,
+        default_value = None,
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub price_improvement_factor: Option<f64>,
```
fleupold, line 316: Given that these are flattened, I think they should be prefixed with fee_policy or the like so that when we set or inspect the env variables of the pod it's clearer what `PRICE_IMPROVEMENT_FACTOR` relates to (cf. e.g. `ethrpc::Arguments`).

This goes for all of the variables below...

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +303,51 @@ impl std::fmt::Display for Arguments {
...
+
+    /// How much of the order's volume should be taken as a protocol fee.
+    #[clap(
+        long,
+        env,
+        default_value = None,
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub volume_factor: Option<f64>,
```
fleupold, line 325: Are you planning to use this as the cap if set? In this case I would make it clear that this is the max_volume_factor

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +303,51 @@ impl std::fmt::Display for Arguments {
...
+
+    /// Should protocol fees be collected or skipped for limit orders with
+    /// in-market price at the time of order creation.
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
+    pub skip_in_market_orders: bool,
+
+    /// Should protocol fees be collected or skipped for TWAP limit orders.
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
+    pub skip_twap_orders: bool,
```
fleupold, line 334: Can we remove this for now? I'd still argue that since TWAP isn't a concept in the core CoW Protocol domain, we shouldn't define a fee policy for it. Instead, we can consider the TWAP UI an integration on top of CoW Protocol that defines the fee.

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +303,51 @@ impl std::fmt::Display for Arguments {
...
+        env,
+        default_value = None,
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub volume_factor: Option<f64>,
+
+    /// Should protocol fees be collected or skipped for limit orders with
+    /// in-market price at the time of order creation.
+    #[clap(long, env, action = clap::ArgAction::Set, default_value = "true")]
+    pub skip_in_market_orders: bool,
```
fleupold, line 330:
```suggestion
    pub skip_market_orders: bool,
```

crates/autopilot/src/arguments.rs
```diff
@@ -298,6 +303,51 @@ impl std::fmt::Display for Arguments {
...
+    #[clap(
+        long,
+        env,
+        default_value = None,
+        value_parser = shared::arguments::parse_percentage_factor
+    )]
+    pub volume_factor: Option<f64>,
+
+    /// Should protocol fees be collected or skipped for limit orders with
+    /// in-market price at the time of order creation.
```
fleupold, line 328: It might not be obvious what "in-market price" means.

```suggestion
    /// Should protocol fees be collected or skipped for orders whose
    /// limit price at order creation time suggests they can be immediately filled.
```

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +153,36 @@ pub mod solve {
         pub submission_address: H160,
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub enum FeePolicy {
+        /// If the order receives more than expected (positive deviation from
+        /// quoted amounts) pay the protocol a factor of the achieved
+        /// improvement. The fee is taken in `sell` token for `buy`
+        /// orders and in `buy` token for `sell` orders.
```
fleupold, line 162: @fhenneke, @olgafetisova is this comment true? I think protocol fees should be collected in CoW using some exchange rate (either the pre auction prices, post auction price, tbd)?

Note, that here we are documenting the semantics from the protocol's perspective (our driver implementation may collect the required amounts in buy or sell token, but that's an implementation detail of the driver). In what currency will the protocol later request the quote deviation from the solvers that receive the auction?

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +153,36 @@ pub mod solve {
...
+            /// Price improvement is the difference between executed price and
+            /// limit price or quoted price (whichever is better)
+            ///
+            /// E.g. if a user received 2000USDC for 1ETH while having been
+            /// quoted 1990USDC, their price improvement is 10USDC.
+            /// A factor of 0.5 requires the solver to pay 5USDC to
+            /// the protocol for settling this order.
+            factor: f64,
+            /// Cap protocol fee with a percentage of the order's volume.
+            volume_cap_factor: Option<f64>,
```
fleupold, line 174: Option not needed, since we can always cap at 1.0

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +153,36 @@ pub mod solve {
...
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub enum FeePolicy {
+        /// If the order receives more than expected (positive deviation from
+        /// quoted amounts) pay the protocol a factor of the achieved
+        /// improvement. The fee is taken in `sell` token for `buy`
+        /// orders and in `buy` token for `sell` orders.
+        QuoteDeviation {
```
fleupold, line 163: nit: can we either call this `PositiveQuoteDeviation` or `PriceImprovement` ?

crates/autopilot/src/driver_model.rs
```diff
@@ -150,6 +153,36 @@ pub mod solve {
...
+            /// quoted 1990USDC, their price improvement is 10USDC.
+            /// A factor of 0.5 requires the solver to pay 5USDC to
+            /// the protocol for settling this order.
+            factor: f64,
+            /// Cap protocol fee with a percentage of the order's volume.
+            volume_cap_factor: Option<f64>,
+        },
+        /// How much of the order's volume should be taken as a protocol fee.
+        /// The fee is taken in `sell` token for `sell` orders and in `buy`
+        /// token for `buy` orders.
```
fleupold, line 178: ditto - what currency does the protocol charge solvers the fee in?

crates/autopilot/src/run_loop.rs
```diff
@@ -513,6 +517,28 @@ pub fn solve_request(
                             .collect()
                     };
                 let order_is_untouched = remaining_order.executed_amount.is_zero();
+                let fee_policy = match (
+                    fee_policy.price_improvement_factor,
+                    fee_policy.volume_factor,
```
fleupold, line 522: Let's not semantically overload this config parameter please. No one will guess that volume factor means different things depending on whether or not another parameter is set.

crates/autopilot/src/run_loop.rs
```diff
@@ -513,6 +517,28 @@ pub fn solve_request(
                             .collect()
                     };
                 let order_is_untouched = remaining_order.executed_amount.is_zero();
+                let fee_policy = match (
```
fleupold, line 520: This block of code will explode in complexity in the next PRs and is conveniently embedded in an already very long method. I think we should create a "domain" style fee policy type which is instantiated from the fee_policy arguments and exposes a method `fn apply(&self, &mut order)`. This allows us to encapsulate the logic of
1. How to parse our variety of different arguement combinations into a coherent fee policy
2. How to apply a concrete fee policy to a concrete order.

crates/autopilot/src/arguments.rs
sunce86: It's used for both capping of quote deviation fee and also for volume based fee. Capping actually means having another fee and choose the smaller between them. You can check the db PR where I use the same logic.

crates/autopilot/src/arguments.rs
sunce86: This parameters does not define if we should use fee policies defined in TWAP order. It defines if we should apply the general limit order fee policies to twap orders.

I can remove it but anyway I need to skip twap orders when forwarding fee policies to the driver.

crates/autopilot/src/driver_model.rs
sunce86: Note that we also save external prices for these tokens for each auctions so it is easy to convert.

crates/autopilot/src/arguments.rs
sunce86: prefixed

crates/autopilot/src/arguments.rs
sunce86: Refactored to enum.

crates/autopilot/src/arguments.rs
sunce86: removed.

crates/autopilot/src/arguments.rs
sunce86: renamed

crates/autopilot/src/driver_model.rs
sunce86: removed.

crates/autopilot/src/driver_model.rs
sunce86: renamed

crates/autopilot/src/run_loop.rs
sunce86: After refactoring the argument to enum, now the conversion is 1 to 1

crates/autopilot/src/run_loop.rs
sunce86: refactored the arguments into enum, and then created a "to_domain" function for conversion, similar to what we have in `driver`. It should be a step forward to ddd in autopilot.

crates/autopilot/src/arguments.rs
sunce86: fixed.

crates/autopilot/src/driver_model.rs
sunce86: As you mentioned in the comment ☝️ , the protocol collects the protocol fees in different tokens. The question is, whether the conversion to ETH/COW/else should be done in the autopilot or later in the script for accounting.

fleupold: Sure, but it will still be charged in some currency right? Solvers will receive a bill from us at the end of the week saying: order x got 100 USDT price improvement, please pay us `y`.

`y` will be some percentage of 100 USDT, but in which currency (USDT, COW, ETH)? This is important for someone consuming this model to know so that they can make sure they convert and keep the fees in the right currency if they want to be neutral.

crates/autopilot/src/arguments.rs
fleupold: Actually, with the refactoring to enum, it now seems to be resolved.

crates/autopilot/src/run_loop.rs
```diff
@@ -513,6 +517,17 @@ pub fn solve_request(
...
+
+                let fee_policies = match order.metadata.class {
+                    OrderClass::Market => vec![],
+                    OrderClass::Liquidity => vec![],
+                    // todo https://github.com/cowprotocol/services/issues/2092
+                    // skip protocol fee for limit orders with in-market price
+
+                    // todo https://github.com/cowprotocol/services/issues/2115
+                    // skip protocol fee for TWAP limit orders
+                    OrderClass::Limit(_) => vec![fee_policy_to_domain(&fee_policy)],
```
fleupold, line 529: Shouldn't this somehow take `fee_policy_skip_market_orders` into account and include the match statement on `order.metadata.class`?

I'd create a new protocol_fee module and have a struct apply the fee. Note that in the future we will likely have logic that based on app_data adds additional fees here.

fleupold (commented):
image

Configuration looks good, last concern is about factoring the logic that adds the few on the order struct.

crates/autopilot/src/driver_model.rs
```diff
@@ -150,11 +154,54 @@ pub mod solve {
...
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub enum FeePolicy {
+        /// If the order receives more than expected (positive deviation from
+        /// quoted amounts) pay the protocol a factor of the achieved
+        /// improvement. The fee is taken in `sell` token for `buy`
+        /// orders and in `buy` token for `sell` orders.
+        PriceImprovement {
```
MartinquaXD, line 164:
```suggestion
        #[serde(rename_all = "camelCase")]
        PriceImprovement {
```

crates/autopilot/src/driver_model.rs
```diff
@@ -150,11 +154,54 @@ pub mod solve {
...
+            /// A factor of 0.5 requires the solver to pay 5USDC to
+            /// the protocol for settling this order.
+            factor: f64,
+            /// Cap protocol fee with a percentage of the order's volume.
+            max_volume_factor: f64,
+        },
+        /// How much of the order's volume should be taken as a protocol fee.
+        /// The fee is taken in `sell` token for `sell` orders and in `buy`
+        /// token for `buy` orders.
+        Volume {
```
MartinquaXD, line 180:
```suggestion
        #[serde(rename_all = "camelCase")]
        Volume {
```

MartinquaXD (approved):
Only nits at this point.

crates/autopilot/src/run_loop.rs
sunce86: This will be implemented as a separate task: https://github.com/cowprotocol/services/issues/2092

Let's also move this logic into a separate file as part of ☝️ task, to unblock this PR and merge it.

crates/autopilot/src/driver_model.rs
```diff
@@ -150,11 +154,54 @@ pub mod solve {
...
+            /// quoted 1990USDC, their price improvement is 10USDC.
+            /// A factor of 0.5 requires the solver to pay 5USDC to
+            /// the protocol for settling this order.
+            factor: f64,
+            /// Cap protocol fee with a percentage of the order's volume.
+            max_volume_factor: f64,
+        },
+        /// How much of the order's volume should be taken as a protocol fee.
+        /// The fee is taken in `sell` token for `sell` orders and in `buy`
+        /// token for `buy` orders.
```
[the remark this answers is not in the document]
sunce86: My initial idea was that we need to take the fee as some percent of the `executed_amount` which is always expressed in the sell token for sell orders and in buy token for buy orders, but this is still open and could change during implementation if we figure out we actually need something else.

crates/autopilot/src/driver_model.rs
```diff
@@ -150,11 +154,54 @@ pub mod solve {
...
+        },
+    }
+
     #[derive(Clone, Debug, Default, Deserialize)]
     #[serde(rename_all = "camelCase", deny_unknown_fields)]
     pub struct Response {
         pub solutions: Vec<Solution>,
     }
+
+    pub fn fee_policy_to_domain(fee_policy: &arguments::FeePolicy) -> FeePolicy {
```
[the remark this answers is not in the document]
sunce86: Renamed. I believe it could be both, but the fact is that we use the same struct for both domain and dto purpose.

crates/autopilot/src/driver_model.rs
```diff
@@ -150,11 +154,54 @@ pub mod solve {
         pub submission_address: H160,
     }
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
sunce86: You are right, also removed from `Order`.

crates/autopilot/src/driver_model.rs
sunce86: Fixed.

crates/autopilot/src/driver_model.rs
sunce86: Fixed.

crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -287,3 +304,10 @@ enum Class {
     Limit,
     Liquidity,
 }
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+enum FeePolicy {
+    PriceImprovement { factor: f64, max_volume_factor: f64 },
+    Volume { factor: f64 },
```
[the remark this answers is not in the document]
sunce86: Fixed.

sunce86:
Merging this PR to unblock @narcis96 so that he could work on other tasks in parallel.

PR #2099, opened 2023-11-29
Reduce memory used for balancer v2 liquidity

MartinquaXD:
# Description
The new driver is using a huge amount of memory. One low hanging fruit is to slightly redesign the internals of balancer v2 liquidity. Currently all those stable and weighted pools contain big pieces of data that don't change and are duplicated.
For other liquidity sources this was handled by having an `Inner` component that contains that data which is shared with `Arc`s across all liquidity components.

# Changes
This PR applies that same pattern for balancer stable and weighted pools.

## How to test
Just the compiler since it's only a minor refactor

### DIFF
--- a/crates/driver/src/boundary/liquidity/balancer/v2/mod.rs
+++ b/crates/driver/src/boundary/liquidity/balancer/v2/mod.rs
@@ -57,7 +57,7 @@ fn to_interaction(
         // change this assumption, we would need to change it there as well.
         GPv2Settlement::at(&web3, receiver.0),
         BalancerV2Vault::at(&web3, pool.vault.into()),
-        Arc::new(Allowances::empty(receiver.0)),
+        Allowances::empty(receiver.0),
     );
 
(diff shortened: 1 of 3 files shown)


fleupold (approved):
Love your branch names.

Do you have a slight indication/benchmark of how big this reduction is (e.g. on your local machine) that you could add to the PR description?

squadgazzz (approved):

sunce86 (approved):

MartinquaXD:
Went back over this optimization to supply a reasonable visualization of this improvement. For these graphs I applied the optimization on top of https://github.com/cowprotocol/services/pull/2102 because it makes it easier to see its impact.

1. With optimizations of https://github.com/cowprotocol/services/pull/2102
2_cache_eviction

2. With this PR's optimization in addition we can see that the big but slim spikes are gone now. This is actually not the exact outcome I would have predicted but I take it. 😅 
3_balancer_arcs

PR #2117, opened 2023-12-04
Store market order events only pt1

squadgazzz:
# Description
Partly closes pt3 of the issue https://github.com/cowprotocol/services/issues/1977:
> only store events for market orders as these are the orders for which we want to guarantee very fast execution anyway

# Changes

- [ ] Inserts order events for newly created market orders only.
- [ ] For the time being, the system will continue to update order events for both the newly implemented logic (market orders) and the existing logic (non-market orders). This is a transitional measure. A new function `insert_order_event_if_exists` has been introduced to facilitate this dual support. The need for maintaining updates on old non-market orders persists until the database is purged of entries older than the default threshold of 30 days. This cleanup process is detailed in the [autopilot crate](https://github.com/cowprotocol/services/blob/02bfbbfbfe1a14bddc4b3ad4dca70a980492b803/crates/autopilot/src/arguments.rs#L216-L219).

## How to test
New database tests have been introduced to verify the correct implementation of the changes.

## Further implementation

After the 30-day period, following the release of this PR, the optimizations should be considered:

- Limit the insertion of order event updates strictly to market orders.
- Optimize Order Identification: To further streamline the process, either modify the driver's API to include the order `class` with the order `uid` in the `/reveal` endpoint([link](https://github.com/cowprotocol/services/blob/02bfbbfbfe1a14bddc4b3ad4dca70a980492b803/crates/autopilot/src/driver_model.rs#L186-L191)) or continue utilizing `insert_order_event_if_exists` where it is required.

### DIFF
--- a/crates/database/src/order_events.rs
+++ b/crates/database/src/order_events.rs
@@ -67,6 +67,34 @@ pub async fn insert_order_event(
         .map(|_| ())
 }
 
+/// Insert a row into the `order_events` table if a previous event for the order
+/// `uid` exists.
+pub async fn insert_order_event_if_exists(
+    ex: &mut PgConnection,
+    event: &OrderEvent,
+) -> Result<(), sqlx::Error> {
+    const QUERY: &str = r#"
+        INSERT INTO order_events (
+            order_uid,
+            timestamp,
+            label
+        )
+        SELECT $1, $2, $3
+        WHERE EXISTS (
+            SELECT 1
+            FROM order_events
+            WHERE uid = $1
+        )
+    "#;
+    sqlx::query(QUERY)
+        .bind(event.order_uid)
+        .bind(event.timestamp)
+        .bind(event.label)
+        .execute(ex)
+        .await
+        .map(|_| ())
+}
+
 /// Deletes rows before the provided timestamp from the `order_events` table.
 pub async fn delete_order_events_before(
     pool: &PgPool,
--- a/crates/database/src/order_events.rs
+++ b/crates/database/src/order_events.rs
@@ -83,7 +83,7 @@ pub async fn insert_order_event_if_exists(
         WHERE EXISTS (
             SELECT 1
             FROM order_events
-            WHERE uid = $1
+            WHERE order_uid = $1
         )
     "#;
     sqlx::query(QUERY)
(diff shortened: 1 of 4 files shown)


crates/database/src/order_events.rs
```diff
@@ -67,6 +67,34 @@ pub async fn insert_order_event(
...
+        INSERT INTO order_events (
+            order_uid,
+            timestamp,
+            label
+        )
+        SELECT $1, $2, $3
+        WHERE EXISTS (
+            SELECT 1
+            FROM order_events
+            WHERE order_uid = $1
```
MartinquaXD, line 86: The concept of this function seems pretty implicit.
Since you already do another query to determine whether an insert should actually happen couldn't you do this?
```sql
SELECT $1, $2, $3
        WHERE EXISTS (
            SELECT 1
            FROM orders
            WHERE uid = $1 AND class = 'market'
        )
```
That way we only need a single function `insert_event_if_market_order()`.

squadgazzz: As for now, we need to update the existing non-market orders until all the non-market orders are deleted by the periodic cleaning mechanism. `AND class = 'market'` means that we stop inserting events for all the existing non-market orders, which might not be an expected behavior.

crates/orderbook/src/database/orders.rs
```diff
@@ -1015,4 +1133,24 @@ mod tests {
         let interactions = db.single_order(&uid).await.unwrap().unwrap().interactions;
         assert_eq!(interactions, order.interactions);
     }
+
+    async fn order_event_labels_by_id(pool: &PgPool) -> HashMap<OrderUid, Vec<OrderEventLabel>> {
+        const QUERY: &str = r#"
+                SELECT order_uid, label
+                FROM order_events
```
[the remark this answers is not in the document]
squadgazzz: Thanks, added the sorting, which simplified some test assertions.

crates/database/src/order_events.rs
MartinquaXD: Nah, it's totally fine to completely stop inserting events for non-market orders. The `order_events` table only contains debugging data that is used for this [grafana query](https://g-0263500beb.grafana-workspace.eu-central-1.amazonaws.com/d/Osrh8t_4k/cow-slis?orgId=1&editPanel=8) which already ignores non-market orders.

squadgazzz: > Nah, it's totally fine to completely stop inserting events for non-market orders.

In that case, I need to change the implementation for the entire PR. Will open a separate one.

PR #2123, opened 2023-12-05
Filter in-flight orders

MartinquaXD:
# Description
Although the `autopilot` currently waits until the `driver` reports that the settlement was mined the auction of the next run loop might not yet be aware that the orders of the previous settlement were solved.
If a solver then tries to solve those orders again it will run into simulation failures.

# Changes
We now filter out in-flight orders from the auction if it was built before the block where the in-flight orders were settled.

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -20,7 +20,7 @@ use {
     model::{
         auction::{Auction, AuctionId, AuctionWithId},
         interaction::InteractionData,
-        order::OrderClass,
+        order::{OrderClass, OrderUid},
         solver_competition::{
             CompetitionAuction,
             Order,
@@ -30,15 +30,16 @@ use {
         },
     },
     number::nonzero::U256 as NonZeroU256,
-    primitive_types::{H160, U256},
+    primitive_types::{H160, H256, U256},
     rand::seq::SliceRandom,
     shared::{remaining_amounts, token_list::AutoUpdatingTokenList},
     std::{
         collections::{BTreeMap, HashSet},
-        sync::Arc,
+        sync::{Arc, Mutex},
         time::{Duration, Instant},
     },
     tracing::Instrument,
+    web3::types::TransactionReceipt,
 };
 
 pub struct RunLoop {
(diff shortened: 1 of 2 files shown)


crates/autopilot/src/run_loop.rs
```diff
@@ -54,6 +55,7 @@ pub struct RunLoop {
     pub score_cap: U256,
     pub max_settlement_transaction_wait: Duration,
     pub solve_deadline: Duration,
+    pub in_flight_orders: Arc<Mutex<InFlightOrders>>,
```
sunce86, line 58: Why do we need synchronisation here?

crates/autopilot/src/run_loop.rs
```diff
@@ -443,6 +452,34 @@ impl RunLoop {
     async fn save_competition(&self, competition: &Competition) -> Result<()> {
         self.database.save_competition(competition).await
     }
+
+    /// Removes orders that are currently being settled to avoid solvers trying
+    /// to fill an order a second time.
+    async fn remove_in_flight_orders(&self, mut auction: Auction) -> Auction {
+        let prev_settlement = self.in_flight_orders.lock().unwrap().tx_hash;
+        let tx_receipt = self.web3.eth().transaction_receipt(prev_settlement).await;
```
sunce86, line 460: What happens if in the first iteration the `tx_hash` is initialized with `Default::default()`?

crates/autopilot/src/run_loop.rs
MartinquaXD: Because we populate the data in a method that takes a shared reference to `self`.

crates/autopilot/src/run_loop.rs
MartinquaXD: Then `transaction_receipt()` would try to find the block that settled tx hash `0x0000` which does not exist so we fallback to assuming the previous auction was settled in block `u64::MAX` so we discard all in-flight orders we have at the time (which are 0 due to `Default::default()`).

sunce86 (approved):
lg

fleupold (approved):
LGTM

PR #2130, opened 2023-12-06
Update available networks in network_name

fedgiac:
# Description
We get the name of the current network from a list of networks based on chain id and network id.
The list is updated so that Sepolia is also included.

# Changes

- Updated command to generate the match list.
  This  is done because the current command only takes the _first_ word in the network name. So for example mainnet would be "Ethereum" instead of "Ethereum Mainnet".
  I also sorted the output, so that in the future the diff will be more meaningful.

  - `Ethereum / Mainnet` becomes `Ethereum Mainnet` (Note that it had more than one word because it was [manually changed](https://github.com/cowprotocol/services/commit/20ef3d5de8d46121c547cb6fe9a0d44e5fdce29d) before) 
  - `Ethereum / Goerli` becomes `Goerli`.
  - `xDAI` becomes `Gnosis`
  
  I'm not sure which effects this change is going to have, this value isn't directly needed by the services afaict. The name is used when storing instances, for example, so other parties might rely on it.
- Added an ignored test to confirm that the test works.

## How to test

New ignored test.


fleupold (approved):
Looks good, I'm a bit nervous about the xDAI rename as we recently concluded to stick with xDAI instead of changing to Gnosis Chain everywhere, but if you say it's mainly used for instance names it's probably ok.

Otherwise, you could consider reverting the changes on existing networks) (as the main value of this PR seems to come from adding Sepolia and not changing other names).

MartinquaXD:
I think at least `enso` is using the network name in the auction metadata for some logic (https://github.com/cowprotocol/services/pull/2001).
Not sure what might break on their end if we change the mainnet identifier. cc @devanoneth

fedgiac:
Closed in favor of #2136.

PR #2134, opened 2023-12-07
Speed up token meta data look up

MartinquaXD:
# Description
We've noticed weird delays between the autopilot sending an auction and the driver being ready to start working on it.
The linked issue suggests that we spend a lot of time converting the auction and enriching it with token meta data.
This should be super fast because all the data should be cached at all times.
However, was one token that caused it all to break down: `0xeeeee...` (the address we use to denote the chains native token)

This is roughly what happened:
1. every driver receives auction
2. every driver collects all the addresses for which we don't have cached metadata
3. this always yielded `0xeeee...` (it's not a real token so we never get data when calling ERC20 functions on it)
4. every driver wasted some time sending RPC requests for `0xeeee`
5. every driver tried to take an exclusive lock to write no data (empty vec) to the cache

Overall this wasted time with unnecessary network requests and serially taking an exclusive lock for all drivers.

# Changes
* most importantly filter out `0xeee`
* avoid sending duplicate RPC requests by holding an exclusive lock while fetching missing token meta data
* don't send RPC requests if some other driver cached the data in the mean time
* moved logging of token balance update task out of the critical section

All changes combined results in us now spending a couple microseconds on getting cached balances instead of seconds.
And because this was the bulk of the work captured by the log `auction task execution time` we reduced that time to ~11ms.

## How to test
e2e tests should still pass and manual test confirmed the reduced latency.

Fixes [#](https://github.com/cowprotocol/services/issues/2133)

### DIFF
--- a/crates/driver/src/infra/tokens.rs
+++ b/crates/driver/src/infra/tokens.rs
@@ -157,6 +157,8 @@ impl Inner {
             return;
         }
 
+        // Take exclusive lock because everybody will call this function for the same tokens
+        // so this a simple way to avoid sending duplicate requests.
         let mut cache = self.cache.write().await;
         if tokens.iter().all(|address| cache.contains_key(address)) {
             // Somebody else might have already cached all the data in the meantime.
--- a/crates/driver/src/infra/tokens.rs
+++ b/crates/driver/src/infra/tokens.rs
@@ -5,10 +5,9 @@ use {
     },
     ethrpc::current_block::{self, CurrentBlockStream},
     futures::StreamExt,
-    std::{
-        collections::{HashMap, HashSet},
-        sync::{Arc, RwLock},
-    },
+    itertools::Itertools,
+    std::{collections::HashMap, sync::Arc},
+    tokio::sync::RwLock,
     tracing::Instrument,
 };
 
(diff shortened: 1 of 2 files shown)


crates/driver/src/infra/tokens.rs
```diff
@@ -140,30 +149,43 @@ impl Inner {
         futures::future::join_all(futures).await
     }
 
+    /// Ensures that all the missing tokens are in the cache afterwards while
+    /// taking into account that the function might be called multiple times
+    /// for the same tokens.
```
fleupold, line 154: This is a pretty strong assumption (which may not be true e.g. for concurrent quote requests). Would it be simple/performant enough to fetch token info one by one (should be batched on transport level) and use shared requests?

crates/driver/src/infra/tokens.rs
```diff
@@ -140,30 +149,43 @@ impl Inner {
...
+        }
+
+        // Take exclusive lock because everybody will call this function for the same
+        // tokens so this a simple way to avoid sending duplicate requests.
+        let mut cache = self.cache.write().await;
+        if tokens.iter().all(|address| cache.contains_key(address)) {
+            // Somebody else might have already cached all the data in the meantime.
+            return;
+        }
+        let fetched = self.fetch_token_infos(tokens).await;
```
fleupold, line 167: image

crates/driver/src/infra/tokens.rs
MartinquaXD: I did not try out `RequestSharing` because I was worried the implementation would end up more complex than what I had.
But it turns out that as readable IMO + it should have a higher throughput when different drivers are fetching metadata for different tokens in parallel.
The latency for the case where we fetch Metadata for all tokens in an auction didn't really change with `RequestSharing` btw.

PR #2135, opened 2023-12-07
Optimize rows count metrics

squadgazzz:
# Description
Noticed that the rows count query for the `order_events` table takes **150-200+ seconds** both in staging and prod. The proposal is to try out using rough row count estimation for large tables.

# Changes
`SELECT COUNT(*) ...` is replaced with `SELECT reltuples FROM pg_class WHERE relname='{table}';` for the `order_events` table.

## How to test
Will analyze metrics to understand how often Postgres updates the statistics without manually triggering `ANALYZE`. 

## Further optimization

In case statistics update happens often enough, the same approach could be considered for other tables where very often row count update is not necessary.

### DIFF
--- a/crates/autopilot/src/database.rs
+++ b/crates/autopilot/src/database.rs
@@ -104,7 +104,7 @@ pub async fn database_metrics(db: Postgres) -> ! {
         if let Err(err) = db.update_database_metrics().await {
             tracing::error!(?err, "failed to update table rows metric");
         }
-        tokio::time::sleep(Duration::from_secs(60)).await;
+        tokio::time::sleep(Duration::from_secs(60 * 5)).await;
     }
 }
 
--- a/crates/autopilot/src/database.rs
+++ b/crates/autopilot/src/database.rs
@@ -28,9 +28,16 @@ impl Postgres {
         let metrics = Metrics::get();
 
         // update table row metrics
-        for &table in database::ALL_TABLES {
+        for &table in database::TABLES {
             let mut ex = self.0.acquire().await?;
             let count = count_rows_in_table(&mut ex, table).await?;
+            metrics.table_rows.with_label_values(&[table]).set(count);
+        }
+
+        // update table row metrics
+        for &table in database::LARGE_TABLES {
+            let mut ex = self.0.acquire().await?;
+            let count = count_rough_rows_in_table(&mut ex, table).await?;
             metrics.table_rows.with_label_values(&[table]).set(count);
         }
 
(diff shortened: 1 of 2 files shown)


crates/database/src/lib.rs
```diff
@@ -69,10 +68,20 @@ pub const ALL_TABLES: &[&str] = &[
...
 
+/// The names of potentially big volume tables we use in the db.
+pub const LARGE_TABLES: &[&str] = &["order_events"];
+
+pub fn all_tables() -> Vec<&'static str> {
+    let mut combined = Vec::with_capacity(TABLES.len() + LARGE_TABLES.len());
+    combined.extend_from_slice(TABLES);
+    combined.extend_from_slice(LARGE_TABLES);
+    combined
+}
```
MartinquaXD, line 79: FYI: This function can be written such that it doesn't have to allocate. Obviously doesn't matter here because this function is not performance critical but I guess it's useful to learn about iterator like that.

```suggestion
pub fn all_tables() -> impl Iterator<Item = &'static str> {
    TABLES.iter().copied().chain(LARGE_TABLES.iter().copied())
}
```

crates/autopilot/src/database.rs
```diff
@@ -50,6 +57,11 @@ async fn count_rows_in_table(ex: &mut PgConnection, table: &str) -> sqlx::Result
     sqlx::query_scalar(&query).fetch_one(ex).await
 }
 
+async fn count_rough_rows_in_table(ex: &mut PgConnection, table: &str) -> sqlx::Result<i64> {
```
MartinquaXD, line 60: nit: Nice TIL about fetching estimated table sizes.
Just the function name seems a bit weird. Also maybe a small comment would be nice explaining that actually counting rows can be too slow for very big tables.
```suggestion
async fn estimate_rows_in_table(ex: &mut PgConnection, table: &str) -> sqlx::Result<i64> {
```

MartinquaXD (approved):
I'm a bit surprised that grafana is able to even update the dashboards if fetching the count takes so long.

crates/database/src/lib.rs
squadgazzz: Thanks, I was looking for that :)

crates/autopilot/src/database.rs
```diff
@@ -92,7 +104,7 @@ pub async fn database_metrics(db: Postgres) -> ! {
         if let Err(err) = db.update_database_metrics().await {
             tracing::error!(?err, "failed to update table rows metric");
         }
-        tokio::time::sleep(Duration::from_secs(60)).await;
+        tokio::time::sleep(Duration::from_secs(60 * 5)).await;
```
squadgazzz, line 107: One more proposal is to make updates less regular.

MartinquaXD: Since the estimation query is in the order of 200ms now I think having more recent metrics is more important than saving a few DB resources.

squadgazzz: Counting rows in other tables takes a couple of seconds(2-7).

PR #2136, opened 2023-12-07
Simplify network name function

fedgiac:
# Description
An alternative to PR #2130. It simplifies the network name function to only include networks we actually use.
If a network is unknown, the string name includes both network and chain id.
Notably compared to #2130, there are no naming changes.

# Changes

- Removed unused networks.
- Added Sepolia
- Added extra information to unknown networks.

## How to test

Unknown, as I don't know any protocol using other networks but those I left here.
Specifically for Sepolia, you can compare with the autogenerated entry [here](https://github.com/cowprotocol/services/pull/2130/files#diff-6fe1d095c8577cf6c68e84fa7d84e110cad2d1a38b83a3421513e475e9692007R1088).


crates/shared/src/network.rs
```diff
@@ -1,163 +1,19 @@
...
-        ("97", 97) => "Binance",
-        ("977", 977) => "Nepal",
-        ("99", 99) => "POA",
-        ("99415706", 99415706) => "Joys",
-        ("999", 999) => "Wanchain",
-        _ => "<unknown>",
-    }
+        ("100", 100) => "xDAI",
+        ("11155111", 11155111) => "Ethereum / Sepolia",
+        _ => return format!("<unknown network network_id={network_id} chain_id={chain_id}>"),
```
MartinquaXD, line 14: Should this be an error?
Not sure if we'd ever want to have this as the network name we use in the backend.
If we don't want to change the signature of the function I think it's reasonable to `panic!()` here and keep returning a `&'static str` from the function.

fedgiac: I thought we would be using unknown chains for testing, but this is indeed not the case. I committed this suggestion.

squadgazzz (approved):

MartinquaXD (approved):

PR #2142, opened 2023-12-07
Fix large table count query

MartinquaXD:
# Description
The current query returns a value in an unexpected format.

# Changes
Cast the returned value to a type that can be converted to `i64`.

## How to test
Ran the modified query in `pgadmin` to check that the type returned from it is the same as is returned from the regular `count(*)` query.


MartinquaXD:
FYI: Enabled auto merge since this is a trivial change -- can't break things more than they already are ;)

PR #2166, opened 2023-12-13
Configurable order_events insert batch size

squadgazzz:
# Description
https://github.com/cowprotocol/services/pull/2138 showed a reduction in the insert time. The batch size config should now be configurable.

# Changes
Adds a config to the `Postgres` struct.

## How to test
Passes e2e tests

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -93,7 +93,7 @@ jobs:
       CARGO_PROFILE_DEV_DEBUG: 0
       CARGO_PROFILE_TEST_DEBUG: 0
       CARGO_TERM_COLOR: always
-      FORK_URL: https://eth.merkle.io
+      FORK_URL: ${{ secrets.FORK_URL }}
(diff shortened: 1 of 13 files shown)


crates/autopilot/src/database.rs
```diff
@@ -148,8 +164,10 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn postgres_count_rows_in_table_() {
-        let db = Postgres::new("postgresql://").await.unwrap();
-        let mut ex = db.0.begin().await.unwrap();
+        let db = Postgres::new("postgresql://", NonZeroUsize::new(500).unwrap())
```
MartinquaXD, line 167: For testing purposes it will probably nicer long term if we have a constructor `with_defaults()`. That way we don't have to touch all the e2e tests in case we extend the config in the future.

fleupold (approved):

MartinquaXD (approved):

PR #2087, opened 2023-11-27
Store quotes for limit orders

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2078

# Changes

- [ ] Limit orders are no longer skipped when the quote is determined and stored into database.

## How to test
Will observe the behavior on staging to make sure the quotes are stored.

PR #2173, opened 2023-12-15
Don't delete expired quotes

sunce86:
# Description
Follow up for https://github.com/cowprotocol/services/pull/2087

I did not realize that we delete expired quotes. This PR removes this, so quotes are now saved permanently. This is needed for at least two reasons:
1. Long term, we want to rely on quotes at order creation time to determine if the order was in/out of market price.
2. Protocol fee calculation is done in postprocessing - we want this component to access quotes few blocks after order being filled, and since we aim to make the `OnSettlementEventHandler` reindexable (in case, for example, we notice a bug in reward/payout/fee calculation and want to recalculate all data ) this component would need the quotes from long ago.

Note that the deletion doesn't mess with finding the quote when order is created, since the expiration is handled with `FindQuoteError::Expired`, while `OrderStoring` implementation return both expired and non-expired quotes, which is how it should be.


MartinquaXD:
We have 2 quote related tables:
1. [quotes](https://github.com/cowprotocol/services/blob/main/database/README.md#order_quotes) contains all the short lived quotes the get created for the price estimates issued by people.
2. [order_quotes](https://github.com/cowprotocol/services/blob/main/database/README.md#order_quotes) contains all the quotes that were actually used to create an order with and does not get cleaned up.

PR #132, opened 2022-04-05
Cow-ify repository

nlordell:
This PR just "Cow-ifies" the repository. It tries to:
- Replace references to GPv2 into Cow Protocol
- Replace URLs to the new `cow.fi` ones
- Replace issue links to the ones created in this repo.
- Update dependencies to use the github.com/cowprotocol org forks
- Contract vendoring with new repo pacakges (**Note**, the `@cowprotocol/token@1.1.0` package is not deployed on NPM pending https://github.com/cowprotocol/token/pull/1 - feel free to review that PR as well 😄).

### Test Plan

CI - no code or logic changes, just metadata things.

<details><summary>Run the vendoring script to make sure that the contract sources are correct.</summary>

```
% cargo run -p contracts --features bin --bin vendor
[2022-04-07T09:52:59Z INFO  vendor] vendoring contract artifacts to '/Users/nlordell/Developer/cowprotocol/services/crates/contracts/artifacts'
[2022-04-07T09:52:59Z INFO  vendor] retrieving Http(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("raw.githubusercontent.com")), port: None, path: "/balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/deployed/mainnet/Authorizer.json", query: None, fragment: None })
[2022-04-07T09:52:59Z INFO  vendor] retrieving Http(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("raw.githubusercontent.com")), port: None, path: "/balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/deployed/mainnet/Vault.json", query: None, fragment: None })
[2022-04-07T09:52:59Z INFO  vendor] retrieving Http(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("raw.githubusercontent.com")), port: None, path: "/balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/deployed/mainnet/WeightedPoolFactory.json", query: None, fragment: None })
[...]

PR #2175, opened 2023-12-15
Use regular `prometheus-metric-storage` crate

squadgazzz:
# Description
In order to reduce `cowprotocol` dependencies in the https://github.com/cowprotocol/solvers, we switch to the regular `prometheus-metric-storage` crate.

# Changes
Migrate from https://github.com/cowprotocol/prometheus-metric-storage to https://github.com/taminomara/prometheus-metric-storage/

## How to test
Staging metrics observation

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/base:bullseye
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    # Remove imagemagick due to https://security-tracker.debian.org/tracker/CVE-2019-10131
+    && apt-get purge -y imagemagick imagemagick-6-common
+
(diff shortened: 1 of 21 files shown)


fleupold (approved):
@MartinquaXD this dependency was introduced in https://github.com/cowprotocol/services/commit/e23b37bf1c6e45a0288ae2c97908b9c0b27d714a if I'm not mistaken. Any ideas why we switched from the vendored 0.4.0 version?

MartinquaXD:
The first PR that imported that module by git URL was https://github.com/cowprotocol/services/pull/132 which change a bunch of links to point to the newly forked cow organization repo instead of the old gnosis repo.
There is no mention why we now import by git url instead of crates.io link however. When Ilya and I discussed this dependency we also checked if our version has some additional commits on top of the regular dependency but that also does not seem to be the case.

MartinquaXD (approved):

PR #2201, opened 2023-12-21
Enforce all available api arguments are printed

fedgiac:
# Description
On startup, we want to print all arguments that were passed to any of our binaries.

However, sometimes we add new arguments but we forget to add them to the `Display` implementation.

This PR refactors the code so that the compiler throws an error if some arguments are missing.

# Changes

- Enforce usage of all arguments when implementing `Display` in the autopilot.
- Add a few missing arguments to the `Display` implementation.

I'll do the same for all other binaries if this PR is approved and merged.

## How to test

Code should compile.

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -259,8 +259,12 @@ impl std::fmt::Display for Arguments {
             fee_policy,
             order_events_cleanup_interval,
             order_events_cleanup_threshold,
-            ..
-        } = self;
+            db_url,
+            order_events_insert_batch_size,
+            native_price_estimation_results_required,
+            auction_update_interval,
+            max_settlement_transaction_wait,
+        }: &Arguments = self;
 
         write!(f, "{}", shared)?;
         write!(f, "{}", order_quoting)?;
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Nice one! Very clever use of the compiler

crates/autopilot/src/arguments.rs
```diff
@@ -225,79 +225,129 @@ pub struct Arguments {
...
+            drivers,
+            submission_deadline,
+            additional_deadline_for_rewards,
+            score_cap,
+            shadow,
+            solve_deadline,
+            fee_policy,
+            order_events_cleanup_interval,
+            order_events_cleanup_threshold,
+            db_url,
```
MartinquaXD, line 262: What a big brain move! 😄 
BTW if you prefer you can ignore the DB url right inside the destructured `Arguments`.
```suggestion
            db_url: _intentionally_ignored_db_url,
```

MartinquaXD (approved):

crates/autopilot/src/arguments.rs
fedgiac: Nice!
Thinking about it, I prefer it as it is right now because of the line `writeln!(f, "db_url: SECRET")?;`: if we ignore the database near that line, I hope we remember to edit this line if we edit `db_url`.

Issue #2204, opened 2023-12-21
feat: Filter out useless Balancer pools


sunce86:
### Problem
On colocation activation, we started seeing logs https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/r/s/aTHp4
Specifically, two bad tokens are involved `0xbd1bd5c956684f7eb79da40f582cbe1373a1d593`  and `0x1b808f49add4b8c6b5117d9681cf7312fcf0dc1d`, where the latter one is even in the `unsupported_tokens` list right now.

After investigation, I realized that Balancer liquidity fetcher (which has a function `get_registered_pools()` is indexing pool https://etherscan.io/address/0x94ffd754170b741c9a22bf66885681ef4ce1ea07. This pool is useless and we need to figure out how to filter it out based on some additional `liqudity` field of Balancer subgraph or similar.

### Suggested solution
Adjust the get_registered_pools to only return meaningul pools.

### Acceptance criteria
No more "error retrieving allowance for token" errors.

MartinquaXD:
We have `--balancer-pool-deny-list` for that.

PR #2205, opened 2023-12-21
Filter out useless pools

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2204

By filtering out pools that basically don't have meaningful liquidity we reduce the number of pools from ~1300 to ~500 and also eliminate problematic pools (with old problematic tokens).

## How to test
Play with unit test `balancer_subgraph_query`


MartinquaXD (approved):
Nice 👍

PR #2208, opened 2023-12-22
Remove colocation flag

sunce86:
# Description
Initiative to simplify autopilot parameters so we can move them to `toml` format.

Removes colocation flag as we no longer see we will revert to non-colocated driver.

### DIFF
--- a/crates/autopilot/src/run.rs
+++ b/crates/autopilot/src/run.rs
@@ -567,7 +567,6 @@ pub async fn run(args: Arguments) {
         args.limit_order_price_factor
             .try_into()
             .expect("limit order price factor can't be converted to BigDecimal"),
-        !args.enable_colocation,
         args.fee_objective_scaling_factor,
     );
     solvable_orders_cache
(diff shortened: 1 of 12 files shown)


crates/autopilot/src/run.rs
```diff
@@ -607,43 +606,34 @@ pub async fn run(args: Arguments) {
             .instrument(tracing::info_span!("order_events_cleaner")),
     );
 
-    if args.enable_colocation {
-        if args.drivers.is_empty() {
-            panic!("colocation is enabled but no drivers are configured");
-        }
```
MartinquaXD, line 613: We should still keep this panic.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -239,14 +235,6 @@ impl SolvableOrdersCache {
 
         self.database.store_order_events(&order_events).await;
 
-        let id = if self.store_in_db {
-            let id = self.database.replace_current_auction(&auction).await?;
-            tracing::info!(auction = %id, %block, "stored new auction in database");
```
MartinquaXD, line 244: I think we should still write the auction to the DB. Otherwise the `/auction` endpoint would break can be a useful tool for debugging.

MartinquaXD (commented):
@mfw78 looks like you were too slow. 😄 
Yesterday evening we talked about deleting this flag.

crates/autopilot/src/solvable_orders.rs
sunce86: We do it here for the colocated setup https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/run_loop.rs#L93

crates/autopilot/src/run.rs
sunce86: Reverted.
Initially I removed it because the check should be done on config loading but anyway, let's keep it until we are there.

mfw78 (approved):

MartinquaXD (approved):

PR #2235, opened 2024-01-02
Enforce all arguments are printed

fedgiac:
# Description
An extension of PR  #2201 that applies to all printed arguments instead of just for the orderbook. Same description as for that one:

On startup, we want to print all arguments that were passed to any of our binaries.
However, sometimes we add new arguments but we forget to add them to the Display implementation.
This PR refactors the code so that the compiler throws an error if some arguments are missing.

# Changes

- 12549444b947d06dabf11e8ac9ac767ecd8e0838: destructure `Arguments` and remove `self.` in the display implementation.
- 88a0f4517ea2cdab75912e55d4eaebf1b1612934: remove `..` and manually add back all missing arguments with the right display format.

## How to test

The code compiles, so no arguments are missing.
Because of the autoformatter, this PR is a bit annoying to review. I use [difftastic](https://github.com/Wilfred/difftastic) in git for a semantic diff that makes it easy to see that all changes are as expected.


fleupold (approved):

sunce86 (approved):
Nice trick! Although changes for `solver` not needed anymore, fine to keep them for completness.

PR #2245, opened 2024-01-04
Announce Sepolia availability

fedgiac:
# Description

Now that we have running services on Sepolia, we can add information on where to find them.

## How to test

```
for base_url in {barn.,}api.cow.fi/sepolia; do echo "Checking $base_url:"; curl -X 'GET' "https://$base_url/api/v1/auction" -H 'accept: application/json'; echo; done
```


fleupold (approved):
Does this require changes to the dev docs as well (cc @mfw78 )

PR #2255, opened 2024-01-09
Support new order type in alerter

sunce86:
# Description
Fixes alerter so it works with two different order types:
- one from api/v1/order
- other from api/v1/auction

Consequence of https://github.com/cowprotocol/services/pull/2246

### DIFF
--- a/crates/alerter/src/lib.rs
+++ b/crates/alerter/src/lib.rs
@@ -22,7 +22,7 @@ use {
 #[serde_as]
 #[derive(Debug, serde::Deserialize, Eq, PartialEq)]
 #[serde(rename_all = "camelCase")]
-struct BaseOrder {
+struct Order {
     kind: OrderKind,
     buy_token: H160,
     #[serde_as(as = "HexOrDecimalU256")]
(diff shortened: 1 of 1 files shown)


crates/alerter/src/lib.rs
```diff
@@ -23,30 +22,105 @@ use {
...
+struct BaseOrder {
     kind: OrderKind,
     buy_token: H160,
     #[serde_as(as = "HexOrDecimalU256")]
     buy_amount: U256,
     sell_token: H160,
     #[serde_as(as = "HexOrDecimalU256")]
     sell_amount: U256,
-    #[serde_as(as = "HexOrDecimalU256")]
-    fee_amount: U256,
```
sunce86, line 35: `fee_amount removed as not used anywhere`

crates/alerter/src/lib.rs
```diff
@@ -221,14 +295,10 @@ impl Alerter {
             .await
             .context("solvable_orders")?
             .into_iter()
-            .filter(|order| {
-                !order.is_liquidity_order()
-                    && !order.partially_fillable
-                    && !matches!(order.status, OrderStatus::PresignaturePending)
```
sunce86, line 227: Safe to skip removing `OrderStatus::PresignaturePending` since those are already filtered by autopilot [here](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/solvable_orders.rs#L158-L159)

crates/alerter/src/lib.rs
```diff
@@ -23,30 +22,105 @@ use {
...
     }
 }
 
+#[serde_as]
+#[derive(Debug, serde::Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+struct ModelOrder {
+    #[serde(flatten)]
+    base: BaseOrder,
+    status: OrderStatus,
```
MartinquaXD, line 51: Not sure if the differentiation between these 2 different order types is so significant.
If you had a single order type with `status: Option<OrderStatus>` you could get rid of a bunch of boiler plate logic.

sunce86: Exactly. I removed the remaining fields at the end so completely became ignorant that I was left with only one field of difference :)
Anyway, updated.

crates/alerter/src/lib.rs
```diff
@@ -250,7 +244,7 @@ impl Alerter {
             tracing::debug!(order =% uid, "found closed order");
             let start = Instant::now();
             let api_order = self.orderbook_api.order(uid).await.context("get order")?;
-            if api_order.status == OrderStatus::Fulfilled {
+            if api_order.status == Some(OrderStatus::Fulfilled) {
```
MartinquaXD, line 247: nit: I think here it's probably a good idea to use `.unwrap()` or `.expect()` because we always expect this to be populated and would like to be informed in case this changes at some point in the future, right?

MartinquaXD (approved):

crates/alerter/src/lib.rs
sunce86: Sure, I also thought about it.

Issue #2257, opened 2024-01-09
chore: Single sql query for read_quotes


sunce86:
### Background
https://github.com/cowprotocol/services/pull/2250#discussion_r1444470949

PR #2258, opened 2024-01-09
Single sql query for reading quotes

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2257

# Changes

- [ ] Added general `read_quotes` function on the db level
- [ ] Refactor autopilot to use this new function

## How to test
Added db test.

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -429,8 +429,7 @@ pub async fn read_quotes(
     for order_id in order_ids {
         query = query.bind(order_id);
     }
-    let quotes = query.fetch_all(ex).await?;
-    Ok(quotes)
+    query.fetch_all(ex).await
 }
 
 pub async fn cancel_order(
@@ -1156,7 +1155,7 @@ mod tests {
 
     #[tokio::test]
     #[ignore]
-    async fn read_quotes_roundtrip() {
+    async fn postgres_read_quotes_roundtrip() {
         let mut db = PgConnection::connect("postgresql://").await.unwrap();
         let mut db = db.begin().await.unwrap();
         crate::clear_DANGER_(&mut db).await.unwrap();
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -14,6 +14,7 @@ use {
             BigDecimal,
         },
         PgConnection,
+        QueryBuilder,
     },
 };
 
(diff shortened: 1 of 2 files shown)


squadgazzz (approved):

crates/database/src/orders.rs
```diff
@@ -402,6 +402,37 @@ WHERE order_uid = $1
...
 
+pub async fn read_quotes(
+    ex: &mut sqlx::PgConnection,
+    order_ids: &[OrderUid],
+) -> Result<Vec<Quote>, sqlx::Error> {
+    if order_ids.is_empty() {
+        return Ok(vec![]);
+    }
+
+    let placeholders = order_ids
```
MartinquaXD, line 413: There is a [query builder](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) that seems to accommodate this use case.

As a bonus the builder + taking `order_ids: impl IntoIterator<Item = &OrderUid>` would allow you to create the query without having to allocate a vector for the ids to pass them into the function.
If you still want to have the early return in case the iterator is empty you could do:
```rust
let mut order_ids = order_ids.into_iter().peekable();
if order_ids.peek().is_none() {
    return Ok(vec![]);
}
order_ids
```
Trying to optimize the allocation away is probably a bit overkill but the query builder seems to be reasonable.

squadgazzz: True, forgot about this before approving the PR. It is already being utilized in the project https://github.com/cowprotocol/services/blob/main/crates/database/src/order_events.rs#L76

MartinquaXD: Ahh, right. This is what I wanted to link: https://docs.rs/sqlx/latest/sqlx/struct.QueryBuilder.html#method.separated

crates/database/src/orders.rs
```diff
@@ -402,6 +402,37 @@ WHERE order_uid = $1
...
+    "#,
+        placeholders
+    );
+
+    let mut query = sqlx::query_as(&query);
+    for order_id in order_ids {
+        query = query.bind(order_id);
+    }
+    let quotes = query.fetch_all(ex).await?;
+    Ok(quotes)
```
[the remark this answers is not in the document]
sunce86: Cleaned

crates/database/src/orders.rs
```diff
@@ -1123,6 +1154,29 @@ mod tests {
         assert_eq!(quote2, quote_);
     }
 
+    #[tokio::test]
+    #[ignore]
+    async fn read_quotes_roundtrip() {
```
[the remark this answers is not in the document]
sunce86: Indeed. Didn't know this, however I did run in locally. Thanks.

crates/database/src/orders.rs
sunce86: refactored. seems a bit nicer now.

PR #2270, opened 2024-01-12
Fixes openapi for new Auction

sunce86:
# Description
Follow up for https://github.com/cowprotocol/services/pull/2246

now the `/api/v1/auction` refers to the `AuctionOrder` which has it's own list of fields.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -54,13 +54,13 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       # Start the build process in the background. The following cargo test command will automatically
       # wait for the build process to be done before proceeding.
-      - run: cargo build -p orderbook -p database --tests &
+      - run: cargo build -p orderbook -p database -p autopilot --tests &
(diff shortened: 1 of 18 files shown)


crates/orderbook/openapi.yml
```diff
@@ -1448,3 +1557,39 @@ components:
         totalSurplus:
           type: string
           description: The total surplus.
+    InteractionData:
+      type: object
+      properties:
+        target:
+          $ref: "#/components/schemas/Address"
+        value:
+          type: number
```
MartinquaXD, line 1566: Since this is a `U256` it should probably be a decimal string.

fleupold (approved):
We should really find a way to autogenerate those 🙈

MartinquaXD:
> We should really find a way to autogenerate those 🙈

https://github.com/cowprotocol/services/issues/2203

crates/orderbook/openapi.yml
sunce86: Right, fixed!

Issue #2256, opened 2024-01-09
chore: Rename fee_policies to protocol_fee


sunce86:
### Background
https://github.com/cowprotocol/services/pull/2250#discussion_r1444442728

PR #2274, opened 2024-01-12
Rename fee policies

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2256

Simple renaming from `fee_policies` to `protocol_fees` to make sure the reader knows that the policies are used for protocol fee calculation.


fleupold (approved):
🙏

PR #2279, opened 2024-01-13
OnSettlementEventUpdater break blockchain dependency

sunce86:
# Description
First small step towards moving OnSettlementEventUpdater to `domain`.

This PR breaks the blockchain dependency and uses `infra::Ethereum` object.

Some of the following steps:
1. Break db dependency
2. Move `DecodedSettlement` to `domain`
3. Refactor `OnSettlementEventUpdater::update`: extract CIP20 data construction, create `domain::Transaction` object that will contain subset of onchain transaction data and used to build `DecodedSettlement` using `eth::Ethereum` etc...

# Changes

- [ ] Extracted contracts functionality into separate infra module
- [ ] Using `infra::Ethereum` instead of `web3` and contracts

### DIFF
--- a/crates/autopilot/src/infra/blockchain/mod.rs
+++ b/crates/autopilot/src/infra/blockchain/mod.rs
@@ -72,6 +72,7 @@ pub struct Ethereum {
     web3: DynWeb3,
     network: Network,
     current_block: CurrentBlockStream,
+    settlement: contracts::GPv2Settlement,
 }
 
 impl Ethereum {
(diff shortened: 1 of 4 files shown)


crates/autopilot/src/infra/blockchain/mod.rs
```diff
@@ -5,9 +5,12 @@ use {
     primitive_types::{H256, U256},
     std::{sync::Arc, time::Duration},
     thiserror::Error,
-    web3::types::TransactionReceipt,
 };
 
+pub mod contracts;
+
+pub use self::contracts::Contracts;
```
MartinquaXD, line 12: Since the entirety of the `contracts` module is also supported I think we might as well not have this re-export of the `Contracts` struct here.

MartinquaXD (approved):

crates/autopilot/src/infra/blockchain/mod.rs
sunce86: 👍  resolved

PR #2304, opened 2024-01-18
Fix check for market price

sunce86:
# Description
For an order with signed data
sell_amount = 100
buy amount = whatever

we get the quote response back (that get stored into db):
sell_amount = 90
buy_amount = whatever
fee_amount = 10

When checking if the order is within market price for fee policy, we decided to compare with quote that includes the fee_amount, so that we avoid a case where user submits an order which is out-of-market (meaning can't be fulfilled at the moment of creation) but, because of the fee_amount that is ignored, is considered in-market and therefore excluded from protocol fee taking.

To summarize, if order signed sell_amount is expressed as sell_amount_before_fee, then we also need to compare it to quote sell amount expressed as before fee.

### DIFF
--- a/crates/autopilot/src/infra/persistence/dto/quote.rs
+++ b/crates/autopilot/src/infra/persistence/dto/quote.rs
@@ -1,6 +1,7 @@
 use {
     crate::{boundary, domain},
     number::conversions::big_decimal_to_u256,
+    primitive_types::U256,
 };
 
 pub fn into_domain(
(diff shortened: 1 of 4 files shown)


crates/autopilot/src/infra/persistence/dto/quote.rs
```diff
@@ -1,18 +1,31 @@
...
     Ok(domain::Quote {
         order_uid: domain::OrderUid(quote.order_uid.0),
-        sell_amount: big_decimal_to_u256(&quote.sell_amount).ok_or(AmountOverflow)?,
-        buy_amount: big_decimal_to_u256(&quote.buy_amount).ok_or(AmountOverflow)?,
+        sell_amount: big_decimal_to_u256(&quote.sell_amount).ok_or(Error::AmountOverflow)?,
+        buy_amount: big_decimal_to_u256(&quote.buy_amount).ok_or(Error::AmountOverflow)?,
+        fee: {
+            let gas_amount = U256::from_f64_lossy(quote.gas_amount);
+            let gas_price = U256::from_f64_lossy(quote.gas_price);
+            let sell_token_price = U256::from_f64_lossy(quote.sell_token_price);
```
MartinquaXD, line 15: nit: Not sure if the values are such that it would make a difference but intuitively I would say it should be more accurate and easier to read to do the computation using floats and only convert the final result to `U256`.

MartinquaXD (approved):
Seems reasonable.

crates/autopilot/src/infra/persistence/dto/quote.rs
sunce86: Yeah, can do both. Working with f64 would silently lead to wrong values in case of overflow or division by zero, but don't expect this to happen ever in practice.

PR #2309, opened 2024-01-20
Always enable custom interactions and paying ETH to SCs

MartinquaXD:
# Description
We no longer need to be able to disable creating orders with custom interactions of orders paying ETH to smart contracts.

# Changes
Delete dead code

## How to test
CI, tests should continue to pass


sunce86 (approved):
🧹

PR #2316, opened 2024-01-23
Report only unique unsettled orders

sunce86:
# Description
We had one oncall alert where a lot of unsettled solutions were reported.

I noticed that this metric contains duplicated orders, meaning, if one solver solves solution A and 10 other solvers solve solution B we will have 10 unsettled solutions. 

Now, only unique solutions will be reported.


fleupold (approved):
👏

MartinquaXD (approved):

PR #2324, opened 2024-01-24
Fix bug for fee market price check

sunce86:
# Description
Quote sell amount and quote buy amount were substituted.

🤷

### DIFF
--- a/crates/autopilot/src/domain/fee/mod.rs
+++ b/crates/autopilot/src/domain/fee/mod.rs
@@ -51,8 +51,8 @@ impl ProtocolFee {
                     &order.data.sell_amount,
                     &order.data.buy_amount,
                     &order.data.fee_amount,
-                    &quote.buy_amount,
                     &quote.sell_amount,
+                    &quote.buy_amount,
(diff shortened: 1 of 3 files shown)


crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -51,8 +51,8 @@ impl ProtocolFee {
                     &order.data.sell_amount,
                     &order.data.buy_amount,
                     &order.data.fee_amount,
-                    &quote.buy_amount,
                     &quote.sell_amount,
+                    &quote.buy_amount,
```
MartinquaXD, line 55: Could have been caught easier with 2 structs with named fields like Felix suggested. Maybe it makes sense to make that change as well in this PR.

MartinquaXD (approved):
Approving since this appears to be the only case where args were not correct but named args make sense to me anyway.

crates/autopilot/src/domain/fee/mod.rs
sunce86: added

PR #2325, opened 2024-01-24
Allow custom timeout for e2e services' http client

devanoneth:
# Description
Allow custom timeout for the http client in `e2e`'s `services`. This means we can use the helper functions in `services` without needing to copy them into our own codebase and change the timeout.

# Changes
 -  `new_with_timeout` constructor added to `Services` impl. 

## How to test
Tested in Enso's e2e tests.

### DIFF
--- a/crates/e2e/src/setup/services.rs
+++ b/crates/e2e/src/setup/services.rs
@@ -31,6 +31,37 @@ pub const VERSION_ENDPOINT: &str = "/api/v1/version";
 pub const SOLVER_COMPETITION_ENDPOINT: &str = "/api/v1/solver_competition";
 const LOCAL_DB_URL: &str = "postgresql://";
 
+pub struct ServicesBuilder {
+    timeout: Duration,
+}
+
+impl Default for ServicesBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ServicesBuilder {
+    pub fn new() -> Self {
+        Self {
+            timeout: Duration::from_secs(10),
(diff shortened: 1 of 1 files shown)


crates/e2e/src/setup/services.rs
```diff
@@ -51,6 +51,14 @@ impl<'a> Services<'a> {
         }
     }
 
+    pub async fn new_with_timeout(contracts: &'a Contracts, timeout: Duration) -> Services<'a> {
```
MartinquaXD, line 54: Functions like `with_something()` usually indicate a builder pattern which also seems more appropriate in case we ever have to add more optional configurations you want to opt into.
So specifically I would recommend a signature like:
```
pub fn with_http_timeout(self, timeout: Duration) -> Self;
```
That way you can easily chain function calls modifying `Self`.

devanoneth: I was going to do something like this originally but then realized that `reqwest::Client` does not allow you to change the timeout after it is built. So, I would need to build a small `ServicesBuilder` struct and impl, which I don't mind doing of course if you prefer!

FWIW, secondary constructors with something like `with_foo` are part of the official API guidelines: https://rust-lang.github.io/api-guidelines/predictability.html#constructors-are-static-inherent-methods-c-ctor

But I understand if you prefer to have a proper `ServicesBuilder` already. :)

MartinquaXD (approved):

sunce86 (approved):

PR #2326, opened 2024-01-24
Baseline solver limit orders support

squadgazzz:
# Description
Adds support of 0x LimitOrder to Baseline solver in order to test 0x liquidity properly in e2e tests.

## How to test
Added unit tests.

## Related Issues
https://github.com/cowprotocol/services/issues/1666


crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+fn calculate_amount_out(
+    in_amount: U256,
+    maker_amount: U256,
+    taker_amount: U256,
+    fee: &TakerAmount,
+) -> Option<U256> {
+    let fee_adjusted_amount = in_amount.checked_sub(fee.0)?;
+    let fee_adjusted_amount_bfp = Bfp::from_wei(fee_adjusted_amount);
```
fleupold, line 36: I'm a bit surprised we are using balance Bfp logic here. Why is that?

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,125 @@
...
+    let scaled_maker_amount = Bfp::from_wei(maker_amount).mul_down(Bfp::exp10(18)).ok()?;
+    let scaled_price_ratio = scaled_maker_amount
+        .div_down(Bfp::from_wei(taker_amount))
+        .ok()?;
+    let scaled_out_amount_bfp = fee_adjusted_amount_bfp.mul_down(scaled_price_ratio).ok()?;
+    scaled_out_amount_bfp
```
fleupold, line 42: Don't we need to check here if the 0x LimitOrder is partially fillable and if `taker_amount >= in_amount`?

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,125 @@
...
+            token: eth::TokenAddress(addr!("c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2")),
+        };
+        let fee = TakerAmount(U256::from(fee_amount));
+
+        LimitOrder { maker, taker, fee }
+    }
+
+    #[test]
+    fn test_amount_out_in_round_trip() {
+        let maker_amount: u32 = 321;
```
fleupold, line 86: Are limit orders really using these type of amounts? Aren't they expressed in wei? Maybe using an actual example order could help (but I guess we are doing this in the e2e test).

fleupold (commented):
Approach in general looks correct! Excited to get 0x support in baseline...

crates/solvers/src/boundary/liquidity/limit_order.rs
squadgazzz: Switched to `wei`

crates/solvers/src/boundary/liquidity/limit_order.rs
squadgazzz: Is it possible to understand from the current fields if it is partial-fillable or not? https://0x.org/docs/0x-orderbook-api/api-references/overview#signed-order

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,125 @@
+use {
+    crate::domain::liquidity::limit_order::{LimitOrder, TakerAmount},
+    contracts::ethcontract::{H160, U256},
+    shared::{baseline_solver::BaselineSolvable, sources::balancer_v2::swap::fixed_point::Bfp},
+};
+
+impl BaselineSolvable for LimitOrder {
+    fn get_amount_out(&self, out_token: H160, (in_amount, in_token): (U256, H160)) -> Option<U256> {
+        if in_token == self.taker.token.0 && out_token == self.maker.token.0 {
```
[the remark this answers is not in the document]
squadgazzz: Looks like it is already taken into account: https://github.com/cowprotocol/services/blob/2e18b3699bba7599e4721b89e90137dd74526386/crates/solvers/src/boundary/baseline.rs#L47-L96

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,125 @@
...
+            calculate_amount_out(in_amount, self.maker.amount, self.taker.amount, &self.fee)
+        } else {
+            None
+        }
+    }
+
+    fn get_amount_in(&self, in_token: H160, (out_amount, out_token): (U256, H160)) -> Option<U256> {
+        if out_token == self.maker.token.0 && in_token == self.taker.token.0 {
+            calculate_amount_in(out_amount, self.maker.amount, self.taker.amount, &self.fee)
+        } else {
```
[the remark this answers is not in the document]
squadgazzz: Refactored a little bit

crates/solvers/src/boundary/baseline.rs
```diff
@@ -187,6 +187,17 @@ fn to_boundary_amms(liquidity: &[liquidity::Liquidity]) -> HashMap<TokenPair, Ve
                         }
                     }
                 }
+                liquidity::State::LimitOrder(limit_order) => {
+                    if let Some(token_pair) =
+                        TokenPair::new(limit_order.maker.token.0, limit_order.taker.token.0)
+                    {
+                        amms.entry(token_pair).or_default().push(Amm {
```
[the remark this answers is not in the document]
squadgazzz: Makes sense, renamed.

crates/solvers/src/boundary/liquidity/limit_order.rs
MartinquaXD: IIRC all 0x limit orders are partially fillable.

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn get_amount_out(&self, out_token: H160, (in_amount, in_token): (U256, H160)) -> Option<U256> {
+        if in_token != self.taker.token.0
+            || out_token != self.maker.token.0
+            || in_amount > self.taker.amount
+        {
+            return None;
+        }
+
+        let fee_adjusted_amount = in_amount.checked_sub(self.fee.0)?;
+        if self.maker.amount > self.taker.amount {
```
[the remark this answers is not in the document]
squadgazzz: We do not operate with floating point numbers here. Consider an example where the ratio is 0.5. Then, we should multiply the amounts that could lead to U256 overflow.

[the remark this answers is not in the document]
MartinquaXD: You could also get around that by not computing the ratio first and instead do all operations in order which is what we usually do:
```rust
fee_adjusted_amount * make.amount / taker.amount
```
If you want to be super careful in order to avoid overflows on the multiplication you could even do `a.full_mul(b)` which promotes the result to `U512`.
One thing to note throughout all this is that we should try to mimic what the actual smart contract does as closely as possible. Imagine they have an error in their equation or even some rounding issues. Then it does us no good if we have a theoretically better implementation if a solution generated by using it ends up reverting.

crates/solvers/src/boundary/baseline.rs
```diff
@@ -13,7 +13,7 @@ use {
 
 pub struct Solver<'a> {
     base_tokens: BaseTokens,
-    amms: HashMap<TokenPair, Vec<Amm>>,
+    amms: HashMap<TokenPair, Vec<OnchainLiquidity>>,
```
fleupold, line 16: What's the difference between amms and liquidity? Does it make sense to combine the two somehow (maybe in a separate PR)? Also `amms` should probably be renamed.

crates/solvers/src/boundary/liquidity/limit_order.rs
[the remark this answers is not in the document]
fleupold: Also let's add a small comment in the code explaining this non trivial rationale...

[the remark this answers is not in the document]
squadgazzz: True, tried to follow the contract's implementation: https://github.com/0xProject/protocol/blob/e66307ba319e8c3e2a456767403298b576abc85e/contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersSettlement.sol#L469-L509

crates/solvers/src/boundary/baseline.rs
squadgazzz: Renamed and created an issue https://github.com/cowprotocol/services/issues/2332

fleupold (approved):

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,192 @@
...
+            out_amount.checked_div(inverse_price_ratio)?
+        } else {
+            let price_ratio = self.taker.amount.checked_div(self.maker.amount)?;
+            out_amount.checked_mul(price_ratio)?
+        };
+        amount_before_fee.checked_add(self.fee.0)
+    }
+
+    fn gas_cost(&self) -> usize {
+        0
```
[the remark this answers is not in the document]
squadgazzz: Makes sense. Since currently we have liquidity limit orders only from 0x, adding the gas cost fetching could be done as a separate activity, right?

crates/solvers/src/boundary/liquidity/limit_order.rs
```diff
@@ -0,0 +1,192 @@
...
+impl BaselineSolvable for LimitOrder {
+    fn get_amount_out(&self, out_token: H160, (in_amount, in_token): (U256, H160)) -> Option<U256> {
+        if in_token != self.taker.token.0
+            || out_token != self.maker.token.0
+            || in_amount > self.taker.amount
+        {
+            return None;
+        }
+
+        let fee_adjusted_amount = in_amount.checked_sub(self.fee.0)?;
```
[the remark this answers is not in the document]
squadgazzz: Ah, does it mean we have to fetch the gas price each time we fetch 0x's limit orders from the orderbook?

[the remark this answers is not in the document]
squadgazzz: Btw, taker's token fee is not scaled AFAIU. And the protocol fee should not be considered while calculating in/out amounts.

crates/solvers/src/boundary/liquidity/limit_order.rs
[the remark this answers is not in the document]
MartinquaXD: I think we don't need a separate activity for that.
IIRC when we fetch liquidity sources we already assign a `cost` to each one which refers to their gas cost. For 0x orders we could simply multiply their fee by the current gas price (cached and already updated in the background). Then when the limit order liquidity source could simply return the gas cost previously assigned to it by the liquidity fetching logic.

crates/solvers/src/boundary/liquidity/limit_order.rs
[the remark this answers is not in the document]
MartinquaXD: If I understand the contract code correctly the `takerTokenFeeAmount` gets deducted after the order gets partially filled and does not reduce the available fill amount so I think it does not even have any influence on the math here.
Additionally the protocol fee was decided to be 0 according to the page I linked so I believe the logic is actually just in_amount * maker_amount / taker_amount.

MartinquaXD (approved):
Still not super sure about this code.
If you want to merge it as is I recommend watching the staging `baseline` simulations closely because an error in this code can lead to failing solutions or wait for the e2e test to 100% confirm the correctness of this code before merging both PRs together.

crates/solvers/src/boundary/liquidity/limit_order.rs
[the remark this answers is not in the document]
squadgazzz: Ok, simplified the formula to align with the one used in the smart contract. Added [this temp test](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c54d546b1bd978befb9f70a928de9b94) with the values from [this tx](https://phalcon.blocksec.com/explorer/tx/eth/0x9040f3901c2199efb7ab830e11c27516a6c039df5121fad439a8520015f5f51d?line=19&debugLine=19). The second assertion fails due to precision issue, which is expected, I believe.

crates/e2e/tests/e2e/liquidity.rs
```diff
@@ -206,47 +196,46 @@ async fn zero_ex_liquidity(web3: Web3) {
...
     if query.sender == Some(gpv2_addr) {
         let typed_order = Eip712TypedZeroExOrder {
-            maker_token: order_creation.sell_token,
-            taker_token: order_creation.buy_token,
-            maker_amount: order_creation.sell_amount.as_u128() * 2,
-            taker_amount: order_creation.buy_amount.as_u128() * 2,
+            maker_token: order_creation.buy_token,
+            taker_token: order_creation.sell_token,
+            maker_amount: order_creation.buy_amount.as_u128() * 3,
+            taker_amount: order_creation.sell_amount.as_u128() * 2,
```
squadgazzz, line 226: This multiplication is required to pass [this check](https://github.com/cowprotocol/services/blob/baseline-solver-limit-orders-support/crates/solvers/src/domain/solution.rs#L203-L205). Not sure it is 100% valid.

squadgazzz:
Apologies for re-requesting review. Would appreciate double-checking the `forked_node_zero_ex_liquidity_mainnet` e2e test correctness of the overall logic and values.

crates/e2e/src/nodes/forked_node.rs
```diff
@@ -61,4 +61,16 @@ impl<T: Transport> ForkedNodeApi<T> {
...
+    // whitelisted.
+    pub fn set_mocked_settle(&self) -> CallFuture<(), T::Out> {
+        let gpv2_auth_address_json =
+            serde_json::json!("0x2c4c28DDBdAc9C5E7055b4C863b72eA0149D8aFE");
+        let bytecode_json = serde_json::json!("0x600160005260206000F3");
+        CallFuture::new(
+            self.transport
+                .execute("anvil_setCode", vec![gpv2_auth_address_json, bytecode_json]),
+        )
+    }
```
fleupold, line 75: Why is this necessary now? Can't we just use an existing solver address or whitelist one we need? How is the [existing e2e test](https://github.com/cowprotocol/services/blob/3a6b13e0669e629f9a0f36c5f10bd85f648076a9/crates/e2e/tests/e2e/limit_orders.rs#L491) achieving this?

crates/e2e/tests/e2e/liquidity.rs
fleupold: 🤔 to me it seems that the inequality in the check should be a `<=` instead of `<`.

Why even multiply with 2 (maybe add a comment)?

squadgazzz:
> Why even multiply with 2 (maybe add a comment)?

Otherwise, the solution is discarded since the simulation is reverted on-chain. Gonna need to reproduce it on Tenderly.

crates/e2e/tests/e2e/liquidity.rs
```diff
@@ -207,49 +195,80 @@ async fn zero_ex_liquidity(web3: Web3) {
...
+            expiry: NaiveDateTime::MAX.timestamp() as u64,
+            salt: U256::from(Utc::now().timestamp()),
+        };
+        let usdt_weth_order = Eip712TypedZeroExOrder {
+            maker_token: weth_address,
             taker_token: order_creation.buy_token,
-            maker_amount: order_creation.sell_amount.as_u128() * 2,
-            taker_amount: order_creation.buy_amount.as_u128() * 2,
+            // add sufficient liquidity
+            maker_amount: 1_000_000_000_000_000_000u128,
```
MartinquaXD, line 241: FYI this amount comes from `--amount-to-estimate-prices-with`. Probably a good idea to add a comment for the purpose of each order. (e.g. main quote, native price sell_token, native price buy_token).

crates/solvers/src/boundary/baseline.rs
```diff
@@ -137,10 +137,12 @@ impl<'a> Solver<'a> {
     }
 }
 
-fn to_boundary_amms(liquidity: &[liquidity::Liquidity]) -> HashMap<TokenPair, Vec<Amm>> {
+fn to_boundary_amms(
```
MartinquaXD, line 140: Now with most things renamed to onchain liquidity maybe `to_boundary_liquidity` is more appropriate.

crates/e2e/tests/e2e/liquidity.rs
MartinquaXD: Maybe instead of adjusting the amount with these factors it's a better idea to make the order partially fillable. That way we don't have to have a comment for all the factors but only when we assert that the traded amounts look good in the end.

crates/e2e/src/setup/colocation.rs
```diff
@@ -172,6 +171,7 @@ http-timeout = "10s"
 
 [submission]
 gas-price-cap = "1000000000000"
+max-confirm-time= "2s"
```
MartinquaXD, line 174: Is this actually needed?
I would assume in a forked test we'd bring the tx on-chain immediately and don't even need this upper limit.

MartinquaXD (commented):
The diff is a bit tricky to review as it is. Maybe it makes sense to have this plus the PR introducing the 0x liquidity as one for reviewing purposes?

squadgazzz:
Indeed, it's hard to review at the moment, I will merge it into the e2e test original branch.

Issue #2310, opened 2024-01-22
chore: Optionally enable quote verification


MartinquaXD:
### Background
Currently the frontend has to tell us that the backend should verify a quote by passing `priceQuality: "verified"` in the `/quote` request.
To easily enforce in the future that all quotes have been verified we need to make sure the backend is able to decide on its own which quotes to verify.

### Details
We should start by verifying all quotes where `from` is not `0` and the `price_quality` is not `fast`.
This should only happen conditionally based on a CLI flag in case we encounter any issues and have to quickly revert back to the status quo.

PR #2327, opened 2024-01-25
Move TradeVerifier into separate file and add trait

MartinquaXD:
# Description
Mostly moves code around and introduces trait for trade verification which will later be used for unit tests.
Also adds the `verified` flag to quotes but at this point this flag is not yet used for anything.

## How to test
Originally part of https://github.com/cowprotocol/services/pull/2271 and changes were manually tested there.
CI

Fixes: https://github.com/cowprotocol/services/issues/2310

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -487,7 +487,7 @@ pub struct Estimate {
     pub gas: u64,
     /// Address of the solver that provided the quote.
     pub solver: H160,
-    /// Did we verify the correctness of this estimates properties?
+    /// Did we verify the correctness of this estimate's properties?
     pub verified: bool,
 }
 
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -40,6 +40,7 @@ pub mod oneinch;
 pub mod paraswap;
 pub mod sanitized;
 pub mod trade_finder;
+pub mod trade_verifier;
 pub mod zeroex;
 
 #[derive(Clone, Debug)]
@@ -486,6 +487,8 @@ pub struct Estimate {
     pub gas: u64,
     /// Address of the solver that provided the quote.
     pub solver: H160,
+    /// Did we verify the correctness of this estimates properties?
+    pub verified: bool,
 }
 
 impl Estimate {
--- a/crates/shared/src/price_estimation/trade_verifier.rs
+++ b/crates/shared/src/price_estimation/trade_verifier.rs
@@ -29,8 +29,8 @@ use {
 
 #[async_trait::async_trait]
 pub trait TradeVerifying: Send + Sync + 'static {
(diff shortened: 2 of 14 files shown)


crates/shared/src/price_estimation.rs
```diff
@@ -486,6 +487,8 @@ pub struct Estimate {
     pub gas: u64,
     /// Address of the solver that provided the quote.
     pub solver: H160,
+    /// Did we verify the correctnes of this estimates properties?
```
sunce86, line 490: `correctness`

crates/shared/src/price_estimation/trade_verifier.rs
```diff
@@ -0,0 +1,338 @@
...
+        signature::{Signature, SigningScheme},
+    },
+    number::nonzero::U256 as NonZeroU256,
+    std::sync::Arc,
+    web3::{ethabi::Token, types::CallRequest},
+};
+
+#[async_trait::async_trait]
+pub trait TradeVerifying: Send + Sync + 'static {
+    /// Verifies that the proposed [`Trade`] acually fulfills the [`PriceQuery`]
```
sunce86, line 32: `actually`

sunce86 (approved):

crates/shared/src/price_estimation.rs
MartinquaXD: Oh, no. Not at the base of my PR stack! :sweat_smile:

crates/shared/src/price_estimation/trade_verifier.rs
```diff
@@ -0,0 +1,338 @@
...
+            );
+        }
+
+        let output = self.simulator.simulate(call, overrides).await?;
+        let summary = SettleOutput::decode(&output)?;
+        let verified = Estimate {
+            out_amount: summary.out_amount(query.kind)?,
+            gas: summary.gas_used.as_u64(),
+            solver: trade.solver,
+            verified: true,
```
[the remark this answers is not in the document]
MartinquaXD: > Do we even need the Noop one (couldn't we just keep TradeVerifier optional wherever it's used?)

At some point I want to write tests for the logic that prefers verified quotes over unverified ones. Because I think e2e test will be tricky to pull unit test would be the only option.
In order to mock a trade verifier for a unit test I will have to use a trait.
This plus the fact that we eventually want to enforce that we always have a verifier available (and simply don't use it for some quotes) lead to me to model optional trade verification via the NoopVerifier instead of the optional verifier.
When we are ready to always enforce the availability of a trade verifier we can simply drop the NoopVerifier.

> I find it a bit weird that the Noop tradeverifier returns verified: false and the real one returns an error in case it's unable to verify.

Hmm, true. Being able to return an error or `verified: false` is ambiguous. Would be cleaner to return something like `Result<VerifiedQuote, Error>` instead and always return `anyhow!("verification disabled")` from the NoopVerifier.

Will think more about the design.

PR #2329, opened 2024-01-25
fix: extend anvil timeout

devanoneth:
# Description
Depending on the RPC provider, anvil can sometimes take a bit longer than 5 seconds to start up.

# Changes
 - Increase anvil URL timeout from 5 seconds to 15.
 - Parse and store anvil URL in `Node` struct. I am doing this now as eventually I'd like to be able to run multiple anvil instances on different ports to parallelize fork tests.

## How to test
CI will run the fork tests, also tested locally.

P.S: This should be the last of my e2e PRs for a bit, I have just been adding a lot more e2e tests to the Enso solver. Appreciate the reviews as always! :)


MartinquaXD (approved):
> ...as eventually I'd like to be able to run multiple anvil instances on different ports to parallelize fork tests.

You are not the only one. :smile: 
This will not be straight forward. I already spent a bunch of time on that. The big problem is that we need a DB for the e2e tests and since we are using `postgres` we can't spawn multiple instances in memory. In case you are curious here is my [PR](https://github.com/cowprotocol/services/pull/2023) trying to get it done but the tests are still not 100% reproducible despite using docker. :/

devanoneth:
Ooft, that's quite the PR! Appreciate you sending that on before I went down a rabbit hole of my own. Will take a look at that before going further myself, thanks!

squadgazzz (approved):

devanoneth:
Closing as this was also resolved by #2336.

PR #2339, opened 2024-01-30
Add block_number to order_execution

sunce86:
# Description
Adds `block_number` column to db table `order_execution`.

This is to make the table reorg friendly, meaning, whenever reorg happens, event updater will also delete all entries in this table that are from reorged block.

This enables us to remove the delay of 64 blocks in the `OnSettlementEventUpdater` component. Removal will be implemented in a separate PR so that it can be easily reverted if need be.

### DIFF
--- a/crates/database/src/events.rs
+++ b/crates/database/src/events.rs
@@ -72,7 +72,7 @@ pub async fn delete(
     ex.execute(sqlx::query(QUERY_PRESIGNATURES).bind(delete_from_block_number))
         .await?;
 
-    // Observations and order executions are not events but data deducted from the
+    // Observations and order executions are not events but data derived from the
     // onchain data. The reason we delete them here is that we want to keep the
     // database state consistent.
 
--- a/crates/database/src/events.rs
+++ b/crates/database/src/events.rs
@@ -68,13 +68,21 @@ pub async fn delete(
     ex.execute(sqlx::query(QUERY_SETTLEMENTS).bind(delete_from_block_number))
         .await?;
 
+    const QUERY_PRESIGNATURES: &str = "DELETE FROM presignature_events WHERE block_number >= $1;";
+    ex.execute(sqlx::query(QUERY_PRESIGNATURES).bind(delete_from_block_number))
+        .await?;
+
+    // Observations and order executions are not events but data deducted from the
+    // onchain data. The reason we delete them here is that we want to keep the
+    // database state consistent.
+
     const QUERY_OBSERVATIONS: &str =
         "DELETE FROM settlement_observations WHERE block_number >= $1;";
     ex.execute(sqlx::query(QUERY_OBSERVATIONS).bind(delete_from_block_number))
         .await?;
 
-    const QUERY_PRESIGNATURES: &str = "DELETE FROM presignature_events WHERE block_number >= $1;";
-    ex.execute(sqlx::query(QUERY_PRESIGNATURES).bind(delete_from_block_number))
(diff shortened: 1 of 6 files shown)


crates/database/src/events.rs
```diff
@@ -68,13 +68,21 @@ pub async fn delete(
     ex.execute(sqlx::query(QUERY_SETTLEMENTS).bind(delete_from_block_number))
         .await?;
 
+    const QUERY_PRESIGNATURES: &str = "DELETE FROM presignature_events WHERE block_number >= $1;";
+    ex.execute(sqlx::query(QUERY_PRESIGNATURES).bind(delete_from_block_number))
+        .await?;
+
+    // Observations and order executions are not events but data deducted from the
```
MartinquaXD, line 75: The word should probably be deduced or derived.

database/sql/V059__add_block_number_order_execution.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE order_execution
+    ADD COLUMN block_number bigint NOT NULL DEFAULT 0;
```
MartinquaXD, line 2: The column should probably drop the `DEFAULT` modifier after the migration populated the old entries.

database/sql/V059__add_block_number_order_execution.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE order_execution
```
MartinquaXD, line 1: Please also update the DB readme.md. 🙏

MartinquaXD (commented):
Should this PR also already reduce the reorg safety threshold used in the OnSettlementEventUpdater?

database/sql/V059__add_block_number_order_execution.sql
sunce86: Aaaaaa I will never remember this 😮‍💨

sunce86:
> Should this PR also already reduce the reorg safety threshold used in the OnSettlementEventUpdater?

Removal will be implemented in a separate PR so that it can be easily reverted if need be.

crates/database/src/events.rs
sunce86: fixed

database/sql/V059__add_block_number_order_execution.sql
sunce86: dropped `DEFAULT` but added `COALESCE(settlements.block_number, 0)` so that we have `0` inserted if there is no corresponding auction_id in `settlements`.

database/sql/V059__add_block_number_order_execution.sql
sunce86: added

squadgazzz (approved):

database/sql/V059__add_block_number_order_execution.sql
```diff
@@ -0,0 +1,8 @@
...
+    ADD COLUMN block_number bigint NOT NULL;
```
MartinquaXD, line 2: IIRC you have to:
1. add the new column with some default value (existing columns don't have a value yet)
2. populate existing rows
3. drop the default modifier

Otherwise postgres will probably complain when you try to add the new column. Or did you try to run the migration locally while you already had some order_execution rows in the DB?

sunce86: I ran a migration locally and it did not complain, however it was done on an empty db.
I adjusted the migration to what you suggest and also did an example run with non-empty db and it works as expected, so I believe this version is the way to go.

fleupold (approved):

MartinquaXD (approved):

PR #2373, opened 2024-02-06
Update Flyway to 10.7.1

squadgazzz:
# Description
The current version doesn't work on ARM architecture. I had to update to v8, but we can probably jump directly to the latest version.

## How to test
Local/CI runs


fleupold (approved):
I use it in docker, but ok

MartinquaXD:
It should also be more reliable. When I worked on dockerizing the e2e tests I had to bump the version as well because sometimes it didn't apply the migrations. :/

MartinquaXD (approved):

PR #2377, opened 2024-02-06
Price improvemenet fee policy autopilot argument

squadgazzz:
# Description
Closes task n1 from https://github.com/cowprotocol/services/issues/2287


crates/autopilot/src/arguments.rs
```diff
@@ -380,6 +387,9 @@ impl FeePolicy {
 pub enum FeePolicyKind {
     /// How much of the order's surplus should be taken as a protocol fee.
     Surplus { factor: f64, max_volume_factor: f64 },
+    /// How much of the order's price improvement should be taken as a protocol
```
MartinquaXD, line 390: Would be good to define `price improvement` since this is a bit nuanced.

crates/autopilot/src/arguments.rs
```diff
@@ -407,6 +417,22 @@ impl FromStr for FeePolicyKind {
                     max_volume_factor,
                 })
             }
+            "price-improvement" => {
```
MartinquaXD, line 420: I believe we currently tend to prefer `camelCase` over all so it makes sense to also use it here IMO.

MartinquaXD (approved):

fleupold (approved):

sunce86 (approved):

PR #2388, opened 2024-02-08
Make legacy http estimators compatible with quote verification

MartinquaXD:
# Description
So far quote verification was heavily geared towards price estimation via drivers. But if we want to try out verification in prod to get a good idea on how well the feature works we should also add support for the legacy http price estimators.

# Changes
Applies existing pattern of putting a `TradeFinder` into a `TraderEstimator`. This design seems overall pretty convoluted to me but this is just the last stretch before we can finally move towards quotes via drivers so it doesn't seem reasonable to introduce yet another adapter pattern to make the price estimation hierarchy work somehow.

The only estimators remaining that are incompatible with quote verification is the baseline and balancer solver. I think it's reasonable to not implement the same thing for those and instead use the data from all the other estimators, polish quote verification, transition completely to driver based price estimation and drop all the legacy price estimation code.

## How to test
CI

### DIFF
--- a/crates/shared/src/price_estimation/factory.rs
+++ b/crates/shared/src/price_estimation/factory.rs
@@ -4,7 +4,7 @@ use {
         baseline::BaselinePriceEstimator,
         competition::CompetitionEstimator,
         external::ExternalPriceEstimator,
-        http::HttpPriceEstimator,
+        http::{HttpPriceEstimator, HttpTradeFinder},
         instrumented::InstrumentedPriceEstimator,
         native::{self, NativePriceEstimator},
         native_price_cache::CachingNativePriceEstimator,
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Looks fine, although I'm surprised we are shipping verified quotes without drivers. What is missing to turn this on (I was under the impression this was a requirement for verified quotes)?

MartinquaXD:
Initially I thought adding support for the legacy estimators was more involved than this so I only focused on the driver based verification.
I think figuring out a nice way to report gas costs from the driver based solver (currently we use 290K by default) would delay the rollout more than just adding support for the legacy solver. But in general I still want to kill the old price estimator setup with fire ASAP.

PR #2391, opened 2024-02-08
More logging for verified quotes

MartinquaXD:
# Description
It's possible that we see either failed simulations or simulations that return vastly different `out_amount` or `gas` costs. For these it would be useful to manually check with tenderly to make sure that this was not a logic error in our code.

Initially I just wanted to just log the calldata and block so you can manually recreate the simulation with tenderly but then I remembered that we also use 2 fake contracts in the simulation which would be super annoying (maybe impossible?) to set up using the tenderly UI.

We already had code that immediately simulated a tx using tenderly and only logged a link to the simulation but IIRC indiscriminately creating simulations with tenderly was pretty expensive and we don't really know upfront which simulations might be interesting to look at so we would have to log everything.
Instead I adjusted the `Web3ThenTenderly` code simulation implementation such that it first simulates using web3 and afterwards logs a shell command that creates the simulation on tenderly and spits out a link afterwards.
This is free until we actually manually trigger the simulation when debugging something. The only unfortunate thing is that these logs can get quite long so it could be that we run into issues with Kibana.

# Changes
- reorders logged data from most relevant to least relevant
- logs latency introduced by verification
- command to recreate simulation with tenderly

## How to test
Used a local test setup to produce the new log and checked it out on [tenderly](https://dashboard.tenderly.co/devcow/project/simulator/fcb67f32-50e6-4e55-a9d0-a59fd651bcf7?trace=0.1.17.0).

Screenshot 2024-02-09 at 00 40 11

### DIFF
--- a/crates/shared/src/code_simulation.rs
+++ b/crates/shared/src/code_simulation.rs
@@ -24,6 +24,7 @@ pub trait CodeSimulating: Send + Sync + 'static {
         &self,
         call: CallRequest,
         overrides: StateOverrides,
+        block: Option<u64>,
     ) -> Result<Vec<u8>, SimulationError>;
 }
 
@@ -63,10 +64,14 @@ impl CodeSimulating for Web3 {
         &self,
         call: CallRequest,
         overrides: StateOverrides,
+        block: Option<u64>,
     ) -> Result<Vec<u8>, SimulationError> {
+        let block = block
+            .map(|b| BlockNumber::Number(b.into()))
+            .unwrap_or(BlockNumber::Latest);
         Ok(self
             .eth()
-            .call_with_state_overrides(call, BlockNumber::Latest.into(), overrides)
+            .call_with_state_overrides(call, block.into(), overrides)
             .await?
             .0)
     }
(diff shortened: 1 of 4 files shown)


crates/shared/src/tenderly_api.rs
```diff
@@ -102,6 +102,24 @@ impl TenderlyApi for TenderlyHttpApi {
...
     }
 
+    fn log(&self, simulation: SimulationRequest) -> Result<()> {
+        let request_url = crate::url::join(&self.api, "simulate");
+        let simulation_url =
+            crate::url::join(&self.dashboard, "simulator/$SIMULATION_ID").to_string();
+        let body = serde_json::to_string(&simulation)?;
+
+        #[rustfmt::skip]
+        tracing::error!(
```
squadgazzz, line 112: Since this is going to be logged for both `err` and `ok` cases, why do we need it on the `error` level?

MartinquaXD: Oops, should be `trace` or `debug`. During development I tend to emit new logs as `error` because that is easy to see in our see of logs.
Good catch!

crates/shared/src/code_simulation.rs
```diff
@@ -101,6 +106,46 @@ impl TenderlyCodeSimulator {
...
+        call: CallRequest,
+        overrides: StateOverrides,
+        block: Option<u64>,
+    ) -> Result<SimulationRequest> {
+        Ok(SimulationRequest {
+            block_number: block,
+            network_id: self.network_id.clone(),
+            from: call.from.unwrap_or_default(),
+            to: call.to.unwrap_or_default(),
+            input: call.data.unwrap_or_default().0,
```
sunce86, line 121: any reason why we unwrap here? I assume you expect we always have this data so no point in handling this case?

crates/shared/src/tenderly_api.rs
```diff
@@ -102,6 +102,24 @@ impl TenderlyApi for TenderlyHttpApi {
         Ok(serde_json::from_str(&body)?)
     }
 
+    fn log(&self, simulation: SimulationRequest) -> Result<()> {
+        let request_url = crate::url::join(&self.api, "simulate");
+        let simulation_url =
+            crate::url::join(&self.dashboard, "simulator/$SIMULATION_ID").to_string();
+        let body = serde_json::to_string(&simulation)?;
```
sunce86, line 109: nit: any test that this is properly formatted?

sunce86 (approved):
lg

crates/shared/src/tenderly_api.rs
MartinquaXD: This would only work with a test that gets ignored by default (since we need a secret and don't want to waste limited tenderly resources on tests) and those tend to be dead code anyway because people usually don't rerun them regularly AFAIK.
I tested the code manually and ran the suggested command which produced the expected output, though.

PR #2392, opened 2024-02-08
Log call data as hex string

MartinquaXD:
# Changes
Some of the logs still contained calldata logged as an array of integers instead of a hex string. This is fixed by using `Derivative` to only overwrite the `Debug` implementation of these fields specifically.
For this I copied and slightly adjusted 1 helper function from the `shared` crate to not make the `model` crate depend on `shared` just for that.

## How to test
Look into the logs of the local node tests.


squadgazzz (approved):

fleupold (approved):
nit: This method could maybe be moved into `model::debug` or `model::format`

PR #2397, opened 2024-02-10
Prioritize newer orders in auction

MartinquaXD:
# Description
When we build the auction in the autopilot we filter out orders that are unlikely to be filled to shrink the auction to a more manageable size.
One step is to filter out fill-or-kill orders where the owner doesn't have enough sell_token for all of the open orders. The new logic sorts orders by `creation_timestamp` (desc) to first use the user's existing balance on recent orders rather than old ones. 

Fixes: #1997


sunce86 (approved):
Lg.

squadgazzz (approved):

PR #2399, opened 2024-02-12
Refactor surplus calculation

sunce86:
# Description
Only refactor, no new functionalities added.

I've extracted the calculation of surplus **in surplus token** into separate function. Since this new function is not strictly related to `solution::fee` and since it will be reused for rank by surplus feature, I moved it to the `solution::trade` file where it belongs.

The rest of surplus functionalities needed strictly for `solution::fee`, like `surplus_in_sell_token` is kept in that file.

## How to test
Existing tests.

### DIFF
--- a/crates/driver/src/domain/competition/solution/fee.rs
+++ b/crates/driver/src/domain/competition/solution/fee.rs
@@ -111,25 +111,6 @@ impl Fulfillment {
         }
     }
 
-    /// Returns the surplus denominated in the sell token.
-    fn surplus_in_sell_token(
-        &self,
-        sell_amount: eth::U256,
-        buy_amount: eth::U256,
-        prices: ClearingPrices,
-    ) -> Result<eth::U256, Error> {
-        let surplus = self.surplus(sell_amount, buy_amount, prices)?;
-        let surplus_in_sell_token = match self.order().side {
-            Side::Buy => surplus,
-            Side::Sell => surplus
-                .checked_mul(prices.buy)
-                .ok_or(trade::Error::Overflow)?
-                .checked_div(prices.sell)
-                .ok_or(trade::Error::DivisionByZero)?,
-        };
-        Ok(surplus_in_sell_token)
-    }
-
     /// Computes protocol fee compared to the given reference amounts taken from
     /// the order or a quote.
     fn calculate_fee(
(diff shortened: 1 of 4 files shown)


squadgazzz (approved):

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -127,6 +134,76 @@ impl Fulfillment {
...
+    /// Returns the surplus denominated in the surplus token.
+    ///
+    /// The surplus token is the buy token for a sell order and sell token for a
+    /// buy order.
+    pub fn surplus(
+        &self,
+        sell_amount: eth::U256,
+        buy_amount: eth::U256,
+        prices: ClearingPrices,
+    ) -> Result<eth::U256, Error> {
```
MartinquaXD, line 147: The signature looks a bit odd now. Since `Trade` should already know its own limit price and executed amount we should only have to pass in a clearing price to get a surplus.
This is because we want to compute surplus of a trade compared to some price other than the limit price as well for the different fee mechanisms.
Maybe this would be slightly nicer with 2 functions like `Trade::surplus_over_reference_price()` which takes the reference `buy_price` and `sell_price` (currently `sell_amount` and `buy_amount` respectively) and `Trade::surplus_over_limit_price()` which simply calls the other function with the orders limit price.

crates/driver/src/domain/competition/solution/fee.rs
```diff
@@ -105,6 +111,25 @@ impl Fulfillment {
         }
     }
 
+    /// Returns the surplus denominated in the sell token.
+    fn surplus_in_sell_token(
```
MartinquaXD, line 115: Sounds like this should also be a function on `Trade`. Additionally one could consider changing the signature to only take `surplus` and `prices`.
This together with the other suggestion might make the individual fee implementations more explicit.

MartinquaXD (commented):
Moving the logic into the trade makes sense to me. I think we should just be careful to not throw around `sell_amount` and `buy_amount` unnecessarily because they can have so many different meanings in different contexts.

crates/driver/src/domain/competition/solution/trade.rs
sunce86: I renamed the function to `surplus_over_reference_price`.

crates/driver/src/domain/competition/solution/fee.rs
sunce86: moved to `Fulfillment`

MartinquaXD (approved):

PR #2402, opened 2024-02-12
Cow amm mitigation

MartinquaXD:
# Description
There are some rough edges in the current CoW AMM implementation. This PR addresses those until they are fixed on the SC level.

# Changes
Made `autopilot` and `driver` aware of `CoW AMM` orders and enforced at most 1 order per contract in each auction and solution.


crates/driver/src/domain/competition/mod.rs
```diff
@@ -103,6 +103,33 @@ impl Competition {
...
+                        order.trader().0
+                    }
+                    solution::Trade::Jit(j) => j.order().signature.signer,
+                };
+                let is_cow_amm_order = self.eth.contracts().cow_amm_contracts().contains(&owner);
+                if is_cow_amm_order {
+                    let entry = amm_orders.entry(owner).or_default();
+                    *entry += 1;
+                    if *entry > 1 {
+                        tracing::warn!("solution contains more than 1 order for the same CoW AMM");
```
sunce86, line 125: can we log solution id?

crates/driver/src/infra/blockchain/contracts.rs
```diff
@@ -10,12 +11,14 @@ pub struct Contracts {
     vault_relayer: eth::ContractAddress,
     vault: contracts::BalancerV2Vault,
     weth: contracts::WETH9,
+    cow_amm_contracts: HashSet<eth::Address>,
```
sunce86, line 14: nit: rename `cow_amms`

sunce86 (approved):
lg, just nits feel free to ignore for temporary PR

crates/driver/src/domain/competition/mod.rs
```diff
@@ -103,6 +103,37 @@ impl Competition {
...
+                            .unwrap();
+                        order.trader().0
+                    }
+                    solution::Trade::Jit(j) => j.order().signature.signer,
+                };
+                let is_cow_amm_order = self.eth.contracts().cow_amms().contains(&owner);
+                if is_cow_amm_order {
+                    let entry = amm_orders.entry(owner).or_default();
+                    *entry += 1;
+                    if *entry > 1 {
```
squadgazzz, line 124: We can probably replace `HashMap::<eth::Address, usize>` with `HashSet::<eth::Address>`

MartinquaXD: True. Initially collected the total counts in the map and returned false when some count was greater than 1.

squadgazzz (approved):

fleupold (commented):
Not important since we will revert soon, but I was expecting to also check in the autopilot solution inspection (where we decide which solver to call `settle` on) to check the executed order IDs to make sure only one order per ERC1271 contract is used.

MartinquaXD:
Felt like that would not have been needed given the fact that we are currently running all drivers and we also have a check for that in the driver.

PR #2404, opened 2024-02-13
Driver tests refactoring for protocol fees

### LINKED ISSUE #2267: chore: Add driver tests for protocol fees
### Background
Only e2e tests are added for protocol fees.

In order to more effectively test the full matrix of possible cases (sell/buy order, fee policy kinds, existing order_fee etc), we should
1. Add possibility to test protocol fees in the driver 
2. Implement tests

More context here: https://github.com/cowprotocol/services/pull/2213#discussion_r1443751990

squadgazzz:
# Description
Refactors driver tests to be able to use them in UT fashion by providing custom quotes and explicit expected results without replicating original protocol fees logic in the tests.

The easiest approach is to continue using Uniswap on-chain interactions, but it turned out that it has some limitations where the pool's executed amounts ratio should be the same as the pool's reserves ratio. The proposed approach addresses this limitation by adjusting the pool's reserves by providing precalculated executed amounts using a reversed function. The precision issue with calculations without a floating point adds complexity to the implemented functions.

# Changes
- [ ] Adds ability to provide custom pool's executed amounts
- [ ] Results are now compared with precalculated values
- [ ] Minor refactoring to operate with WEI values for better readability
- [ ] Pool's reserve token `a` or `b` adjustment functions could be used interchangeably

## Related Issues
Fixed #2267


sunce86 (approved):
I think reusing the pool with adjusted reserves is the way to go. Happy to see this works.

crates/driver/src/tests/setup/mod.rs
```diff
@@ -428,22 +556,40 @@ pub fn ab_pool() -> Pool {
...
+        liquidity_provider: LiquidityProvider::Amm,
+    }
+}
+
+pub fn ab_pmm_pool(quote: &OrderQuote) -> Pool {
+    Pool {
+        liquidity_provider: LiquidityProvider::Pmm,
+        ..ab_pool()
     }
+    .adjusted_reserve_b(quote)
```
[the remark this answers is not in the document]
squadgazzz: It doesn't depend on the order side. Just wanted to keep the `adjusted_reserve_a ` until all fees are covered with tests.

crates/driver/src/tests/cases/mod.rs
```diff
@@ -18,30 +21,55 @@ pub mod solver_balance;
 /// Example solver name.
 const SOLVER_NAME: &str = "test1";
 
+// All values below in WEI
```
fleupold, line 24: Aren't all values now in e18?

crates/driver/src/tests/cases/mod.rs
```diff
@@ -18,30 +21,55 @@ pub mod solver_balance;
...
 
 /// The default solver fee for limit orders.
-pub const DEFAULT_SOLVER_FEE: u128 = 100u128;
+pub const DEFAULT_SOLVER_FEE: f64 = 1e-16;
 
 /// The default maximum value to be payout out to solver per solution
-pub const DEFAULT_SCORE_CAP: u128 = 10000000000000000u128;
+pub const DEFAULT_SCORE_CAP: f64 = 0.01;
+
+pub trait ToWei {
```
fleupold, line 54: nit: from the signature this looks more like an `into_wei`. Also, is this change even needed now (i don't see where we are using the float representation)

crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -0,0 +1,204 @@
...
+        .name(test_name)
+        .pool(pool)
+        .order(order)
+        .solution(ab_solution())
+        .done()
+        .await;
+    let expected = ExpectedOrder {
+        name: ab_order().name,
+        executed_sell_amount: 100u32.to_wei(),
+        executed_buy_amount: 40u32.to_wei(),
```
fleupold, line 38: Can we make it so that from reading the test it's easy to understand the math? Ie, here I would expect us to set the limit price and executed price explicitly (which then makes it clear why we expect the following executed amounts).

crates/driver/src/tests/setup/driver.rs
```diff
@@ -76,12 +79,7 @@ pub fn solve_req(test: &Test) -> serde_json::Value {
             "protocolFees": match quote.order.kind {
                 order::Kind::Market => json!([]),
                 order::Kind::Liquidity => json!([]),
-                order::Kind::Limit { .. } => json!([{
-                    "surplus": {
-                        "factor": 0.0,
-                        "maxVolumeFactor": 0.06
-                    }
-                }]),
+                order::Kind::Limit { .. } => json!([quote.order.fee_policy.to_json_value()]),
```
fleupold, line 82: Don't we expect this in all variants? How will this work once we have fees for market orders?

crates/driver/src/tests/setup/blockchain.rs
```diff
@@ -51,11 +51,18 @@ pub struct Interaction {
...
+    Amm,
+    Pmm,
+}
+
 /// A uniswap pool deployed as part of the blockchain setup.
 #[derive(Debug, Clone, Copy)]
 pub struct Pool {
     pub reserve_a: Asset,
     pub reserve_b: Asset,
+    pub liquidity_provider: LiquidityProvider,
```
fleupold, line 65: This nesting doesn't seem coherent. `LiquidityProvider:PMM` doesn't seem to care about reserves. Shouldn't Pool become a higher level `Liquidity` enum which can be either a `Pool` or `PMM` (both of which can implement `Into<Liquidity>`

crates/driver/src/tests/setup/mod.rs
```diff
@@ -428,22 +558,40 @@ pub fn ab_pool() -> Pool {
...
-        amount_b: DEFAULT_POOL_AMOUNT_B.into(),
+        amount_a: DEFAULT_POOL_AMOUNT_A.to_wei(),
+        amount_b: DEFAULT_POOL_AMOUNT_B.to_wei(),
+        liquidity_provider: LiquidityProvider::Amm,
+    }
+}
+
+pub fn ab_pmm_pool(quote: &OrderQuote) -> Pool {
+    Pool {
+        liquidity_provider: LiquidityProvider::Pmm,
```
fleupold, line 569: Sounds like PMM should hold onto the quote instead of requiring it on the order? This should be able to let you get rid of the branch that panics below.

fleupold (commented):
Looks good overall, but I don't think PMM should be subfield on Pool but rather there should be a Liquidity type which can be Pool (or AMM) and PMM.

crates/driver/src/tests/cases/mod.rs
squadgazzz: Right, forgot to remove this comment.

crates/driver/src/tests/cases/mod.rs
squadgazzz: I was experimenting with test values and made mistakes a couple of times with extra or missing zeros. Really liked the approach from e2e tests and tried to use it here. It can be reverted if it looks redundant at the moment. The float representation is used in constants only.

crates/driver/src/tests/setup/blockchain.rs
squadgazzz: Uniswap is still being used which requires reserves in both cases to get the env properly configured.

crates/driver/src/tests/setup/mod.rs
squadgazzz: Initially, just wanted to add flexibility to use a quote for each order. Currently, don't see if it is required anymore. Moved to the PMM.

crates/driver/src/tests/setup/driver.rs
squadgazzz: Enabled for all variants.

squadgazzz: upd: reverted the change for now since it affects other tests.

crates/driver/src/tests/cases/mod.rs
fleupold: ok, could be its own PR then 😄

crates/driver/src/tests/cases/protocol_fees.rs
fleupold: nit: In fact you can probably create a macro or TestCase struct which contains all fields (order type, limit, quote, executed amounts and expectations).

crates/driver/src/tests/setup/driver.rs
fleupold: So how will this work once we have fees for market orders as well?

sunce86: Market order are planned to be deprecated by the 15th of March, as planned by BD. We also won't test the fee for in-market limit orders by then.

crates/driver/src/tests/setup/blockchain.rs
squadgazzz: Because on-chain interactions still use uniswap contract: https://github.com/cowprotocol/services/blob/2bf814045e45d723597bf3648d4a43a2fc4e075b/crates/driver/src/tests/setup/blockchain.rs#L642-L653

And to simulate this interaction, reserves are required:
https://github.com/cowprotocol/services/blob/2bf814045e45d723597bf3648d4a43a2fc4e075b/crates/driver/src/tests/setup/blockchain.rs#L424-L434

That was the simplest way to put it into practice. Right now, we're able to use any limit price and quote, but this means we have to adjust the reserve values to ensure the simulation works out. If the reserves ratio isn't in the right proportion to the quote amounts, it causes the simulation to fail.

crates/driver/src/tests/cases/mod.rs
squadgazzz: Reverted the change

crates/driver/src/tests/cases/protocol_fees.rs
squadgazzz: Added explicit values and extracted all the common parts into a single function.

crates/driver/src/tests/setup/blockchain.rs
fleupold: I see, but isn't the concept of PMM misleading in that case (it has nothing to do with a private market maker providing a quote).

Instead, could we just have a method to create a pool that satisfies a certain quote which is called in the main test case and leave the existing struct untouched?

squadgazzz: Yeah, now it makes sense to remove it. Had quite a lot of experiments, and post refactoring is still required. Replaced the `LiquidityProvider` struct with the quote.

crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -0,0 +1,143 @@
...
+        .sell_amount(50000000000000000000u128.into())
+        .buy_amount(40000000000000000000u128.into());
+    let pool = ab_adjusted_pool(quote);
+    let order = ab_order()
+        .kind(order::Kind::Limit)
+        .sell_amount(50000000000000000000u128.into())
+        .side(side)
+        .solver_fee(Some(10000000000000000000u128.into()))
+        .fee_policy(fee_policy)
+        .executed(40000000000000000000u128.into());
```
fleupold, line 33: Can we test partial fills as well?

crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -0,0 +1,143 @@
...
+    expected_order: ExpectedOrder,
+) {
+    let test_name = format!("Protocol Fee: {side:?} {fee_policy:?}");
+    let quote = ab_liquidity_quote()
+        .sell_amount(50000000000000000000u128.into())
+        .buy_amount(40000000000000000000u128.into());
+    let pool = ab_adjusted_pool(quote);
+    let order = ab_order()
+        .kind(order::Kind::Limit)
+        .sell_amount(50000000000000000000u128.into())
```
fleupold, line 29: What's the limit buy amount of the order? I'd slightly prefer if these values were defined in each test case, we don't care too much about receptive code in tests, the most important thing is that a future maintainer can quickly understand the assumptions this test is making when it breaks.

crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -0,0 +1,143 @@
...
+async fn surplus_protocol_fee_buy_order_not_capped() {
+    let side = order::Side::Buy;
+    let fee_policy = FeePolicy::Surplus {
+        factor: 0.5,
+        // high enough so we don't get capped by volume fee
+        max_volume_factor: 1.0,
+    };
+    let expected = ExpectedOrder {
+        name: ab_order().name,
+        executed_sell_amount: 100000000000000000000u128.into(),
```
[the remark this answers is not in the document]
squadgazzz: Initially, the quote was part of the order. Now, it is part of the Pool since the quote affects the reserve amounts. Do we really want the quote to be part of the order? If so, different orders could potentially have different quotes while working with a single pool adjusted with a random quote?

Expected sell/buy amounts are now part of the order.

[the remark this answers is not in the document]
fleupold: They don't have to be part of the order, you can create a new struct `TestCase` in which you define all the variables you need for your test (limit amounts, quote amounts, executed amounts, order type etc). This way you co-locate your assumptions into a single method and don't require a poor developer that wants to understand the test 3 months from now to jump into multiple methods to get the values they need to understand the test case.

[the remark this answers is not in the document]
squadgazzz: Ah, apologies, I didn't get the initial idea correctly. Created a TestCase struct.

crates/driver/src/tests/cases/protocol_fees.rs
squadgazzz: That requires some extra refactoring. Wdyt about implementing it in a separate PR since the current one is already too big?

crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -0,0 +1,195 @@
...
+        .done()
+        .await;
+
+    test.solve().await.ok().orders(&[order]);
+}
+
+#[tokio::test]
+#[ignore]
+async fn surplus_protocol_fee_buy_order_not_capped() {
+    let side = order::Side::Buy;
```
fleupold, line 64: nanonit: inline for more readability

fleupold (commented):
I'm still missing why the quote is needed on the Pool if we already create the pool with adjusted reserves.

crates/driver/src/tests/setup/blockchain.rs
```diff
@@ -56,6 +56,9 @@ pub struct Interaction {
 pub struct Pool {
     pub reserve_a: Asset,
     pub reserve_b: Asset,
+    /// Specifies a predefined quote. Original AMM formula is used in case the
+    /// quote is empty.
+    pub liquidity_quote: Option<LiquidityQuote>,
```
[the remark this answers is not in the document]
squadgazzz: The initial idea was to get rid of the AMM formula and to explicitly provide a quote here:
https://github.com/cowprotocol/services/blob/9db38e05afa265cb40dcffef265f9679becb4c48/crates/driver/src/tests/setup/blockchain.rs#L79-L85
That frees us from caring about the AMM formula at all. The pool is adjusted "under the hood" just to make the simulation pass with our custom quotes since we continue using the uniswap contract. A single pool can be used for multiple orders. Don't really see other options here.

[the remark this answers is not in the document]
squadgazzz: Another option is to introduce an optional `buy_amount` field in `Order` struct. If the value is specified, then `Pool` uses this value instead of calculating with AMM formula. But in this case, the pool can be used only for a single order. If any other order is added with a different sell/buy ratio, on-chain simulation will fail.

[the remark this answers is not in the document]
fleupold: Why can we not use `order.sell_amount` like we do when no quote is specified? Assuming that the reserves have been adjusted for the quote, when trying to trade `sell_amount` with the pool we should get exactly quote.buy_amount & quote.sell_amount back.

What am I missing?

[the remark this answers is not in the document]
squadgazzz: Like that? https://github.com/cowprotocol/services/pull/2404/commits/c4810f6aa1c255892b1876d8805ca5ed2046c52e
Initially, I just wanted to avoid using the AMM formula when the adjusted pool reserves. Do we now need to specify the order's `buy_amount` explicitly? If yes, how should it be used? The AMM's `out` function result should be compared with the `order.buy_amount`?

fleupold (approved):
Looks good. I think the actual definition of the test case could be made a bit more easy on the eye (by using descriptive nested fields like `quote: {sell: ..., buy:...}` and to `ToWei` helper you initially introduced or even using macros but that's a voluntary exercise.

squadgazzz:
> Looks good. I think the actual definition of the test case could be made a bit more easy on the eye (by using descriptive nested fields like quote: {sell: ..., buy:...} and to ToWei helper you initially introduced or even using macros but that's a voluntary exercise.

I'll create a separate PR for this.

PR #2405, opened 2024-02-13
Clean up solver

MartinquaXD:
# Description
There was an external contribution working on a TODO comment inside the solver crate. This code actually was no longer used anywhere so they effectively wasted their time.
We also regularly get comments asking about the `solver` crate and explaining the current state is always a bit weird.

# Changes
I tried to delete everything from the `solver` crate that was not used in the other crates.

## How to test
CI


sunce86 (approved):
🧹 

I hoped that we keep the solver crate code around for some more time since I sometimes look into how things were implemented before, but the reason you mentioned in the PR description seems like good enough to get rid of it.

squadgazzz (approved):
Awesome

squadgazzz:
> I hoped that we keep the solver crate code around for some more time since I sometimes look into how things were implemented before

Maybe we could create a private archive in a separate repo.

PR #2410, opened 2024-02-14
Track tokens lost in settlement contract

MartinquaXD:
# Description
Currently we don't really know if the interactions suggested in a quote were able to pay out the advertised amount because the interactions actually produced the required tokens or if the settlement contract happened to have buffers big enough to pay out any difference that might arise.

# Changes
Track sell and buy token balance of the settlement contract before and after the settlement.
If the settlement contract has less fewer tokens than before we at least partly paid the quote our of our own pocket.
This PR only adds tracking and logging of these balances but does not yet act on the difference.

Theoretically a solver could try to sell any token the settlement contract has in its buffers (not just the sell_token) so the check is not bullet proof but it seems to be a fair assumption for now that they will only use the sell token balance.

## How to test
CI should continue to work and we should see the new logs.


squadgazzz (approved):

sunce86 (commented):
How would we act on the difference?

I am worried that we would exclude solutions that give a slightly different amount compared to advertised values, because, for example, we simulated the quote on a different block.

MartinquaXD:
My plan is to analyze quotes for a while to get an idea how often this is actually an issue. If there many quotes that have slightly wrong values I would probably add a configurable difference tolerance before a quote should be discarded.

sunce86 (approved):

Issue #2086, opened 2023-11-27
feat: Wrap eth::U256 into a type


sunce86:
### Problem
Default serde implementation for `U256` type encodes/decodes in hexadecimal format, while we always want to use decimal format. We need to make sure that every time we use `U256`, we add an annotation `#[serde_as(as = "serialize::U256")]` to the field.

This has been forgotten numerous times, cf #2075, #1958, etc...  

### Suggested solution
Wrap U256 in its own type that implements default serialization using `HexOrDecimalU256`. 

Additionally, investigate if there is a way to forbid using raw `U256` in our system (with [clippy](https://rust-lang.github.io/rust-clippy/master/#/disallowed_type) or something similar) so we don't accidentally continue to `U256` instead of our own type.

duguorong009:
@sunce86 
Can I pick the issue?

duguorong009:
@sunce86 @fleupold 
Sorry, but I think I can't resolve the issue.

I've tried several times, in order to implement the suggested solution.
But, unfortunately, I didn't manage to finish it.
The reason is that the codebase is HUGE(at least for myself), and 
the introduction of new wrapper type brings too many updates which I can't handle.

From my side, I think it is better way that the team should be more diligent when reviewing the PR which includes `U256` type.
Or, hopefully, another guy can resolve this issue as you expect.

MartinquaXD:
Thanks for trying, though.
I think the biggest problem is probably that in the orderbook and autopilot we use the same types that define the API and the deserialization behavior in the business logic. This makes any change to the API type also a change to the business logic which is as you pointed out quite a lot.
We didn't make this mistake in the `driver` and `solvers` crates so there it should be a reasonably small change.
In the medium term future we should probably apply the same pattern to the orderbook and autopilot so we can easily change the deserialization without affecting the whole code base.

cakevm:
Hey, I just like to mention that there has been some developments in the Rust ecosystem for EVM-based chains with [alloy-rs](https://github.com/alloy-rs/alloy) as the new kid on the block. I also saw that [ethcontract-rs](https://github.com/cowprotocol/ethcontract-rs) hasn’t received many updates over the past year. 

Have you maybe considered switching over to the Alloy type system to maybe improve this issue mentioned in this issue?

Potential benefits:
- No need for custom serialization to/from hex string for byte types (e.g., U256, FixedBytes<56>, ...)
- Less code and no custom `serde_as` implementations required
[...]

PR #2416, opened 2024-02-20
Draft: Create a wrap for eth::U256 in order to ensure correct serde #2086

Mateo-mro:
# Description
Remove the `eth::U256` type and create a new type `U256` in `crates/number`. The idea is to enforce specific (de)serialization for for the `U256` type as per https://github.com/cowprotocol/services/issues/2086.

# Changes
The new type as been converted into `primitive_types::U256` when necessary when needed for the business logic. Ideal solution would be to propagate the new type not only as an `API` type, but also as an internal type for the business logic operations (but it requires many more changes, it could be potentially done in a followup issue).

## How to test

 #2086


crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -223,8 +224,8 @@ pub struct Solutions {
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Solution {
     id: u64,
-    #[serde_as(as = "HashMap<_, serialize::U256>")]
-    prices: HashMap<eth::H160, eth::U256>,
+    #[serde_as(as = "HashMap<_, _>")]
```
Mateo-mro, line 227: I am no so sure we need this here.

crates/driver/src/domain/eth/gas.rs
```diff
@@ -2,6 +2,7 @@ use {
     super::{Ether, U256},
     bigdecimal::Zero,
     num::zero,
+    number::U256 as NumberU256,
```
Mateo-mro, line 5: Not idea to have the two types `primite_type::U256` and `number::U256`. ideally everything would be moved to `number::U256`.

Mateo-mro:

crates/driver/src/infra/solver/dto/solution.rs
MartinquaXD: Since key and value now both use their regular deserialize implmentations this can be dropped AFAIK.

crates/driver/src/util/serialize/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! Serialization utilities for use with [`serde_with::serde_as`] macros.
 
 mod hex;
-mod u256;
+// mod u256;
 
-pub use {self::hex::Hex, u256::U256};
+pub use {self::hex::Hex /* , u256::U256 */};
```
MartinquaXD, line 6: Left over comments.

crates/number/src/serialization.rs
```diff
@@ -66,6 +73,112 @@ where
     deserializer.deserialize_str(Visitor {})
 }
 
+/// Serialize and deserialize [`U256`] as a decimal string.
```
MartinquaXD, line 76: I think the plan was to always serialize to a decimal string but still be able to deserialize from a decimal or hex string for backwards compatibility reasons.

MartinquaXD (commented):
Maybe I misunderstood the original issue but I thought we only want to have a new type that handles the (de)serialization without having to remember to use macros.

If we want to completely replace `primitve_types::U256` with our new type I think we should look into implementing `std::ops::Deref` for the new type. That way we probably wouldn't have to reimplement a bunch of functions for the new type that we are currently using on the existing type.

Alternatively if we want to actually completely replace `primitive_types::U256` with the new type it might also make sense to add a new crate that re-exports all the primtive_types but replaces the regular `U256` with the new `U256` type. But if we already go that far could also consider forking the `primitive_types` crate or even patching the behavior upstream.

crates/driver/src/boundary/mempool.rs
```diff
@@ -51,7 +52,7 @@ pub enum Kind {
     /// The MEVBlocker private mempool.
     MEVBlocker {
         url: reqwest::Url,
-        max_additional_tip: eth::U256,
+        max_additional_tip: primitive_types::U256,
```
Mateo-mro, line 55: I did this to keep the same behavior.

squadgazzz:
Would it be possible to split this activity into separate PRs moving from crate to crate? Maybe, to also add tracking tasks to the original issue.

crates/number/src/serialization.rs
```diff
@@ -1,69 +1,165 @@
...
+    }
+}
+
+impl From<u64> for U256 {
+    fn from(value: u64) -> Self {
+        Self(value.into())
+    }
+}
+
+impl From<i32> for U256 {
```
MartinquaXD, line 121: Converting an signed type to an unsigned one without a way to notice that we transformed a negative number seems dangerous. I think if there is ever the need to convert from a singed type it's okay to make the developer jump through some hoops to proof to the compiler that this is safe.
Alternatively you could implement `TryFrom` instead.

crates/number/src/serialization.rs
```diff
@@ -1,69 +1,165 @@
...
+    }
+}
+
+impl From<u16> for U256 {
+    fn from(value: u16) -> Self {
+        Self(value.into())
+    }
+}
+
+impl Hash for U256 {
```
MartinquaXD, line 145: I think this and some other traits could be derived.

crates/autopilot/src/driver_model.rs
```diff
@@ -0,0 +1,171 @@
+//! Types for communicating with drivers as defined in
```
MartinquaXD, line 1: Why do we need these types here? The autopilot already knows how to talk to a driver. :thinking:

MartinquaXD (commented):
The `driver` and `solvers` crate export a `U256` type as part of the `eth` module (e.g. [here](https://github.com/cowprotocol/services/blob/main/crates/driver/src/domain/eth/mod.rs#L16)). Might be a good idea to export the new `number::U256` instead of the current `primitive_types::U256` instead to keep people from using the old type.

Overall I'm still not sure what the best solution is for this. :/
Maybe @fleupold, @sunce86 and @squadgazzz have an opinion as well?

crates/autopilot/src/driver_model.rs
Mateo-mro: How does it enter here? I must have overlooked it with so many changes, sorry.

Mateo-mro: I think it was because of my latest rebase.

crates/number/src/serialization.rs
Mateo-mro: I totally agree with you. That implementation is the one coming from the `primitive_types`. The `From<i32>` was only used for testing purposes. I changed the tests to use `unsigned`. Thanks!

squadgazzz:
> Might be a good idea to export the new number::U256 instead of the current primitive_types::U256 instead to keep people from using the old type.

That looks convenient.

Mateo-mro:
> The `driver` and `solvers` crate export a `U256` type as part of the `eth` module (e.g. [here](https://github.com/cowprotocol/services/blob/main/crates/driver/src/domain/eth/mod.rs#L16)). Might be a good idea to export the new `number::U256` instead of the current `primitive_types::U256` instead to keep people from using the old type.
> 
> Overall I'm still not sure what the best solution is for this. :/ Maybe @fleupold, @sunce86 and @squadgazzz have an opinion as well?

It is a good idea, but I think the best is to do it in a followup issue, not to have a lot of more changes in this one, wdyt?

MartinquaXD (approved):
Looks alright to me. Let's give it a try.

Mateo-mro:
> Unfortunately not. There is an open [issue](https://github.com/rust-lang/rust-clippy/issues/8581) for warning on trait implementation on specific types but and the open draft [PR](https://github.com/rust-lang/rust-clippy/pull/12194) doesn't seem like it would get merged any time soon. The options I see are:
> 
> 1. fork primitive types and patch it
> 2. try to upstream that patch. They are aware of the [issue](https://github.com/paritytech/parity-common/issues/643) but refuse to change the default behavior. I asked if they would be fine with a feature flag that switches to the saner logic and they are open to this option. So depending on how tricky that patch is (probably requires a patch to a macro) this might be the best way out.
> 3. completely replace `U256` type but I think it will not be possible to make it work seamlessly with the regular `primitive_types` during computations.
> 4. push the clippy PR over the finish line

I think 1 may be complicated to maintain. We would need to rebase the fork every time we want to bump to the latest version of `primitive_types`.

The option 3 is not complicated, it should be a smaller PR than this one I believe. I could take care of that as a follow up issue (I'll create the issue once confirmed).

sunce86:
> The `driver` and `solvers` crate export a `U256` type as part of the `eth` module (e.g. [here](https://github.com/cowprotocol/services/blob/main/crates/driver/src/domain/eth/mod.rs#L16)). Might be a good idea to export the new `number::U256` instead of the current `primitive_types::U256` instead to keep people from using the old type.
> 
> Overall I'm still not sure what the best solution is for this. :/ Maybe @fleupold, @sunce86 and @squadgazzz have an opinion as well?

Long term, I would prefer to forbid using `primitive_types::U256` in the `domain`, and instead have a wrapper around it for every object, because every U256 represents something meaningful to the `domain`. Not sure if it should be number::U256, because we don't need serde in `domain`.

crates/e2e/tests/e2e/ethflow.rs
```diff
@@ -135,7 +135,7 @@ async fn eth_flow_tx_zero_fee(web3: Web3) {
             .get_order(&ethflow_order.uid(onchain.contracts()).await)
             .await
             .unwrap();
-        order.metadata.executed_surplus_fee > U256::zero()
+        order.metadata.executed_surplus_fee > U256::zero().into()
```
sunce86, line 138:
```suggestion
        order.metadata.executed_surplus_fee > number::U256::zero()
```

crates/e2e/tests/e2e/limit_orders.rs
```diff
@@ -376,10 +376,10 @@ async fn mixed_limit_and_market_orders_test(web3: Web3) {
 
     let order_b = OrderCreation {
         sell_token: token_b.address(),
-        sell_amount: to_wei(5),
-        fee_amount: to_wei(1),
+        sell_amount: to_wei(5).into(),
```
sunce86, line 379: nit: what happens if we change `to_wei()` to return `number::U256`?

crates/number/src/nonzero.rs
```diff
@@ -30,6 +30,14 @@ impl TryFrom<ZeroU256> for U256 {
     }
 }
 
+impl TryFrom<crate::U256> for U256 {
+    type Error = anyhow::Error;
+
+    fn try_from(value: crate::U256) -> Result<Self, Self::Error> {
+        ZeroU256::from(value).try_into()
+    }
+}
```
sunce86, line 39: This should be removed as not used.

crates/number/src/serialization.rs
```diff
@@ -1,69 +1,134 @@
...
 
-pub fn serialize<S>(value: &U256, serializer: S) -> Result<S::Ok, S::Error>
-where
-    S: Serializer,
-{
-    serializer.serialize_str(&value.to_string())
+impl U256 {
+    pub fn zero() -> Self {
+        Self(primitive_types::U256::zero())
+    }
```
sunce86, line 83: nit: Instead, how about implementing an existing trait for this:

```
impl num::Zero for U256 {
    fn zero() -> Self {
        Self(primitive_types::U256::zero())
    }

    fn is_zero(&self) -> bool {
        self.0.is_zero()
    }
}

impl std::ops::Add for U256 {
    type Output = Self;

    fn add(self, rhs: Self) -> Self::Output {
        Self(self.0 + rhs.0)
    }
}
```

sunce86 (approved):
Mostly nits.

PR #2419, opened 2024-02-21
Rank by surplus switch with config

sunce86:
# Description
Temporary config to deterministically switch all drivers to rank by surplus at the same time.

# Changes
Added configuration parameter to driver.

## How to test
Existing tests passing. 

Also tested locally e2e test by adding parameter `cip38-activation = "2022-02-21T12:34:56Z"` to driver configuration to test the cip38 activation and proper score calculation.

### DIFF
--- a/crates/driver/src/infra/config/file/mod.rs
+++ b/crates/driver/src/infra/config/file/mod.rs
@@ -180,6 +180,10 @@ struct SolverConfig {
     /// Timeout configuration for the solver.
     #[serde(default, flatten)]
     timeouts: Timeouts,
+
+    /// Datetime when the CIP38 should be activated.
+    #[serde(default)]
+    cip38_activation: Option<chrono::DateTime<chrono::Utc>>,
 }
 
 #[serde_as]
--- a/crates/driver/src/domain/competition/mod.rs
+++ b/crates/driver/src/domain/competition/mod.rs
@@ -173,10 +173,13 @@ impl Competition {
             .map(|settlement| {
                 observe::scoring(&settlement);
                 (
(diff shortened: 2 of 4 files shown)


squadgazzz (approved):

fleupold (approved):

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -180,6 +180,10 @@ struct SolverConfig {
     /// Timeout configuration for the solver.
     #[serde(default, flatten)]
     timeouts: Timeouts,
+
+    /// Datetime when the CIP38 should be activated.
+    #[serde(default)]
+    cip38_activation: Option<chrono::DateTime<chrono::Utc>>,
```
MartinquaXD, line 186: Is this value only optional for sth like testing purposes (i.e. it's annoying having to add that value)?
Or do we foresee a scenario where we would have to revert back to the old scoring mechanism without breaking any potential solver accounting infra?

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -180,6 +180,10 @@ struct SolverConfig {
     /// Timeout configuration for the solver.
     #[serde(default, flatten)]
     timeouts: Timeouts,
+
+    /// Datetime when the CIP38 should be activated.
+    #[serde(default)]
```
MartinquaXD, line 185: Can be omitted when dealing with `Option`s
```suggestion
```

MartinquaXD (approved):
Nothing wrong with the PR; just curious how we envision the roll out based on the code.

sunce86:
Closed in favor https://github.com/cowprotocol/services/pull/2483

PR #2422, opened 2024-02-22
Inline function `with_protocol_fees`

sunce86:
# Description
Inlines `with_protocol_fees` to prevent unintentionally calling the function multiple times and changing the state of the struct. Protocol fees are meant to be applied only once and be done with it, so doing it in constructor feels natural.

## How to test
Existing tests.


fleupold (approved):
👏

squadgazzz (approved):

PR #2428, opened 2024-02-23
fix: dockerfile missing slash

mfw78:
# Description

Missing slash in the deploy docker.

# Changes

- [x] Fixes missing slash

## How to test
1. Check the CI/CD builds the deploy correctly.


MartinquaXD (approved):
Actually surprised that this seems to be needed but let's try to unblock autodeploy.

PR #2439, opened 2024-02-27
Add estimator name to trace

MartinquaXD:
# Description
Currently it's a bit finicky to figure out which logs where emitted from which estimator in the orderbook/autopilot.
This PR adds the estimator name to all logs emitted by an instrumented estimator.

### DIFF
--- a/crates/shared/src/price_estimation/instrumented.rs
+++ b/crates/shared/src/price_estimation/instrumented.rs
@@ -3,6 +3,7 @@ use {
     futures::future::FutureExt,
     prometheus::{HistogramVec, IntCounterVec},
     std::{sync::Arc, time::Instant},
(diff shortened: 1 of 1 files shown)


fleupold (approved):

sunce86 (approved):
Which exactly log is this fixing? I know I saw a lot of price estimation logs all containing the estimator name...

MartinquaXD:
You are probably referring to the logs in the competition estimator that prints the estimator name with each incoming estimate. The info span I added also gives us those names for logs that happen while we are computing that estimate. This is currently just useful for quickly finding out which estimator a trade verification log refers to.
It's already possible to know this but you would have to look up the public address of the solver that gets printed with it which is not great.

PR #2441, opened 2024-02-27
Allow providing custom appData when replacing order #2365

### LINKED ISSUE #2365: chore: Allow providing custom appdata when cancelling an order
### Background
Support for cancelling an open order and submitting a new one (effectively replacing the old one) was introduced in https://github.com/cowprotocol/services/pull/239.
The idea is that the new order has to have an appdata value that is equal to the hash of a cancellation request for the old order uid.
Because we have to be able to recover the owner from the signature of the new order and the appdata matches the cancellation of the old order uid we prove that it was the actual owner that intended to cancel the old order and not some random person messing with us.

Since we introduced that feature we also extended possible use cases for appdata so it's reasonable to expect that someone could also replace an old order with a new order that also contains some custom appdata you care about.
But because we currently expect some specific value for the cancellation appdata this is not possible.

### Acceptance criteria
A mechanism that allows you to cancel an old order while also providing additional appdata for the new order.
This probably requires you to provide the appdata content (i.e. the unhashed JSON string) which contains a new field with the order uid to cancel in addition to all the appdata values we expect by now.

This endpoint is used very rarely so we might get away with a breaking change and not introducing a new endpoint for this.

Mateo-mro:
# Description
Allow users to cancel an old order while also providing additional appData for the new order.
There is a new field in the metadata part called `replacedOrder` which specifies the order uid of the order to replace

# Changes
- remove the replace order endpoint
- modify the create order endpoint in order to allow providing an appData which contains the replacedOrder UID field in the metadata

These are breaking changes to the API, but since the FE didn't integrate it yet, it should be ok.

## How to test
1. Unit tests
2. e2e tests

<!--
## Related Issues

Fixes #2365


crates/orderbook/src/orderbook.rs
```diff
@@ -340,11 +374,7 @@ impl Orderbook {
         // Verify that the new order is a valid replacement order by checking
         // that the `app_data` encodes an order cancellation and that both the
         // old and new orders have the same signer.
-        let cancellation = OrderCancellation {
-            order_uid: old_order.metadata.uid,
-            ..Default::default()
-        };
-        if new_order.data.app_data != cancellation.hash_struct()
+        if app_data.metadata.original_order != old_order.metadata.uid
```
MartinquaXD, line 377: According to grafana this endpoint was never called successfully in the last 90 days. Since nobody is using it I think we might as well break the API completely and remove the `order_uid` from the path.
Since we now require the uid to be present in the appdata there is no need to also have it in the path and that also removes 1 extra error case.

crates/orderbook/src/orderbook.rs
```diff
@@ -317,6 +324,16 @@ impl Orderbook {
         old_order: OrderUid,
         new_order: OrderCreation,
     ) -> Result<OrderUid, ReplaceOrderError> {
+        #[derive(Deserialize)]
```
MartinquaXD, line 327: This function is getting a bit big. Maybe we could move the verification that it's actually okay to replace the order into a separate function.

crates/orderbook/src/api/replace_order.rs
```diff
@@ -42,6 +42,18 @@ impl IntoWarpReply for ReplaceOrderError {
                 super::error("InvalidReplacement", err.to_string()),
                 StatusCode::UNAUTHORIZED,
             ),
+            err @ ReplaceOrderError::InvalidAppDataHash => reply::with_status(
```
MartinquaXD, line 45: New error cases and changed API need to be document in `orderbook/openapi.yml`.

MartinquaXD (commented):
Would be nice to have an e2e test as well.
I think a nice test could be to:
1. create order selling permittable token (while not setting any approvals)
2. replace that order with an order selling the same token but with a permit pre-interaction
3. assert that order 2 was settled

If the feature doesn't work we will not be able to settle the second order in the test.
Feel free to reach out for more context.

crates/e2e/tests/e2e/limit_orders.rs
```diff
@@ -695,3 +701,119 @@ async fn forked_gnosis_single_limit_order_test(web3: Web3) {
...
+        ..Default::default()
+    }
+    .sign(
+        EcdsaSigningScheme::Eip712,
+        &onchain.contracts().domain_separator,
+        SecretKeyRef::from(&SecretKey::from_slice(trader_a.private_key()).unwrap()),
+    );
+    let order_id = services.create_order(&order).await.unwrap();
+    let limit_order = services.get_order(&order_id).await.unwrap();
+    assert_eq!(limit_order.metadata.class, OrderClass::Limit);
```
MartinquaXD, line 781: This is not needed to test replacements.

crates/e2e/tests/e2e/limit_orders.rs
```diff
@@ -36,6 +36,12 @@ async fn local_node_mixed_limit_and_market_orders() {
     run_test(mixed_limit_and_market_orders_test).await;
 }
 
+#[tokio::test]
+#[ignore]
+async fn local_node_replace_order() {
```
MartinquaXD, line 41: I think this test can live in its own file since the other tests are more concerned with verifying various settlement behaviors.

crates/e2e/tests/e2e/limit_orders.rs
```diff
@@ -695,3 +701,119 @@ async fn forked_gnosis_single_limit_order_test(web3: Web3) {
...
+            ),
+        },
+        ..Default::default()
+    }
+    .sign(
+        EcdsaSigningScheme::Eip712,
+        &onchain.contracts().domain_separator,
+        SecretKeyRef::from(&SecretKey::from_slice(trader_a.private_key()).unwrap()),
+    );
+    services.replace_order(&new_order).await.unwrap();
```
MartinquaXD, line 804: Let's assert that the status of the other order is now cancelled.

crates/e2e/tests/e2e/limit_orders.rs
```diff
@@ -695,3 +701,119 @@ async fn forked_gnosis_single_limit_order_test(web3: Web3) {
...
+    wait_for_condition(TIMEOUT, || async { services.solvable_orders().await == 1 })
+        .await
+        .unwrap();
+
+    wait_for_condition(TIMEOUT, || async { services.solvable_orders().await == 0 })
+        .await
+        .unwrap();
+
+    let balance_after = token_b.balance_of(trader_a.address()).call().await.unwrap();
+    assert!(balance_after.checked_sub(balance_before).unwrap() >= to_wei(8));
```
MartinquaXD, line 818: I believe this should probably be `balance >= 3`, right?
Since the first order is probably not supposed to be executed.

crates/e2e/tests/e2e/replace_order.rs
```diff
@@ -0,0 +1,135 @@
...
+#[ignore]
+async fn local_node_replace_order() {
+    run_test(single_replace_order_test).await;
+}
+
+async fn single_replace_order_test(web3: Web3) {
+    let mut onchain = OnchainComponents::deploy(web3.clone()).await;
+
+    let [solver] = onchain.make_solvers(to_wei(1)).await;
+    let [trader_a] = onchain.make_accounts(to_wei(1)).await;
```
squadgazzz, line 23:
```suggestion
    let [trader] = onchain.make_accounts(to_wei(1)).await;
```

crates/orderbook/src/orderbook.rs
```diff
@@ -312,9 +319,57 @@ impl Orderbook {
...
+
+        let app_data: AppData = serde_json::from_str(full_app_data.as_str())
+            .map_err(ReplaceOrderError::InvalidAppData)?;
+
+        self.find_order_for_cancellation(&app_data.metadata.replaced_order)
+            .await
+            .map_err(Into::into)
+    }
+
+    async fn replace_order_verification(
```
squadgazzz, line 357: Why this function is async?

crates/orderbook/src/orderbook.rs
```diff
@@ -312,9 +319,57 @@ impl Orderbook {
...
+        &self,
+        validated_new_order: &Order,
+        validated_old_order: &Order,
+    ) -> Result<(), ReplaceOrderError> {
+        // Verify that the new order is a valid replacement order by checking
+        // that the `app_data` encodes an order cancellation and that both the
```
squadgazzz, line 363: > that the `app_data` encodes an order cancellation

Is that the case for this function? Currently, it checks owners only.

crates/orderbook/src/orderbook.rs
```diff
@@ -312,9 +319,57 @@ impl Orderbook {
...
+        // old and new orders have the same signer.
+        if validated_new_order.metadata.owner != validated_old_order.metadata.owner {
```
MartinquaXD, line 365: nit: I think inlining this part would be fine. Slightly less mental overhead jumping around and also having this part in its own function adds a lot of code (relatively) just for the signature.

crates/e2e/tests/e2e/replace_order.rs
```diff
@@ -0,0 +1,135 @@
...
+    wait_for_condition(TIMEOUT, || async { services.solvable_orders().await == 1 })
+        .await
+        .unwrap();
+
+    wait_for_condition(TIMEOUT, || async { services.solvable_orders().await == 0 })
+        .await
+        .unwrap();
+
+    let balance_after = token_b.balance_of(trader_a.address()).call().await.unwrap();
+    assert!(balance_after.checked_sub(balance_before).unwrap() >= to_wei(3));
```
MartinquaXD, line 134: One more comment on this: Initially I thought the `>= 8` should lead to an error but I guess the liquidity pools are configured such that you get loads of surplus. That means using the buy token to verify that only the second trade happened doesn't really work.
What you can do instead is to verify that the new sell token balance is exactly `to_wei(3)` less than at the start. For this to be bullet proof you need to create these orders as fill-or-kill (i.e. `partiallyFillable: false`).

MartinquaXD (commented):
Looks good to me. Just 2 nits remaining. 👍

crates/orderbook/src/orderbook.rs
Mateo-mro: It is the case, it checks the owner but the replaced order is found by using the `app_data` order it. I rephrased it.

crates/e2e/tests/e2e/replace_order.rs
MartinquaXD: The assertion to test that only 1 of the orders got executed still uses the buy token balance instead of the sell token balance.
Using the sell token balance for sell orders is easier to grasp because we know for a fact that exactly this amount should be used. Whereas if we assert something about the buy token balance the exact amounts depend on how much surplus the solution generated.

crates/orderbook/src/orderbook.rs
```diff
@@ -483,62 +513,85 @@ mod tests {
...
-                        from: Some(H160([2; 20])),
-                        signature: Signature::Eip712(Default::default()),
-                        app_data: AppDataHash(cancellation.hash_struct()).into(),
-                        ..Default::default()
+                .replace_order(OrderCreation {
+                    from: Some(H160([2; 20])),
+                    signature: Signature::Eip712(Default::default()),
+                    app_data: OrderCreationAppData::Full {
+                        full: format!(
+                            r#"{{"version":"0.9.0","metadata":{{"replacedOrder":"{}"}}}}"#,
```
alfetopito, line 533: Not really a must as the backend doesn't care about the version AFAICT, but just for the correctness sake, use version `1.1.0` where `repleaceOrder` is introduced.

crates/e2e/tests/e2e/replace_order.rs
Mateo-mro: Oh, I misunderstood you. Clear now! thank you!

MartinquaXD (approved):

MartinquaXD:
BTW to auto-close an issue when your PR gets merged the description has to be `Fixes #issue` and not just `#issue`.

squadgazzz (approved):

sunce86 (approved):

crates/orderbook/src/orderbook.rs
```diff
@@ -312,9 +319,43 @@ impl Orderbook {
...
+        new_order: &OrderCreation,
+    ) -> Result<Order, ReplaceOrderError> {
+        #[derive(Deserialize)]
+        struct AppData {
+            metadata: Metadata,
+        }
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct Metadata {
+            replaced_order: OrderUid,
```
alfetopito, line 333: Actually, I have a concern regarding using a flat value for this, which affects this PR too

https://github.com/cowprotocol/app-data/pull/60#discussion_r1505764953

crates/orderbook/src/orderbook.rs
```diff
@@ -312,9 +319,43 @@ impl Orderbook {
         Ok(())
     }
 
+    async fn get_replaced_order(
+        &self,
+        new_order: &OrderCreation,
+    ) -> Result<Order, ReplaceOrderError> {
+        #[derive(Deserialize)]
+        struct AppData {
```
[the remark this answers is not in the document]
Mateo-mro: Oh, I didn't see this! nice, using it! thanks!

crates/e2e/src/setup/services.rs
```diff
@@ -344,6 +344,30 @@ impl<'a> Services<'a> {
         }
     }
 
+    /// Replace an [`Order`].
+    /// If the response status code is not `201`, return the status and the
+    /// body.
+    pub async fn replace_order(
+        &self,
+        order: &OrderCreation,
```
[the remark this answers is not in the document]
Mateo-mro: I don't think this will work, the order has to be signed, and I alter it within the call, the signature is not valid anymore.

crates/e2e/tests/e2e/replace_order.rs
```diff
@@ -0,0 +1,251 @@
...
+        ..Default::default()
+    }
+    .sign(
+        EcdsaSigningScheme::Eip712,
+        &onchain.contracts().domain_separator,
+        SecretKeyRef::from(&SecretKey::from_slice(trader_b.private_key()).unwrap()),
+    );
+    let response = services.replace_order(&new_order).await;
+    let (error_code, _) = response.err().unwrap();
+    assert_eq!(error_code, StatusCode::UNAUTHORIZED);
```
MartinquaXD, line 131: Let's assert that we are still able to settle the original order. To prevent the case where we actually modified the DB but returned an error anyway.

crates/shared/src/app_data.rs
```diff
@@ -18,10 +21,17 @@ pub struct ValidatedAppData {
 }
 
 #[derive(Debug, Default, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
 pub struct ProtocolAppData {
     #[serde(default)]
     pub hooks: Hooks,
     pub signer: Option<H160>,
+    pub replaced_order: Option<ReplacedOrder>,
```
MartinquaXD, line 29: Should we maybe throw an error if this value is populated while placing an order with the regular endpoint?
Actually if this we are now always parsing this field from the metadata we could even consider ditching the `PATCH /orders` endpoint altogether and only implement the replacement mechanism on `POST /orders` with that special appdata. That would get rid of the ambiguity.

MartinquaXD (approved):
Only nits or discussion points that can also be discussed at a later point in time (since nobody is using this endpoint, yet).
Feel free to ignore.

alfetopito:
> Only nits or discussion points that can also be discussed at a later point in time (since nobody is using this endpoint, yet). Feel free to ignore.

We'll start using it soon! Better hurry if you want to break things :)

crates/orderbook/openapi.yml
```diff
@@ -133,50 +136,6 @@ paths:
...
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/UID"
-        400:
-          description: Error cancelling and replacing new order with an old one.
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/ReplaceOrderError"
```
MartinquaXD, line 170: `ReplaceOrderError` is no longer in use so that sub schema can also be deleted.

MartinquaXD (approved):
Logic looks good. The openapi thing should be cleaned up before merging but the appdata validation can happen in a different PR.

crates/orderbook/src/orderbook.rs
```diff
@@ -480,68 +514,78 @@ mod tests {
...
-            new_order_uid,
-        );
+        let (order_id, _) = orderbook
+            .add_order(OrderCreation {
+                from: Some(old_order.metadata.owner),
+                signature: Signature::Eip712(Default::default()),
+                app_data: OrderCreationAppData::Full {
+                    full: format!(
+                        r#"{{"version":"1.1.0","metadata":{{"replacedOrder":{{"uid":"{}"
+        }}}}}}"#,
```
alfetopito, line 581: Nitpick: that's a weird line split. Is this rust fmt's fault?

crates/orderbook/openapi.yml
```diff
@@ -25,6 +25,9 @@ paths:
   /api/v1/orders:
     post:
       summary: Create a new order.
+        If the appData contains a valid replacement order UID, then the indicated order is cancelled. 
+        This allows an old order to be cancelled AND a new order to be created in an atomic operation with a single signature.
+        This may be useful for replacing orders when on-chain prices move outside of the original order's limit price.
```
alfetopito, line 30: Can this be somehow made clearer as to what and how needs to be provided?
It mentions the replacement order UID in the appData, but doesn't say which fields or how it should be presented.

Unless one knows what this is about, there's little guidance as to how actually perform the order replacement.

alfetopito (approved):
👍

squadgazzz (approved):

PR #2445, opened 2024-02-28
Introduce mechanism to discard inaccurate quotes

MartinquaXD:
# Description
A price estimator can claim they can achieve a certain out_amount with their quote but when we actually simulate the calldata we might see that the suggested calldata actually sells more sell token or pays out less buy token than advertised.
If the settlement contract has enough balance in the buffers for these tokens the bad calldata would actually simulate without reverting but just because the settlement contract pays the difference.

# Changes
This PR introduces another CLI argument to configure how much of a difference we can tolerate paying out of our own pocket. Picking a threshold instead of completely forbidding any buffer usage is necessary since many solutions have rounding issues in the neighborhood of a couple of wei which doesn't matter in the grand scheme of things.

Note that this logic can be disabled by configuring a high inaccuracy tolerance in case this causes issues for some reason.

## How to test
Added a unit test for a free function implementing this check since adding an e2e test would be quite a bit of work. (Probably requires a token that uses fee on transfer logic)


fleupold (approved):
Code looks good, although I'm still not fond of that logic as it means we won't be able to use buffers when quoting.
If we are worried about solvers accidentally using buffer trades, I think we should think about a way for solvers to indicate that they are intentionally using buffers.

MartinquaXD:
> not fond of that logic as it means we won't be able to use buffers when quoting

Correct, but this was actually agreed upon and communicated to solvers [here](https://github.com/cowprotocol/services/blob/main/crates/driver/src/domain/quote.rs#L143-L151) with `trusted: false` for the buy and sell token (which are the only ones this logic checks).

> If we are worried about solvers accidentally using buffer trades, I think we should think about a way for solvers to indicate that they are intentionally using buffers.

Absolutely, that is my longer term plan.

squadgazzz (approved):

PR #2447, opened 2024-02-29
Legacy API translation inside driver

MartinquaXD:
# Description
Currently we need the `legacy` solvers instance to convert a new API call into the old format and convert the response back. This means at the moment we have to spawn 1 pod per legacy solver endpoint (e.g. `/solve` and `/quote` can require separate instances).

# Changes
Exports the conversion logic from the `solvers` crate and makes use of it in the `driver` to not require a whole new process to do the conversion.

## How to test
TODO: manual test and maybe testing in staging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,6 +1668,7 @@ dependencies = [
  "serde_with",
  "shared",
  "solver",
+ "solvers",
  "tap",
  "tempfile",
  "thiserror",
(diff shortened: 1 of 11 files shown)


fleupold (commented):
Didn't we discuss that we are ok with spawning an extra pod and don't want to spend any ressources on supporting the legacy http interface which is going away in 4 weeks?

MartinquaXD:
Well, 90% of this PR was done by the time you wrote on Slack since this was so little work I figured I might as well finish it in case people change their opinion.

PR #2453, opened 2024-02-29
Don't underestimate gas cost in verification

MartinquaXD:
# Description
There are a few operations that we add to a verified quote which increase the overall gas cost but would not be there during an actual settlement. That's why gas costs for those operations should be discounted from the `settle()` call gas cost.
Since we started detecting internal buffer costs we also added more of those requests which happen before and after the `settle()` call. Since we only track the gas costs of the `settle()` call discounting those gas costs will decrease the estimated gas count incorrectly leaving us with too small gas estimates.

# Changes
Adds a flag to the `store_balance()` helper call that controls which calls should actually be discounted.

## How to test
CI


sunce86 (approved):

MartinquaXD:
Merging with 1 approval since this is an easy fix and is needed to unblock verified quotes.

PR #2460, opened 2024-03-01
Settlement merge flag

nuts-rice:
# Description
Addresses #2418 by a flag in `driver` config file as well as checks for settlement merge from `solver`

- `skip-merge` value  in config file
- member functions for checking config value 
- checks if should merge solutions or skip merge using `Fetch`  or  `Skip`
- flags if solution is `MergeSkippable` or not

## How to test
1. Should be possible to skip merging solution when driver is flagged to Skip
2. Should be impossible to skip merging solution when driver is flagged to Fetch

### DIFF
--- a/crates/driver/example.toml
+++ b/crates/driver/example.toml
@@ -1,9 +1,11 @@
 [[solver]]
-name = "mysolver" # Arbitrary name given to this solver, must be unique
-endpoint = "http://0.0.0.0:7872"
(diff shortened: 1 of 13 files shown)


nuts-rice:

sunce86:
Dependent on https://github.com/cowprotocol/services/pull/2471

I suggest first merging the ☝️  and then continue.

PR #2463, opened 2024-03-04
Metric for token list update

sunce86:
# Description
Oncall alert fix.

# Changes

- [ ] Changes `tracing::error` to `tracing::warn`
- [ ] Instead, adds metric so we can create alert


fleupold (approved):

MartinquaXD (approved):
Unrelated to this PR:
I just noticed this component takes an `Option<Url>` and does nothing if `None` was passed. I think we should probably refactor this eventually to make the url non-optional and simply not instantiate this component if we don't have a url.

squadgazzz (approved):

sunce86:
> Unrelated to this PR: I just noticed this component takes an `Option<Url>` and does nothing if `None` was passed. I think we should probably refactor this eventually to make the url non-optional and simply not instantiate this component if we don't have a url.

Leaving up to you to create issue for this.

PR #2474, opened 2024-03-05
[EASY] Domain separator is domain object

sunce86:
# Description
Part 2 of https://github.com/cowprotocol/services/pull/2436

# Changes

- [ ] Adds domain separator as a domain object in autopilot
- [ ] Removes boundary::DomainSeparator export from boundary


crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -215,7 +215,7 @@ impl Inner {
         let surplus = settlement.total_surplus(&external_prices);
         let (fee, order_executions) = {
             let domain_separator = self.eth.contracts().settlement_domain_separator();
-            let all_fees = settlement.all_fees(&external_prices, domain_separator);
+            let all_fees = settlement.all_fees(&external_prices, &model::DomainSeparator(domain_separator.0));
```
sunce86, line 218: Will be thrown away as seen in https://github.com/cowprotocol/services/pull/2436

crates/autopilot/src/domain/eth/mod.rs
```diff
@@ -0,0 +1,5 @@
+/// Domain separator used for signing.
+///
+/// https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator
+#[derive(Debug, Clone, Copy)]
+pub struct DomainSeparator(pub [u8; 32]);
```
Mateo-mro: nitpick: have you considered making the internal type not public and derive `Into`/`From` (`derive_more` crate)?

Mateo-mro (approved):

crates/autopilot/src/domain/eth/mod.rs
sunce86: Not gaining much with it, I believe. We can probably add it later if it becomes ugly.

PR #2476, opened 2024-03-05
[trivial] Fix quote verification when receiver is 0 address

MartinquaXD:
# Description
When the an order has `receiver: 0x000..000` the smart contract will send the proceeds to the owner of the order (see [here](https://github.com/cowprotocol/contracts/blob/main/src/contracts/libraries/GPv2Order.sol#L102-L107)).
This was currently not respected in the simulation logic when tracking the funds of the `receiver`.

# Changes
Track balances of `owner` when `receiver == 0x0000`.
I specifically did not want to overwrite the owner in the `Verification` struct because that would diverge the execution of the simulation from what would actually happen more.

Example simulation from xdai: https://dashboard.tenderly.co/gp-v2/production/simulator/fe44ac0d-21a6-46ba-b0a8-b7edebe059ad
Noticed via alert: https://cowservices.slack.com/archives/C0371SB243E/p1709656851339479


squadgazzz (approved):
One of my eyes still twitches when such a change doesn't brake any existing test 😅

MartinquaXD:
That's fair. The test coverage of this feature is still very bad.

Issue #2469, opened 2024-03-05
chore: Eliminate the order class in post order


Mateo-mro:
### Background
Since all the orders are limit orders, eliminate the order class from the post order.

Keep the order class in the database as an internal class with: `limit` and `market`. By doing this, we can easily count the number of real `limit` orders in order to limit the number of orders per user.

### Details
It has to be done:
- remove the order class in post order
- classify properly internally if an order is `market` or `limit`
- eliminate `liquidity` class order from the database (migration needed to move them as `limit` order)
- change the way we count of "in-market` order to use the database class

### Acceptance criteria
The changes are backwards compatible in behavior, but the API contract is changed. All the current tests must pass.

PR #2477, opened 2024-03-05
Hotfix: Make OrderWithQuote as inner join in order not to get nullable quotes

Mateo-mro:
# Description
Do an `inner join` instead of `left join` in order not to have nullable quotes in case they do not exist.

This query will be removed with https://github.com/cowprotocol/services/issues/2469 

# Changes
Do an `inner join` instead of `left join` in order not to have nullable quotes in case they do not exist. This way it does not error in case of missing quote (it won't be counted, as it is a liquidity order anyway).

## How to test
1. CI
2. e2e


MartinquaXD (approved):
Looks alright to me. Let's see if the e2e tests agree.

Issue #2485, opened 2024-03-06
chore: Insert appData into the database when replacing an order


Mateo-mro:
### Background
The appData is not stored in the database when a replaced order is executed.

### Details
The appData has to be stored in the database when an order is replaced.

### Acceptance criteria
e2e test: The appData was successfully inserted into the database,

PR #2486, opened 2024-03-06
Insert appData into the database when replacing an order #2485

Mateo-mro:
# Description
The appData has to be stored in the database when an order is replaced.

# Changes
The full appData is stored in the database when an order is replaced.

## How to test
1. e2e
## Related Issues

Fixes 2485


crates/orderbook/src/database/orders.rs
```diff
@@ -305,6 +305,17 @@ impl OrderStoring for Postgres {
                     if let Some(quote) = new_quote {
                         insert_quote(&new_order.metadata.uid, &quote, ex).await?;
                     }
+                    if let Some(full_app_data) = new_order.metadata.full_app_data {
```
alfetopito, line 308: Isn't `fullAppData` mandatory for the replace order?

Mateo-mro: @alfetopito it is indeed, but it will always be `Some(_)` here. In order to change this I would need to create new structs and change the signature of a couple of functions.

MartinquaXD, line 308: This is a 1:1 copy of the same logic inside `insert_order()`, right?
In that case moving that into a separate function would make sense to me.

crates/e2e/tests/e2e/replace_order.rs
```diff
@@ -270,3 +276,124 @@ async fn single_replace_order_test(web3: Web3) {
         to_wei(3)
     );
 }
+
+async fn single_replace_order_test_with_custom_app_data(web3: Web3) {
```
MartinquaXD, line 280: Since this is 99% the same test as `replace_single_order_test` AFAICS and `e2e` tests can already be a bit slow at times I would just adjust `replace_single_order_test` by adding the custom data and asserting that the `fullappdata` of the new order matches your expectation.

fleupold (approved):

MartinquaXD (approved):

squadgazzz (approved):

PR #2515, opened 2024-03-14
External solver support in e2e tests

devanoneth:
# Description
A new function `start_protocol_external_solver` is created which allows e2e tests to easily setup the protocol with an external solver that uses the new solver API. 

# Changes
- Add a `start_protocol_external_solver` function 
- Add a `run_baseline` flag to the `start_protocol` functions that use external solvers, which if set to true will also run a baseline solver alongside the external solver.

## How to test
 - CI will run, but as I'm not a contributor, forked tests can't run until merge. I've run them locally and they pass.
 - Enso CI tests run and pass against this commit hash.


devanoneth:
> LG. However, can't currently approve it since the forked node tests fail locally for some reason(even on the `main` branch).

Thanks! Forked node tests are working locally for me on the latest commit on my branch (just merged main in).

crates/e2e/src/setup/services.rs
```diff
@@ -193,13 +193,15 @@ impl<'a> Services<'a> {
     }
 
     /// Starts a basic version of the protocol with a single legacy solver and
-    /// quoter.
+    /// quoter. Optionally starts a baseline solver and uses it for price
+    /// estimation.
     pub async fn start_protocol_legacy_solver(
```
MartinquaXD, line 198: This function is not used in our code base. Given that we are transitioning towards dropping the legacy API are you still using this function?
If not we should just delete it.

MartinquaXD (approved):
Code looks correct although a bit sketchy.
I think eventually we need to find a way to configure these CLI arguments more reasonably but this is probably a bigger endeavor.

crates/e2e/src/setup/services.rs
devanoneth: You're right, thanks! We don't need it anymore. Deleted in my new commit.

PR #2517, opened 2024-03-14
Separate logging of gpv2 order filled reason

sunce86:
# Description
In cases when solvers try to solve already filled order, we want to have a separate reason so we can build alerts on top of it.


squadgazzz (approved):

fleupold (changes requested):
This solution strikes me as very hacky (matching on a magic string in the code that actually reports the error). Shouldn't this be part of the classification logic and result in a separate error type instead?

Also, I'm not so sure if this should be a structurally different error than a simulation revert as solvers could include already filled orders and it wouldn't be our fault.

PR #2521, opened 2024-03-15
Update Solver/driver api

harisang:
It seems there is a typo in the specification of the new solver/driver api. For custom interactions, I believe the driver is expecting an entry that is `callData` while in the api description we write `calldata`. This has already caused errors with some solver migrating to the new api.

Not sure if this is the only change that is needed so would appreciate some reviews.


fleupold (commented):
This is the API spec between the autopilot and the driver. I believe you also want to update 

https://github.com/cowprotocol/services/blob/027c8e6eeca33be45eb87ab20c30471b5edc1ac6/crates/solvers/openapi.yml#L739

PR #2523, opened 2024-03-15
[EASY] Separate log for disabled mempool

sunce86:
# Description
A separate log for when the mempool is disabled is nice to have - to avoid printing WARN log and the whole settlement data, and since it's not an actual error.


fleupold (approved):
Noice 👏

squadgazzz (approved):

PR #2538, opened 2024-03-19
PriceImprovement policy fee API

squadgazzz:
# Description
Adds a missing variant to the API


crates/orderbook/openapi.yml
```diff
@@ -1538,8 +1538,20 @@ components:
           type: number
       required:
         - factor
+    PriceImprovement:
+      description: The protocol fee is taken as a percent of the order price improvement which is a difference between the executed price and the best quote.
```
MartinquaXD, line 1542: nit: Should we specify the range of the values [0, 1)?

crates/orderbook/src/dto/order.rs
```diff
@@ -43,6 +43,15 @@ pub struct Order {
     pub signature: Signature,
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Quote {
+    pub sell_amount: U256,
```
MartinquaXD, line 50: Should use `#[serde_as(as = "HexOrDecimalU256")]`.

crates/orderbook/openapi.yml
squadgazzz: Added

crates/orderbook/src/dto/order.rs
squadgazzz: Thanks, added

MartinquaXD (approved):

crates/autopilot/src/infra/persistence/dto/order.rs
```diff
@@ -281,8 +281,11 @@ pub enum FeePolicy {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Quote {
+    #[serde_as(as = "HexOrDecimalU256")]
```
sunce86, line 284: If you changed this, then you also have to update the `driver` side: https://github.com/cowprotocol/services/blob/main/crates/driver/src/infra/api/routes/solve/dto/auction.rs#L329-L332

[Tests](https://github.com/cowprotocol/services/pull/2539) probably not catching this since they are not yet merged.
Driver tests apparently have their own (also wrong) serialization that needs to be updated: https://github.com/cowprotocol/services/blob/main/crates/driver/src/tests/setup/fee.rs#L33

sunce86 (approved):

PR #2547, opened 2024-03-20
Add onchain order creation error for non-zero-fee orders

fedgiac:
# Description

We had a missing refund for an eth-flow order because it tried to use nonzero fees on order creation, which is rejected by the backend since about two weeks (see 75f6e6092f305f3c5705ba7ce25b15b1e6fc78a9).

When this happens, our logs report `Could not retrieve a quote for order [..]: unspecified`. This PR should change the log message to read `Could not retrieve a quote for order [..]: non_zero_fee`.

## How to test

I did not test these changes outside of CI.


sunce86 (approved):
Can you create the issue for fixing the refunder?

PR #2552, opened 2024-03-21
Add cap for partner fee

### LINKED ISSUE #2541: feat: Add cap for partner fee
### Problem
We should validate the `partner_fee.bps` and make sure the value is smaller than some configuration parameter (default value 1%).

In case an order is created with a higher value than 1%, we should reject the creation of order with the appropriate message (right?).

### Acceptance criteria
Creating an order with partner fee higher than 1% should be rejected.

Mateo-mro:
# Description
Add a maximum cap (0.01) for the partner fee in case it is 0.01 or above.

# Changes
- Introduce a new type called `Factor` to enforce the limits of the value [0, 1) and provides a function to cap the partner fee

## How to test
1. e2e tests
2. unit test

## Related Issues
Fixes #2541


crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -93,10 +95,41 @@ pub enum Policy {
...
     },
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Into)]
+pub struct Factor(f64);
+
+impl Factor {
+    /// Convert into a `Factor` capping the value to fit [0, 1)
+    pub fn capped_from(value: f64) -> Self {
+        Self(value.max(0.0).min(0.99999))
```
MartinquaXD, line 108: I think it makes sense to not have `capped_from` and only use `try_from()` instead. Usually it's better long term to catch unexpected values early instead of trying to make them work somehow.

Mateo-mro: @MartinquaXD yes, but according to @olgafetisova (see https://github.com/cowprotocol/services/issues/2541#issuecomment-2009211011), we should cap the partner fee.

I agree that capping the partner fee but "errorring" for the other fees is a little bit confusing. What should we do?

sunce86: I would agree with Martin here. Specifically, we don't want to cap all factor values for all variants. We want to cap only partnerFee bps, so IMO this should be done only at the place where we extract partnerFee from appData and nowhere else. 

All other `factor`s should be rejected if outside of allowed range.

sunce86: Also, the caps are different:
1. Generally, for all `factor`s =  1.0
2. For bps is 100bps =  0.01

Mateo-mro: Oh I see, I misunderstood it then. Perfect! :ok_hand: thank you guys!

sunce86: And also, we want the 100bps = 0.01 to be configurable (as autopilot config parameter)

MartinquaXD: > but according to Olga (see https://github.com/cowprotocol/services/issues/2541#issuecomment-2009211011), we should cap the partner fee

Capping the values for partner fee and bps is a different concern, though. I thought the type introduced here is just here to make sure that we have reasonable factors. If we want to cap partner fees and bps there should be separate functions handling the individual capping values that return a `Factor`.

Mateo-mro: > And also, we want the 100bps = 0.01 to be configurable (as autopilot config parameter)

@sunce86 is it ok if I do that as a follow up task?

@MartinquaXD yeah, I agree, I was actually doing the changes as you said, so we think the same :))  the overuse of `from_capped` came from my misunderstanding that the max value for the partner fee was `1.0` instead of `0.01`

Mateo-mro:
https://github.com/cowprotocol/services/issues/2554

crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -93,10 +95,41 @@ pub enum Policy {
...
     },
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Into)]
+pub struct Factor(f64);
+
+impl Factor {
+    /// Convert a partner fee into a `Factor` capping its value
+    pub fn partner_fee_capped_from(value: f64) -> Self {
+        Self(value.max(0.0).min(0.01))
```
MartinquaXD, line 108:
```suggestion
    pub fn partner_fee_capped_from(bps: f64) -> Self {
        Self(bps.max(0.0).min(0.01))
```

crates/autopilot/src/domain/fee/policy.rs
```diff
@@ -55,8 +61,8 @@ impl Surplus {
             boundary::OrderClass::Liquidity => None,
             boundary::OrderClass::Limit => {
                 let policy = domain::fee::Policy::Surplus {
-                    factor: self.factor,
-                    max_volume_factor: self.max_volume_factor,
+                    factor: Factor::try_from(self.factor).unwrap(),
```
MartinquaXD, line 64: The domain types itself should also use the `Factor`. That way we don't need unwraps here.

crates/autopilot/src/domain/fee/mod.rs
Mateo-mro: @MartinquaXD what you propose is incorrect, it is parsing directly `f64` after the `bpd / 10_000`. We either keep it as it is or move the `bps` conversion inside this function (which I would avoid not to mix `bps` with `f64`.

MartinquaXD: Ahh, true. Actually after thinking more about this the `Factor` type is very general purpose and the conversion from `bps` to a factor (**and capping that**) is very specific to the partner fee stuff and probably doesn't belong into the `Factor` implementation to begin with.
I mean conceptually a `Factor` can be used in lots of different contexts which are completely unrelated to fees.
WDYT?

Mateo-mro: @MartinquaXD I agree with you, but in this case `Factor` is already particular for the fees (for example the fact that it cannot be `1.0` is very fee specific). I'd propose to rename it to `FeeFactor`, and then the capping function will have another input parameter to cap it with a max value (see https://github.com/cowprotocol/services/issues/2554). WDYT?

crates/driver/src/infra/simulator/tenderly/mod.rs
```diff
@@ -102,6 +102,7 @@ pub struct Simulation {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
```
Mateo-mro, line 105: Linter was complaining and I don't know if I should remove it.

crates/observe/src/future.rs
```diff
@@ -32,6 +32,7 @@ pin_project! {
 }
 
 #[derive(Debug)]
+#[allow(dead_code)]
```
Mateo-mro, line 35: Linter was complaining and I don't know if I should remove it.

crates/driver/src/infra/simulator/tenderly/mod.rs
MartinquaXD: Will open a PR to address this.

MartinquaXD (approved):

crates/autopilot/src/database/fee_policies.rs
```diff
@@ -77,20 +82,22 @@ mod tests {
 
         // surplus fee policy without caps
         let fee_policy_1 = domain::fee::Policy::Surplus {
-            factor: 0.1,
-            max_volume_factor: 1.0,
+            factor: FeeFactor::try_from(0.1).unwrap(),
```
sunce86, line 85: nit: this should work like 0.1.try_into().unwrap() but feel free to ignore

crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -116,3 +151,22 @@ impl From<domain::Quote> for Quote {
...
+            (2.0, 0.01),
+            (0.0, 0.0),
+            (-1.0, 0.0),
+            (0.001, 0.001),
+        ];
+        for (from, result) in values {
+            assert_eq!(f64::from(FeeFactor::partner_fee_capped_from(from)), result);
+        }
+    }
+}
```
sunce86, line 172: fine to test it out locally but I am not sure we want to use unit tests in the `domain`. I think the current decision is to not use it unless really necessary.

crates/e2e/tests/e2e/protocol_fee.rs
```diff
@@ -164,23 +165,32 @@ async fn partner_fee_sell_order_test(web3: Web3) {
     //
     // With protocol fee:
     // Expected executed_surplus_fee is 167058994203399 +
-    // 0.1*(10000000000000000000 - 167058994203399) = 1000150353094783059
+    // 0.01*(10000000000000000000 - 167058994203399) = 100165388404261365
     //
-    // Final execution is 10000000000000000000 GNO for 8884273887308040129 DAI, with
+    // Final execution is 10000000000000000000 GNO for 9772701276038844388 DAI, with
     // executed_surplus_fee = 1000150353094783059 GNO
```
sunce86, line 171: `executed_surplus_fee = 100165388404261365 GNO`

crates/e2e/tests/e2e/protocol_fee.rs
```diff
@@ -164,23 +165,32 @@ async fn partner_fee_sell_order_test(web3: Web3) {
...
     //
     // Settlement contract balance after execution = 1000150353094783059 GNO =
```
sunce86, line 173: `executed_surplus_fee = 100165388404261365 GNO`

sunce86 (approved):
lgtm, few nits

squadgazzz (approved):
LG

Issue #2555, opened 2024-03-21
bug: Rounding errors lead to discrepancy between score and surpus


fhenneke:
### Problem
With CIP-38, scores should deterministically depend on solutions. If there are not protocol fees, the score of the winning solution should be equal to the surplus of the solution. For tokens with few decimals, this does not always seem to be the case.

### Impact
This is probably not super important, but it indicates that the logic for computing scores from an encoded solution in the driver is different from the logic to compute surplus in the autopilot. This will lead to an observable difference when we move the driver logic to the autopilot.

At the moment, the solver rewards scripts use both surplus and score and this might lead to some inconsistencies. 

### To reproduce
For the transaction with hash `0x2733e73bba9a97b093270a36fa1677ddfb7e8ab2b415e29fac775fc50f418cbc`, the `settlement_scores` shows a winning score of `66264322342847212`. The entry in `settlement_observations` for that auction shows a surplus of `66309509871089417`. The score is _smaller_ than surplus.

The [competition endpoint](https://api.cow.fi/mainnet/api/v1/solver_competition/by_tx_hash/0x2733e73bba9a97b093270a36fa1677ddfb7e8ab2b415e29fac775fc50f418cbc) shows slightly different numbers for buy amounts than [etherscan](https://etherscan.io/tx/0x2733e73bba9a97b093270a36fa1677ddfb7e8ab2b415e29fac775fc50f418cbc).

### Expected behaviour
The logic in the autopilot is consistent with what I obtain using traded amounts and external prices from our database directly and with what happens onchain.

PR #2566, opened 2024-03-26
[EASY] Fix ceil_div in scoring

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2555

[Previously](https://github.com/cowprotocol/services/pull/2464) we fixed the ceil_div in SettlementEncoder, however the similar fix is needed in scoring module.

## How to test
Added temporary unit tests just to prove the fix is right. 
https://github.com/cowprotocol/services/pull/2566/commits/8ec61b259711ab14994d71042ff0848981f8e94c

Two multiplication unit tests prove that there is a difference in the sell price => difference between `21487` and `21488` when using the problematic settlement.

Then, added scoring test for the problematic settlement which proves that this difference causes the final score difference `66264322342847212` and `66309509871089417`.


crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -413,3 +414,27 @@ pub mod error {
...
+    #[test]
+    fn multiplication_div() {
+        let result = eth::U256::from(992158370040532752u128)
+            .checked_mul(eth::U256::from(1000713976565140939u128))
+            .unwrap()
+            .checked_div(eth::U256::from(46207788330881468994906599317053u128))
+            .unwrap();
+        assert_eq!(result, eth::U256::from(21487u128));
+    }
+}
```
sunce86, line 440: Will remove completely before merging.

crates/driver/src/domain/competition/solution/scoring.rs
```diff
@@ -354,3 +354,90 @@ pub enum Error {
...
+            (
+                eth::TokenAddress(ContractAddress(H160::from_slice(&hex!(
+                    "0001A500A6B18995B03f44bb040A5fFc28E45CB0"
+                )))),
+                competition::auction::Price::new(U256::from(1341581552361326u128).into()).unwrap(),
+            ),
+        ]);
+        println!("{:?}", scoring.score(&prices).unwrap());
+    }
+}
```
sunce86, line 443: Will remove completely before merging.

MartinquaXD (approved):

Mateo-mro (approved):

squadgazzz (approved):
That is a shame we can't catch it either in the e2e or driver tests.

PR #2569, opened 2024-03-27
[EASY] Bug: Count the protocol fees for the scoring in all the cases

Mateo-mro:
# Description
We need to count the protocol fees for the scoring in all the cases. Good catch from @fleupold .

See @fleupold 's comment: https://github.com/cowprotocol/services/pull/2544#discussion_r1541118981

# Changes
- Count the protocol fees for the scoring in all the cases

## How to test
1. Regression tests

### DIFF
--- a/crates/driver/src/tests/cases/protocol_fees.rs
+++ b/crates/driver/src/tests/cases/protocol_fees.rs
@@ -174,7 +174,7 @@ async fn protocol_fee_calculated_on_the_solver_side() {
                 buy: 40.ether().into_wei(),
             },
         },
-        expected_score: 20.ether().into_wei(),
+        expected_score: 35.ether().into_wei(),
         fee_handler: FeeHandler::Solver,
     };
 
--- a/crates/driver/src/domain/competition/solution/mod.rs
+++ b/crates/driver/src/domain/competition/solution/mod.rs
@@ -48,7 +48,6 @@ pub struct Solution {
     score: SolverScore,
     weth: eth::WethAddress,
     gas: Option<eth::Gas>,
-    fee_handler: FeeHandler,
 }
 
 impl Solution {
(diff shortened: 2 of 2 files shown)


crates/driver/src/tests/cases/protocol_fees.rs
```diff
@@ -174,7 +174,7 @@ async fn protocol_fee_calculated_on_the_solver_side() {
                 buy: 40.ether().into_wei(),
             },
         },
-        expected_score: 20.ether().into_wei(),
+        expected_score: 35.ether().into_wei(),
```
Mateo-mro, line 177: Now even if the `Solver` calculates the protocol fee, we still need to count it for the score, as per:

```rust
    /// CIP38 score defined as surplus + protocol fee
    ///
    /// Denominated in NATIVE token
    fn score(&self, prices: &auction::Prices) -> Result<eth::Ether, Error> {
        tracing::debug!("Scoring trade {:?}", self);
        Ok(self.native_surplus(prices)? + self.native_protocol_fee(prices)?)
    }
```

fleupold (approved):
Let's see what @squadgazzz and @sunce86 the masters of protocol fee related scoring say, but this looks correct to me.

crates/driver/src/tests/cases/protocol_fees.rs
sunce86: Not too obvious from the numbers in the test that we get capped by volume, but the math seems correct.

sunce86 (approved):
Lg.

good catch @fleupold

crates/driver/src/tests/cases/protocol_fees.rs
squadgazzz: Would it be possible to add comments to the test case explaining the expected values?

PR #2574, opened 2024-03-28
`non-zero-fee` DB enum variant

squadgazzz:
https://cowservices.slack.com/archives/C0361CDD1FZ/p1711653554068709

Should be covered by tests


MartinquaXD:
> Should be covered by tests

The execution of the migration is covered by tests but not the onchain creation of a malformed order.

MartinquaXD (approved):

fleupold (approved):

sunce86 (commented):
Is it possible to add a postgres unit test that iterates over all variants of error and inserts them into db, so we can catch this error next time?

PR #2576, opened 2024-03-28
[EASY] Iterator for OnchainOrderPlacementError

sunce86:
# Description
Inspired by error https://github.com/cowprotocol/services/pull/2574, now the unit test will fail if we forget to add the new variant to database.

## How to test
Existing tests

### DIFF
--- a/crates/database/Cargo.toml
+++ b/crates/database/Cargo.toml
@@ -12,7 +12,7 @@ const_format = "0.2"
 futures = { workspace = true }
 hex = { workspace = true }
 sqlx = { workspace = true }
-strum = { version = "0.25", features = ["derive"] }
+strum = { workspace = true }
 
 [dev-dependencies]
 hex-literal = { workspace = true }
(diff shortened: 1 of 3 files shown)


crates/database/Cargo.toml
```diff
@@ -12,6 +12,7 @@ const_format = "0.2"
 futures = { workspace = true }
 hex = { workspace = true }
 sqlx = { workspace = true }
+strum = { version = "0.25", features = ["derive"] }
```
MartinquaXD, line 15: We also have `strum` in the workspace `cargo.toml`.

MartinquaXD (approved):
Neat 👍

crates/database/Cargo.toml
sunce86: fixed

squadgazzz (approved):

fleupold (approved):
noice

PR #2580, opened 2024-03-29
Avoid building 2 auctions after restart

MartinquaXD:
# Description
The autopilot periodically builds new auctions via a background task that gets [spawned](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/solvable_orders.rs#L126-L129) immediately after starting the autopilot.
However, it also ["manually"](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/run.rs#L564-L567) builds 1 auction right at the start. This leads to 2 auctions being built in parallel whenever we restart the autopilot.

# Changes
Delete "manual" auction building process to do less work on restarts.

## How to test
e2e tests


squadgazzz (approved):
Nice!

sunce86 (approved):
nit: SolvableOrdersCache::update can be a private function now.

PR #2581, opened 2024-03-29
[EASY] Delay DB metrics until after restart is done

MartinquaXD:
# Description
Restarting the autopilot (especially trying out a new config) can be risky because it can be extremely slow on startup.
If you have to revert your temporary change because something it broken it can basically be a mini-outage.

# Changes
Don't immediately update DB metrics (table sizes) but rather start by waiting.
This effectively delays updating the DB metrics until **after** the most important work is done after a restart (i.e. building the new auction / filling native price cache).

## How to test
I built a hacky benchmark that simply starts the autopilot connected to the read replica of the prod mainnet DB (to get the most realistic data). That showed that the query to fetch open orders takes ~10s - 20s when we immediately start updating the DB metrics in parallel but only ~2s-3s when we delay updating those metrics.


squadgazzz (approved):
LGTM

sunce86 (approved):

PR #2584, opened 2024-03-31
Filter Empty Liquidity Sources

### LINKED ISSUE #2349: chore: Filter out "empty" liquidity sources
### Background
Some solvers report `Ignoring ill-defined AMM "3": Attempt to create an AMM with non-positive token balance.` indicating that we are sending "empty" AMMs.

There are definitely some constant product AMMs that have either 0 or very little liquidity.
It probably makes sense to filter out any pool that has `< X$` worth of tokens in them.

Examples:
* https://etherscan.io/address/0x4db44282701e28e0c2ac76188b8489298aabbc04
* https://etherscan.io/address/0xe2deb0d3578f99fb1eeaaf4c5f25a441bfd75d9f
* https://etherscan.io/address/0xa65ab9b0d2f070eb9c68615b723d9ed95b36ba22

[instance.json](https://github.com/cowprotocol/services/files/14116820/instance.json)

### Acceptance criteria
If an AMM does not actually have any liquidity we should not send it to the solver.
Additionally we should not attempt to fetch its state again on the next block if it's empty.

bh2smith:
If approved, this PR closes #2349. This is intended to provide the routes necessary to filter empty liquidity sources. It really just amounts to implementing `is_empty` on `Liquidity` but each liquidity source will have its own definition of emptiness. 

In this draft we naively implement the method for every source as "NOT EMPTY" except for `ConstantProduct` pools where we filter out any pool having one or more token reserves equal to 0 WEI. The logic of each of these would benefit from passing relevant token prices to the method calls along with a configurable USD value associated with 
"emptiness".

I was thinking that prices might already exist somewhere in the program's runtime so we could just route them into here (seeking further guidance on how the backend devs plan to approach this). 

cc @MartinquaXD (as the creator of the linked issue)

### DIFF
--- a/crates/solver/src/liquidity.rs
+++ b/crates/solver/src/liquidity.rs
@@ -73,6 +73,17 @@ impl Liquidity {
             Liquidity::Concentrated(amm) => Some(amm.pool.address),
         }
     }
+
+    pub fn is_empty(&self) -> bool {
+        match self {
+            Liquidity::ConstantProduct(amm) => amm.is_empty(),
+            Liquidity::BalancerWeighted(amm) => amm.is_empty(),
+            Liquidity::BalancerStable(amm) => amm.is_empty(),
+            // This should probably check balances of the limit order...
+            Liquidity::LimitOrder(_) => false,
+            Liquidity::Concentrated(_) => false,
+        }
+    }
 }
 
 /// A trait associating some liquidity model to how it is executed and encoded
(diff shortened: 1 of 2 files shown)


MartinquaXD:
The approach outlined in the PR seems reasonable to me. However, the more important part of:
> Additionally we should not attempt to fetch its state again on the next block if it's empty.

would not yet be addressed by this PR (even with correct implementations for `is_empty()`).

PR #2585, opened 2024-04-01
Cleanup score

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2493
Fixes https://github.com/cowprotocol/services/issues/1494

Important notes:
1. ~Removed risk parameter config from solvers - requires cleanup of "riskParameters" in infrastructure repo~ - will be done in a separate PR
2. Solvers api "score" field is optional until solvers update their responses - will be removed completely https://github.com/cowprotocol/services/issues/2588.

Follow ups:
1. cleanup of "riskParameters" in infrastructure repo
3. Update gnosis solvers response https://github.com/gnosis/solvers/blob/main/src/api/routes/solve/dto/solution.rs - https://github.com/gnosis/solvers/pull/10
4. Add more score errors to notify endpoint - https://github.com/cowprotocol/services/issues/2597

## How to test
Existing tests. No change in behaviour is expected.


MartinquaXD (commented):
My understanding is that unfortunately the correct bidding behavior still requires the risk parameters to adjust the computed solution / reported score based on the revert risk (see [slack](https://cowservices.slack.com/archives/C036JAGRQ04/p1711444426038049?thread_ts=1711399703.435729&cid=C036JAGRQ04)).

crates/driver/src/infra/notify/mod.rs
```diff
@@ -27,29 +27,13 @@ pub fn scoring_failed(
...
-        score::Error::RiskAdjusted(score::risk::Error::ObjectiveValueNonPositive(
-            quality,
-            gas_cost,
-        )) => notification::Kind::ScoringFailed(
-            notification::ScoreKind::ObjectiveValueNonPositive(*quality, *gas_cost),
-        ),
-        score::Error::RiskAdjusted(score::risk::Error::Boundary(_)) => return,
-        score::Error::Boundary(_) => return,
-        score::Error::Scoring(_) => return, // TODO: should we notify?
+        _ => return,
```
squadgazzz, line 36: Can we keep listing all other variants explicitly? That would help to avoid mistakes in the future if a new variant is added, for example.

crates/driver/src/infra/notify/notification.rs
```diff
@@ -49,26 +49,8 @@ pub enum Kind {
...
-    /// [0, 1]
-    /// [ONLY APPLICABLE TO SCORES BASED ON SUCCESS PROBABILITY]
-    SuccessProbabilityOutOfRange(f64),
-    /// Objective value is defined as quality (surplus + fees) - gas costs.
-    /// Protocol doesn't allow solutions that cost more than they bring to
-    /// the users and protocol.
-    /// [ONLY APPLICABLE TO SCORES BASED ON SUCCESS PROBABILITY]
-    ObjectiveValueNonPositive(Quality, GasCost),
+    /// No clearing prices are present for all trades.
+    InvalidClearingPrices,
```
squadgazzz, line 53: Do we still need the enum? Can we use `Kind::ScoringFailed` directly without inner value?

MartinquaXD:
> My understanding is that unfortunately the correct bidding behavior still requires the risk parameters to adjust the computed solution / reported score based on the revert risk

Since the consensus on slack seems to be that the risk parameters can be replaced by offsetting the settlement gas cost it can be removed as well after all. Sorry for the back and forth. 😅

MartinquaXD (approved):
Actually the `risk` parameters can also be removed in a separate PR. Might be nicer for reviewing purposes and in case one of these has to be reverted.

crates/driver/src/infra/notify/notification.rs
sunce86: Actually we don't need it but I guessed it's less work to leave it like this since we will add more variants in https://github.com/cowprotocol/services/issues/2597

crates/driver/src/infra/notify/mod.rs
sunce86: I simplified the error enum and listed all of them. Can be additionally cleaned up as part of
https://github.com/cowprotocol/services/issues/2597

sunce86: Actually forgot to commit 😶‍🌫️  will do it in a follow up PR

PR #2591, opened 2024-04-03
Test branch

MartinquaXD:
## Description
Copy of https://github.com/cowprotocol/services/pull/2253 that directly merges main to see if this would run the CI as expected.

## TODO
delete after test

PR #2592, opened 2024-04-03
Run CI on HEAD commit instead of merge commit

MartinquaXD:
# Description
When running CI on PR github does something unexpected. It does actually **NOT** run on the last commit of the PR (i.e. HEAD) but rather checks out the target branch (usually `main`) and merges the HEAD into it.
Most of the time this either works without issues or results in a merge conflict causing the CI to not start at all.

However, when the HEAD commit can be merged into the target branch without merge conflicts **BUT WITH BREAKING CHANGES NEVERTHELESS** you get build or test execution issues that can not be reproduced by just using your branch locally.

For more info check out the investigation on [slack](https://cowservices.slack.com/archives/C0375NV72SC/p1712179673286249?thread_ts=1712177157.374909&cid=C0375NV72SC).

# Changes
Configures checkout action to pull HEAD instead of the merge commit.

## How to test
Check the job history of this https://github.com/cowprotocol/services/pull/2591.
First it didn't work because the `main` branch made `start_autopilot` async without causing a merge conflict. But with this patch it actually ran the CI on the commit one would intuitively expect.
Note that github still doesn't let you merge the PR if it's not up-to-date so I don't really see any downside using the HEAD commit here.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


squadgazzz (approved):
Neat! That actually the first time I see that param is explicitly required to work with the HEAD. Never faced this issue before.

PR #2602, opened 2024-04-09
[EASY] Should we quote using limit orders - config

sunce86:
# Description
Introduces per solver configuration which defines whether the limit or market orders should be used for quoting.

Instead of making the switch for all solvers at once (or doing it in a hacky way internally in each solver implementation), now we are able to make the switch one by one and test each one (otherwise it would be too risky IMO). Also I am not sure external solvers are ready to quote using limit orders, so this PR allows us to not be blocked on introducing this feature to subset of solvers (baseline, dex solvers etc).

Once merged, baseline will be moved to quote using limit orders (via infrastructure PR) and tested.

## How to test
Existing tests. No change in behaviour should exist with this PR.


crates/driver/src/domain/quote.rs
```diff
@@ -87,7 +87,12 @@ impl Order {
         };
 
         let solutions = solver
-            .solve(&self.fake_auction(eth, tokens).await?, &liquidity)
+            .solve(
+                &self
+                    .fake_auction(eth, tokens, solver.quote_using_limit_orders())
```
MartinquaXD, line 92: This call is getting a bit weird. Maybe the code would be slightly nicer if `fake_auction` would be a function on the solver instead that takes the order? 🤔

sunce86: I expect `quote_using_limit_orders` to be removed once all solvers are able to quote using limit orders, which would mean we would be left without the only reason to put the `fake_auction` on `solver` if we do what you suggest.

MartinquaXD: Yeah, fair point. Overall I find how the quoting code is currently structured a bit weird regardless but this could then be a refactoring outside of this PR.

MartinquaXD (approved):
Makes sense to me to be able to migrate to the new quoting scheme bit by bit. :+1:

PR #2603, opened 2024-04-09
Fix decimals for OneInchSpotPrice estimator

sunce86:
# Description
OneInchSpotPrice estimator assumes that all tokens have 18 decimals. In order for `NativePriceEstimateResult` to be populated as expected, we need to divide the response by number of the decimals for each specific token.

USDC example: https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/r/s/EFIxv
Here we can see that result was divided by 1e18 instead of 1e6, since USDC has 6 decimals.

# Changes

- [ ] Added token info fetching to OneInchSpotPrice estimator
- [ ] ...

## How to test
Existing tests

### DIFF
--- a/crates/shared/src/price_estimation/native/oneinch.rs
+++ b/crates/shared/src/price_estimation/native/oneinch.rs
@@ -63,18 +63,19 @@ impl OneInch {
         tokio::task::spawn(async move {
             let mut block_stream = into_stream(current_block);
             loop {
-                match update_prices(
+                let current_prices = get_current_prices(
                     &client,
                     base_url.clone(),
                     api_key.clone(),
                     chain_id,
                     token_info.as_ref(),
                 )
-                .await
-                {
-                    Ok(new_prices) => {
+                .await;
+
+                match current_prices {
+                    Ok(current_prices) => {
                         tracing::debug!("OneInch spot prices updated");
-                        *prices.lock().unwrap() = new_prices;
+                        *prices.lock().unwrap() = current_prices;
                     }
                     Err(err) => {
                         tracing::warn!(?err, "OneInch spot price update failed");
@@ -99,7 +100,7 @@ impl NativePriceEstimating for OneInch {
     }
 }
 
-async fn update_prices(
+async fn get_current_prices(
     client: &Client,
     base_url: Url,
     api_key: Option<String>,
@@ -147,7 +148,10 @@ async fn update_prices(
 mod tests {
     use {
         super::*,
-        crate::price_estimation::oneinch::BASE_URL,
+        crate::{
+            price_estimation::oneinch::BASE_URL,
+            token_info::{MockTokenInfoFetching, TokenInfo},
+        },
         std::{env, str::FromStr},
     };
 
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/native/oneinch.rs
```diff
@@ -105,16 +125,33 @@ async fn update_prices(
...
+        .0;
+
+    let token_decimals = token_info
+        .get_token_infos(&prices.keys().copied().collect::<Vec<_>>())
+        .await;
+
+    let normalized_prices = prices
+        .into_iter()
+        .filter_map(|(token, price)| {
+            let decimals = token_decimals.get(&token)?.decimals?;
```
squadgazzz, line 139: How about adding more context on which exactly token is failed and for what reason?

squadgazzz (approved):

crates/shared/src/price_estimation/native/oneinch.rs
```diff
@@ -105,16 +125,36 @@ async fn update_prices(
...
         .context("Failed to fetch Native 1Inch prices")?;
-    let result = serde_json::from_str::<Response>(&response)
-        .with_context(|| format!("Failed to parse Native 1inch prices from {response:?}"))?;
-    Ok(result.0)
+    let prices = serde_json::from_str::<Response>(&response)
+        .with_context(|| format!("Failed to parse Native 1inch prices from {response:?}"))?
+        .0;
+
+    let token_infos = token_info
+        .get_token_infos(&prices.keys().copied().collect::<Vec<_>>())
```
Mateo-mro, line 133: why do you need to copied it if you are passing it by ref?

sunce86: that's the type required by `get_token_infos` function.

MartinquaXD: Yeah, I have to pass a reference to a slice but it has to be `&[H160]` not `&[&H160]`.

crates/shared/src/price_estimation/native/oneinch.rs
```diff
@@ -105,16 +125,36 @@ async fn update_prices(
...
+        .await;
+
+    let normalized_prices = prices
+        .into_iter()
+        .filter_map(|(token, price)| {
+            let Some(decimals) = token_infos.get(&token).and_then(|info| info.decimals) else {
+                tracing::debug!(?token, "could not fetch decimals; discarding spot price");
+                return None;
+            };
+            let normalized_price = price.to_f64_lossy() / 10f64.powf(decimals.into());
```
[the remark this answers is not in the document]
MartinquaXD: Delaying the rounding error makes sense. But we have to do the computation with `BigRationals` instead of `U256`. Otherwise we'll lose even more precision given that `U256` can't encode fractional numbers.

crates/shared/src/price_estimation/native/oneinch.rs
```diff
@@ -22,8 +22,9 @@ struct Response(#[serde_as(as = "HashMap<_, HexOrDecimalU256>")] HashMap<Token,
 
 type Token = H160;
 type PriceInWei = U256;
+
 pub struct OneInch {
-    prices: Arc<Mutex<HashMap<Token, PriceInWei>>>,
+    prices: Arc<Mutex<HashMap<Token, f64>>>,
```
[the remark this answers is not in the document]
MartinquaXD: Added a type alias for `f64`.

PR #2604, opened 2024-04-10
Hotfix v2.253.2

squadgazzz:
Fix decimals for OneInchSpotPrice estimator (#2603)


MartinquaXD (changes requested):
Hotfix looks good. (previous commit is the last commit of the regular release and diff checks out).
Just marking as "requesting changes" so we don't merge it by accident.

MartinquaXD:
I guess in the future it's enough to link the created release in the infra hotfix PR if you want us to sanity check it since the diff is also visible there.

PR #2612, opened 2024-04-11
Add more notify call for scoring

### LINKED ISSUE #2597: chore: Add more notify call for scoring
### Background
This issue captures follow up task after https://github.com/cowprotocol/services/pull/2585 is merged.

Few more notify types should be added: `InvalidExecutedAmount` and `MissingPrices` are probably worth adding.

Mateo-mro:
# Description
Add more notify call for scoring

# Changes
Add notify calls for the following score errors:
- `InvalidExecutedAmount`
- `MissingPrice(TokenAddress)`
- `CustomPrice(String)`

## How to test
1. e2e
2. unit test

## Related Issues
Fixes #2597

### DIFF
--- a/crates/driver/src/domain/competition/solution/mod.rs
+++ b/crates/driver/src/domain/competition/solution/mod.rs
@@ -151,7 +151,9 @@ impl Solution {
                     .get(&trade.order().buy.token.wrap(self.weth))
                     .ok_or(error::Scoring::InvalidClearingPrices)?,
             };
-            let custom_prices = trade.calculate_custom_prices(&uniform_prices)?;
+            let custom_prices = trade
+                .calculate_custom_prices(&uniform_prices)
+                .map_err(error::Scoring::CalculateCustomPrices)?;
             trades.push(scoring::Trade::new(
                 trade.order().sell,
                 trade.order().buy,
@@ -163,7 +165,7 @@ impl Solution {
         }
 
         let scoring = scoring::Scoring::new(trades);
-        Ok(scoring.score(prices)?)
+        scoring.score(prices).map_err(error::Scoring::from)
     }
 
     /// Approval interactions necessary for encoding the settlement.
@@ -499,7 +501,29 @@ pub mod error {
         InvalidClearingPrices,
         #[error(transparent)]
         Math(#[from] Math),
+        #[error("failed to calculate custom prices")]
+        CalculateCustomPrices(#[source] Trade),
+        #[error("missing native price for token {0:?}")]
+        MissingPrice(TokenAddress),
+    }
+
+    impl From<scoring::Error> for Scoring {
+        fn from(value: scoring::Error) -> Self {
+            match value {
+                scoring::Error::MissingPrice(e) => Self::MissingPrice(e),
+                scoring::Error::Math(e) => Self::Math(e),
+                scoring::Error::Scoring(e) => e,
+            }
+        }
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Trade {
+        #[error("orders with non solver determined gas cost fees are not supported")]
+        ProtocolFeeOnStaticOrder,
+        #[error("invalid executed amount")]
+        InvalidExecutedAmount,
         #[error(transparent)]
-        Score(#[from] scoring::Error),
+        Math(#[from] Math),
     }
 }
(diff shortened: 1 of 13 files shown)


crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -497,7 +499,31 @@ pub mod error {
         InvalidClearingPrices,
         #[error(transparent)]
         Math(#[from] Math),
+        #[error("failed to calculate custom prices")]
+        CalculateCustomPrices(#[source] Trade),
+        #[error("missing native price for token {0:?}")]
+        MissingPrice(TokenAddress),
+        #[error("failed to calculate custom price {0:?}")]
```
MartinquaXD, line 506: What makes this error different from `CalculateCustomPrices`? From the perspective of the solver they seem pretty indistinguishable and not self explanatory.

MartinquaXD (commented):
I think those new error types also have to be added to the parsing logic on the solver side.

crates/driver/src/domain/competition/solution/mod.rs
Mateo-mro: Fair enough, that error could be squashed into its upper level error `error::Scoring` without a problem.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -497,7 +499,29 @@ pub mod error {
...
+        fn from(value: scoring::Error) -> Self {
+            match value {
+                scoring::Error::MissingPrice(e) => Self::MissingPrice(e),
+                scoring::Error::Math(e) => Self::Math(e),
+                scoring::Error::Scoring(e) => e,
+            }
+        }
+    }
+
+    #[derive(Debug, thiserror::Error)]
```
MartinquaXD, line 518: Why move this here?
Looks like the error is mainly used in a different file and we only need it to know about it here to convert from it.

crates/driver/src/infra/notify/notification.rs
```diff
@@ -51,6 +51,10 @@ pub enum Kind {
 pub enum ScoreKind {
     /// No clearing prices are present for all trades.
     InvalidClearingPrices,
+    /// Trade: invalid executed amount
+    InvalidExecutedAmount,
+    /// missing native price for token {0:?}
```
MartinquaXD, line 56: These doc comments could be more explicit. Also why use the debug notation here?

crates/driver/src/infra/solver/dto/notification.rs
```diff
@@ -108,6 +118,10 @@ pub enum Kind {
         succeeded_once: bool,
     },
     InvalidClearingPrices,
+    MissingPrice {
```
MartinquaXD, line 121:
```suggestion
    #[serde(rename_all = "camelCase")]
    MissingPrice {
```

MartinquaXD (commented):
I think we still need a change in the `solvers-dto` crate. Otherwise the solver will not know how to deserialize the new error. This will then also require an update in the gnosis `solvers` repo.

crates/driver/src/infra/solver/dto/notification.rs
Mateo-mro: This is redundant. The enum `Kind` already derives it.

MartinquaXD: The `derive` on the enum kind only causes the enum variant **names** to be camelcased but not the **fields** of the structs inside the variant.

Mateo-mro: ah, I misunderstood, in that case yes.

squadgazzz (approved):

MartinquaXD (approved):
Looks okay to me.
@sunce86 could you also give it a look. I think since you worked quite a bit on the notification system you are probably more likely to spot errors than anyone of use.

crates/driver/src/domain/competition/solution/scoring.rs
```diff
@@ -419,6 +419,6 @@ pub enum Error {
     MissingPrice(eth::TokenAddress),
     #[error(transparent)]
     Math(#[from] Math),
-    #[error("failed to calculate custom price {0:?}")]
-    CustomPrice(String),
+    #[error("scoring: failed to calculate custom price {0:?}")]
```
sunce86, line 422: i believe this comment needs to be updated

sunce86 (approved):

PR #2614, opened 2024-04-12
Add per solver settlement merging flag

MartinquaXD:
# Description
Fixes: https://github.com/cowprotocol/services/issues/2418

# Changes
Instead of always allowing the driver to merge multiple solutions proposed by the solver it now has to be enabled via a flag.

## How to test
Adjusted existing tests and added new test where merging is theoretically possible but doesn't happen because the feature was not enabled (mostly copied from another test).


crates/driver/src/infra/config/file/load.rs
```diff
@@ -77,6 +85,10 @@ pub async fn load(chain: eth::ChainId, path: &Path) -> infra::Config {
                 request_headers: config.request_headers,
                 fee_handler: config.fee_handler,
                 quote_using_limit_orders: config.quote_using_limit_orders,
+                merge_solutions: match config.merge_solutions {
```
Mateo-mro, line 88: nit: why not implement the `From<>` trait and just then do
```rust
config.merge_solutions.into()
```

MartinquaXD: 1. this conversion only happens in one location so here it would just add mental overhead to jump to the function definition IMO
2. implementing `From<bool>` loses all of the surrounding context so I think it's more error prone than this more explicit conversion in place

Mateo-mro: why would it be more error prone if the logic to convert from `bool` to the custom type is unified in one place? :thinking:

Mateo-mro: The logic would be the same:

```rust
impl From<bool> for SolutionMerging {
  fn from(value: bool) -> Self {
      match config.merge_solutions {
            true => SolutionMerging::Allowed,
             false => SolutionMerging::Forbidden,
        },
  }
}
```

it would be the **exact** same logic, and in case in the future it is needed to do a `from` or `into` the logic will be consistent and not all over the place.

MartinquaXD: > it would be the exact same logic

Nope, because you cheated. 😛 
Either you implement `From<bool>` then you don't get the context that the bool is actually read from `config.merge_solutions` (like you have in your code) or you implement `From<Config>` which would allow your implementation to actually read `config.merge_solutions` but then the `From` implementation is just weird. 😅

squadgazzz: To me, `From<bool>` looks too generic.

MartinquaXD: Exactly. If you actually implement `From<bool>` nothing is stopping you from calling `config.destroy_world.into()` or `config.protect_users.into()` or whatever else precisely because there is no context that could be checked by the compiler that you are actually doing the right thing.

Mateo-mro: > Nope, because you cheated. 😛

haha I see what you mean, I copied/pasted :laughing: 

I understand your point guys!, but I think the whole issue comes with this:

well, as it is the field:
```rust
    #[serde(default)]
    merge_solutions: bool,
```
I believe maybe the best would be to have a custom type (even tho underneath the type is just a bool), and then do the transformation from that type to `SolutionMerging`. But I won't be very insistent here since it is "annoying" to do so. My concern in the end is the same as yours guys: to misuse the bool by accident.

Mateo-mro: Also guys, you are casting a `bool` into an `enum`, why not to use the `enum` directly in the configuration, making it easier to add more in the future? (if it is needed at all, maybe it's not needed :sweat_smile: )

sunce86 (approved):

PR #2615, opened 2024-04-12
Index Traded order event labels based on on-chain data

### LINKED ISSUE #2530: bug: index `Traded` order event labels based on on-chain data
### Problem
Currently we are indexing the `Traded` events after the driver reports that it settled the solution. However, it can happen that the solver thinks it failed to submit a solution but actually succeeded.

### Impact
In those cases we currently don't add the `Traded` order events label to the DB.

### Fix
Instead we should be indexing those order event labels when we are indexing the on-chain events.

Mateo-mro:
# Description
Currently we are indexing the Traded events after the driver reports that it settled the solution. However, it can happen that the solver thinks it failed to submit a solution but actually succeeded.

# Changes
Index order event labels when we are indexing the on-chain events.

## How to test
1. regression tests

## Related Issues
Fixes #2530

### DIFF
--- a/crates/autopilot/src/database/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/database/on_settlement_event_updater.rs
@@ -75,9 +75,7 @@ impl super::Postgres {
                 .iter()
                 .map(|(order_uid, _)| auction::order::OrderUid::from(*order_uid))
                 .collect::<Vec<_>>();
-            store_order_events(ex, order_uids, OrderEventLabel::Traded, Utc::now())
-                .await
-                .context("store_order_events")?;
+            store_order_events(ex, order_uids, OrderEventLabel::Traded, Utc::now()).await;
             for (order, executed_fee) in auction_data.order_executions {
                 database::order_execution::save(
                     ex,
--- a/crates/autopilot/src/database/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/database/on_settlement_event_updater.rs
@@ -73,9 +73,8 @@ impl super::Postgres {
             let order_uids = auction_data
                 .order_executions
                 .iter()
-                .map(|(order_uid, _)| *order_uid)
-                .map(Into::into)
-                .collect::<Vec<auction::order::OrderUid>>();
+                .map(|(order_uid, _)| auction::order::OrderUid::from(*order_uid))
+                .collect::<Vec<_>>();
             store_order_events(ex, order_uids, OrderEventLabel::Traded, Utc::now())
                 .await
                 .context("store_order_events")?;
--- a/crates/autopilot/src/database/on_settlement_event_updater.rs
+++ b/crates/autopilot/src/database/on_settlement_event_updater.rs
@@ -1,6 +1,12 @@
 use {
+    crate::{database::order_events::store_order_events, domain::auction},
     anyhow::{Context, Result},
-    database::{byte_array::ByteArray, settlement_observations::Observation},
+    chrono::Utc,
+    database::{
+        byte_array::ByteArray,
+        order_events::OrderEventLabel,
+        settlement_observations::Observation,
+    },
     ethcontract::U256,
     model::order::OrderUid,
     number::conversions::u256_to_big_decimal,
@@ -64,6 +70,15 @@ impl super::Postgres {
             .await
             .context("insert_settlement_observations")?;
 
+            let order_uids = auction_data
+                .order_executions
+                .iter()
+                .map(|(order_uid, _)| *order_uid)
+                .map(Into::into)
+                .collect::<Vec<auction::order::OrderUid>>();
+            store_order_events(ex, order_uids, OrderEventLabel::Traded, Utc::now())
+                .await
+                .context("store_order_events")?;
             for (order, executed_fee) in auction_data.order_executions {
                 database::order_execution::save(
                     ex,
--- a/crates/autopilot/src/database/order_events.rs
+++ b/crates/autopilot/src/database/order_events.rs
@@ -1,13 +1,13 @@
 pub use database::order_events::OrderEventLabel;
 use {
     crate::domain,
-    anyhow::{Context, Result},
+    anyhow::Result,
     chrono::{DateTime, Utc},
     database::{
         byte_array::ByteArray,
         order_events::{self, OrderEvent},
     },
-    sqlx::Error,
+    sqlx::{Error, PgConnection},
 };
 
 impl super::Postgres {
@@ -18,12 +18,11 @@ impl super::Postgres {
 }
 
 pub async fn store_order_events(
-    db: &super::Postgres,
+    ex: &mut PgConnection,
     order_uids: Vec<domain::OrderUid>,
     label: OrderEventLabel,
     timestamp: DateTime<Utc>,
 ) -> Result<()> {
-    let mut ex = db.pool.begin().await.context("begin transaction")?;
     for uid in order_uids {
         let event = OrderEvent {
             order_uid: ByteArray(uid.0),
@@ -31,8 +30,7 @@ pub async fn store_order_events(
             label,
         };
 
-        order_events::insert_order_event(&mut ex, &event).await?
+        order_events::insert_order_event(ex, &event).await?
     }
-    ex.commit().await?;
     Ok(())
 }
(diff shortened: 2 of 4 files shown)


crates/autopilot/src/database/on_settlement_event_updater.rs
```diff
@@ -64,6 +70,15 @@ impl super::Postgres {
             .await
             .context("insert_settlement_observations")?;
 
+            let order_uids = auction_data
+                .order_executions
+                .iter()
+                .map(|(order_uid, _)| *order_uid)
+                .map(Into::into)
```
MartinquaXD, line 77: Can be collapsed into one `.map()`.

crates/autopilot/src/database/order_events.rs
```diff
@@ -18,21 +18,19 @@ impl super::Postgres {
 }
 
 pub async fn store_order_events(
-    db: &super::Postgres,
+    ex: &mut PgConnection,
     order_uids: Vec<domain::OrderUid>,
     label: OrderEventLabel,
     timestamp: DateTime<Utc>,
 ) -> Result<()> {
-    let mut ex = db.pool.begin().await.context("begin transaction")?;
```
MartinquaXD, line 26: What's the reason to no longer build a transaction here?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -100,16 +104,25 @@ impl Persistence {
...
+                let store_order_events_inner = async {
+                    let start = Instant::now();
+                    let events_count = order_uids.len();
+                    let mut ex = db.pool.begin().await.context("begin transaction")?;
+                    store_order_events(&mut ex, order_uids, label, Utc::now()).await?;
+                    ex.commit().await?;
+                    Ok::<(Instant, usize), anyhow::Error>((start, events_count))
+                };
+
+        match store_order_events_inner.await {
```
MartinquaXD, line 117: Formatting is weird.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -100,16 +104,25 @@ impl Persistence {
...
-                match boundary::store_order_events(&db, order_uids, label, Utc::now()).await {
-                    Ok(_) => {
-                        tracing::debug!(elapsed=?start.elapsed(), ?events_count, "stored order events");
-                    }
-                    Err(err) => {
-                        tracing::warn!(?err, "failed to insert order events");
-                    }
-                }
+
+                let store_order_events_inner = async {
```
MartinquaXD, line 108: Is this intermediate async closure needed here?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -100,16 +104,25 @@ impl Persistence {
...
+                    let events_count = order_uids.len();
+                    let mut ex = db.pool.begin().await.context("begin transaction")?;
+                    store_order_events(&mut ex, order_uids, label, Utc::now()).await?;
+                    ex.commit().await?;
+                    Ok::<(Instant, usize), anyhow::Error>((start, events_count))
+                };
+
+        match store_order_events_inner.await {
+            Ok((start, events_count)) => {
+                tracing::debug!(elapsed=?start.elapsed(), ?events_count, "stored order events");
```
MartinquaXD, line 119: Since the order events labels are not critical I think we could also make `store_order_events` handle and log the error internally to make the call sites nicer.

crates/autopilot/src/database/order_events.rs
Mateo-mro: becuase in order to make this function usable outside the `persistence` module I needed to pass to it the `ex: &mut PgConnection` already, so it is not needed to begin the tx.

crates/autopilot/src/database/on_settlement_event_updater.rs
Mateo-mro: sure! I like it more with two `.maps` because I will put it with one.

crates/autopilot/src/infra/persistence/mod.rs
Mateo-mro: @MartinquaXD I can do that, but then the behavior won't be the same as it used to, are you OK with that?

crates/autopilot/src/database/order_events.rs
Mateo-mro: btw, why don't we have the database access unified within the `autopilot`? is there any reason for it?

crates/autopilot/src/infra/persistence/mod.rs
MartinquaXD: I think this is fine. In the future please feel free to suggest these slight changes to the behavior if it leads to more readable code. If you are not sure if a change is okay just point it out and it can be discussed in the PR.

crates/autopilot/src/database/order_events.rs
```diff
@@ -18,21 +19,29 @@ impl super::Postgres {
...
         let event = OrderEvent {
             order_uid: ByteArray(uid.0),
             timestamp,
             label,
         };
 
-        order_events::insert_order_event(&mut ex, &event).await?
+        match order_events::insert_order_event(ex, &event).await {
+            Ok(_) => {
+                tracing::debug!(elapsed=?start.elapsed(), ?events_count, "stored order events");
```
MartinquaXD, line 38: You are printing the log for each order in the loop (claiming that it stored `events_count` things).
Over all I would say `store_order_events` should:
* take a `&mut PgConnection`
* start a fresh transaction
* handle errors internally by logging
* return nothing (an error in this logic is not critical and already got logged internally)

Then the caller does not have to care about transactions or the function throwing errors at all making all call sites nicer to read without compromising anything (since we didn't handle any errors before anyway).

crates/autopilot/src/database/order_events.rs
```diff
@@ -18,21 +19,31 @@ impl super::Postgres {
 }
 
 pub async fn store_order_events(
-    db: &super::Postgres,
+    ex: &mut PgConnection,
     order_uids: Vec<domain::OrderUid>,
     label: OrderEventLabel,
     timestamp: DateTime<Utc>,
 ) -> Result<()> {
```
Mateo-mro, line 26: We need to keep returning error here because this function is used by different kind of callers.

MartinquaXD: The point of my remark was that there is no reasonable error handling besides logging stuff. That's why we can do all the error handling internally and don't return any result from this function at all.
That way the call site looks clean and the logging happens in a single location instead of at every call site.

crates/autopilot/src/database/order_events.rs
MartinquaXD: There is still one log like `events_count=12 stored order events` for every event that gets inserted.
What makes more sense would be something like `order_count=10 stored order events` and `order_count=3 failed to store order events` (once instead of for every single insert).

crates/autopilot/src/database/order_events.rs
Mateo-mro: I removed the `Result` and made it panic for any error.

MartinquaXD: Even if we expect this case to happen extremely rarely panicking is not okay here. This would disrupt everything and not just the surrounding DB operation that this change is partly intended to protect. I was thinking about something along those lines:
```rust
pub async fn store_order_events(
    ex: &mut PgConnection,
    order_uids: Vec<domain::OrderUid>,
    label: OrderEventLabel,
    timestamp: DateTime<Utc>,
) {
    let start = Instant::now();
    let count = order_uids.len();

    let insert = async move {
        let mut ex = ex.begin().await?;

        for uid in order_uids {
            let event = OrderEvent {
                order_uid: ByteArray(uid.0),
                timestamp,
                label,
            };

            order_events::insert_order_event(&mut ex, &event).await?;
        }

        ex.commit().await
    };

    match insert.await {
        Ok(_) => tracing::debug!(?label, count, elapsed = ?start.elapsed(), "stored order events"),
        Err(err) => tracing::warn!(?label, count, ?err, "failed to insert order events"),
    }
}
```

MartinquaXD (approved):

squadgazzz (approved):
LG

PR #2618, opened 2024-04-14
Delete more solver code

MartinquaXD:
# Description
Deletes code that is no longer used outside of the `solver` crate.
Also reduces the visibility of some items to help `clippy` find unused items.


crates/driver/src/boundary/settlement.rs
```diff
@@ -171,12 +169,18 @@ impl Settlement {
...
             settlement::Internalization::Disable => InternalizationStrategy::EncodeAllInteractions,
         });
-        let builder = settle_method_builder(
-            contract,
-            encoded_settlement,
-            ethcontract::Account::Local(self.solver.into(), None),
-        );
-        let tx = builder.into_inner();
+
+        let account = ethcontract::Account::Local(self.solver.into(), None);
```
MartinquaXD, line 173: Moved from `settlement_simulation` module since this was the only code that was still in use.

Mateo-mro (approved):
LGTM, nice clean up!

PR #2620, opened 2024-04-15
Include necessary data for simulations within the auction json sent to solvers

### LINKED ISSUE #2606: feat: Include necessary data for simulations within the auction json sent to solvers
Currently, solvers need to use the /auction endpoint to recover missing info, such as hooks, for orders in a batch auction in order to be able to simulate their proposed solution themselves. We could instead ensure that all necessary info is included in the auction json.

Mateo-mro:
# Description
Currently, solvers need to use the /auction endpoint to recover missing info, such as hooks, for orders in a batch auction in order to be able to simulate their proposed solution themselves. We could instead ensure that all necessary info is included in the auction json.

# Changes
Data added to the auction json:
- all values signed as part of the order data: buy_amount, sell_amount, valid_to, owner, receiver
- pre- and post-hooks

## How to test
1. e2e tests
2. regression tests

## Related Issues
Fixes #2606


Mateo-mro:
I need to open a PR for the gnosis solver too.

crates/solvers-dto/Cargo.toml
```diff
@@ -6,10 +6,11 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
+model = { path = "../model" }
```
Mateo-mro, line 9: I don't know if this is the best way. But all the `Signature` type is huge, and it would be a lot of functions and structs to copy. Maybe having a crate for the `Signature`?

crates/solvers-dto/src/auction.rs
```diff
@@ -33,11 +33,22 @@ pub struct Order {
...
+    #[serde_as(as = "HexOrDecimalU256")]
+    pub full_buy_amount: U256,
+    pub fee_policies: Option<Vec<FeePolicy>>,
+    pub valid_to: u32,
     pub kind: Kind,
+    pub receiver: Option<H160>,
+    pub owner: H160,
     pub partially_fillable: bool,
+    pub pre_interactions: Vec<model::interaction::InteractionData>,
+    pub post_interactions: Vec<model::interaction::InteractionData>,
```
MartinquaXD, line 48: This is still missing `app_data`, fee_amount` (😱), `sell_token_source`, `buy_token_destination`.

crates/solvers-dto/src/auction.rs
```diff
@@ -33,11 +33,22 @@ pub struct Order {
...
     pub class: Class,
-    pub fee_policies: Option<Vec<FeePolicy>>,
+    #[serde(flatten)]
+    pub signature: model::signature::Signature,
```
MartinquaXD, line 51: Since we only need to serialize the data here I think it's fine to not import the `model` crate here and instead copy the few signature related types.

crates/solvers/src/tests/baseline/bal_liquidity.rs
```diff
@@ -53,11 +53,19 @@ async fn weighted() {
...
+                    "validTo": 0,
                     "kind": "sell",
+                    "owner": "0x5b1e2c2762667331bc91648052f646d1b0d35984",
                     "partiallyFillable": false,
+                    "preInteractions": [],
+                    "postInteractions": [],
                     "class": "market",
-                    "feePolicies": [],
+                    "signingScheme": "presign",
+                    "signature": "0x",
```
MartinquaXD, line 68: Quasilabs stumbled over this previously. Technically there is no additional signature associated with the order but in the SC presign order will use the owner address as the signature. Might be nice to just pre-fill this accordingly.

crates/solvers-dto/src/auction.rs
Mateo-mro: @MartinquaXD it is not "few" types. It is all the deserialization, with `SignatureSchema` and everything. I did copy it and it was horrible. I am totally against it.

crates/solvers-dto/src/auction.rs
Mateo-mro: aha, so everything is needed? got it

Mateo-mro: btw what is `buy_token_destination`, don't you mean `buy_token_balance`?

MartinquaXD: Yeah, in the old code it's called `buy_token_balance` but in the new code `buy_token_destination` IIRC. `buy_token_destination` makes more sense IMO.

crates/solvers/src/tests/baseline/bal_liquidity.rs
MartinquaXD: Yes, if we do that we should also update the orderbook format. We can leave the old format for this PR and discuss the change further but overall I like the concept that whatever is in the `signature` field is what you have to encode on-chain as well (even if it's technically duplicate data like for pre-sign).

crates/solvers-dto/src/auction.rs
MartinquaXD: > It is all the deserialization

But we don't really have to deserialize the signatures into anything fancy.
To me it seems like we actually just need:
```rust
#[derive(Serialize, Deserialize)]
enum SigningScheme {
    Eip712,
    EthSign,
    Eip1271,
    PreSign,
}
```

And the struct we send over the wire would be:
```
```

Mateo-mro: ah, so you mean to do it like that, ok ok. Won't it be needed in the solvers the whole `Signature` functionality in the future? :thinking:

crates/solvers/src/tests/baseline/bal_liquidity.rs
Mateo-mro: I completely understand. I created an issue in order to not forget and discuss it further: https://github.com/cowprotocol/services/issues/2626

crates/solvers-dto/src/auction.rs
MartinquaXD: So far none of the solvers we have to maintain are using the signature at all (or any other field you are adding in this PR). It's exclusively external solvers that use those and they have to do their own deserialization anyway. Also I think even for their purposes this simple version should be good enough given that this is how they currently "consume" the signature from the `/auction` endpoint.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -28,6 +30,7 @@ pub struct Order {
     pub kind: Kind,
     pub app_data: AppData,
     pub partial: Partial,
+    pub executed: U256,
```
MartinquaXD, line 33: 1. this information is already covered by `partial`
2. if this were to be added it should use the new types like `eth::TokenAmount`.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -333,6 +336,26 @@ pub enum SellTokenBalance {
...
+    fn from(value: SellTokenBalance) -> Self {
+        match value {
+            SellTokenBalance::Erc20 => Self::Erc20,
+            SellTokenBalance::Internal => Self::Internal,
+            SellTokenBalance::External => Self::External,
+        }
+    }
+}
+
+impl From<SellTokenSource> for SellTokenBalance {
```
MartinquaXD, line 349: Not needed.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -352,6 +375,24 @@ pub enum BuyTokenBalance {
...
+impl From<BuyTokenBalance> for BuyTokenDestination {
+    fn from(value: BuyTokenBalance) -> Self {
+        match value {
+            BuyTokenBalance::Erc20 => Self::Erc20,
+            BuyTokenBalance::Internal => Self::Internal,
+        }
+    }
+}
+
+impl From<BuyTokenDestination> for BuyTokenBalance {
```
MartinquaXD, line 387: Not needed.

crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -15,7 +15,14 @@ use {
             serialize,
         },
     },
+    app_data::AppDataHash,
     indexmap::IndexMap,
+    model::{
+        interaction::InteractionData,
+        order::{BuyTokenDestination, SellTokenSource},
+        signature::Signature,
```
MartinquaXD, line 23: What happened to not using the `model` crate here? Were there issues with that after all?

crates/driver/src/tests/cases/internalization.rs
```diff
@@ -29,7 +38,10 @@ async fn valid_internalization() {
 async fn untrusted_internalization() {
     let test = setup()
         .pool(ab_pool())
-        .order(ab_order().internalize())
+        .order(ab_order()
+            .owner(eth::H160::from_str(TRADER_ADDRESS).unwrap())
```
MartinquaXD, line 42: Since the testing framework seems to assume the same owner for most orders anyway (otherwise we probably wouldn't have a constant for this) I think making this the default and not having to add `.owner()` everywhere would be nicer.
When it's needed `.owner()` could still be used to overwrite the default owner, though.

crates/driver/src/domain/competition/order/mod.rs
Mateo-mro: Ok, being that the case, I'll remove it! Thanks.

crates/driver/src/infra/solver/dto/auction.rs
Mateo-mro: No issues, just I thought it was needed to be removed only from `solvers-dto` create. Removing it now, no probs.

crates/solvers-dto/src/auction.rs
```diff
@@ -33,11 +34,69 @@ pub struct Order {
...
+    pub full_buy_amount: U256,
+    pub fee_policies: Option<Vec<FeePolicy>>,
+    pub valid_to: u32,
     pub kind: Kind,
+    pub receiver: Option<H160>,
+    pub owner: H160,
     pub partially_fillable: bool,
+    pub pre_interactions: Vec<InteractionData>,
+    pub post_interactions: Vec<InteractionData>,
+    pub sell_token_balance: SellTokenSource,
```
MartinquaXD, line 50: There are still a few cases of `sell_token_balance` instead of `sell_token_source`.

crates/driver/src/tests/cases/internalization.rs
```diff
@@ -10,7 +10,9 @@ use crate::tests::{
 async fn valid_internalization() {
     let test = setup()
         .pool(ab_pool())
-        .order(ab_order().internalize())
+        .order(ab_order()
```
MartinquaXD, line 13: nit: to make the diff smaller and slightly easier to review it would be nice if these changes that are no longer necessary could be reverted.
* formatting (not sure why clippy is seemingly fine with both versions)
* referring to `base_order` instead of `ab_order()`.

MartinquaXD (commented):
Just minor nits remaining

MartinquaXD (approved):
All my comments have been addressed.
@cowprotocol/backend could somebody else please give this a review with a fresh set of eyes.

squadgazzz (approved):
LGTM

Issue #2627, opened 2024-04-17
chore: reduce time spent in synchronous code


MartinquaXD:
### Background
Since we are routing almost all the quotes through the driver we are seeing a few performance issues.
Currently my best guess is that we are spending a bunch of time in synchronous code which can clog the async runtime somewhat.
I built a very crude test setup which has all the major liquidity sources enabled and a colocated baseline solver as a quoter. Then I built a script that sends USDC -> USDT quote requests to the API with random sell amounts (to avoid caching).
I ran the script for 5 minutes and generated flamegraphs to show which optimizations could the most impactful.

I identified 3 major issues:
#### Creating contract instances using `ethcontract-rs`
To call contract methods we currently have to instantiate contract instances. This is done with `Contract::at(web3, address)` and can be extremely costly. The reason is that although there is 0 dynamic data associate with correctly calling a contract method `ethcontract-rs` relies on runtime values a lot. And creating those runtime values (contract instances) is very costly for contracts with many methods.

Ideally we would replace `ethcontract-rs` with something like `alloy` for calling contract methods in a type safe manor because `alloy` does not involve any runtime logic.

#### Converting from old types to new types
In the driver we currently still have the `boundary` layer which converts legacy cold to types which fit nicely into our new domain layer. This conversion is pure overhead since eventually the old legacy code should not exist anymore. My tests showed that converting the uniswapv3 liquidity is the most costly so that is the one we should convert first.

#### Serializing requests
Currently we send liquidity to solvers which rely on that. For auctions this is not a huge deal since auctions happen relatively rarely but quotes can happen 20 times per second so the serialization overhead hurts way more for those requests.
The most basic solution would probably be to introduce a new serialization format which is more performant but this obviously a huge breaking change so it would have to be opt-in but at least for the baseline solver it could be a huge gain.
This change obviously needs to be discussed extensively but could result in very significant performance improvements.

![flamegraph_original](https://github.com/cowprotocol/services/assets/19190235/2df8df41-58c1-4efd-811b-c14ecd0bb169)

PR #2628, opened 2024-04-17
Reuse contract instance for better performance

MartinquaXD:
# Description
First task of https://github.com/cowprotocol/services/issues/2627

Creating contract instances with `ethcontract-rs` is very costly (especially in tight loops).
A very significant amount of time is spent creating instances to query allowances.

# Changes
Instead of creating 1 contract instance per token to fetch balances for we now create a single erc20 instance and reuse it for all `allowance()` calls. This is unfortunately pretty ugly since `ethcontract-rs` doesn't really expose the generated calldata very nicely and doesn't really compose with `web3`.

## How to test
Correctness should be covered by e2e tests
Performance was verified using the benchmark mentioned in https://github.com/cowprotocol/services/issues/2627

It's a bit hard to see but fetching allowances went from ~20% to ~7.6% of the entire runtime.
The overall runtime of the axum request handler went from ~74% -> 63% of the runtime.
Overall I think the performance gains are significant enough to warrant the slightly ugly but very local performance hack until we replace `ethcontract-rs` with `alloy` entirely.

Before
Screenshot 2024-04-17 at 21 32 46

Optimized
Screenshot 2024-04-17 at 21 28 35

### DIFF
--- a/crates/solver/src/interactions/allowances.rs
+++ b/crates/solver/src/interactions/allowances.rs
@@ -389,27 +389,21 @@ mod tests {
         let web3 = mock::web3();
         web3.transport()
             .mock()
-            .expect_execute_batch()
-            .returning(move |calls| {
-                Ok(calls
-                    .into_iter()
-                    .map(|(method, params)| {
-                        assert_eq!(method, "eth_call");
+            .expect_execute()
+            .returning(move |method, params| {
+                assert_eq!(method, "eth_call");
 
-                        let call =
-                            serde_json::from_value::<CallRequest>(params[0].clone()).unwrap();
-                        assert_eq!(call.data.unwrap(), allowance_call_data(owner, spender));
+                let call = serde_json::from_value::<CallRequest>(params[0].clone()).unwrap();
+                assert_eq!(call.data.unwrap(), allowance_call_data(owner, spender));
+                let to = call.to.unwrap();
 
-                        let to = call.to.unwrap();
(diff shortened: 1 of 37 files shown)


crates/solver/src/interactions/allowances.rs
```diff
@@ -397,27 +389,21 @@ mod tests {
...
-                        if to == addr!("1111111111111111111111111111111111111111") {
-                            Ok(allowance_return_data(1337.into()))
-                        } else if to == addr!("2222222222222222222222222222222222222222") {
-                            Err(web3::Error::Decoder("test error".to_string()))
-                        } else {
-                            panic!("call to unexpected token {to:?}")
-                        }
-                    })
-                    .collect())
+            .expect_execute()
```
MartinquaXD, line 392: The web3 instance we use for real auto batches requests under the hood such that all requests can be made with `eth_call` instead.

crates/solver/src/interactions/allowances.rs
```diff
@@ -440,24 +426,4 @@ mod tests {
             },
         );
     }
-
-    #[tokio::test]
-    async fn fetch_fails_on_batch_errors() {
```
MartinquaXD, line 445: Since our web3 instance auto-batches under the hood the new code is no longer aware of any special cases related to batching so this test can be dropped.

Mateo-mro (approved):

sunce86 (approved):
lg

squadgazzz (approved):

PR #2629, opened 2024-04-17
Make UniswapV3 state clones cheap

MartinquaXD:
# Description
Uniswap V3 liquidity contains a lot of state and currently we deep clone the entire state of a pool that is needed as liquidity for a quote or an auction.

# Changes
Store univ3 pool state in an Arc. That way cloning becomes dirt cheap. The only slight annoyance is that updating the cached state becomes a bit weird.
Arcs can be tricky to update because theoretically any number of threads might have access to the same memory. Luckily `Arc::make_mut()` is a thing. That allows us to get mutable access to the Arc if we are the only one using it at the time or it will clone the content and give us mutable access to that. That way updating the pool state at least stays very cheap whenever it can.

## How to test
Correctness is covered by existing unit tests
Performance was measured with a local benchmark.
As you can see in the images the `partition_map` which cloned the relevant uni v3 pools is basically gone now.

Before
Screenshot 2024-04-21 at 20 26 37

After
Screenshot 2024-04-21 at 20 26 45

### DIFF
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -397,11 +397,18 @@ fn append_events(pools: &mut HashMap<H160, Arc<PoolInfo>>, events: Vec<Event<Uni
             .address;
 
         if let Entry::Occupied(mut entry) = pools.entry(address) {
+            let old_allocation = Arc::as_ptr(entry.get());
+
             // Clones and updates the current state if something still has strong references
             // to that Arc otherwise it will just mutate the original in place.
             let current = Arc::make_mut(entry.get_mut());
-            let pool = &mut current.state;
 
+            match old_allocation == current as *const _ {
+                true => metrics().uniswapv3_pool_updates_fast.inc(),
+                false => metrics().uniswapv3_pool_updates_slow.inc(),
+            }
+
+            let pool = &mut current.state;
             match event.data {
                 UniswapV3Event::Burn(burn) => {
                     let tick_lower = BigInt::from(burn.tick_lower);
@@ -494,6 +501,19 @@ impl Maintaining for UniswapV3PoolFetcher {
     fn name(&self) -> &str {
         "UniswapV3PoolFetcher"
     }
+}
+
+#[derive(prometheus_metric_storage::MetricStorage, Debug)]
+struct Metrics {
+    /// Counts how often uniswap v3 pools got updates with the fast path
+    uniswapv3_pool_updates_fast: prometheus::IntCounter,
+    /// Counts how often uniswap v3 pools got updates with the slow path
+    uniswapv3_pool_updates_slow: prometheus::IntCounter,
+}
+
+fn metrics() -> &'static Metrics {
+    Metrics::instance(observe::metrics::get_storage_registry())
+        .expect("unexpected error getting metrics instance")
 }
 
 #[cfg(test)]
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -397,16 +397,9 @@ fn append_events(pools: &mut HashMap<H160, Arc<PoolInfo>>, events: Vec<Event<Uni
             .address;
 
         if let Entry::Occupied(mut entry) = pools.entry(address) {
-            // Because contents of `Arc`s can't be modified we:
-            // 1. swap out the current state with a dummy
-            // 2. update current state
-            // 3. put updated state into `Arc` and swap back into the cache
-            let mut current = {
-                let mut dummy = Arc::default();
-                std::mem::swap(entry.get_mut(), &mut dummy);
-                // use unwrap_or_clone to make updates cheaper when nobody is using the state
-                Arc::unwrap_or_clone(dummy)
-            };
+            // Clones and updates the current state if something still has strong references
+            // to that Arc otherwise it will just mutate the original in place.
+            let current = Arc::make_mut(entry.get_mut());
             let pool = &mut current.state;
 
             match event.data {
@@ -474,9 +467,6 @@ fn append_events(pools: &mut HashMap<H160, Arc<PoolInfo>>, events: Vec<Event<Uni
                     pool.sqrt_price = swap.sqrt_price_x96;
                 }
             }
-
-            // `current` has now been updated and can be swapped back into the cache
-            std::mem::swap(entry.get_mut(), &mut Arc::new(current));
         }
     }
 }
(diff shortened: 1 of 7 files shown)


crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -388,13 +389,26 @@ impl PoolFetching for UniswapV3PoolFetcher {
...
             .expect("metadata must exist for mined blocks")
             .address;
-        if let Some(pool) = pools.get_mut(&address).map(|pool| &mut pool.state) {
+
+        if let Entry::Occupied(mut entry) = pools.entry(address) {
+            // Because contents of `Arc`s can't be modified we:
+            // 1. swap out the current state with a dummy
+            // 2. update current state
+            // 3. put updated state into `Arc` and swap back into the cache
+            let mut current = {
```
Mateo-mro, line 404: I see the caller of this functions is an `async` function. Why not to use a tokio `Mutex`, `RwLock`, `parking_lot` or other solution to do this write? `Arc` can be accessed by multiple readers, and create a race conditions. You can argue that you are very sure this is not accessed multiple times by many readers, but then it forces new developers or developers who don't know this to be aware of it, if the functionality were to be extended (let's say we add another parallel function for the maintenance of Uniswap, etc)

MartinquaXD: > Arc can be accessed by multiple readers, and create a race conditions.

Arc can be accessed by multiple readers but it can not be modified which is why you can't cause race conditions with it. That is precisely the reason I have to jump through these hoops. The only thing that can happen is that:
1. somebody fetches liquidity (by cloning the pool)
2. I update the cache
3. they don't see the update
However, this is already the case so in that sense from the perspective of an outside observer nothing about this function changed (Except that it's now returning the data inside an `Arc`).

> You can argue that you are very sure this is not accessed multiple times by many readers, but then it forces new developers or developers who don't know this to be aware of it

No, other developers don't have to be careful about anything here.
The state I'm modifying is guaranteed to be accessed only by me during the update.
First of all we get an exclusive reference to the HashMap so we know for sure nobody else is modifying that map at the same time.
Secondly because I use `Arc::unwrap_or_clone()` either nobody else is holding a copy of that particular `Arc` (then it's safe to unwrap) or somebody is currently using it then we simply clone and modify the cloned version. Modifying the cloned version is also safe because nobody can get access to it until I'm done with it and put it back into the map.

> Why not to use a tokio Mutex, RwLock, parking_lot or other solution to do this write?

They don't really help here. The thing I'm trying to solve is expensive copies of the pool state. That can only be solved with some sort of reference counting. So the only thing I could be doing is to use `Mutex<Arc<Data>>`. However, since we already have an exclusive access to the map the `Mutex` is not needed to protect against race conditions so it would just be overhead.

Mateo-mro: > Arc can be accessed by multiple readers but it can not be modified which is why you can't cause race conditions with it.

This is not correct, you are swapping the memory of the `Arc`, therefore modifying its content in order to modify it, and substituting the `Arc` pointing to a `dummy` which is not even containing an object the same structure as `PoolInfo`. Making any access by another party to that `Arc` extremely dangeous.

> The state I'm modifying is guaranteed to be accessed only by me during the update.
First of all we get an exclusive reference to the HashMap so we know for sure nobody else is modifying that map at the same time.

This is partially true. Because you are passing `&mut HashMap<>` doesn't mean that someone has references of the `Arc`s inside the `HashMap` (that's the whole point of the `Arc` after all). Indeed you are making the `HashMap` mutable only for your function, but the access to the `Arc` is still shared among everyone who wants to clone it.

> Secondly because I use Arc::unwrap_or_clone() either nobody else is holding a copy of that particular Arc (then it's safe to unwrap) or somebody is currently using it then we simply clone and modify the cloned version. Modifying the cloned version is also safe because nobody can get access to it until I'm done with it and put it back into the map.

This is true, but this only guarantees you that the data you modify is either not used by someone else, or in that case, it is clone. But we can have this scenario:

Someone else clones the `Arc` before that function. If another tokio thread/task cloned the entry Arc before this code executes, it retains a separate reference to the original data. Swapping out the contents of the entry Arc with a dummy value does not affect this separate reference. If the other thread concurrently accesses or modifies the data through its reference to the entry `Arc`, it can lead to a race condition. This can definitely result in unpredictable behavior, data corruption, or crashes.

> They don't really help here. The thing I'm trying to solve is expensive copies of the pool state. That can only be solved with some sort of reference counting. So the only thing I could be doing is to use Mutex<Arc<Data>>. However, since we already have an exclusive access to the map the Mutex is not needed to protect against race conditions so it would just be overhead.

It should be the opposite `Arc<Mutex<>>` :thinking:  But @MartinquaXD , we have exclusive access to the map, but we don't have exclusive access to the `Arc` contained in the map :pray:

Mateo-mro (approved):

crates/shared/src/sources/uniswap_v3/pool_fetching.rs
MartinquaXD: We resolved the race condition concerns in a call. It's correct that using `RwLock` around each pool might be faster overall depending on how often we actually try to update the state while somebody else still has access to the data. I will add metrics to see how often that actually happens. The more often it happens the more likely using `RwLock` would be the right choice here.

squadgazzz (approved):
LGTM

sunce86 (approved):

MartinquaXD:
While looking further into this I noticed that we are actually never running the maintenance for the uni v3 pools in the driver (i.e. their state never gets updated). After fixing this it indeed turned out that using `Arc<RwLock<Pool>>` is indeed faster than just `Arc<Pool>`.
Will update the PR accordingly.

crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -169,7 +192,7 @@ impl PoolsCheckpointHandler {
             .get_pools_with_ticks_by_ids(&pool_ids, registered_pools.fetched_block_number)
             .await?
             .into_iter()
-            .filter_map(|pool| Some((pool.id, pool.try_into().ok()?)))
+            .filter_map(|pool| Some((pool.id, PoolInfo::try_from(pool).ok()?.into())))
```
Mateo-mro, line 195: Why not to implement `TryFrom` for `PoolInfoHandler` directly instead of doing two conversions? same applies to [L269](https://github.com/cowprotocol/services/pull/2629/files#diff-8a0be017c847413b5a422604316ed606c59b2294e551106ea4aea93c95bde758R269).

crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -491,229 +520,233 @@ impl Maintaining for UniswapV3PoolFetcher {
...
-        assert_eq!(
-            pools[&address].state.liquidity_net,
-            BTreeMap::from([
-                (BigInt::from(100_000), BigInt::from(12345)),
-                (BigInt::from(105_000), BigInt::from(54321)),
-                (BigInt::from(110_000), BigInt::from(-66666))
-            ])
-        );
-    }
+    // use {
```
Mateo-mro, line 523: why all these tests have been commented?

PR #2638, opened 2024-04-19
Bump crates to the latest versions

Mateo-mro:
# Description
Bump crates to the latest versions

# Changes
- Bump crates to the latest versions
- Unify in the `workspace` all the crates which are used in at least twice

## How to test
1. Regression tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,46 +3,53 @@ resolver = "2"
 members = ["crates/*"]
 
 [workspace.dependencies]
-anyhow = "1"
-async-trait = "0.1"
+anyhow = "1.0.82"
+async-trait = "0.1.80"
+axum = "0.6"
 bigdecimal = "0.3"
-cached = { version = "0.44", default-features = false }
-chrono = { version = "0.4", default-features = false }
-clap = { version = "4", features = ["derive", "env"] }
-derivative = "2"
+cached = { version = "0.49.3", default-features = false }
+chrono = { version = "0.4.38", default-features = false }
+clap = { version = "4.5.4", features = ["derive", "env"] }
+derivative = "2.2.0"
 ethcontract = { version = "0.25.6", default-features = false, features = ["aws-kms"] }
 ethcontract-generate = { version = "0.25.6", default-features = false }
 ethcontract-mock = { version = "0.25.6", default-features = false }
-ethereum-types = "0.14"
-futures = "0.3"
+ethereum-types = "0.14.1"
+flate2 = "1.0.28"
+futures = "0.3.30"
 gas-estimation = { git = "https://github.com/cowprotocol/gas-estimation", tag = "v0.7.3", features = ["web3_", "tokio_"] }
-hex = { version = "0.4", default-features = false }
-hex-literal = "0.4"
+hex = { version = "0.4.3", default-features = false }
+hex-literal = "0.4.1"
 humantime = "2.1.0"
 humantime-serde = "1.1.1"
+hyper = "0.14.28"
 indexmap = "2.2.5"
-itertools = "0.11"
-lazy_static = "1"
-maplit = "1"
-mockall = "0.11"
-num = "0.4"
-once_cell = "1"
+itertools = "0.12.1"
+lazy_static = "1.4.0"
+maplit = "1.0.2"
+mockall = "0.12.1"
+num = "0.4.2"
+once_cell = "1.19.0"
 primitive-types = "0.12"
-prometheus = "0.13"
+prometheus = "0.13.3"
 prometheus-metric-storage = "0.5.0"
-rand = "0.8"
-regex = "1"
-reqwest = "0.11"
-secp256k1 = "0.27"
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
-serde_with = "3"
+rand = "0.8.5"
+regex = "1.10.4"
+reqwest = "0.11.27"
+secp256k1 = "0.27.0"
+serde = { version = "1.0.198", features = ["derive"] }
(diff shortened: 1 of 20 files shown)


Cargo.toml
```diff
@@ -3,46 +3,53 @@ resolver = "2"
...
-regex = "1"
-reqwest = "0.11"
-secp256k1 = "0.27"
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
-serde_with = "3"
+rand = "0.8.5"
+regex = "1.10.4"
+reqwest = "0.11.27"
+secp256k1 = "0.27.0"
```
Mateo-mro, line 39: We cannot update to `0.29.0` until `web3` crate is not updated to the aforementioned version.

Cargo.toml
```diff
@@ -3,46 +3,53 @@ resolver = "2"
 members = ["crates/*"]
 
 [workspace.dependencies]
-anyhow = "1"
-async-trait = "0.1"
+anyhow = "1.0.82"
+async-trait = "0.1.80"
+axum = "0.6"
```
Mateo-mro, line 8: Updating `axum` is a rabbit hole, but we need to do it sooner or later. The reason is that in order to update `axum` we need to update altogether:
- `hyper`
- every `http` crate
- `reqwest`
- `warp` (we are using our own forked version 3 years old...)

I'd recommend to fully delete `warp` dependency (latest versions of `axum` basically do everything which `warp` does anyway). And then update `axum` + `hyper` which both have decent breaking changes.

What do you guys think? should I create a task for it since it will be tricky to do?

I need to do this sooner or later, because we cannot update any of those crates.

Cargo.toml
```diff
@@ -3,46 +3,53 @@ resolver = "2"
...
 bigdecimal = "0.3"
```
Mateo-mro, line 9: We cannot update `bigdecimal` so easily, they removed the `sqlx` immediate support, very likely it is solved by just converting it into a primitive, but it needs more research to really check this, otherwise we can make break our database. What's more, this crate update is really not that needed at the moment.

crates/database/src/orders.rs
```diff
@@ -1248,7 +1248,7 @@ mod tests {
         assert_eq!(time, order.cancellation_timestamp.unwrap());
 
         // Cancel again and verify that cancellation timestamp was not changed.
-        let irrelevant_time = Utc.timestamp_opt(1234567890, 1_000_000_000).unwrap();
+        let irrelevant_time = Utc.timestamp_opt(1234564319, 1_000_000_000).unwrap();
```
Mateo-mro, line 1251: From `chrono`:

```
The nanosecond part can exceed 1,000,000,000 in order to represent a leap second, but only when secs % 60 == 59. (The true "UNIX timestamp" cannot represent a leap second unambiguously.)
```

So I needed to change the main timestamp to have as the second part `% 60 = 59`

Cargo.toml
Mateo-mro: The PR for the crate `ethceontract-rs` : https://github.com/cowprotocol/ethcontract-rs/pull/969

Cargo.toml
```diff
@@ -3,46 +3,53 @@ resolver = "2"
...
-cached = { version = "0.44", default-features = false }
-chrono = { version = "0.4", default-features = false }
-clap = { version = "4", features = ["derive", "env"] }
-derivative = "2"
-ethcontract = { version = "0.25.4", default-features = false, features = ["aws-kms"] }
+cached = { version = "0.49.3", default-features = false }
+chrono = { version = "0.4.38", default-features = false }
+clap = { version = "4.5.4", features = ["derive", "env"] }
+derivative = "2.2.0"
+ethcontract = { version = "0.25.5", default-features = false, features = ["aws-kms"] }
```
MartinquaXD: Needs to be updated to 0.25.6 by now.

crates/driver/Cargo.toml
```diff
@@ -16,36 +16,36 @@ path = "src/main.rs"
...
 tap = "1.0.1"
-thiserror = "1.0"
-tokio = { version = "1.22", features = ["macros", "rt-multi-thread", "signal", "time"] }
-toml = "0.7"
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal", "time"] }
+toml = { workspace = true }
 tower = "0.4"
 tower-http = { version = "0.4", features = ["limit", "trace"] }
 url = { version = "2.3", features = ["serde"] }
```
MartinquaXD, line 51: url is also used in more than 1 instance, right?

crates/driver/Cargo.toml
```diff
@@ -16,36 +16,36 @@ path = "src/main.rs"
...
-hex-literal = "0.4"
+futures = { workspace = true }
+hex = { workspace = true }
+hex-literal = { workspace = true }
 humantime = { workspace = true }
 humantime-serde = { workspace = true }
-hyper = "0.14"
+hyper = { workspace = true }
 lazy_static = { workspace = true }
 indexmap = { version = "2", features = ["serde"] }
```
MartinquaXD, line 34: could also use the workspace

MartinquaXD (commented):
Nice initiative.
Looks good overall just a bit concerned with updating the AWS dependencies. Could you please do a test run for those. I think none of that is covered by any integration tests.

MartinquaXD (approved):
Since testing the AWS change it pretty elaborate without merging this PR we agreed in DMs that testing in staging (by settling a test trade) and reverting if it doesn't work is fine.

squadgazzz (approved):

PR #2643, opened 2024-04-22
0x liquidity cache improvements

squadgazzz:
# Description
Addresses Martin's suggestions: [1](https://github.com/cowprotocol/services/pull/2639#discussion_r1574214480), [2](https://github.com/cowprotocol/services/pull/2639#discussion_r1574210427)

## Changes

- `RwLock` is replaced with `arc_swap::ArcSwap` to avoid blocking.
- `OrderRecord` fields are Arc'ed.
- Use already grouped `OrderBuckets` in the cache to avoid redundant computation on each `get_liquidity` function call.

## How to test
Existing tests.


crates/solver/src/liquidity/zeroex.rs
```diff
@@ -48,7 +48,7 @@ impl ZeroExLiquidity {
     ) -> Self {
         let gpv2_address = gpv2.address();
         let allowance_manager = AllowanceManager::new(web3, gpv2_address);
-        let orderbook_cache = Arc::new(RwLock::new(vec![]));
+        let orderbook_cache = Arc::new(ArcSwap::from_pointee(Vec::new()));
```
MartinquaXD, line 51: Using `Default::default()` somewhere here should work.
```suggestion
        let orderbook_cache = Default::default();
```

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -136,8 +135,8 @@ impl LiquidityCollecting for ZeroExLiquidity {
         pairs: HashSet<TokenPair>,
         _block: Block,
     ) -> Result<Vec<Liquidity>> {
-        let zeroex_orders = self.orderbook_cache.read().await.clone();
-        let order_buckets = generate_order_buckets(zeroex_orders, pairs);
+        let zeroex_orders = self.orderbook_cache.load().clone();
+        let order_buckets = generate_order_buckets((*zeroex_orders).clone(), pairs);
         let filtered_zeroex_orders = get_useful_orders(order_buckets, 5);
```
MartinquaXD, line 140: I just noticed now that we are not getting a fresh set of `OrderRecord`s for every request `generate_order_buckets()` and `get_useful_orders() should be rewritten such that they only operate on references and only clone the values that will actually get returned.

crates/shared/src/zeroex_api.rs
```diff
@@ -260,8 +261,8 @@ pub struct Order {
 #[derive(Debug, Default, Clone, Deserialize, Serialize, Eq, PartialEq)]
 pub struct OrderRecord {
     #[serde(rename = "metaData")]
-    pub metadata: OrderMetadata,
-    pub order: Order,
+    pub metadata: Arc<OrderMetadata>,
+    pub order: Arc<Order>,
```
MartinquaXD, line 265: These struct only make sense together so they should be stored inside the same `Arc`.

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -68,24 +68,24 @@ impl ZeroExLiquidity {
...
+            buy_token: record.order().taker_token,
             sell_amount,
-            buy_amount: record.metadata.remaining_fillable_taker_amount.into(),
+            buy_amount: record.metadata().remaining_fillable_taker_amount.into(),
             kind: OrderKind::Buy,
             partially_fillable: true,
             user_fee: U256::zero(),
             settlement_handling: Arc::new(OrderSettlementHandler {
-                order: record.order,
+                order: record.order().clone(),
```
MartinquaXD, line 88: Now we should store the entire `OrderRecord` in the `OrderSettlementHandler`.

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -159,33 +155,37 @@ impl LiquidityCollecting for ZeroExLiquidity {
...
+            })
+        })
+        .into_group_map()
 }
 
 /// Get the `orders_per_type` best priced and biggest volume orders.
-fn get_useful_orders(order_buckets: OrderBuckets, orders_per_type: usize) -> Vec<OrderRecord> {
+fn get_useful_orders(
+    order_buckets: &OrderBuckets,
+    relevant_pairs: HashSet<TokenPair>,
```
MartinquaXD, line 176:
```suggestion
    relevant_pairs: &HashSet<TokenPair>,
```

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -159,33 +155,37 @@ impl LiquidityCollecting for ZeroExLiquidity {
...
+    orders_per_type: usize,
+) -> Vec<OrderRecord> {
     let mut filtered_zeroex_orders = vec![];
-    for mut orders in order_buckets.into_values() {
+    for orders in order_buckets
+        .iter()
+        .filter_map(|((token_a, token_b), record)| {
+            TokenPair::new(*token_a, *token_b)
+                .is_some_and(|pair| relevant_pairs.contains(&pair))
+                .then_some(record)
```
MartinquaXD, line 185: The relevant pairs (<=13) will likely be fewer than all the buckets so iterating over the relevant pairs and indexing into the `order_buckets` should be faster and nicer to read.
This also shows that the key of the buckets should now probably be `TokenPair` instead of `(H160, H160)`.

MartinquaXD (commented):
Looks already very good. 👍

crates/solver/src/liquidity/zeroex.rs
squadgazzz: > This also shows that the key of the buckets should now probably be TokenPair instead of (H160, H160).

Well, the TokenPair is sorted inside. Previously the logic was the following:
```
        .filter(|record| {
            match TokenPair::new(record.order().taker_token, record.order().maker_token) {
                Some(pair) => relevant_pairs.contains(&pair),
                None => false,
            }
        })
        .for_each(|order| {
            let bucket = buckets
                .entry((order.order().taker_token, order.order().maker_token))
                .or_default();
            bucket.push(order);
        })
```

Initially, I used the TokenPair as a key but reverted the change since the tests started to fail.

crates/solver/src/liquidity/zeroex.rs
squadgazzz: Fixed

crates/solver/src/liquidity/zeroex.rs
squadgazzz: Fixed

crates/solver/src/liquidity/zeroex.rs
squadgazzz: IMO, altering the logic might be worth a separate PR.

MartinquaXD: Yeah, that's fair. 👌

MartinquaXD (approved):

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -194,21 +194,21 @@ fn get_useful_orders(order_buckets: OrderBuckets, orders_per_type: usize) -> Vec
...
+            let price_1 = order_1.order().maker_amount as f64 / order_1.order().taker_amount as f64;
+            let price_2 = order_2.order().maker_amount as f64 / order_2.order().taker_amount as f64;
             price_1.total_cmp(&price_2)
         });
         filtered_zeroex_orders.extend(orders.drain(orders.len() - orders_per_type..));
 
-        orders.sort_by_key(|order| order.metadata.remaining_fillable_taker_amount);
-        filtered_zeroex_orders.extend(orders.into_iter().rev().take(orders_per_type));
+        orders.sort_by_key(|order| order.metadata().remaining_fillable_taker_amount);
+        filtered_zeroex_orders.extend(orders.iter().rev().take(orders_per_type).cloned());
```
Mateo-mro, line 204: Why to do this change in this line? eve tho the clone is cheap, but there is no need for it :thinking:

Mateo-mro (approved):

crates/solver/src/liquidity/zeroex.rs
squadgazzz: Looks like a leftover from one of the previous states of the PR. Reverted, thanks.

PR #2645, opened 2024-04-22
Reduce cloning in 0x liquidity fetching

MartinquaXD:
# Description
Only clone `OrderRecord`s that actually get returned as the response instead of cloning everything and then discarding orders that don't matter.

## How to test
unit tests
CI

### DIFF
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -136,9 +136,11 @@ impl LiquidityCollecting for ZeroExLiquidity {
         pairs: HashSet<TokenPair>,
         _block: Block,
     ) -> Result<Vec<Liquidity>> {
-        let zeroex_orders = self.orderbook_cache.read().await;
-        let order_buckets = generate_order_buckets(zeroex_orders.as_slice(), pairs);
-        let filtered_zeroex_orders = get_useful_orders(order_buckets, 5);
+        let filtered_zeroex_orders = {
+            let zeroex_orders = self.orderbook_cache.read().await;
+            let order_buckets = generate_order_buckets(zeroex_orders.as_slice(), pairs);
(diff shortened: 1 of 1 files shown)


crates/solver/src/liquidity/zeroex.rs
```diff
@@ -36,7 +36,7 @@ pub struct ZeroExLiquidity {
     pub orderbook_cache: Arc<OrderbookCache>,
 }
 
-type OrderBuckets = HashMap<(H160, H160), Vec<OrderRecord>>;
+type OrderBuckets<'a> = HashMap<(H160, H160), Vec<&'a OrderRecord>>;
```
Mateo-mro, line 39: Why not to wrap it in an `Arc` like:

```rust
type OrderBuckets = HashMap<(H160, H160), Vec<Arc<OrderRecord>>>;
```

That way it is avoided the hard clones of the `OrderRecord`  here:
- [L203](https://github.com/cowprotocol/services/pull/2645/files#diff-9fc8dfc6f735cce917d647b7f913cc65d3d314a76f8a7e9f31214cabc661518cR203)
- [L206](https://github.com/cowprotocol/services/pull/2645/files#diff-9fc8dfc6f735cce917d647b7f913cc65d3d314a76f8a7e9f31214cabc661518cR206)

The signature of the function could change to;
```rust
fn get_useful_orders(order_buckets: OrderBuckets, orders_per_type: usize) -> Vec<Arc<OrderRecord>> {
```

That way we avoid making any copy of `OrderRecord` but always pointing to the reference in memory :thinking: (In case we want to fully avoid making hard cloning)

MartinquaXD: Yeah, brought this up already with @squadgazzz and he's currently working on that. Actually he came up with some other improvement which will make this PR obsolete so it can be closed.

PR #2649, opened 2024-04-23
[EASY] Remove deny_unknown_fields for auction order in solvers-dto

Mateo-mro:
# Description
Remove `deny_unknown_fields` for auction order in solvers-dto. This way new changes in the payload does not break the solvers. It is needed to merge https://github.com/cowprotocol/services/pull/2620 without having downtime for the solvers.

# Changes
Remove `deny_unknown_fields` for auction order in solvers-dto.

## How to test
1. Regression tests


crates/solvers-dto/src/auction.rs
```diff
@@ -24,7 +24,7 @@ pub struct Auction {
 
 #[serde_as]
 #[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
MartinquaXD, line 27: I think it even makes sense to remove `deny_unknown_fields` altogether given that any update to these fields became more painful when we migrated the solvers to the gnosis organization.

squadgazzz (approved):

MartinquaXD (approved):

PR #2650, opened 2024-04-23
Playground startup fix

squadgazzz:
Fixes playground execution and driver's config. The main problem was about the absence of container dependency where the driver failed to start while the anvil was not fully initialized.
More details are in the comments.


playground/docker-compose.fork.yml
```diff
@@ -127,6 +128,9 @@ services:
       - ./driver.toml:/driver.toml
     ports:
       - 9000:80
+    depends_on:
+      chain-check:
+        condition: service_healthy
```
squadgazzz, line 133: This condition is required since there is a time gap between the chain's container `created` status and when the anvil starts the service. This avoids adding a startup timeout to the driver container.

playground/docker-compose.fork.yml
```diff
@@ -163,5 +167,28 @@ services:
         - ETH_RPC_URL=http://127.0.0.1:8545
     ports:
       - 8001:80
+
+  chain-check:
+    image: curlimages/curl:latest
+    depends_on:
+      - chain
+    command: >
+      sh -c "while true; do
```
squadgazzz, line 176: I had to use `sh -c` explicitly because, for some reason, the container exists right after the startup with the explicitly provided entry point.

playground/driver.toml
```diff
@@ -26,6 +25,3 @@ base-tokens = [
 
 [[liquidity.uniswap-v2]] # Uniswap V2 configuration
 preset = "uniswap-v2"
-
-[[liquidity.balancer-v2]]
-preset = "balancer-v2"
```
squadgazzz, line 31: This is removed since a graph URL is required.

playground/docker-compose.fork.yml
```diff
@@ -123,10 +126,15 @@ services:
     environment:
       - ETHRPC=http://chain:8545
       - ADDR=0.0.0.0:80
+      - RUST_BACKTRACE=1
```
MartinquaXD, line 129: `RUST_BACKTRACE` can also be added to the orderbook, autopilot and solver.

playground/docker-compose.fork.yml
```diff
@@ -163,5 +171,31 @@ services:
         - ETH_RPC_URL=http://127.0.0.1:8545
     ports:
       - 8001:80
+
+  chain-check:
```
MartinquaXD, line 175: This whole section seems like it should rather be part of the chain directly.
Actually I'm a bit surprised that apparently it's possible to set the allow list code before anvil can return a block. But maybe the easiest solution would be to have a readiness check in the form of:
```
if no_current_block {
    return false;
}
anvil_setCode()
```

MartinquaXD (commented):
I see the need for synchronizing the startup sequence correctly but the `chain-check` seems out of place.
Also since this was broken for quite some time it would be cool if we could have a cronjob that tries to spin up the playground once per day and settles a trade or so.

playground/docker-compose.fork.yml
squadgazzz: Yeah, true. A separate service was the fastest option to test it. Replaced with an inner health check.

playground/docker-compose.fork.yml
MartinquaXD: We should keep `TOML_TRACE_ERROR` here. It makes it easier to debug when you have parsing issues with your config file.

MartinquaXD (approved):

mfw78 (approved):

PR #2652, opened 2024-04-23
Downgrade anyhow

Mateo-mro:
# Description
After a long session of debugging together, @MartinquaXD discovered that the latest anyhow is always printing the full trace.

# Changes
Revert back to anyhow 1.0.76

## How to test
1. regression tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ resolver = "2"
 members = ["crates/*"]
 
 [workspace.dependencies]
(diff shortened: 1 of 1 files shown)


MartinquaXD (approved):

MartinquaXD:
For more context the additional logging overhead caused us the pods to use so many more resources that even sepolia pods regularly got evicted because they ran out of memory.

PR #2656, opened 2024-04-24
Do not trust driver's settle result

squadgazzz:
Addresses a [suggestion](https://github.com/cowprotocol/services/pull/2630#pullrequestreview-2019761096) not trusting the driver's settle result in a colocated environment.


crates/autopilot/src/run_loop.rs
```diff
@@ -445,8 +445,17 @@ impl RunLoop {
         .await
         {
             futures::future::Either::Left((res, _)) => res,
-            futures::future::Either::Right((res, _)) => {
-                res.map_err(SettleError::Failure).map(|tx| tx.tx_hash)
+            futures::future::Either::Right((driver_result, onchain_task)) => {
+                let driver_tx_hash = driver_result.map_err(SettleError::Failure)?.tx_hash;
+                let onchain_tx_hash = onchain_task.await?;
+                if onchain_tx_hash != driver_tx_hash {
```
MartinquaXD, line 451: Completely overlooked this in the review of the original PR but my initial idea was to only allow solvers to report when they gave up submitting a tx on-chain and always take the tx hash from the autopilot. This PR would essentially accomplish the same with the additional overhead of making the solver report back the tx hash that will ultimately get ignored.

MartinquaXD:
@sunce86 I'm very suspicious about getting fewer details to associate an auction with onchain data. For example if we only associate an auction with a block we could run into issues with multiple tx associated to the same auction happening on the same block (might be relevant for combinatorial auction if want to go down that route). Also we should probably be able to detect when a block gets reorged and replaced with a block containing the same auction.
Overall I think allowing the driver to only report submission aborts but not successes is the safer route while requiring almost no changes.

crates/autopilot/src/run_loop.rs
squadgazzz: Removed `/settle` response body.

crates/autopilot/src/run_loop.rs
```diff
@@ -446,16 +446,8 @@ impl RunLoop {
...
-                if onchain_tx_hash != driver_tx_hash {
-                    tracing::warn!(
-                        ?driver_tx_hash,
-                        ?onchain_tx_hash,
-                        "driver and onchain tx hashes do not match"
-                    );
-                }
-                Ok(onchain_tx_hash)
+                driver_result.map_err(SettleError::Failure)?;
+                onchain_task.await
```
MartinquaXD, line 450: Now that a solver can only signal abort we no longer have to await the autopilot loop to compare transaction hashes. Technically a solver could still play games by pretending it aborted submitting while submitting anyways but the disruption caused by that would be very minor. However, we should still be able to track these cases to inform/disable solvers that repeatedly violate this so I think this should be logged here.

crates/autopilot/src/infra/solvers/mod.rs
```diff
@@ -42,9 +42,31 @@ impl Driver {
         &self,
         request: &settle::Request,
         timeout: std::time::Duration,
-    ) -> Result<settle::Response> {
-        self.request_response("settle", request, Some(timeout))
+    ) -> Result<()> {
+        let url = util::join(&self.url, "settle");
```
Mateo-mro, line 46: This is a little bit dangerous because you assume the `self.url` will always have a slash, but it is not enforced anywhere (I didn't see it at least :thinking: ). If a driver config is added without the slash this will crash. Maybe we could check if the last character is a slash to ensure this :thinking:

crates/autopilot/src/infra/solvers/mod.rs
```diff
@@ -42,9 +42,31 @@ impl Driver {
...
+            .json(request)
+            .timeout(timeout)
+            .send()
             .await
+            .context("send")?;
+        let status = response.status().as_u16();
+
+        tracing::trace!(%status, "solver response");
+
+        if status != 200 {
```
Mateo-mro, line 65: Why not to use the custom types from `reqwest`?

```suggestion
        let status = response.status();
 
         tracing::trace!(%status, "solver response");
 
        if status != StatusCode::OK {
```

crates/driver/src/infra/api/routes/settle/mod.rs
```diff
@@ -15,15 +15,17 @@ pub(in crate::infra::api) fn settle(router: axum::Router<State>) -> axum::Router
...
     let handle_request = async {
         observe::settling();
         let result = competition.settle().await;
         observe::settled(state.solver().name(), &result);
-        let calldata = result?;
-        Ok(axum::Json(dto::Settled::new(calldata)))
+        match result {
+            Err(err) => Err(err.into()),
+            Ok(_) => Ok(()),
+        }
```
Mateo-mro, line 28: Unnecessary match statement:
```suggestion
result.map_err(Into::into).map(|_| ())
```

crates/driver/src/tests/setup/mod.rs
```diff
@@ -1307,8 +1250,13 @@ impl<'a> Settle<'a> {
 
     /// Expect the /settle endpoint to return a 400 BAD REQUEST response.
     pub fn err(self) -> SettleErr {
```
Mateo-mro, line 1252: I know this is not your change, but why this is not `Option<SettleErr>` so it is consistent with Rust std function `.err()` (and the panic is handled by the caller)? having two kind of `.err()` outcomes could be confusing :thinking:

crates/autopilot/src/infra/solvers/mod.rs
MartinquaXD: `util::join` actually enforces exactly that because we ran into the issues you mentioned in the past.

Mateo-mro: aaah, I didn't see it! I am blind! thanks for the clarification @MartinquaXD :ok_hand: :pray:

crates/driver/src/tests/setup/mod.rs
squadgazzz: This code is used in tests only.

Mateo-mro (approved):

crates/autopilot/src/run_loop.rs
MartinquaXD: But we are still awaiting the autopilot deadline despite the solver already telling us it gave up.

squadgazzz: But there is a question mark operator which should early return in case of error.

MartinquaXD (approved):

crates/driver/openapi.yml
```diff
@@ -129,10 +129,6 @@ paths:
       responses:
         200:
           description: Execution accepted.
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/SettleResponse"
```
sunce86, line 135: Should we then also remove SettleResponse object?

sunce86 (approved):

Issue #2624, opened 2024-04-16
bug: Baseline sometimes unexpectedly returning fee=0


sunce86:
### Problem
Example case: https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/r/s/iHbXX

Quote, side::Sell, trading `0xbe9895146f7af43049ca1c1ae358b0541ea49704` for `0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2`. For some reason, inner baseline request that checks for native price of `0xbe9895146f7af43049ca1c1ae358b0541ea49704` to determine the fee, is failing and therefore the fee is 0. But, it should not fail if basically the same original order ☝️ is trading the same tokens as the inner request and it returns some exchange rate between `0xbe9895146f7af43049ca1c1ae358b0541ea49704` and `0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2`

### Impact
No impact at the moment since we use gas, so lower priority.

### Expected behaviour
Fee !=0 returned.

### Additional context
Functionality implemented here: https://github.com/cowprotocol/services/pull/2590

PR #2662, opened 2024-04-25
[EASY] Move out fee calculation from "into_solution"

sunce86:
# Description
Baseline solver. Moves out fee calculation from the `into_solution` function. 

This is done as a preparation for a follow up PR, where I will not calculate the sell token price, but instead build `native_price_request` to buy weth in an exact amount needed to cover the gas costs (instead of using 1ETH as a default buy value). The output sell amount of `native_price_request` will directly give us the amount of sell token needed to cover the gas costs - fee.

Related to https://github.com/cowprotocol/services/issues/2624, where the `native_price_request` uses 1ETH as buy amount and fails. I don't see any good reason to use 1ETH here as buy amount, as it is not needed. It is perfectly fine to use, for example, 0.01ETH amount that actually covers the gas cost and it's more accurate.

### DIFF
--- a/crates/solvers/src/domain/eth/mod.rs
+++ b/crates/solvers/src/domain/eth/mod.rs
@@ -45,6 +45,16 @@ impl From<U256> for Ether {
     }
 }
 
+/// A token amount in wei, always representing the sell token of an order.
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct SellTokenAmount(pub U256);
+
+impl From<U256> for SellTokenAmount {
+    fn from(value: U256) -> Self {
+        Self(value)
+    }
+}
+
 /// Like [`Gas`] but can be negative to encode a gas discount.
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub struct SignedGas(i64);
--- a/crates/solvers/src/domain/eth/mod.rs
+++ b/crates/solvers/src/domain/eth/mod.rs
@@ -1,5 +1,4 @@
-use {crate::util::bytes::Bytes, web3::types::AccessList};
-
+use {crate::util::bytes::Bytes, derive_more::From, web3::types::AccessList};
 mod chain;
 
 pub use {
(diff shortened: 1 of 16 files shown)


crates/solvers/src/domain/eth/mod.rs
```diff
@@ -45,6 +45,16 @@ impl From<U256> for Ether {
     }
 }
 
+/// A token amount in wei, always representing the sell token of an order.
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct SellTokenAmount(pub U256);
+
+impl From<U256> for SellTokenAmount {
```
Mateo-mro, line 52: This `From` implementation can be avoided by using the crate `derive_more`, I added it to the crates workspace, so it shouldn't be any size overhead to add it to this create too. Then we can just import:
```rust
use derive_more::{Into, From};

#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd, From, Into)]
pub struct SellTokenAmount(pub U256);
```

crates/solvers/src/domain/solution.rs
```diff
@@ -299,15 +295,15 @@ pub enum Fee {
...
+    Surplus(eth::SellTokenAmount),
 }
 
 impl Fee {
     /// Returns the dynamic component for the fee.
     pub fn surplus(&self) -> Option<U256> {
         match self {
             Fee::Protocol => None,
-            Fee::Surplus(fee) => Some(*fee),
+            Fee::Surplus(fee) => Some(fee.0),
```
Mateo-mro, line 306: If the `derive_more` is added to `SellTokenAmount, here we could do:

```suggestion
Fee::Surplus(fee) => Some(fee.into()),
```

squadgazzz (approved):

MartinquaXD:
> to buy weth in an exact amount needed to cover the gas costs (instead of using 1ETH as a default buy value)

The native token price is completely unrelated to the traded volume, though. It's a spot price and needs to be normalized across all solvers equally because this is an important part of the surplus calculation. If all solvers use spot prices derived from wildly different buy amounts they will optimize their solutions differently which leads to an unpredictable competition.

sunce86:
> > to buy weth in an exact amount needed to cover the gas costs (instead of using 1ETH as a default buy value)
> 
> The native token price is completely unrelated to the traded volume, though. It's a spot price and needs to be normalized across all solvers equally because this is an important part of the surplus calculation. If all solvers use spot prices derived from wildly different buy amounts they will optimize their solutions differently which leads to an unpredictable competition.

I find it hard to understand your concern. Sell token price is only used to determine the fee, which solvers are anyway free to calculate as they think is appropriate to cover gas costs. 
If Baseline receives a native token price request (quote) from driver (buy 1ETH for sell token), the result will still be the same, it's just the fee will be slightly adjusted to be more accurate.
For solve request, all solvers receive the same sell token price (it will exist for all tokens in auction), so no change will happen at all. But even in this case, if solver needs to pay 0.01ETH for gas cost, why solver needs to use the same (reference) price to determine the amount of sell token to charge as fee? It's the inner responsibility of solver to correctly estimate gas costs.

MartinquaXD (approved):

crates/solvers/src/domain/eth/mod.rs
sunce86: Used, thanks

crates/solvers/src/domain/solution.rs
sunce86: Would have to be `Some((*fee).into())`, so I just kept it as is for now.

PR #2666, opened 2024-04-26
More precise fee calculation in baseline solver

sunce86:
# Description
For quote requests (aka reference token price is None), fee is now calculated using the actual gas cost needed to cover the network expense, instead of using the random 1ETH amount which might not be appropriate for some cases: for example, when liquidity of the sell token is small and 1ETH significantly moves the price (or it is not even achievable) while actual realistic 0.01ETH gas cost moves the price far less.

Additional change is that we no longer return solutions with fee=0 (related to https://github.com/cowprotocol/services/issues/2624). This was happening for quotes where sell amounts are smaller than the network fee. For example, quote selling 5 usdc for weth while the network fee is 20$ should not return a valid quote.

## How to test
Existing quote with limit orders tests. Note how the `fee` and `executed` fields changed only slightly due to high liquidity of the tokens.

Fixes https://github.com/cowprotocol/services/issues/2624

### DIFF
--- a/crates/e2e/src/setup/colocation.rs
+++ b/crates/e2e/src/setup/colocation.rs
@@ -12,7 +12,6 @@ weth = "{weth:?}"
 base-tokens = []
 max-hops = 1
 max-partial-attempts = 5
-native-token-price-estimation-amount = "100000000000000000"
         "#,
     ));
 
--- a/crates/solvers/config/example.baseline.toml
+++ b/crates/solvers/config/example.baseline.toml
@@ -4,5 +4,4 @@ chain-id = "1"
 base-tokens = []
 max-hops = 0
 max-partial-attempts = 5
-native-token-price-estimation-amount = "100000000000000000"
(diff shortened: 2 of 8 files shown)


Mateo-mro (approved):

MartinquaXD (approved):
Good change 👍

squadgazzz:
> For example, quote selling 5 usdc for weth while the network fee is 20$ should not return a valid quote.

Do we have a test case covering this?

MartinquaXD:
Is there anything blocking this PR from being merged?

sunce86:
Fixing the failing test and in depth testing require non-trivial time so I decided to put in on stand-by for now because of tight schedule for colocation features (next two weeks I won't be able to contribute much for colocation).

squadgazzz:
> Fixing the failing test and in depth testing require non-trivial time so I decided to put in on stand-by

Could you convert it to a draft until then?

sunce86:
Low priority, will reopen when we find time for it.

PR #2672, opened 2024-04-29
Add arbitrum contract addresses

MartinquaXD:
# Changes
Adds arbitrum address for all contracts that are available there to `build.rs`.
Updates tests that assert which contract is available on which networks.

## How to test
This change was already deployed on an arbitrum staging stack and I was able to settle a solution with it (plus some incoming changes).

### DIFF
--- a/crates/contracts/build.rs
+++ b/crates/contracts/build.rs
@@ -14,6 +14,7 @@ const MAINNET: &str = "1";
 const GOERLI: &str = "5";
 const GNOSIS: &str = "100";
 const SEPOLIA: &str = "11155111";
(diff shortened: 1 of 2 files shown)


fleupold (approved):

squadgazzz (approved):
LG

mfw78:
Did we want to keep the habit of appending the TX hash for the deployments, if not, perhaps erase all? (can be recovered simply through an explorer if need be)

MartinquaXD:
Yeah, I think documenting the tx hash of the contract creation doesn't really add anything here. I'm fine with deleting all of those comments.

PR #2682, opened 2024-05-01
JIT orders surplus support for cow amm

### LINKED ISSUE #2635: feat: JIT orders surplus support
Excerpt from https://github.com/cowprotocol/services/issues/2607
JIT orders scoring should be supported.

Potential prerequisite: https://github.com/cowprotocol/services/issues/2215

mateo-mro:
# Description
JIT orders surplus support for cow amm

# Changes
- Support only surplus fee for cow amm addresses
- Calculate score for ALL the JIT orders !! 

## How to test
Tests are in another PR: https://github.com/cowprotocol/services/pull/2699

Fixes #2635

### DIFF
--- a/crates/driver/src/domain/competition/auction.rs
+++ b/crates/driver/src/domain/competition/auction.rs
@@ -42,7 +42,7 @@ impl Auction {
         tokens: impl Iterator<Item = Token>,
         deadline: time::Deadline,
         eth: &Ethereum,
-        surplus_capturing_jit_order_owners: &HashSet<eth::Address>,
+        surplus_capturing_jit_order_owners: HashSet<eth::Address>,
     ) -> Result<Self, Error> {
         let tokens = Tokens(tokens.map(|token| (token.address, token)).collect());
 
@@ -66,10 +66,7 @@ impl Auction {
             tokens,
             gas_price: eth.gas_price().await?,
             deadline,
-            surplus_capturing_jit_order_owners: surplus_capturing_jit_order_owners
-                .iter()
-                .cloned()
-                .collect(),
+            surplus_capturing_jit_order_owners,
         })
     }
 
--- a/crates/driver/src/domain/competition/solution/mod.rs
+++ b/crates/driver/src/domain/competition/solution/mod.rs
@@ -133,7 +133,7 @@ impl Solution {
         self.gas
     }
 
-    fn trade_count_for_scoring(
+    fn trade_count_for_scorable(
         &self,
         trade: &Trade,
         surplus_capturing_jit_order_owners: &HashSet<eth::Address>,
@@ -156,11 +156,9 @@ impl Solution {
         surplus_capturing_jit_order_owners: &HashSet<eth::Address>,
     ) -> Result<eth::Ether, error::Scoring> {
         let mut trades = Vec::with_capacity(self.trades.len());
-        for trade in self
-            .trades()
-            .iter()
-            .filter(|trade| self.trade_count_for_scoring(trade, surplus_capturing_jit_order_owners))
-        {
+        for trade in self.trades().iter().filter(|trade| {
+            self.trade_count_for_scorable(trade, surplus_capturing_jit_order_owners)
+        }) {
             // Solver generated fulfillment does not include the fee in the executed amount
             // for sell orders.
             let executed = match trade.side() {
@@ -218,7 +216,7 @@ impl Solution {
         !self
             .trades
             .iter()
-            .any(|trade| self.trade_count_for_scoring(trade, surplus_capturing_jit_order_owners))
+            .any(|trade| self.trade_count_for_scorable(trade, surplus_capturing_jit_order_owners))
     }
 
     pub fn merge(&self, other: &Self) -> Result<Self, error::Merge> {
--- a/crates/driver/src/infra/api/routes/solve/dto/auction.rs
+++ b/crates/driver/src/infra/api/routes/solve/dto/auction.rs
@@ -152,9 +152,7 @@ impl Auction {
             }),
             time::Deadline::new(self.deadline, timeouts),
             eth,
-            &self
-                .surplus_capturing_jit_order_owners
-                .unwrap_or_default()
+            self.surplus_capturing_jit_order_owners
                 .into_iter()
                 .map(Into::into)
                 .collect::<HashSet<_>>(),
@@ -203,7 +201,8 @@ pub struct Auction {
     tokens: Vec<Token>,
     orders: Vec<Order>,
     deadline: chrono::DateTime<chrono::Utc>,
-    surplus_capturing_jit_order_owners: Option<Vec<eth::H160>>,
+    #[serde(default)]
+    surplus_capturing_jit_order_owners: Vec<eth::H160>,
 }
 
 impl Auction {
--- a/crates/driver/src/domain/quote.rs
+++ b/crates/driver/src/domain/quote.rs
@@ -178,7 +178,7 @@ impl Order {
             .into_iter(),
             self.deadline,
             eth,
-            &HashSet::default(),
+            HashSet::default(),
         )
         .await
         .map_err(|err| match err {


crates/driver/src/tests/setup/solver.rs
```diff
@@ -168,75 +168,154 @@ impl Solver {
...
+                                "appData": jit.quoted_order.order.app_data,
+                                "feeAmount": jit.quoted_order.order.solver_fee.unwrap_or_default().to_string(),
+                                "kind": match jit.quoted_order.order.side {
+                                            order::Side::Sell => "sell",
+                                            order::Side::Buy => "buy",
+                                },
+                                "partiallyFillable": matches!(jit.quoted_order.order.partial, Partial::Yes { .. }),
+                                "sellTokenBalance": jit.quoted_order.order.sell_token_source,
+                                "buyTokenBalance": jit.quoted_order.order.buy_token_destination,
+                                "signature": if config.quote { "0x".to_string() } else { format!("0x{}", hex::encode(jit.quoted_order.order_signature(config.blockchain))) },
```
MartinquaXD, line 309: This currently signs the JIT orders with the trader's private key (taken from `quoted_order`) meaning the JIT order's sell_token balance will be taken from the original trader.

crates/driver/src/boundary/settlement.rs
```diff
@@ -326,13 +326,20 @@ fn to_boundary_jit_order(domain: &DomainSeparator, order: &order::Jit) -> Order
         ..Default::default()
     };
     let signature = order.signature.to_boundary_signature();
+    let signer = signature
+        .recover_owner(&order.signature.data.0, domain, &data.hash_struct())
+        .expect("Invalid signer");
 
-    Order {
+    if signer != order.signature.signer.0 {
```
Mateo-mro, line 333: I thought of adding this here so we avoid placing a TX which will 100 % fail

crates/driver/src/domain/competition/solution/fee.rs
```diff
@@ -27,29 +27,115 @@ use {
...
+            eth::{self, Asset},
         },
-        eth::{self},
+        util::conv::u256::U256Ext,
     },
     bigdecimal::Zero,
 };
 
-impl Fulfillment {
+impl Trade {
```
Mateo-mro, line 47: My idea was to use the full `Trade` to calculate the protocol fees, independently of if it is a `Jit` or `Fulfillment` order, this was it should be very straightforward to support more fee policies for either type of order.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -96,14 +98,27 @@ impl Solution {
...
+                    if cow_amm_addresses.contains(&jit.order().receiver.into()) {
+                        let prices = ClearingPrices {
+                            sell: solution.prices[&jit.order().sell.token.wrap(solution.weth)],
+                            buy: solution.prices[&jit.order().buy.token.wrap(solution.weth)],
+                        };
+                        let trade = trade.with_protocol_fees(prices)?;
+                        trades.push(trade);
+                    } else {
+                        let mut jit = jit.clone();
+                        jit.remove_protocol_fees();
```
Mateo-mro, line 118: I took the decision of directly removing the protocol fees for the corner case in which a non CoW AMM address places an JIT order with protocol fees (which should never happen but better to be safe). The reason is because if I don't do this, we would need to check again against the CoW AMM addresses when calculating the `score`, making it more complex. Another option would be to reject any JIT order with protocol fees which aren't coming from CoW AMM addresses. What do you guys think?

Mateo-mro:
The tests can be found in another PR: https://github.com/cowprotocol/services/pull/2699

crates/driver/src/boundary/settlement.rs
MartinquaXD: We definitely need to be able to figure out the actual signer of the order because that is the only way to know whether a JIT order is a CoW AMM order but it should probably be done when converting the JIT `dto` into a domain object.

crates/driver/src/domain/competition/order/mod.rs
```diff
@@ -399,6 +399,7 @@ pub struct Jit {
     pub sell_token_balance: SellTokenBalance,
     pub buy_token_balance: BuyTokenBalance,
     pub signature: Signature,
+    pub protocol_fees: Vec<FeePolicy>,
```
MartinquaXD, line 402: Let's not give JIT orders protocol fees.
There usual use case is providing liquidity for a solution and is more akin to an AMM interaction which also doesn't get taxed in some way. Additionally protocol fee is something that should be determined by the protocol and at the moment the driver has no way of knowing whether a fee proposed by a solver for their JIT order is equal to whatever the backend would have decided.

crates/driver/src/domain/competition/solution/fee.rs
```diff
@@ -146,7 +238,7 @@ impl Fulfillment {
         let fee_from_volume = self.fee_from_volume(prices, max_volume_factor)?;
         // take the smaller of the two
         let protocol_fee = std::cmp::min(fee_from_surplus, fee_from_volume);
-        tracing::debug!(uid=?self.order().uid, ?fee_from_surplus, ?fee_from_volume, ?protocol_fee, executed=?self.executed(), surplus_fee=?self.surplus_fee(), "calculated protocol fee");
+        tracing::debug!(?fee_from_surplus, ?fee_from_volume, ?protocol_fee, executed=?self.executed(), surplus_fee=?self.surplus_fee(), "calculated protocol fee");
```
MartinquaXD, line 241: order uid is very important information to have here. And given that JIT orders should not have fees in the first place it's not great that I got removed here.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -96,14 +98,27 @@ impl Solution {
...
+                        let trade = trade.with_protocol_fees(prices)?;
+                        trades.push(trade)
                     }
                     order::Kind::Liquidity => {
                         trades.push(trade);
                     }
                 },
-                Trade::Jit(_) => trades.push(trade),
+                Trade::Jit(jit) => {
+                    if cow_amm_addresses.contains(&jit.order().receiver.into()) {
```
MartinquaXD, line 109: The important part to check is the orders `signer`. I assume the `receiver` of the CoW AMM orders have to be `0x00000` to avoid creating orders which can send the proceeds of a trade outside of the contract.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -96,14 +98,27 @@ impl Solution {
...
+                        let prices = ClearingPrices {
+                            sell: solution.prices[&jit.order().sell.token.wrap(solution.weth)],
+                            buy: solution.prices[&jit.order().buy.token.wrap(solution.weth)],
+                        };
+                        let trade = trade.with_protocol_fees(prices)?;
```
MartinquaXD, line 114: No need to have protocol fees here. If you remember we jumped through a bunch of hoops to make CoW AMM orders exempt from protocol fees in the autopilot, remember?
Would be odd if we'd suddenly want to have fees when the trade is proposed as a JIT order. This PR should only be about computing the surplus for CoW AMM JIT orders and counting it towards the score of the solution.

crates/driver/src/domain/competition/solution/trade.rs
```diff
@@ -263,10 +201,15 @@ pub struct Jit {
     /// partially fillable, the executed amount must equal the amount from the
     /// order.
     executed: order::TargetAmount,
+    fee: Fee,
```
MartinquaXD, line 204: No fee needed.

MartinquaXD (commented):
Not really sure where the miscommunication happened but it was pointed out multiple times (slack, GH issues) that the change should be about counting the order surplus towards the solution score and **NOT** about applying protocol fees to JIT orders.

crates/driver/src/boundary/settlement.rs
Mateo-mro: Sure! but the signer is set automatically to the solver address (regardless of the real signer). I can add this check there.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -136,29 +136,63 @@ impl Solution {
...
+                        fulfillment.order().protocol_fees.clone(),
+                    ))
+                }
+                Trade::Jit(jit) => {
+                    if self
+                        .solver
+                        .cow_amm_addresses()
+                        .contains(&jit.order().signature.signer.into())
+                    {
+                        let executed = match jit.order().side {
```
Mateo-mro, line 173: We can potentially move the functions at `Trade` level and do this operation at such level.

crates/driver/src/boundary/settlement.rs
fleupold: We will never place a tx that will fail (we simulate it before). I think this check is a bit weird here. We don't really need to verify the signer here, so it's not clear why it gets asserted (and adds more error handling that may not be required). I'd be in favor of removing it.

crates/driver/src/infra/config/file/mod.rs
```diff
@@ -265,6 +266,10 @@ struct SolverConfig {
     /// Which `tx.origin` is required to make a quote simulation pass.
     #[serde(default)]
     quote_tx_origin: Option<eth::H160>,
+
+    /// List of CoW AMM addresses
+    #[serde(default)]
+    cow_amm_addresses: Vec<H160>,
```
fleupold, line 272: Can we add this list to the auction that gets sent to the driver (I believe the autopilot already has this list). This way co-located drivers could also settle CoW AMMs.

crates/driver/src/boundary/settlement.rs
MartinquaXD: > but the signer is set automatically to the solver address (regardless of the real signer)

Yeah, this is exactly the part that should change. Instead of always blindly setting the solver address we can just recover the real signer from the signature when we convert the dto instead.

crates/driver/src/infra/config/file/mod.rs
MartinquaXD: I'm a bit torn on this. CoW AMMs are nothing inherent to the protocol so putting them into the data sent to the driver seems odd to me. Also CoW AMMs could be understood as a liquidity source which gets managed by the driver. This might be a bit misguided though because once CoW AMM orders can contribute to the solution score with their surplus it could be perfectly reasonable for a solver to completely ignore user orders and only settle CoW AMM orders (meaning they would not be used to provide liquidity for other orders).

crates/driver/src/boundary/settlement.rs
Mateo-mro: Got it! Doing that! :ok_hand:

crates/driver/src/infra/config/file/mod.rs
fleupold: I wouldn't put "CoW AMMs" in the file, but instead call them "surplus capturing jit-order owners" or the like. It's true that eventually we will need a formal definition of what qualifies as such orders (I assume we will allow-list certain smart order factories for which we know they create "good" contract instances) but whatever the logic is the autopilot needs to implement it to double check the score computation.

For now it's a hardcoded list and so I think it should be sent along with the auction on every batch. How else would we communicate to solvers which addresses are eligible?

MartinquaXD: > For now it's a hardcoded list and so I think it should be sent along with the auction on every batch. How else would we communicate to solvers which addresses are eligible?

I think for now it's a reasonable suggestion to send the list with the request but eventually when we have possibly many CoW AMMs I would just expect the driver to compute the score based on what it thinks the available AMMs are and let the autopilot do the same.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -133,32 +133,46 @@ impl Solution {
...
-            trades.push(scoring::Trade::new(
-                trade.order().sell,
-                trade.order().buy,
-                trade.order().side,
-                executed,
-                trade.custom_prices(&uniform_prices)?,
-                trade.order().protocol_fees.clone(),
-            ))
+        for trade in self.trades() {
+            if self.trade_count_for_scoring(trade) {
```
MartinquaXD, line 150: nit: code is unnecessarily indented. You could either do `for trade in trades.iter().filter()` or `if !count_for_scoring { continue; }`

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -194,7 +208,13 @@ impl Solution {
...
     /// An empty solution has no user trades and a score of 0.
     pub fn is_empty(&self) -> bool {
-        self.user_trades().next().is_none()
+        !self.trades.iter().any(|trade| match trade {
+            Trade::Fulfillment(fulfillment) => match fulfillment.order().kind {
+                order::Kind::Market | order::Kind::Limit { .. } => true,
+                order::Kind::Liquidity => false,
+            },
+            Trade::Jit(_) => true,
+        })
```
MartinquaXD, line 217:
```suggestion
        !self.trades.iter().any(|trade| self.trade_count_for_scoring(&trade))
```

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -133,32 +133,46 @@ impl Solution {
         self.gas
     }
 
+    fn trade_count_for_scoring(&self, trade: &Trade) -> bool {
+        match trade {
+            Trade::Fulfillment(_) => true,
```
MartinquaXD, line 138: liquidity orders also should not generate any surplus. We no longer create these so technically it doesn't matter but it would be better to cover this case explicitly to indicate that we didn't forget about it.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -194,7 +208,13 @@ impl Solution {
 
     /// An empty solution has no user trades and a score of 0.
```
MartinquaXD, line 209: No trade which is allowed to capture surplus is more appropriate now.

crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -204,6 +223,53 @@ impl Solutions {
             })
             .collect()
     }
+
+    // Function to check that a JIT order signature is signed by the solver
+    fn recover_signer_from_jit_trade_order(
+        jit: &JitTrade,
+        signature: &competition::order::Signature,
+        solver: &Solver,
```
MartinquaXD, line 231: Let's pass `eth::DomainSeparator` directly. This can be read from `solver.eth.contracts().settlement_domain_separator()`.

crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -204,6 +223,53 @@ impl Solutions {
             })
             .collect()
     }
+
+    // Function to check that a JIT order signature is signed by the solver
```
MartinquaXD, line 227: nit: It doesn't matter if the order was actually signed by the solver. Again, CoW AMM orders would have the CoW AMM contract address as the signer for example.
Also doc comments need `///`

crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -204,6 +223,53 @@ impl Solutions {
...
+    fn recover_signer_from_jit_trade_order(
+        jit: &JitTrade,
+        signature: &competition::order::Signature,
+        solver: &Solver,
+    ) -> Result<H160, super::Error> {
```
MartinquaXD, line 232: Let's return the appropriate type right away.
```suggestion
    ) -> Result<eth::Address, super::Error> {
```

crates/driver/src/infra/solver/mod.rs
```diff
@@ -129,6 +130,8 @@ pub struct Config {
     pub solver_native_token: ManageNativeToken,
     /// Which `tx.origin` is required to make quote verification pass.
     pub quote_tx_origin: Option<eth::Address>,
+    /// List of surplus capturing JIT-order owners
+    pub surplus_capturing_jit_order_owners: HashSet<H160>,
```
MartinquaXD, line 134: Types in the domain should use the domain types meaning this would be `HashSet<eth::Address>`. But since another comment suggests passing the cow amm addresses in the `/solve` request this would go away anyway.

crates/driver/src/domain/competition/solution/mod.rs
Mateo-mro: @MartinquaXD  Oh, damn, you are right, sorry :pray:

crates/driver/src/infra/config/file/mod.rs
Mateo-mro: Guys @fleupold @MartinquaXD I think I reworked it as suggested! please let me know! :pray:

crates/autopilot/src/infra/solvers/dto/solve.rs
```diff
@@ -47,6 +48,13 @@ impl Request {
                 .unique_by(|token| token.address)
                 .collect(),
             deadline: Utc::now() + chrono::Duration::from_std(time_limit).unwrap(),
+            surplus_capturing_jit_order_owners: (surplus_capturing_jit_order_owners.is_empty())
```
MartinquaXD, line 51: This should encode values if there are values to encode.
```suggestion
            surplus_capturing_jit_order_owners: (!surplus_capturing_jit_order_owners.is_empty())
```
But overall I think it's probably not worth the effort to not serialize empty vectors anyway. So the `Option<Vec<T>>` in the request might as well be `Vec<T>`.
And if we were keen on not serializing empty vectors we could do:
```rust
#[serde(skip_serializing_if = "Vec::is_empty")]
```
This still makes sense because vectors in rust actually only allocate the first time you push into them making the creation of a bunch of empty vectors basically free.
But yeah overall introducing `Option` here and in the deserializing logic just results in more conversion logic for not much benefit.

crates/driver/src/domain/competition/auction.rs
```diff
@@ -41,6 +42,7 @@ impl Auction {
         tokens: impl Iterator<Item = Token>,
         deadline: time::Deadline,
         eth: &Ethereum,
+        surplus_capturing_jit_order_owners: &[eth::Address],
```
MartinquaXD, line 45: This can already take a `HashSet<eth::Address>` to avoid cases where you currently collect values into a vector, pass a reference and then collect into a hashset.

crates/driver/src/domain/competition/solution/mod.rs
```diff
@@ -48,6 +48,7 @@ pub struct Solution {
     solver: Solver,
     weth: eth::WethAddress,
     gas: Option<eth::Gas>,
+    surplus_capturing_jit_order_owners: HashSet<eth::Address>,
```
MartinquaXD, line 51: These addresses are nothing specific to each solution. I makes more sense to pass a reference of this into `Solution::score()` and `Solution::is_empty()` instead.

crates/driver/src/infra/api/routes/solve/dto/auction.rs
```diff
@@ -196,6 +203,7 @@ pub struct Auction {
     tokens: Vec<Token>,
     orders: Vec<Order>,
     deadline: chrono::DateTime<chrono::Utc>,
+    surplus_capturing_jit_order_owners: Option<Vec<eth::H160>>,
```
MartinquaXD, line 206: The autopilot now sends this field unconditionally. Technically we only need `serde(Default)` to avoid errors in the shadow competition which receives requests from prod but uses the code from staging.
```suggestion
    #[serde(Default)]
    surplus_capturing_jit_order_owners: Vec<eth::H160>,
```

crates/driver/src/domain/competition/auction.rs
```diff
@@ -41,6 +42,7 @@ impl Auction {
         tokens: impl Iterator<Item = Token>,
         deadline: time::Deadline,
         eth: &Ethereum,
+        surplus_capturing_jit_order_owners: &HashSet<eth::Address>,
```
MartinquaXD, line 45: There is still currently one instance where you create a new `HashSet` only to pass the reference and have this function clone the reference to the newly created `HashSet` anyway.
```suggestion
        surplus_capturing_jit_order_owners: HashSet<eth::Address>,
```

MartinquaXD (approved):
Code looks good IMO. Just a few nits remaining.

crates/driver/src/infra/solver/dto/solution.rs
```diff
@@ -204,6 +217,48 @@ impl Solutions {
...
+    ) -> Result<eth::Address, super::Error> {
+        let order_data = OrderData {
+            sell_token: jit.order.sell_token,
+            buy_token: jit.order.buy_token,
+            receiver: Some(jit.order.receiver),
+            sell_amount: jit.order.sell_amount,
+            buy_amount: jit.order.buy_amount,
+            valid_to: jit.order.valid_to,
+            app_data: AppDataHash(jit.order.app_data),
+            fee_amount: jit.order.fee_amount,
```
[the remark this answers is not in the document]
sunce86: Should be 0 but I am ok with leaving it as is until we remove the fee from the JIT order completely.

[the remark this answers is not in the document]
Mateo-mro: I understand, thank you!

PR #2690, opened 2024-05-04
Verify quotes based on zeroex RFQ orders even if the simulation fails

MartinquaXD:
# Description
Quoters with market maker integration currently quote by returning an interaction that calls `fillRfqOrder()` on the zeroex contract with an order they just signed for that specific quote.
However, at least 1 relevant quoter currently signs zeroex orders with `txOrigin: 0x0000` which makes the orders invalid ([code](https://github.com/0xProject/protocol/blob/development/contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersInfo.sol#L77-L80)). The proper solution would be to underwrite orders that set `txOrigin` to literally any other value and adjust the test for that but since the solvers will be slow to adjust to this (signaled on telegram) we need a temporary workaround if we want to rollout quote verification on prod mainnet again without losing essential quoters.

The actual long term fix will come in a follow up PR.

# Changes
If a simulation fails the `TradeVerifier` checks if the trade contains interactions calling `fillRfqOrder()` or `fillOrKillRfqOrder()` and considers it verified anyway.

## How to test
Added a forked node e2e test that replays one of the problematic quotes we [saw in prod](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/context/c0e240e0-d9b3-11ed-b0e6-e361adffce0b/PhgZQ48B0pqFYaDRQAtQ?_g=(filters:!())&_a=(columns:!(log),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!t,index:c0e240e0-d9b3-11ed-b0e6-e361adffce0b,key:kubernetes.labels.app_kubernetes_io%2Fcomponent,negate:!t,params:(query:controller),type:phrase),query:(match_phrase:(kubernetes.labels.app_kubernetes_io%2Fcomponent:controller))),('$state':(store:appState),meta:(alias:!n,disabled:!t,index:c0e240e0-d9b3-11ed-b0e6-e361adffce0b,key:kubernetes.labels.network,negate:!f,params:(query:mainnet),type:phrase),query:(match_phrase:(kubernetes.labels.network:mainnet)))))) and asserts that it would now pass the verification.

### DIFF
--- a/crates/autopilot/src/run.rs
+++ b/crates/autopilot/src/run.rs
@@ -420,6 +420,7 @@ pub async fn run(args: Arguments) {
             gas_price: gas_price_estimator.clone(),
         },
     )
+    .await
     .expect("failed to initialize price estimator factory");
 
     let native_price_estimator = price_estimator_factory
--- a/crates/e2e/tests/e2e/main.rs
+++ b/crates/e2e/tests/e2e/main.rs
@@ -20,6 +20,7 @@ mod partial_fill;
 mod partially_fillable_balance;
 mod partially_fillable_pool;
 mod protocol_fee;
+mod quote_verification;
 mod quoting;
 mod refunder;
 mod replace_order;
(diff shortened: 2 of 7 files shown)


MartinquaXD:
Superseeded by https://github.com/cowprotocol/services/pull/2691.

PR #2700, opened 2024-05-07
[TRIVIAL] Use checked div when construction a quote

MartinquaXD:
# Description
Some solver returned a quote with the buy price being 0 which caused the driver to panic.

# Changes
Used `.checked_div()` to avoid panics when the sell or buy price is zero.

## How to test
compiler

### DIFF
--- a/crates/driver/src/domain/quote.rs
+++ b/crates/driver/src/domain/quote.rs
@@ -15,7 +15,9 @@ use {
         },
         util::{self, conv::u256::U256Ext},
     },
-    std::{collections::HashSet, iter},
+    anyhow::Context,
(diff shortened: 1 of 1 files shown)


Mateo-mro (approved):

sunce86 (commented):
I would prefer for this to be part of the solution validation, and consequently rejecting the solution before we even come to this point in the code.

MartinquaXD:
There is currently no solution validation because the code path for quoting is very much trimmed down compared to the regular `/solve` code path. The reason is that solutions generated for quotes can not be be verified (and hence would revert) for many reasons which might not indicate that the quote itself is bad.

PR #2708, opened 2024-05-08
Unify and simplify block stream creation

MartinquaXD:
# Changes
Simplify creation of `CurrentBlockStream` by basing the poll interval off of the connected chains immutable block frequency. Nothing is really changed for existing networks since the `MAX_POLL_INTERVAL` we fallback to for slow chains is the default value we have configure in the infra repo in forever.

## How to test
e2e tests should still pass
I guess review configured network specific block frequencies carefully


crates/ethrpc/src/current_block/mod.rs
```diff
@@ -134,6 +135,33 @@ pub async fn current_block_stream(url: Url, poll_interval: Duration) -> Result<C
     Ok(receiver)
 }
 
+/// Returns a reasonable poll interval for the given network.
+async fn poll_interval(web3: &Web3) -> Result<Duration> {
+    let chain = web3.eth().chain_id().await?;
+
+    let block_time = match chain.as_u64() {
+        1 => Duration::from_millis(12_000),
```
Mateo-mro, line 143: Where are these numbers coming from? :thinking:

crates/ethrpc/src/current_block/mod.rs
```diff
@@ -134,6 +135,33 @@ pub async fn current_block_stream(url: Url, poll_interval: Duration) -> Result<C
...
+        100 => Duration::from_millis(12_000),
```
sunce86, line 144: gnosis chain is 5s i think https://gnosisscan.io/blocks

sunce86: agree with mateo we should consider documenting the source for all these numbers

MartinquaXD: True, forgot to push the 5s last night.
Will also add comments with links to the block frequency.

fleupold (changes requested):
I'm not a huge fan of this change. First, I think it's bad to only poll for new blocks on mainnet for instance once every 12s. If you just missed a block, you may become up to 24 seconds behind, which is a huge delay.

Also, I think we should decide on where we want to configure the block interval for each chain and do it once. Right now we have this in the infra repo, here we add it to the ethrpc crate, and I wouldn't be surprised if this logic is lingering around elsewhere. I think we should have one place to configure network specifics such as block frequency.

MartinquaXD:
This was intended as an easy way to get correctly configured block stream poll intervals without having to introduce a bunch of config logic into the driver. Since opening the PR I learned that having a block stream that very closely matches the block interval is actually not that important so this PR can be closed.

PR #2712, opened 2024-05-09
[ON HOLD] Reject solution containing outside-of-auction order

sunce86:
ON HOLD since I want to get back to this once I work more on fetching the offchain data from db that will be used in `OnSettlementEventUpdater`. I already see that I will need a different variation of the `orders_do_not_exist` functionality for that as well.

# Description
Implements first part of the https://github.com/cowprotocol/services/issues/2667#issuecomment-2102239321

Rejects the winning solution if it contains an order that is not part of the auction. Since JIT orders do not satisfy this condition, we also need to make sure that those that are not in the auction, are also not in the database.

## How to test
e2e tests

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -372,11 +372,7 @@ impl RunLoop {
         auction: &domain::Auction,
         settlement: &Settlement,
     ) -> bool {
-        let auction_orders = auction
-            .orders
-            .iter()
-            .map(|o| OrderUid(o.uid.0))
-            .collect::<HashSet<_>>();
+        let auction_orders = auction.orders.iter().map(|o| o.uid).collect::<HashSet<_>>();
 
         // all JIT orders will be non_auction_orders
         let non_auction_orders = settlement
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -1,7 +1,7 @@
 use {
     crate::{
         database::competition::Competition,
-        domain::{self, auction::order::Class, OrderUid},
+        domain::{self, auction::order::Class, settlement::Settlement, OrderUid},
         infra::{
             self,
             solvers::dto::{reveal, settle, solve},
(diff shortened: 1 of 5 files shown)


crates/autopilot/src/run_loop.rs
```diff
@@ -141,6 +141,23 @@ impl RunLoop {
...
+                Ok(settlement) => settlement,
+                Err(err) => {
+                    tracing::error!(driver = %driver.name, ?err, "unexpected calldata format");
+                    return;
+                }
+            };
+
+            if !self.validate_settlement(&auction, &settlement).await {
+                tracing::error!(driver = %driver.name, "invalid settlement");
+                return;
```
MartinquaXD, line 158: Shouldn't we still store all the relevant data for this competition when the solution fails some checks?

crates/autopilot/src/run_loop.rs
```diff
@@ -348,6 +365,32 @@ impl RunLoop {
...
+            .filter(|order| !auction_orders.contains(order))
+            .cloned()
+            .collect::<Vec<_>>();
+
+        // make sure all non-auction orders are also non-database
+        self.persistence
+            .orders_do_not_exist(&non_auction_orders)
+            .await
+            .unwrap_or_else(|err| {
+                tracing::warn!(?err, "failed to check if all orders exist");
```
MartinquaXD, line 389: These checks are pretty important so they should fail loudly IMO (e.g. `tracing::error!`)
```suggestion
                tracing::warn!(?err, "failed to check if all unexpected orders are JIT orders");
```

crates/autopilot/src/run_loop.rs
```diff
@@ -141,6 +141,23 @@ impl RunLoop {
...
+                self.eth.contracts().settlement_domain_separator(),
+            ) {
+                Ok(settlement) => settlement,
+                Err(err) => {
+                    tracing::error!(driver = %driver.name, ?err, "unexpected calldata format");
+                    return;
+                }
+            };
+
+            if !self.validate_settlement(&auction, &settlement).await {
```
MartinquaXD, line 156: I think before we enforce any checks like this we should ensure that we await the correct settlement transaction. It's possible that a malicious losing solver would settle a solution with the expected auction_id out of competition.
Currently we just await the next tx that has the expected auction_id attached to its calldata. What we should probably do instead is to make it so that we only await a solution with the expected auction_id that also came from the winning solver.
That way this verification always applies to "real" solution and a malicious solver can't disrupt the main loop.
This disrupting solver would then easily be removed from the next competitions by the circuit breaker. Unfortunately this likely requires a change in the DB because we have a 1:1 mapping between tx_hash and auction_id at the moment whereas it should probably be a 1:1 mapping between tx_hash and (auction_id, solver).

MartinquaXD (commented):
This concept of this change makes sense to me (prepare data that is needed by circuit breaker to remove winning solver) but I think it's currently susceptible to malicious solvers interfering.

Issue #2719, opened 2024-05-11
Add `score` calculation in `domain::Settlement`


sunce86:
Add the `score` function to the `domain::Settlement` object. 

Should be already done here: https://github.com/cowprotocol/services/blob/settlement-auction-struct/crates/autopilot/src/domain/settlement/mod.rs#L46-L60 just add unit test for a real tx containing multiple protocol fees.

sunce86:
Closed with https://github.com/cowprotocol/services/pull/2724 afterall.

I've decided not to update the unit test, as it can't even cover a significant number of edge cases that exist (partial orders, different multiple variations of protocol fees, jit orders etc).
So, I've decided to test it by comparing with existing results, as explained in https://github.com/cowprotocol/services/pull/2730

PR #2724, opened 2024-05-13
Calculate score in autopilot domain

sunce86:
# Description
Related to https://github.com/cowprotocol/services/issues/2719 (will not fix it until unit test is updated 👇 )

Adds score calculation to `domain::Settlement` object, which will be used to calculate the score during solver competition.

## How to test
Will update the `settlement` unit test to test with a settlement transaction that contains multiple fee policies so the whole functionality (surplus, multiple fee policies, score) is tested together.

Will update the test in a separate PR. Now, only the function is merged to unblock progress on other tasks for colocation.


crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -32,6 +34,26 @@ impl Settlement {
...
+        &self,
+        prices: &auction::Prices,
+        policies: &HashMap<OrderUid, Vec<fee::Policy>>,
+    ) -> Result<eth::Ether, trade::Error> {
+        self.trades
+            .iter()
+            .map(|trade| {
+                trade.score(
+                    prices,
+                    &policies.get(trade.order_uid()).cloned().unwrap_or_default(),
```
Mateo-mro, line 51: The clone could be avoided.
```suggestion
                    policies
                        .get(trade.order_uid())
                        .map(|value| value.as_slice())
                        .unwrap_or_default(),
```

Mateo-mro (approved):

crates/autopilot/src/domain/settlement/mod.rs
sunce86: You are right! fixed

PR #2734, opened 2024-05-14
Fetch `Competition` from database

sunce86:
# Description
For a given auction_id, implements fetching a related auction object from and competition solution from database. These  objects should be the last piece of information needed to give the full context to a settlement, enough to do whatever is needed in the autopilot domain.

# Changes

- [ ] Implemented `infra::persistence::get_competition` to get auction object and competition solution from database
- [ ] fetched objects become a part of a `domain::Settlement`

## How to test
Added test log to `OnSettlementEventUpdater` that reports differences in the promised score vs delivered score. This should be enough to observe different settlements in prod. Locally, I've observed the output of different e2e tests and it looks good.

Non-failure of any e2e tests should give a sanity check.


Mateo-mro (approved):

crates/autopilot/src/domain/settlement/auction.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use {
+    crate::domain::{self, competition, eth},
+    std::collections::{HashMap, HashSet},
+};
+
+/// Offchain auction data related to a specific settlement that got settled for
+/// it.
+#[derive(Debug)]
+pub struct Auction {
```
MartinquaXD, line 11: I think this should be called sth like `FinalizedAuction` or `ArbitratedAuction` because the regular auction can't have a winner, solution and other things yet.

crates/autopilot/src/domain/settlement/auction.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub deadline: eth::BlockNo,
+    /// Winning score promised during competition (based on the promised
+    /// `competition::Solution`)
+    pub score: competition::Score,
+    /// Winning solution promised during competition.
+    pub solution: competition::Solution,
+    /// Auction external prices
+    pub prices: domain::auction::Prices,
+    /// Settlement orders that are missing from the orderbook (JIT orders).
+    pub missing_orders: HashSet<domain::OrderUid>,
```
MartinquaXD, line 24: Should this be allowed to include orders that are in the orderbook but were not part of the original auction? IMO these would technically not be JIT orders.
Coming to think of it if a solver settles a user order that was not part of the auction it should owe the protocol the order fees on top of whatever amount it gets slashed for. 🤔

crates/autopilot/src/domain/settlement/transaction.rs
```diff
@@ -3,31 +3,40 @@
...
 /// A transaction that settles a settlement. Interacts with the settlement
 /// contract `settle` function.
+#[derive(Debug)]
 pub struct Tx {
     settlement: Settlement,
     transaction: Transaction,
     #[allow(dead_code)]
     receipt: Receipt,
+    #[allow(dead_code)]
+    auction: super::Auction,
```
MartinquaXD, line 20: Seems odd that a struct that's just aggregating a bunch of data is not higher in the directory hierarchy than the individual components. 🤔

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -171,6 +171,18 @@ impl Inner {
 
         tracing::debug!(?hash, ?update, "updating settlement details for tx");
 
+        {
+            // temporary to debug and compare with current implementation
+            // TODO: use instead of current implementation
+            let settlement =
+                domain::settlement::Tx::new(hash.into(), &self.eth, &self.persistence).await;
+            tracing::info!(?settlement, "settlement object");
```
MartinquaXD, line 179: nit: log message is redundant

crates/autopilot/src/domain/settlement/transaction.rs
```diff
@@ -38,6 +47,20 @@ impl Tx {
     pub fn solver(&self) -> eth::Address {
         self.transaction.solver
     }
+
+    /// Score identical to the one promised during the competition.
+    pub fn check_score(&self) -> bool {
```
MartinquaXD, line 52: I don't think the settlement should make any claims about the correctness of the score. IMO having functions `promised_score()` and `observed_score()` and letting the new circuit breaker (or temporary logs) decide itself what to do.

crates/autopilot/src/infra/persistence/dto/fee_policy.rs
```diff
@@ -48,3 +48,31 @@ pub fn from_domain(
         },
     }
 }
+
+pub fn into_domain(
+    policy: FeePolicy,
+    quote: &domain::quote::Quote,
```
MartinquaXD, line 54: Probably a bigger refactor awaiting but it seems odd to me that we need to have a domain object of a quote to turn the infra FeePolicy into a domain Policy.
Shouldn't the `persistence` layer be capable of recovering the required quote data?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
+
+        let (fee_policies, missing_orders) = {
+            let mut missing_orders = HashSet::new();
+            let mut fee_policies = HashMap::new();
+            for order in settlement.order_uids().cloned() {
+                match database::orders::read_order(&mut ex, &ByteArray(order.0))
+                    .await
+                    .context("fetch order")?
+                {
+                    Some(_) => {
```
MartinquaXD, line 227: Since we only care about `.is_some()` you could safe 1 level of indentation with `if {} else {}` here.
Or even:
```rust
if let found_order.is_none() {
    missing_orders.insert(order);
    continue;
}

// handle case where order exists without extra indentation
```

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
+                        let policies =
+                            database::fee_policies::fetch(&mut ex, auction, ByteArray(order.0))
+                                .await
+                                .context("fetch fee policies")?;
+
+                        for policy in policies {
+                            let policy = dto::fee_policy::into_domain(policy, &quote)?;
+                            fee_policies
+                                .entry(order)
+                                .or_insert_with(Vec::new)
```
MartinquaXD, line 245:
```suggestion
                                .or_default()
```

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
+            let solver_competition = database::solver_competition::load_by_id(&mut ex, auction)
+                .await
+                .context("load solver competition")?
+                .ok_or(error::Auction::SolverCompetition(
+                    "missing solver competition",
+                ))?;
+            let competition: model::solver_competition::SolverCompetitionDB =
+                serde_json::from_value(solver_competition.json)
+                    .map_err(|_| error::Auction::SolverCompetition("json conversion"))?;
+            let winning_solution = competition.solutions.last().unwrap();
```
MartinquaXD, line 270: There is already error handling all over the place. I think there is no need to `.unwrap()` here then.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
-                .map_err(Error::Price)?;
+                .map_err(error::Auction::Price)?;
             prices.insert(token, price);
         }
 
         Ok(prices)
     }
+
+    /// Get auction data related to the given settlement.
+    pub async fn get_settlement_auction(
```
MartinquaXD, line 176: I think this function is quite long but lends itself very well to be split up into multiple private functions that implement the individual steps.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
+            prices
+        };
+
+        let (fee_policies, missing_orders) = {
+            let mut missing_orders = HashSet::new();
+            let mut fee_policies = HashMap::new();
+            for order in settlement.order_uids().cloned() {
+                match database::orders::read_order(&mut ex, &ByteArray(order.0))
+                    .await
+                    .context("fetch order")?
```
squadgazzz, line 225: Since we don't care about the retrieved data, multiple SQL queries can be replaced by a single one like it was done in a separate PR: https://github.com/cowprotocol/services/pull/2688/commits/7da5c49dd4da4494cbab09a31c46643ac0096062#diff-93818c26d2dd1e7d1a300cbcad4d774e0266a7f183708ba7f70a1a8a5a30243cR758-R782

crates/autopilot/src/domain/settlement/auction.rs
sunce86: Agree. Initially when implemented it contained only auction data but I added more, without renaming. IMO shouldn't have auction in its name since it contains auction and other data as well, like winning solution or missing order uids etc. Maybe `Competition`

crates/autopilot/src/domain/settlement/auction.rs
sunce86: [Should this be allowed to include orders that are in the orderbook but were not part of the original auction?](url)

IMO no. 
1. A promised solution (solution returned during competition) that contains out-of-auction orders should be rejected. 
2. If a promised solution doesn't contain any out-of-auction order, but delivered solution does, solver should be removed.

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: I intented to use it for log search... typing just `settlement` might not hit what I want.

crates/autopilot/src/infra/persistence/dto/fee_policy.rs
sunce86: It should, however we don't have it implemented yet. Instead, I reused what we currently have (similar to what we did [here](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/database/auction.rs#L76-L87)). If you think this should be part of the colocation feature and launch blocking, I will create an issue for it.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -2,34 +2,54 @@
...
+        let auction = persistence.get_settlement_auction(&solution).await?;
         Ok(Self {
             solution,
             transaction,
+            auction,
         })
     }
+
+    /// The onchain delivered score of a solution.
+    pub fn score(&self) -> Result<super::competition::Score, solution::error::Score> {
```
MartinquaXD, line 43: `promised_score` and `observed_score` or `delivered_score` seem obvious enough to me that one might not even have to read the doc comment.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -141,38 +145,198 @@ impl Persistence {
...
+                        for policy in policies {
+                            let policy = dto::fee_policy::into_domain(policy, &quote)?;
+                            fee_policies
+                                .entry(order)
+                                .or_insert_with(Vec::new)
+                                .push(policy);
+                        }
+                    }
+                    None => {
+                        missing_orders.insert(order);
```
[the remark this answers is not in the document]
sunce86: done

crates/autopilot/src/domain/settlement/competition.rs
```diff
@@ -0,0 +1,18 @@
+//! Auction data related to the specific settlement.
+
+use {
+    crate::domain::{self, eth},
+    std::collections::HashMap,
+};
+
+#[derive(Debug)]
+pub struct Auction {
```
fleupold, line 9: Don't we still need the orders here to filter out JIT orders in the solution?

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -2,39 +2,151 @@
...
+                        buy: clearing_prices[buy_token_index],
+                    },
+                },
+            ));
+        }
+
+        let auction = persistence.get_auction(auction_id).await?;
+        let solution = persistence.get_competition_solution(auction_id).await?;
+
+        let score = domain::competition::Score::new(
```
fleupold, line 83: nit: would move `Score` type on the top level domain, since it's no longer competition specific (it's just as much competition::Score as it would be settlement::Score)

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -2,39 +2,151 @@
...
-    Solution(#[from] solution::Error),
+mod error {
+    use super::*;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Error {
+        #[error(transparent)]
+        Decoding(#[from] tokenized::error::Decoding),
+        #[error("failed to recover order uid {0} for auction {1}")]
+        OrderUidRecover(tokenized::error::Uid, crate::domain::auction::Id),
```
fleupold, line 119: Why do we need the auction id here but not in the other errors?

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -2,39 +2,151 @@
...
+        #[error(transparent)]
+        Auction(#[from] infra::persistence::error::Auction),
+        #[error(transparent)]
+        Solution(#[from] infra::persistence::error::Solution),
+    }
+    #[derive(Debug, thiserror::Error)]
+    pub enum Score {
+        /// Per CIP38, zero score solutions are rejected.
+        #[error(transparent)]
+        Zero(#[from] domain::competition::ZeroScore),
```
fleupold, line 131: Is there any value in wrapping this empty struct here? Why not just have it `Score::Zero` to be less of a nested error mess?

I see you are frequently creating new error structs that simply wrap the inner error types with an automatic conversion a la `#[from]`. While this is very convenient and avoids us having to think about the errors too much, I think it will cause confusion and a mess down the line. I think we should think as carefully about the error types (what errors make sense at this level and would be handled differently) instead of aggregating all potential suberror types in an somewhat automatically generated super error type.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        auction_id: domain::auction::Id,
+    ) -> Result<domain::settlement::Auction, error::Auction> {
+        let mut ex = self.postgres.pool.begin().await.context("begin")?;
+
+        let deadline = {
+            let scores = database::settlement_scores::fetch(&mut ex, auction_id)
+            .await
+            .context("fetch scores")?
+            // if score is missing, no competition / auction exist for this auction_id
+            .ok_or(error::Auction::MissingDeadline)?;
```
fleupold, line 153: Is `MissingDeadline` the right error for this? Isn't it more a missing solution or missing auction?

I think this method should return errors in ascending severity
1. There is simply no auction with this id
2. Something else about recovering the auction went wrong
3. Auction exists but there is no solution for this auction

crates/autopilot/src/domain/settlement/competition.rs
```diff
@@ -0,0 +1,18 @@
...
+    pub id: domain::auction::Id,
+    /// Auction external prices
+    pub prices: domain::auction::Prices,
+    /// Settlement should appear onchain before this block.
+    pub deadline: eth::BlockNo,
```
fleupold, line 14: Is this really something specific to the auction? 
To me an auction can exist without a solution. However, looking at `get_auction` we can not recover this type unless there is a solution. Therefore, would it make more sense to have this field on the solution (or make it recoverable even if there was no solution?

I guess this would bring us back to creating a `settlement::Solution` type which differs from `competition::Solution` in
1. Not having a solution id
2. Having a deadline
3. (Possibly not needing the prices)

I might actually prefer this setup

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        auction_id: domain::auction::Id,
+    ) -> Result<domain::competition::Solution, error::Solution> {
+        let mut ex = self.postgres.pool.begin().await.context("begin")?;
+
+        let (winner, score) = {
+            let scores = database::settlement_scores::fetch(&mut ex, auction_id)
+                .await
+                .context("fetch scores")?
+                // if score is missing, no competition / auction exist for this auction_id
+                .ok_or(error::Solution::MissingScore)?;
```
fleupold, line 235: Here also it's not really a missing score, but rather a missing solution (it's hard for the callsite to reason that a particular missing property like "score" should imply there has never been a valid solution for this auction).

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        // for now, this is a `good enough` solution to unblock other tasks.
+        let solver_competition = database::solver_competition::load_by_id(&mut ex, auction_id)
+            .await
+            .context("load solver competition")?
+            .ok_or(error::Solution::SolverCompetition(anyhow::anyhow!(
+                "missing solver competition"
+            )))?;
+        let competition: model::solver_competition::SolverCompetitionDB =
+            serde_json::from_value(solver_competition.json).map_err(|_| {
+                error::Solution::SolverCompetition(anyhow::anyhow!("json conversion"))
```
fleupold, line 258: Ignoring the actual error in the error will make it impossible to debug this error.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        }
+        let mut prices = HashMap::new();
+        for (token, price) in winning_solution.clearing_prices.clone().into_iter() {
+            prices.insert(
+                token.into(),
+                domain::auction::Price::new(price.into())
+                    .map_err(|_| error::Solution::Price(domain::auction::InvalidPrice))?,
+            );
+        }
+        Ok(competition::Solution::new(0, winner, score, orders, prices))
```
fleupold, line 301: Using 0 as an id here is 🤡 and an indicator that we maybe shouldn't reuse the competition struct.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        #[error("failed to read data from database")]
+        DbError(#[from] anyhow::Error),
+        #[error("failed dto conversion from database")]
+        DbConversion(&'static str),
+        #[error(transparent)]
+        Price(#[from] domain::auction::InvalidPrice),
+        #[error("deadline not found in the database")]
+        MissingDeadline,
+        #[error("quote not found in the database for an existing order")]
+        MissingQuote,
```
fleupold, line 325: Should this contain the order id for which the quote is missing to make it easier to debug this issue when oncall sees this error?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
     DbError(#[from] anyhow::Error),
-    #[error(transparent)]
-    Price(#[from] domain::auction::InvalidPrice),
+}
+
+pub mod error {
+    use super::*;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Auction {
```
fleupold, line 315: Can you document what those errors mean, to give some help for future maintainers as to how they should be interpreted?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+    #[derive(Debug, thiserror::Error)]
+    pub enum Solution {
+        #[error("failed to read data from database")]
+        DbError(#[from] anyhow::Error),
+        #[error(transparent)]
+        Price(#[from] domain::auction::InvalidPrice),
+        #[error("score not found in the database")]
+        MissingScore,
+        #[error(transparent)]
+        ZeroScore(#[from] domain::competition::ZeroScore),
```
fleupold, line 339: Needless wrap of empty struct

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,207 @@ impl Persistence {
...
+        #[error("deadline not found in the database")]
+        MissingDeadline,
+        #[error("quote not found in the database for an existing order")]
+        MissingQuote,
+        #[error("orders are missing from the auction")]
+        MissingOrders,
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Solution {
```
fleupold, line 331: ditto docs please. From just reading this I have no idea what `error::Solution::Price` is supposed to mean.

fleupold (changes requested):
Sorry I know we spend some time discussing this yesterday, but afterall I think we should
1. Introduce a separate `settlement::Solution` object to avoid shoe-horning things too much. The difference between the settlement module and competition module should be that competition handles everything during the auction runloop and settlement everything after when a settlement is observed (as we can see there are a few differences in the domain's view of the world between these phases). As for code duplication of the score computation, maybe we can create a free method for it which can be easily invoked from both `settlement::Solution` as well as `competition::Solution`
2. Rethink some of the error cases as I feel they aren't really clear and intuitive right now.
3. Double check that we are not regressing jit order surplus calculation.

sunce86:
Reopening. Requires rework though.

sunce86:
Ready for review. 

I've narrowed down the PR as much as possible. Initially, the PR was supposed to implement fetching of ALL data needed for
1. current functionality to work - calculate data for settlement_observations table
2. future functionality to work - circuit breaker rules, for which the full competition data is needed

I've decided to tackle only (1) in this PR. In general, will continue implementing `domain::Settlement` so that it does (1) and finally replaces legacy code (this refactor has taken too long). After that, will proceed with (2)

crates/autopilot/src/domain/settlement/auction.rs
sunce86: This now contains only data related to Auction so kept it as is.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,122 @@ impl Persistence {
...
+                        .context("fetch quote")?
+                        .map(dto::quote::into_domain)
+                        .transpose()
+                        .map_err(|_| error::Auction::DbConversion("quote overflow"))?
+                        .ok_or(error::Auction::MissingQuote)?;
+
+                    let policies =
+                        database::fee_policies::fetch(&mut ex, auction_id, ByteArray(order.0))
+                            .await
+                            .context("fetch fee policies")?;
```
squadgazzz, line 202: 3*N queries are made here, where N is the number of orders. How about making just 3 queries in total by reading all the orders at once, then all the quotes, and, in the end, all the policies? 

nit: Orders and quotes could be joined in the SQL, which remains to be a simple query.

crates/autopilot/src/domain/settlement/auction.rs
fleupold: I'm still a bit confused how this relates to `domain::auction::Auction`. E.g. why does it not need the orders (to later on check which orders in the solution are allowed to count for surplus)?

I also think the things that are on this struct but not on the other (fee_policies, jit_orders) should probably become part of the core auction as it needs to be decided before the runloop is officially started.

Maybe we can add a note/task that eventually these two structs could potentially be combined (or are at least married in some way).

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +137,122 @@ impl Persistence {
...
-    Price(#[from] domain::auction::InvalidPrice),
+}
+
+pub mod error {
+    use super::*;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Auction {
+        #[error("failed to read data from database")]
+        DbError(#[from] anyhow::Error),
```
fleupold, line 246: This blanket conversion seems dangerous (it's inviting to simply handle any error with ? which would be translated into a DB error even if it's unrelated to DB operations). Can you please create the DbError explicitly in the cases where it makes sense (ie. remove the `#[from]` here)?

fleupold (commented):
Looks largely good. Agree that we should reduce the number of roundtrips in `fetch_auction` (we have ~2k orders in the auction)

sunce86:
> I'm still a bit confused how this relates to domain::auction::Auction. E.g. why does it not need the orders (to later on check which orders in the solution are allowed to count for surplus)?

It does contain the list of all orders in domain::Auction in a bit non-obvious field:

```
/// Fee policies for all orders in the auction. For some orders, there may
/// be no fee policies.
pub fee_policies: HashMap<domain::OrderUid, Vec<domain::fee::Policy>>,
```

You are right that both structs represent the same thing. However, last time we discussed this, IIRC, we identified `executed` field from the `Order` as problematic to reconstruct from database, as it represents a screenshot of the order before being executed.

If we had the full domain::Auction saved to db, we could just fetch it and use it instead of domain::settlement::Auction. Instead, we try to reconstruct the Auction from what we have in the db, and since we cant reconstruct everything, we need a different struct 🤷‍♂️

crates/autopilot/src/infra/persistence/mod.rs
squadgazzz: Just realized that due to `sqlx` limitations, fetching all the fee policies at once won't be possible without postgres-specific queries: https://github.com/cowprotocol/services/pull/2809#discussion_r1675527726
But if we might have ~2k orders, IMO, it should still be considered to make 3 queries instead of 2k.

crates/autopilot/src/domain/settlement/auction.rs
```diff
@@ -0,0 +1,22 @@
...
+};
+
+#[derive(Debug)]
+pub struct Auction {
+    pub id: domain::auction::Id,
+    /// Auction external prices
+    pub prices: domain::auction::Prices,
+    /// Fee policies for all orders in the auction. For some orders, there may
+    /// be no fee policies.
+    pub fee_policies: HashMap<domain::OrderUid, Vec<domain::fee::Policy>>,
```
fleupold, line 15: Should we call this orders then (and comment that it also contains the fee policies)?

sunce86: Yes, renamed 👍

crates/autopilot/src/infra/persistence/mod.rs
sunce86: Transformed to 3 queries 👍

sunce86: Joining orders and quotes is a more involving change outside of scope of PR, should be tackled when we refactor solvable orders cache.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: removed

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+                .map_err(error::Auction::DbError)?;
+
+            // compile order data
+            let mut orders = HashMap::new();
+            for order in auction_orders.iter() {
+                let quote = quotes
+                    .get(order)
+                    .ok_or(error::Auction::MissingQuote(*order))?;
+
+                orders.insert(*order, vec![]);
```
Mateo-mro, line 216: it is not needed if you extend L220 to do `.or_default()`

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+                for policy in policies.get(&ByteArray(order.0)).unwrap_or(&vec![]) {
```
Mateo-mro, line 217: Instead of doing `.unwrap_or(&vec![])`, why not to iterate just if it is `Some(T)`?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
-    Price(#[from] domain::auction::InvalidPrice),
+}
+
+pub mod error {
+    use super::*;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Auction {
+        #[error("failed to read data from database")]
+        DbError(anyhow::Error),
```
Mateo-mro, line 260: you can have it with `#[source]`, that way the errors are chained with `anyhow::Error`.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+        #[error("failed dto conversion from database")]
+        DbConversion(anyhow::Error),
+        #[error("auction data not found in the database")]
+        Missing,
+        #[error(transparent)]
+        Price(#[from] domain::auction::InvalidPrice),
+        #[error("quote not found in the database for an existing order")]
```
Mateo-mro, line 267: Consider adding the order UID to the error log, something like:

```rust
#[error("quote not found in the database for an existing order: {0]")]
```

crates/database/src/fee_policies.rs
```diff
@@ -72,6 +73,33 @@ pub async fn fetch(
...
+    WHERE auction_id = $1
+    ORDER BY order_uid, application_order
+    "#;
+
+    let rows: Vec<FeePolicy> = sqlx::query_as::<_, FeePolicy>(QUERY)
+        .bind(auction_id)
+        .fetch_all(ex)
+        .await?;
+
+    let mut grouped_fee_policies: HashMap<OrderUid, Vec<FeePolicy>> = HashMap::new();
```
Mateo-mro, line 91: Shouldn't this logic be in the caller since this is just a database query function? :thinking:

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
             .await
             .map_err(Error::DbError)
     }
+
+    /// Get auction data.
+    pub async fn get_auction(
+        &self,
+        auction_id: domain::auction::Id,
+    ) -> Result<domain::settlement::Auction, error::Auction> {
```
squadgazzz, line 148: DB metrics should probably be added for each SQL query inside this function.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+                .into_iter()
+                .map(|order| domain::OrderUid(order.0))
+                .collect::<HashSet<_>>();
+
+            // get quotes for all auction orders
+            let quotes = self
+                .postgres
+                .read_quotes(auction_orders.iter())
+                .await
+                .map_err(error::Auction::DbError)?;
```
squadgazzz, line 201: As Felix [suggested in a separate PR](https://github.com/cowprotocol/services/pull/2809#discussion_r1679132985), let's not fetch all the quotes. Here is how I tried to address it: https://github.com/cowprotocol/services/pull/2809/files#diff-ba91e794defad0d0d318c65a444c3bd6bfc1c6cee8abc31c5af7eb07ae428093R24-R48

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+                .await
+                .context("fetch fee policies")
+                .map_err(error::Auction::DbError)?;
+
+            // compile order data
+            let mut orders = HashMap::new();
+            for order in auction_orders.iter() {
+                let quote = quotes
+                    .get(order)
+                    .ok_or(error::Auction::MissingQuote(*order))?;
```
squadgazzz, line 214: I think we still have orders without quotes in the DB. `MissingQuote` error should only be returned if there is no quote for the `PriceImprovement` fee policy.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -135,12 +140,133 @@ impl Persistence {
...
+
+                orders.insert(*order, vec![]);
+                for policy in policies.get(&ByteArray(order.0)).unwrap_or(&vec![]) {
+                    let policy = dto::fee_policy::into_domain(policy.clone(), quote)
+                        .map_err(error::Auction::DbConversion)?;
+                    orders.entry(*order).and_modify(|policies| {
+                        policies.push(policy);
+                    });
```
squadgazzz, line 222: Fee policies can be all converted to domain first and only then inserted into the HashMap along with the order using a single operation instead of N HashMap accesses and inserts.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: I want to have all orders from auction saved - even those without any fee policy attached.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: No special reason except it felt more readable this way

crates/database/src/fee_policies.rs
sunce86: I don't think it's meaningful to use this data without grouping in any way, so I put it there.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: timer?

squadgazzz: Yep

crates/autopilot/src/infra/persistence/mod.rs
Mateo-mro: but in this case there is an unneeded allocation :thinking:

crates/autopilot/src/infra/persistence/mod.rs
sunce86: added

crates/autopilot/src/infra/persistence/mod.rs
sunce86: added

crates/autopilot/src/infra/persistence/mod.rs

crates/autopilot/src/infra/persistence/mod.rs
sunce86: The orders are [not included](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/solvable_orders.rs#L268) in the `Auction` if they don't have a `Quote`.

I believe this was intentional decision by the team.

But still fixed, as this part of the code might change.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: true. first converted then inserted

crates/autopilot/src/infra/persistence/mod.rs
sunce86: added one for the whole function, will add more later if it takes too long

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -155,12 +159,164 @@ impl Persistence {
...
-    Price(#[from] domain::auction::InvalidPrice),
+}
+
+pub mod error {
+    use super::*;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Auction {
+        #[error("failed to read data from database: {0}")]
+        DbError(#[source] anyhow::Error),
```
fleupold, line 312: What does `#[source]` do and why is it useful here?

crates/autopilot/src/infra/persistence/dto/fee_policy.rs
```diff
@@ -48,3 +48,42 @@ pub fn from_domain(
...
+            },
+        },
+    };
+    Ok(policy)
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("failed to convert database data to domain data {0}")]
+    Conversion(#[from] anyhow::Error),
```
fleupold, line 86: nit: maybe call `Inconsistency` to cover both data not being there + date not being parseable.

fleupold (commented):
LGTM, mostly nits except for the `unrwap` which I think we should avoid where-ever possible in non startup code (otherwise we could end up crash looping)

crates/autopilot/src/infra/persistence/mod.rs
Mateo-mro: @fleupold `#[source]`  is used primarily for error chaining, helping you to trace back through nested errors. If we were to check an upper level which was the original cause, we couldn't see it, as `#[from}` squashes the original error. You can actually chain both.

@sunce86 implemented `#[source]` because of my suggestion. I personally always like to implement `#[source]` not to lose the original error and have all nested properly.

Mateo-mro: This is extremely helpful/powerful if we use `thiserror` crate all the time (since it forces you to define an enum per error instead of just a string, making it more robust and easy to act on specific errors)

squadgazzz (approved):
LGTM

crates/autopilot/src/infra/persistence/dto/fee_policy.rs
```diff
@@ -48,3 +48,42 @@ pub fn from_domain(
...
     }
 }
+
+pub fn try_into_domain(
+    policy: FeePolicy,
+    quote: Option<&domain::quote::Quote>,
+) -> Result<domain::fee::Policy, Error> {
+    let policy = match policy.kind {
+        FeePolicyKind::Surplus => domain::fee::Policy::Surplus {
+            factor: policy.surplus_factor.unwrap().try_into()?,
```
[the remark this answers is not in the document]
sunce86: You are right. Fixed.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -155,12 +159,164 @@ impl Persistence {
...
+        #[error("failed dto conversion from database: {0}")]
+        FeePolicy(anyhow::Error),
```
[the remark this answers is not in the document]
sunce86: Extra context contains order uid. I refactored error variant to actually contain specific types instead of anyhow::Error

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -155,12 +159,164 @@ impl Persistence {
...
+                        .any(|policy| {
+                            matches!(
+                                policy.kind,
+                                database::fee_policies::FeePolicyKind::PriceImprovement
+                            )
+                        })
+                        .then_some(order_uid)
+                }))
+                .await
+                .map_err(error::Auction::DbError)?;
```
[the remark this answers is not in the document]
sunce86: Context should be taken from inner function call

crates/autopilot/src/infra/persistence/dto/fee_policy.rs
sunce86: Renamed and refactored

sunce86:
@fleupold i've addressed the comments and dismissed your "request for change" as it's not relevant anymore for the current state of PR -> will merge
any additional comments can be addressed in any of the follow up PRs

PR #2737, opened 2024-05-14
Remove code related to old HTTP solver API

MartinquaXD:
# Description
Since we removed the last quoter that used the old API we can finally clean up some more stuff.

# Changes
- removed legacy http price estimator from price estimator factory
- deleted http solver
- deleted most of http solver model (and slightly moved around remaining bits)
- deleted legacy adapter and tests
- removed balancerv2 and univ3 subgraph urls from api and autopilot args
- slightly simplified how we construct price estimators in the factory but this can be cleaned up a lot more; just wanted to keep this PR mostly about deleting stuff instead of changing stuff to make it easier to review

## How to test
compiler and e2e tests


nlordell (approved):

Mateo-mro (approved):

crates/autopilot/src/run.rs
```diff
@@ -240,15 +229,12 @@ pub async fn run(args: Arguments) {
             UniV2BaselineSourceParameters::from_baseline_source(*source, &chain_id.to_string())
         })
         .chain(args.shared.custom_univ2_baseline_sources.iter().copied());
-    let (pair_providers, pool_fetchers): (Vec<_>, Vec<_>) = futures::stream::iter(univ2_sources)
+    let pair_providers: Vec<_> = futures::stream::iter(univ2_sources)
```
fleupold, line 232: This is still required for bad token detection? Do we have a plan somewhere on how to remove that beast as well?

fleupold (approved):

fleupold:
image

Who is this guy?

crates/autopilot/src/run.rs
MartinquaXD: Yeah, the bad token detection still uses it. :/
I don't have a very concrete plan how to move forward with this. My idea was to use our currently strictest quote verification mode `enforce-if-possible` to phase out bad token detection. There are a bunch of caveats around this so I would just try it out on staging and see how it goes. 🙏

PR #2739, opened 2024-05-15
Gas estimator driver config

squadgazzz:
### Background

In the process of adapting for Arbitrum compatibility, we encountered issues with gas price estimation. The default gas price estimator in our driver provided values significantly higher than what the RPC endpoint returns, resulting in orders being created with insufficient fees. This discrepancy led to solvers being unable to find profitable solutions due to excessively high gas prices.

The gas price estimator in the driver is currently hardcoded, while the autopilot/orderbook allows for flexible selection of the gas price estimator. This inconsistency has caused orders to be created with a low fee while solvers receive instructions based on a much higher gas price, preventing them from finding viable solutions.

### Changes

This change will allow us to use a Web3-based estimator, which provides more accurate gas prices for Arbitrum.

### DIFF
--- a/crates/driver/src/infra/blockchain/gas.rs
+++ b/crates/driver/src/infra/blockchain/gas.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 /// Wrapper around the gas estimation library.
 /// Also allows to add additional tip to the gas price. This is used to
 /// increase the chance of a transaction being included in a block, in case
@@ -10,6 +8,7 @@ use {
     ethcontract::dyns::DynWeb3,
     gas_estimation::{nativegasestimator::NativeGasEstimator, GasPriceEstimating},
 };
+use {crate::infra::config::file::GasEstimatorType, std::sync::Arc};
 
 type MaxAdditionalTip = eth::U256;
 type AdditionalTipPercentage = f64;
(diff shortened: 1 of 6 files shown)


crates/driver/src/infra/blockchain/gas.rs
```diff
@@ -10,26 +8,34 @@ use {
     ethcontract::dyns::DynWeb3,
     gas_estimation::{nativegasestimator::NativeGasEstimator, GasPriceEstimating},
 };
+use {crate::infra::config::file::GasEstimatorType, std::sync::Arc};
```
MartinquaXD, line 11: Mmmh, why does `cargo fmt` not complain that these imports are not merged into one? 🤔

crates/driver/src/infra/config/mod.rs
```diff
@@ -6,6 +6,8 @@ use crate::{
 pub mod file;
 pub use file::encoding;
 
+use crate::infra::config::file::GasEstimatorType;
```
MartinquaXD, line 9: Import should also be merged with imports above.

MartinquaXD (approved):
Change looks good to me. 👌

crates/driver/src/infra/blockchain/gas.rs
squadgazzz: Yeah, there is something weird with the gas_estimation::GasPriceEstimating, which breaks the formatter. My IDE, for example, can't resolve that type at all. Manually merged the imports.
image

Mateo-mro (approved):

PR #2746, opened 2024-05-16
Refactor domain::Settlement

sunce86:
# Description
Addresses most of the code review comments from previous PRs: https://github.com/cowprotocol/services/pull/2710 and https://github.com/cowprotocol/services/pull/2725

Lost of moving around and renaming stuff.


crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
+//! This module defines Solution as originated from a mined transaction
```
sunce86, line 1: Renamed a file. No new code.

crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
...
+            OrderUid,
+        },
+    },
+    ethcontract::{common::FunctionExt, tokens::Tokenize, U256},
+    std::collections::HashMap,
+};
+
+/// Settlement originated from a calldata of a settlement transaction.
+#[derive(Debug)]
+pub struct Solution {
```
MartinquaXD, line 25: I thought a "Solution" was supposed to be a solver's abstract plan of settling an auction.

crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+impl From<U256> for TradeFlags {
+    fn from(value: U256) -> Self {
+        Self(value)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
MartinquaXD, line 207: Still exports `Error` type and `error` module publicly. Is there a good reason to do so?

crates/autopilot/src/domain/settlement/tokenized.rs
```diff
@@ -5,7 +5,7 @@ use {
 };
 
 // Original type for input of `GPv2Settlement.settle` function.
-pub(super) type Settlement = (
+pub(super) type Solution = (
```
MartinquaXD, line 8: `pub(super)` seems a bit sketch to me. Looks like this is only used by the `solution` file (again not sure why on-chain stuff is related to a solution).
Instead a folder structure like this could help with that.
```
/solution/tokenized
         /mod
```
Where mod can now use types that are private in `tokenized`.

crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
...
+            .iter()
+            .map(|trade| trade.native_fee(prices))
+            .sum()
+    }
+
+    pub fn new(
+        calldata: &eth::Calldata,
+        domain_separator: &eth::DomainSeparator,
+    ) -> Result<Self, Error> {
+        let function = contracts::GPv2Settlement::raw_contract()
```
fleupold, line 81: Can this logic and the TradeFlag encoding be moved into `tokenized`? Seems more appropriate there since it's actually looking at the tokenised solution.

crates/autopilot/src/infra/blockchain/dto.rs
```diff
@@ -1,24 +1,29 @@
-impl TryFrom<web3::types::Transaction> for crate::domain::settlement::Transaction {
-    type Error = &'static str;
+impl TryFrom<(web3::types::Transaction, web3::types::TransactionReceipt)>
+    for crate::domain::settlement::Transaction
+{
+    type Error = anyhow::Error;
 
-    fn try_from(value: web3::types::Transaction) -> Result<Self, Self::Error> {
+    fn try_from(
+        (transaction, receipt): (web3::types::Transaction, web3::types::TransactionReceipt),
```
MartinquaXD, line 7: Try from looks very sketchy for tuples. Maybe it's time for a `try_new` constructor.

crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Encoding {
+        #[error("unable to decode settlement calldata: {0}")]
+        Decoding(#[from] web3::ethabi::Error),
+        #[error("unable to tokenize calldata into expected format: {0}")]
+        Tokenizing(#[from] ethcontract::tokens::Error),
+        #[error("unable to recover order uid: {0}")]
+        OrderUidRecover(#[from] tokenized::Error),
+    }
```
fleupold, line 227: I think this struct should be moved into tokenized as well as it's 100% encoding related.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -1,366 +1,56 @@
...
-pub mod transaction;
-pub use transaction::{Transaction, Tx};
+pub use solution::Solution;
 
-use super::competition;
-
-/// Settlement originated from a calldata of a settlement transaction.
-#[derive(Debug)]
+/// A transaction that executes a solution. Interacts with the settlement
+/// contract `settle` function.
```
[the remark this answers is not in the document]
MartinquaXD: Also maybe names like `OnChainSolution` or `ExecutedSolution` (executed on-chain bundled with the respective transaction) and `AbstractSolution` or `ProposedSolution` (`/solve` response) make sense here.

crates/autopilot/src/domain/settlement/solution.rs
```diff
@@ -0,0 +1,363 @@
...
+            Error::Encoding(auction, self)
+        }
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Score {
+        #[error(transparent)]
+        Trade(#[from] trade::Error),
+        #[error(transparent)]
+        Zero(#[from] competition::ZeroScore),
```
fleupold, line 240: Can we document these errors? Not sure what a `settlement::solution::error::Score::Trade` should mean. I also don't think they should be collapsed (I know it's convenient to just wrap an existing error type into another one), but I think the correct score error types should be
- Zero
- MissingPrice
- Math

Where trade::Error get translated accordingly.

MartinquaXD (commented):
Naming and descriptions are still lacking.

crates/autopilot/src/domain/settlement/solution.rs
sunce86: Related to https://github.com/cowprotocol/services/pull/2746#discussion_r1603506408

I refactored the code as Felix suggested: solution + transaction = settlement

crates/autopilot/src/infra/blockchain/mod.rs
```diff
@@ -13,7 +13,7 @@ use {
 };
 
 pub mod contracts;
-pub mod dto;
+mod dto;
```
fleupold, line 16: nit: I'm not sure dto is the right name for this (we are basically creating a model for infra vs. domain representations, but not a data transfer object). How do we do infra -> domain conversions in other places of the codebase? 

Maybe we can just lift this code up into the main mod?

crates/autopilot/src/infra/blockchain/mod.rs
```diff
@@ -113,24 +113,15 @@ impl Ethereum {
...
-        &self,
-        hash: eth::TxId,
-    ) -> Result<domain::settlement::transaction::Receipt, Error> {
-        self.web3
-            .eth()
-            .transaction_receipt(hash.0)
-            .await?
-            .map(|receipt| receipt.try_into().map_err(Error::IncompleteTransactionData))
-            .ok_or(Error::TransactionNotFound)?
+        let (transaction, receipt) = tokio::try_join!(
```
fleupold, line 116: I think this is a great improvement!

crates/autopilot/src/domain/settlement/mod.rs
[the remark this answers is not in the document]
MartinquaXD: > For example settlement::Solution versus competition::Solution

Yeah, I guess that could work as well. 👌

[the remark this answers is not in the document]
sunce86: Added a few more comments for important structs.

crates/autopilot/src/domain/settlement/solution.rs
sunce86: changed with the refactor

crates/autopilot/src/domain/settlement/tokenized.rs
sunce86: also resolved with refactor

crates/autopilot/src/infra/blockchain/dto.rs
sunce86: How would that look like? I am pretty happy with the existing option tbh

crates/autopilot/src/domain/settlement/solution.rs
sunce86: moved tokenizing part into `tokenized`

crates/autopilot/src/domain/settlement/solution.rs
sunce86: moved and refactored.

crates/autopilot/src/infra/blockchain/mod.rs
sunce86: We have a very similar module for persistence: https://github.com/cowprotocol/services/tree/main/crates/autopilot/src/infra/persistence/dto

I think `dto` was used to indicate that there is some conversion of data between layers happening. IMO, we should have a separate module for each section of persistence and each conversion to be put together with the functions called in that module, but we decided to put all persistence functions in a single file until it becomes large.

sunce86:
I believe all comments are addressed and PR is ready for another round of reviews.

crates/autopilot/src/domain/settlement/solution/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+/// this struct.
+///
+/// Referenced as [`settlement::Solution`] in the codebase.
+#[derive(Debug)]
+pub struct Solution {
+    trades: Vec<Trade>,
+    /// Data that was appended to the regular call data of the `settle()` call
+    /// as a form of on-chain meta data. This is used to associate a
+    /// solution with an auction for which this solution was picked as a winner.
+    auction_id: auction::Id,
```
fleupold, line 32: Given that we can't really do anything with this struct without having access to the actual auction (all public functions require some aspect of the auction, either prices or orders), I still think we should come up with a plan of how to construct it with an actual auction instance as a field (instead of just the ids). But not super important and definitely not something for this PR.

crates/autopilot/src/domain/settlement/solution/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Score {
+        #[error(transparent)]
+        Zero(#[from] competition::ZeroScore),
+        #[error("missing native price for token {0:?}")]
+        MissingPrice(eth::TokenAddress),
+        #[error(transparent)]
+        Math(trade::error::Math),
```
fleupold, line 148: nit, docs for the error types explaining why e.g. Zero or MissingPrice is a problem.

crates/autopilot/src/domain/settlement/solution/tokenized.rs
```diff
@@ -0,0 +1,192 @@
...
+            auction::{self, order},
+            eth,
+        },
+    },
+    app_data::AppDataHash,
+    ethcontract::{common::FunctionExt, tokens::Tokenize, Address, Bytes, U256},
+};
+
+// Original type for input of `GPv2Settlement.settle` function.
+pub struct Tokenized {
```
fleupold, line 15: nit: visibility `pub(super)`

crates/autopilot/src/infra/blockchain/mod.rs
fleupold: > We have a very similar module for persistence:

Well, you wrote this one very recently, I was more wondering on how things are done in the driver, where we had more fundamental discussions of how to adopt DDD in our codebase. There, dto is only used for the API objects we put in the wire to talk between core service components, which is not the case here.

Maybe this is a case for boundary::transaction and a `to_domain` function a la https://github.com/cowprotocol/services/blob/4a6f155004470414591b3b3e1c86118a8f1176d5/crates/driver/src/boundary/liquidity/balancer/v2/weighted.rs#L17

?

crates/autopilot/src/infra/blockchain/dto.rs
fleupold: The issue with `try_from` is that it makes it hard to find the code that actually implements the logic. If I'm on 

https://github.com/cowprotocol/services/blob/d417d97a197af0f4a15372b5e69b08493c843feb/crates/autopilot/src/infra/blockchain/mod.rs#L120-L125

and click on `try_into`, I'm taken to stdlib impl of `try_into` and it's hard to find this actual code here. This would not happen if this was a custom method that isn't part of any trait.

Is there any reason to use the standard trait here rather than a free function other than "syntactic appeal" ?

fleupold (approved):

crates/autopilot/src/infra/blockchain/dto.rs
MartinquaXD: > How would that look like? I am pretty happy with the existing option tbh

Could be as simple as
```rust
fn recover_transaction(tx: web3::types::Transaction, receipt: web3::types::TransactionReceipt) -> Result<domain::Transaction> {}
```
Also writing it out like this kind of points out that this does not seem like a dto functionality to me.

sunce86: Moved out and created a free function `into_domain` for it.

crates/autopilot/src/infra/blockchain/mod.rs
sunce86: Resolved as a free function under `mod.rs` which is still very small for blockchain so I guess it's fine.

crates/autopilot/src/domain/settlement/solution/mod.rs
sunce86: added

crates/autopilot/src/domain/settlement/solution/mod.rs
sunce86: If we want this, I would prefer to:
1. make `Transaction::input` field of a special `settlement::Calldata` instead of general `eth::Calldata`.
2. have a method `settlement::Calldata::get_auction() -> Option<AuctionId>`
3. Do this as a first thing in the `settlement::Solution` constructor. Then, having an auction id, fetch `Competition` from provided `infra::Persistence` instance.

PR #2759, opened 2024-05-28
Propagate the surplus capturing JIT-order owners to the solvers

Mateo-mro:
# Description
The solvers would need to know the JIT-order surplus capturing addresses (see e.g. https://github.com/cowprotocol/services/issues/2715#issuecomment-2130855593). For that the owner addresses are propagated to the solvers in the auction json.

# Changes
Propagate the list of surplus capturing JIT-order owners to the solvers in the auction json.

## How to test
1. Regression tests

### DIFF
--- a/crates/driver/src/infra/solver/dto/auction.rs
+++ b/crates/driver/src/infra/solver/dto/auction.rs
@@ -21,7 +21,6 @@ use {
         interaction::InteractionData,
         order::{BuyTokenDestination, SellTokenSource},
     },
-    primitive_types::H160,
     serde::Serialize,
     serde_with::serde_as,
     std::collections::{BTreeMap, HashMap},
@@ -309,6 +308,12 @@ impl Auction {
             tokens,
             effective_gas_price: auction.gas_price().effective().into(),
             deadline: auction.deadline().solvers(),
+            surplus_capturing_jit_order_owners: auction
+                .surplus_capturing_jit_order_owners()
+                .iter()
+                .cloned()
+                .map(Into::into)
+                .collect::<Vec<_>>(),
         }
     }
 }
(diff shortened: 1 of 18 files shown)


crates/driver/src/infra/solver/dto/auction.rs
```diff
@@ -349,8 +355,8 @@ struct Order {
     valid_to: u32,
     kind: Kind,
     #[serde(skip_serializing_if = "Option::is_none")]
-    receiver: Option<H160>,
-    owner: H160,
+    receiver: Option<eth::H160>,
+    owner: eth::H160,
```
Mateo-mro, line 359: Removed the `H160` import and make it all as `eth::H160` to keep consistency

crates/solvers/openapi.yml
```diff
@@ -658,6 +659,12 @@ components:
             by the caller.
           allOf:
             - $ref: "#/components/schemas/DateTime"
+        surplusCapturingJitOrderOwners:
+          type: array
+          items:
+            $ref: "#/components/schemas/Address"
+          description: |
+            List of addresses on whose surplus will count towards the objective value of their solution (unlike other orders that were created by the solver).
```
squadgazzz, line 667: We need the same for the `gnosis/solvers` repo

squadgazzz (approved):

sunce86 (commented):
Also solvers-dto crate needs to be updated

crates/solvers/openapi.yml
Mateo-mro: @squadgazzz good point! https://github.com/gnosis/solvers/pull/28

fleupold (approved):
You probably know, but this will also require a change in https://github.com/gnosis/solvers/blob/ca55c3cca15630891f653bca077a8a0c6b364c39/Cargo.toml#L49

Mateo-mro:
> You probably know, but this will also require a change in https://github.com/gnosis/solvers/blob/ca55c3cca15630891f653bca077a8a0c6b364c39/Cargo.toml#L49

@fleupold yup, the PR is ready, I just need to merge this first and then push the tag.

PR #2765, opened 2024-05-31
Implement e2e tests for JIT orders

Mateo-mro:
# Description
Implement e2e tests for JIT orders with a dummy/mocked solver.

The quotes are returned by the baseline solver, but the solution is returned by the mocked solver.


crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl From<&'static str> for Error {
+    fn from(message: &'static str) -> Self {
+        Self { message }
+    }
+}
+
+const REQUEST_BODY_LIMIT: usize = 10 * 1024 * 1024;
+
+pub struct Api {
```
Mateo-mro, line 119: This file is cramped with many things. I would/will split it in a better way. Just did it like that in a rush to have it ready.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,245 @@
...
+    // Kubernetes sends sigterm, whereas locally sigint (ctrl-c) is most common.
+    let mut interrupt = unix::signal(SignalKind::interrupt()).unwrap();
+    let mut terminate = unix::signal(SignalKind::terminate()).unwrap();
+    tokio::select! {
+        _ = interrupt.recv() => (),
+        _ = terminate.recv() => (),
+    };
+}
+
+#[derive(Clone, Debug)]
```
MartinquaXD, line 165: Eventually this should be moved into a separate function.
We could have 1 file implementing the mock solver and another file that contains a bunch of helper functions to build the solution that the solver is supposed to return (like this one).

crates/e2e/tests/e2e/jit_orders.rs
```diff
@@ -0,0 +1,226 @@
...
+                    &onchain.contracts().domain_separator,
+                    SecretKeyRef::from(&SecretKey::from_slice(solver.private_key()).unwrap()),
+                ),
+                executed_amount: to_wei(10),
+            }),
+            solvers_dto::solution::Trade::Fulfillment(solvers_dto::solution::Fulfillment {
+                executed_amount: to_wei(10),
+                fee: Some(0.into()),
+                // Dummy as it will be overwritten in the solver
+                order: [0; 56],
```
MartinquaXD, line 133: I think instead of having logic inside the solver to overwrite dummy data in the test we should first create the user order with the orderbook which returns the actual order uid and then write that into the solution we want to return.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,245 @@
...
+        // Return the same order UID as the order UID in the auction
+        solution
+            .trades
+            .iter_mut()
+            .filter_map(|trade| match trade {
+                Trade::Fulfillment(fulfillment) => Some(fulfillment),
+                Trade::Jit(_) => None,
+            })
+            .zip(auction.orders.iter())
+            .for_each(|(fulfillment, order)| fulfillment.order = order.uid);
```
MartinquaXD, line 48: I think the solver should be as dumb as possible and only return exactly what was configured in the test. The test just has to be structured such that you know the order uids before spawning the solver. Which should be okay because you can spawn the orderbook before the autopilot and the driver for the mock solver. (would require 2 separate drivers which might have inspired this workaround?).

MartinquaXD (commented):
Nice!
This is pretty much what I had in mind. The code still needs to be split up a bit and moved to more appropriate places but it should be enough to build a PoC e2e test for the CoW AMM.

crates/e2e/src/setup/solver.rs
Mateo-mro: Yes, I fully agree, once we have all the tests I want to clean it up for merge :ok_hand:

crates/e2e/Cargo.toml
```diff
@@ -10,6 +10,7 @@ app-data = { path = "../app-data" }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 autopilot = { path = "../autopilot" }
+axum = { workspace = true }
```
squadgazzz, line 13: I'm just curious, why not warp, which is already used in the `ZeroExApi` mock?

crates/e2e/tests/e2e/jit_orders.rs
MartinquaXD: > A way to solve this issue would be implementing some type of mspc

Had this PR marinating in my head and came back to suggest exactly this. That way you can super easily set up the solver and adjust as is needed. This also allows the same mock solver to return multiple different solutions throughout the test which I like! I just would use a `Mutex` instead of a `channel` to make it as simple as possible as a channel would introduce edge cases like:
1. how do I handle a tester sending lots and lots of solutions into the channel
2. should I return every solution exactly once or reuse it until a new solution appears
3. probably more that I currently don't think about

With a mutex it will be very clear what should happen just from reading the test case.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,243 @@
...
+};
+
+pub struct Mock {
+    solution: Solution,
+}
+
+impl Mock {
+    pub fn new(solution: Solution) -> Self {
+        Self { solution }
+    }
```
MartinquaXD, line 33: With the `Mutex` solution this could be replaced by a default implementation.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,243 @@
...
+            .for_each(|(fulfillment, order)| fulfillment.order = order.uid);
+
+        Solutions {
+            solutions: vec![solution],
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Config {
```
MartinquaXD, line 57: With the `Mutex` this could be dropped altogether.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub addr: SocketAddr,
+    pub solution: Solution,
+}
+
+pub async fn run_mock(config: Config, bind: Option<oneshot::Sender<SocketAddr>>) {
```
MartinquaXD, line 62: I think `bind` can be non-optional since there should be no need to spawn a solver of which you don't want to know the port of.

MartinquaXD:
Since I needed the suggested changes for the cow amm test anyway I went ahead and implemented them. The PR can still be cleaned up but at least the functionality of configuring a solution after spawning the solver is there now.

crates/e2e/Cargo.toml
Mateo-mro: I'd like to avoid using `wrap` at all costs. The `wrap` crate made sense 2 years ago when `axum` was in a very early stage and didn't implement many features. Nowadays, `axum` has everything which `wrap` has, but better.

crates/e2e/src/setup/solver.rs
```diff
@@ -0,0 +1,175 @@
...
+        order::{BuyTokenDestination, OrderData, OrderKind, SellTokenSource},
+        signature::EcdsaSigningScheme,
+        DomainSeparator,
+    },
+    reqwest::Url,
+    solvers_dto::{
+        auction::Auction,
+        solution::{Asset, Kind, Solution, Solutions},
+    },
+    std::sync::{Arc, Mutex},
```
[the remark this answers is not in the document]
Mateo-mro: @MartinquaXD oh I see, apparently the improved that, normal mutexes used to block tokio threads time ago, I lived that sadly :sweat: ok in that case let's leave it like this for the test :ok_hand:

crates/e2e/tests/e2e/jit_orders.rs
```diff
@@ -0,0 +1,227 @@
...
+            >= to_wei(5)
+    );
+    // Since the fee is 0 in the custom solution, the balance difference has to be
+    // exactly 10 wei
+    assert_eq!(
+        solver_balance_before
+            .checked_sub(solver_balance_after)
+            .unwrap(),
+        to_wei(10)
+    );
```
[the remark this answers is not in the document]
Mateo-mro: @MartinquaXD should we apply this logic to all the e2e tests? (I could quickly do that in another PR)

crates/e2e/src/setup/solver.rs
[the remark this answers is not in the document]
MartinquaXD: The regular Mutexes still do block the whole thread. It's just that there is now a clippy lint that complains if you hold a regular mutex over an `.await` so the issue is mostly mitigated in our CI.
You are still screwed if your mutex gets locked too long (couple ms or more) but since we usually have relatively little contention on mutexes and mostly clone stuff out of it we can usually use regular mutexes just fine.

crates/e2e/tests/e2e/jit_orders.rs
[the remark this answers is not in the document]
MartinquaXD: I think there is no need for that. Let's keep using best practices for new code and refactor old stuff that gets in our way but I don't see a lot of value in going over all the old test code and fixing that up.

crates/e2e/Cargo.toml
squadgazzz: Yeah, but since `axum` is used only to implement dummy mocks, 99% of its futures won't be used and its configuration seems more involved.

crates/e2e/tests/e2e/jit_orders.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    // Fund trader account
+    token_a.mint(trader_a.address(), to_wei(100)).await;
+    token_b.mint(trader_a.address(), to_wei(100)).await;
+
+    // Fund solver account
+    token_a.mint(solver.address(), to_wei(1000)).await;
+    token_b.mint(solver.address(), to_wei(1000)).await;
+
+    // Create and fund Uniswap pool
```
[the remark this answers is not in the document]
Mateo-mro: @MartinquaXD we need to pick our evil, either we create the pool or I believe we should return `interactions` from the solver response (which also adds overhead).

crates/e2e/Cargo.toml
Mateo-mro: Majority of the configuration in `axum` is not really needed. I just deleted it for simplicity.

crates/e2e/src/setup/solver/mock.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn configure_solution(&self, solution: Option<Solution>) {
+        *self.solution.lock().unwrap() = solution;
+    }
+}
+
+impl Default for Mock {
+    fn default() -> Self {
+        let solution = Arc::new(Mutex::new(None));
+
+        let app = axum::Router::new()
```
Mateo-mro, line 35: I simplified all of this since it wasn't needed.

crates/e2e/src/setup/solver/solution.rs
```diff
@@ -0,0 +1,93 @@
+use {
```
Mateo-mro, line 1: I decided to call this module `solution` since the jit order is part of the solution (as in dto mod)

crates/e2e/tests/e2e/jit_orders.rs
```diff
@@ -0,0 +1,223 @@
...
+    .sign(
+        EcdsaSigningScheme::Eip712,
+        &onchain.contracts().domain_separator,
+        SecretKeyRef::from(&SecretKey::from_slice(trader_a.private_key()).unwrap()),
+    );
+    let order_id = services.create_order(&order).await.unwrap();
+    let limit_order = services.get_order(&order_id).await.unwrap();
+    assert_eq!(limit_order.metadata.class, OrderClass::Limit);
+
+    mock_solver.configure_solution(Some(Solution {
```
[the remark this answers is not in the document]
Mateo-mro: @squadgazzz see https://github.com/cowprotocol/services/pull/2765#discussion_r1622476172

crates/e2e/tests/e2e/jit_orders.rs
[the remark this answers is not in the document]
MartinquaXD: I remember you commented about this causing issues. Is this still in the works?

MartinquaXD (commented):
PR looks good to me.
Just one nit remaining about what seems like an overly complex univ2 pool setup.

Mateo-mro:
> PR looks good to me. Just one nit remaining about what seems like an overly complex univ2 pool setup.

@MartinquaXD yes, I am currently working on that. I plan to push it today :ok_hand:

crates/e2e/tests/e2e/jit_orders.rs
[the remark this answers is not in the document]
Mateo-mro: @MartinquaXD yes, it is. Working on that.

MartinquaXD (approved):
Nice!

crates/e2e/src/setup/solver/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+async fn shutdown_signal() {
+    // Intercept main signals for graceful shutdown.
+    // Kubernetes sends sigterm, whereas locally sigint (ctrl-c) is most common.
+    let mut interrupt = unix::signal(SignalKind::interrupt()).unwrap();
+    let mut terminate = unix::signal(SignalKind::terminate()).unwrap();
+    tokio::select! {
+        _ = interrupt.recv() => (),
+        _ = terminate.recv() => (),
+    };
+}
```
squadgazzz, line 16: Do we really need it? Since the server is started inside the test environment, all threads, including the background ones, are dropped, and neither sigint nor sigkill are sent. Just to make sure, try adding a log inside the tokio::select. I couldn't find any log there.

squadgazzz (approved):

crates/e2e/src/setup/solver/mod.rs
Mateo-mro: Yes, as @MartinquaXD says, if you are running a local test in debug mode and you kill the process because you stop it/overflow/whatever, then the demon keeps running making the port unavailable, and then you need to find the process, kill it, etc.

PR #2771, opened 2024-06-06
Fix CI after foundry update

MartinquaXD:
# Description
https://github.com/foundry-rs/foundry/issues/8035 changed `anvil` to compute the gas price differently. This seems like a breaking change since under some conditions the new logic computes a different gas price than the original one.

This broke 2 of our tests which both involve waiting for the system to flush out cached liquidity data to generate new quotes. For `e2e::e2e quoting::local_node_test` it was enough to just bump the `--base-fee` to `1` (without breaking any other test) and for `e2e::e2e protocol_fee::local_node_combined_protocol_fees` I needed to make the condition to detect flushed liquidity stricter.
Instead of waiting for any change in quotes (which are now triggered too early because of tiny changes in the gas price) we now wait for the quote to be at least twice as good as the old quote.

# Changes
- bump `--base-fee` to `1` in `anvil` args
- stricter synchronization logic for `e2e::e2e protocol_fee::local_node_combined_protocol_fees`
- use nightly version of foundry again in CI

### DIFF
--- a/crates/e2e/src/setup/mod.rs
+++ b/crates/e2e/src/setup/mod.rs
@@ -58,7 +58,7 @@ where
 
 static NODE_MUTEX: Mutex<()> = Mutex::new(());
 
-const DEFAULT_FILTERS: [&str; 9] = [
+const DEFAULT_FILTERS: &[&str] = &[
     "warn",
     "autopilot=debug",
     "driver=debug",
@@ -74,7 +74,7 @@ fn with_default_filters<T>(custom_filters: impl IntoIterator<Item = T>) -> Vec<S
 where
     T: AsRef<str>,
 {
-    let mut default_filters: Vec<_> = DEFAULT_FILTERS.into_iter().map(String::from).collect();
+    let mut default_filters: Vec<_> = DEFAULT_FILTERS.iter().map(|s| s.to_string()).collect();
     default_filters.extend(custom_filters.into_iter().map(|f| f.as_ref().to_owned()));
 
     default_filters
--- a/crates/e2e/tests/e2e/protocol_fee.rs
+++ b/crates/e2e/tests/e2e/protocol_fee.rs
@@ -231,6 +231,9 @@ async fn combined_protocol_fees(web3: Web3) {
         )
         .await
         .unwrap();
+        // Only proceed with test once the quote changes significantly (2x) to avoid
+        // progressing due to tiny fluctuations in gas price which would lead to
+        // errors down the line.
         new_market_order_quote.quote.buy_amount > market_quote_before.quote.buy_amount * 2
     })
     .await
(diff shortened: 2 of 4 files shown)


crates/e2e/src/setup/mod.rs
```diff
@@ -74,7 +74,7 @@ fn with_default_filters<T>(custom_filters: impl IntoIterator<Item = T>) -> Vec<S
 where
     T: AsRef<str>,
 {
-    let mut default_filters: Vec<_> = DEFAULT_FILTERS.into_iter().map(String::from).collect();
+    let mut default_filters: Vec<_> = DEFAULT_FILTERS.iter().map(|s| s.to_string()).collect();
```
Mateo-mro, line 77: why this change? :thinking:

crates/e2e/tests/e2e/protocol_fee.rs
```diff
@@ -231,7 +231,7 @@ async fn combined_protocol_fees(web3: Web3) {
         )
         .await
         .unwrap();
-        new_market_order_quote.quote.buy_amount != market_quote_before.quote.buy_amount
+        new_market_order_quote.quote.buy_amount > market_quote_before.quote.buy_amount * 2
```
Mateo-mro, line 234: from where is coming the `2`?

crates/e2e/src/setup/mod.rs
MartinquaXD: Whenever you want to add or remove a log filter from that list you also have to update its type because it's a statically sized array. With this change I convert the thing into a fat pointer which knows its own length so you don't have to jump through these hoops anymore. Technically not needed for this PR but it annoyed me yet again so I just wanted to finally fix it.

crates/e2e/tests/e2e/protocol_fee.rs
MartinquaXD: This `2` isn't really based in anything. The actual new price is orders of magnitude bigger and the difference in buy amount due to gas prices is way smaller than that. I just wanted to make the smallest possible change without having to deal with floating point math.
If you feel like it's helpful or necessary I can add a comment but I wouldn't want to make this a constants since it's not relevant for anything else.

squadgazzz (approved):
LG, also a comment on the `new_market_order_quote.quote.buy_amount > market_quote_before.quote.buy_amount * 2` would be helpful.

crates/e2e/src/setup/mod.rs
Mateo-mro: oh, I understand your point, got it! :ok_hand: thanks for the explanation!

crates/e2e/tests/e2e/protocol_fee.rs
Mateo-mro: Thanks a lot! :pray: please, could you add a comment for it?

Mateo-mro (approved):

crates/e2e/tests/e2e/protocol_fee.rs
MartinquaXD: Done

fleupold (approved):

PR #2773, opened 2024-06-07
Run `cargo update`

MartinquaXD:
# Description
Updates all dependencies within the constraints of `cargo.toml`.

Also updates forked version of `warp` to latest commit.
Ideally we'd like to get rid of `warp` altogether since it's no longer really maintained but this requires a bunch of changes in the `api` so until then we should at least update the version as long as it's easy to do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.22.0"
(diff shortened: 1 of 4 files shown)


squadgazzz (approved):
Are we going to include it in the release on Monday?

MartinquaXD:
Fair point. 😅 
Can wait with merging until Monday. Not really urgent anyway.

Mateo-mro (approved):
Niiicee :fire:

PR #2781, opened 2024-06-14
Discard deny listed solver solutions

MartinquaXD:
# Description
Currently solvers that get deny listed by the circuit breaker due to rule violations still receive `/solve` and subsequent `/settle` calls by the autopilot.

# Changes
Discards any solution returned by `/solve` where the given submission address is currently deny listed. Since these solutions will no longer get ranked in the competition we'll also not call `/settle` for them anymore.

## How to test
e2e tests should still pass
I didn't add the test case that verifies that a deny listed solver actually gets their solutions removed but if that doesn't work this PR would not make it worse than the status quo.


Mateo-mro (approved):
Nice!!

squadgazzz (approved):
LGTM

PR #2782, opened 2024-06-17
[EASY] Exclude 2 subsequent numbers from the range

squadgazzz:
Follow up to the https://github.com/cowprotocol/services/pull/2750

2 subsequent numbers should not be compiled into a range.

### DIFF
--- a/crates/ethrpc/src/buffered.rs
+++ b/crates/ethrpc/src/buffered.rs
@@ -374,16 +374,7 @@ fn format_indices_as_ranges(indices: BTreeSet<usize>) -> anyhow::Result<String> 
         // Otherwise, there is no need to accumulate the range anymore. Append
         // the range to the result string.
         } else {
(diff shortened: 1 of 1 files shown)


Mateo-mro (approved):

crates/ethrpc/src/buffered.rs
```diff
@@ -388,9 +391,11 @@ fn format_indices_as_ranges(indices: BTreeSet<usize>) -> anyhow::Result<String>
         }
     }
 
-    // Append the last range or a single value.
+    // Append the remaining data.
```
MartinquaXD, line 394: Just noticed that this code it duplicated.
You could add a function like `append_sequence(buffer, start, end)` which encapsulates it.

MartinquaXD (approved):
Change looks correct but makes it apparent that we could profit from reusing code with functions.

PR #2783, opened 2024-06-19
Partial json assertion macro

squadgazzz:
[As suggested in another PR](https://github.com/gnosis/solvers/pull/39#discussion_r1645701146), extracts partial json assertion macro into the `shared` crate to use across crates/repos.

> This pull request's key enhancement is the introduction of macros for partially asserting the contents of POST requests. This method allows us to validate the request structure and static fields while excluding the dynamically generated random values from the assertion checks.

Also, this macro produces meaningful error instead of generic `left json is not equal to right json`, which saves some time when dealing with json structure changes and failing tests.

Once merged, [this PR](https://github.com/gnosis/solvers/pull/39) can be closed.

### DIFF
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -496,6 +496,7 @@ impl Maintaining for UniswapV3PoolFetcher {
 mod tests {
     use {
         super::*,
+        crate::assert_json_matches,
         contracts::uniswap_v3_pool::event_data::{Burn, Mint, Swap},
         ethcontract::EventMetadata,
         serde_json::json,
@@ -570,7 +571,7 @@ mod tests {
         };
 
         let serialized = serde_json::to_value(pool).unwrap();
-        assert_eq!(json, serialized);
+        assert_json_matches!(json, serialized, []);
     }
 
     #[test]
--- a/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
+++ b/crates/shared/src/sources/uniswap_v3/pool_fetching.rs
@@ -571,7 +571,7 @@ mod tests {
         };
 
         let serialized = serde_json::to_value(pool).unwrap();
-        assert_json_matches!(json, serialized, []);
+        assert_json_matches!(json, serialized);
     }
 
     #[test]
--- a/crates/shared/src/test_utils.rs
+++ b/crates/shared/src/test_utils.rs
@@ -246,12 +246,25 @@ mod tests {
 
     #[test]
     #[should_panic(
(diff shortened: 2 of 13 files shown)


Mateo-mro (approved):

crates/shared/src/test_utils.rs
```diff
@@ -0,0 +1,266 @@
...
+///
+/// # Examples
+///
+/// ```
+/// let actual = json!({"user": {"id": 1, "name": "Alice", "email": "alice@example.com"}});
+/// let expected = json!({"user": {"id": 1, "name": "Alice", "email": "bob@example.com"}});
+/// assert_json_matches!(actual, expected, ["user.email"]);
+/// ```
+#[macro_export]
+macro_rules! assert_json_matches {
```
MartinquaXD, line 30: Since the improved error messages are indeed nice could we have 2 macros?
`assert_json_matches` and `assert_json_matches_excluding`

crates/shared/src/sources/uniswap_v3/pool_fetching.rs
```diff
@@ -570,7 +571,7 @@ mod tests {
         };
 
         let serialized = serde_json::to_value(pool).unwrap();
-        assert_eq!(json, serialized);
+        assert_json_matches!(json, serialized, []);
```
MartinquaXD, line 574: Were these the only instances where we tested a JSON for equality so for?
What was your process for finding these?

crates/shared/src/test_utils.rs
```diff
@@ -0,0 +1,266 @@
...
+                },
+                "enabled": false,
+            }
+        });
+        assert_json_matches!(json_a, json_b, ["user.profile.timestamp", "user.enabled"])
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "JSON did not match with the exclusion of specified paths: Mismatch at \
```
MartinquaXD, line 190: Might be nice to also print the 2 JSONs. In case there are a bunch of differences you could spot them all at once instead of having to rerun the test 5 times until you resolved all of them.

squadgazzz: > Might be nice to also print the 2 JSONs.

Added the jsons also.

> In case there are a bunch of differences you could spot them all at once instead of having to rerun the test 5 times until you resolved all of them.

That would probably require the use of an errors accumulator instead of returning a single error from the `json_matches_excluding` function.

MartinquaXD (approved):

crates/shared/src/sources/uniswap_v3/pool_fetching.rs
squadgazzz: I was searching by `json!` macro usage. Added it also into the `models` crate.

PR #2787, opened 2024-06-21
Index native prices of tokens traded by CoW AMMs

### LINKED ISSUE #2774: chore: Automatically index CoW AMMs

Mateo-mro:
# Description
Index native prices of tokens traded by CoW AMMs

# Changes
- Fetch native prices for the tokens the CoW AMM trades
- Send those prices in the auction payload

## How to test
1. Acceptance test

## Related Issues

Fixes #2774

### DIFF
--- a/crates/autopilot/src/run.rs
+++ b/crates/autopilot/src/run.rs
@@ -478,6 +478,7 @@ pub async fn run(args: Arguments) {
             args.protocol_fee_exempt_addresses.as_slice(),
             args.enable_multiple_fees,
         ),
+        cow_amm_registry.clone(),
     );
 
     let liveness = Arc::new(Liveness::new(args.max_auction_age));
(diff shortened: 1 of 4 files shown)


Mateo-mro:
Closing PR since these changes are included in https://github.com/cowprotocol/services/pull/2790

PR #2788, opened 2024-06-24
Blockscout API key

squadgazzz:
Blockscout's public API has 50RPS globally which is rate limited at the moment. To avoid that, an API key can be used with a 10RPS allowance.

### DIFF
--- a/crates/shared/src/bad_token/token_owner_finder/blockscout.rs
+++ b/crates/shared/src/bad_token/token_owner_finder/blockscout.rs
@@ -12,6 +12,7 @@ use {
 pub struct BlockscoutTokenOwnerFinder {
     client: Client,
     base: Url,
+    api_key: Option<String>,
     rate_limiter: Option<RateLimiter>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/shared/src/bad_token/token_owner_finder/blockscout.rs
```diff
@@ -53,8 +60,13 @@ impl BlockscoutTokenOwnerFinder {
             .append_pair("action", "getTokenHolders")
             .append_pair("contractaddress", &format!("{token:#x}"));
 
+        // Don't log the API key!
         tracing::debug!(%url, "Querying Blockscout API");
 
+        if let Some(api_key) = &self.api_key {
```
Mateo-mro, line 66: Are we use the API key is appended to the query with the key `apiKey`?

squadgazzz: As the [API docs](https://docs.blockscout.com/for-users/my-account/api-keys#use-api-key) state.

Mateo-mro (approved):

MartinquaXD (approved):
Good change. I hoped this could be managed in the proxy but that doesn't seem practical if the key gets passed via a query parameter instead of a header.

PR #2789, opened 2024-06-27
(Proposal):  Remove additional_deadline_for_rewards

Mateo-mro:
# Description
It seems the field `additional_deadline_for_rewards` is not used anywhere. It is used to calculate the `block_deadline` which is stored in the database as part of the competition. But it seems it has no other effect. Maybe is it legacy code?

# Changes
Fully remove `additional_deadline_for_rewards`.

## How to test
1. Regression tests

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -47,7 +47,6 @@ pub struct RunLoop {
     pub solvable_orders_cache: Arc<SolvableOrdersCache>,
     pub market_makable_token_list: AutoUpdatingTokenList,
     pub submission_deadline: u64,
-    pub additional_deadline_for_rewards: u64,
     pub max_settlement_transaction_wait: Duration,
     pub solve_deadline: Duration,
     pub in_flight_orders: Arc<Mutex<Option<InFlightOrders>>>,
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):

crates/autopilot/src/run_loop.rs
```diff
@@ -166,9 +165,7 @@ impl RunLoop {
 
             let mut prices = BTreeMap::new();
             let mut fee_policies = Vec::new();
-            let block_deadline = competition_simulation_block
-                + self.submission_deadline
-                + self.additional_deadline_for_rewards;
+            let block_deadline = competition_simulation_block + self.submission_deadline;
```
squadgazzz, line 168: `additional_deadline_for_rewards` has `5` default value. The block deadline is reduced by that value. Is it fine?

Mateo-mro: @squadgazzz the block deadline is not used anywhere as I mentioned in the description. It is used to just store it in the competition. What's more, the overall idea with this change (and the upcoming one) is to reduce it to 3 blocks.

squadgazzz (approved):

PR #2790, opened 2024-06-28
Cow amm driver support

MartinquaXD:
# Description
Adds full support for cow amms in the backend. This includes:
* new contracts in `build.rs`
* `cow-amm` crate which encapsulates indexing of pools and generating template orders
* autopilot adds cow amms to surplus capturing jit order owners and indexes their native prices
* driver generates template orders for all relevant amms
* e2e test based on a fork of mainnet

Kudos to @Mateo-mro for implementing the indexing logic and coming up with the architecture that allowed us to create a single `Registry` that is able to index multiple different contracts to have a simpler API.
Due to some misconceptions on my part the helper contract needed to be adjusted to make indexing of v1 cow amms possible this led to significant changes in the `cow-amm` crate so I offer this PR to give the whole picture instead of refactoring on top of existing PRs.

## How to test
e2e test covers that the driver is able to suggest template orders for imbalanced cow amms and that the driver is able to settle them. Also verifies that the autopilot adds all amms to the surplus capturing jit order owners in the auction sent to the driver.


crates/cow-amm/src/amm.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Converts a successful response of the CowAmmHelper into domain types.
+    /// Can be used for any contract that correctly implements the CoW AMM
+    /// helper interface.
+    fn convert_orders_reponse(
+        &self,
+        order: RawOrder,
+        signature: Bytes<Vec<u8>>,
+        pre_interactions: Vec<RawInteraction>,
+        post_interactions: Vec<RawInteraction>,
+    ) -> Result<(
```
Mateo-mro, line 67: nit: what about returning a struct containing all the fields? big tuples are difficult to track, and easy to mix (e.i. from the signature of the function it is not known which one is pre_interactions and which one post_interactions)

crates/cow-amm/src/factories.rs
```diff
@@ -0,0 +1,46 @@
...
+    anyhow::Result,
+    contracts::{cow_amm_legacy_helper::Event, CowAmmLegacyHelper},
+    ethcontract::common::DeploymentInformation,
+    shared::impl_event_retrieving,
+};
+
+#[async_trait::async_trait]
+pub trait Deployment: Sync + Send {
+    /// Returns the AMM deployed in the given Event.
+    async fn deployed_amm(&self, helper: &CowAmmLegacyHelper) -> Result<Option<Amm>>;
```
Mateo-mro, line 12: Why to have `Result<Option<>>`, it seems to be always `Some()`

crates/cow-amm/src/factories.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+pub mod legacy {
```
Mateo-mro, line 15: Do we plan to add more besides `legacy`?

crates/cow-amm/src/registry.rs
```diff
@@ -0,0 +1,171 @@
...
+            .context("cow amm storage missing for factory")?;
+
+        for event in events {
+            let meta = event
+                .meta
+                .as_ref()
+                .ok_or_else(|| anyhow::anyhow!("Event missing meta"))?;
+            let block_number = meta.block_number;
+            if let Some(cow_amm) = event.data.deployed_amm(&storage.helper).await? {
+                tracing::debug!(?cow_amm, "indexed new cow amm");
```
Mateo-mro, line 145: This is a useful trace, maybe we can make it `info`, wdyt?

crates/cow-amm/src/factories.rs
MartinquaXD: Yes, we'll have to have a separate factory logic for the balancer implementation.

crates/cow-amm/src/factories.rs
MartinquaXD: I believe the balancer implementation of CoW AMMs can issue more events than just the ones we are interested in.
So if we index an event that did not deploy an amm it should return `None` and if we encountered an error during indexing it should return `Err`.

crates/e2e/tests/e2e/cow_amm.rs
```diff
@@ -347,3 +352,184 @@ async fn cow_amm(web3: Web3) {
...
+    let order_id = services.create_order(&order).await.unwrap();
+    let limit_order = services.get_order(&order_id).await.unwrap();
+    assert_eq!(limit_order.metadata.class, OrderClass::Limit);
+
+    // Drive solution
+    tracing::info!("Waiting for trade.");
+    wait_for_condition(TIMEOUT, || async {
+        // Keep mining blocks to trigger the event indexing logic
+        onchain.mint_block().await;
+        tokio::time::sleep(tokio::time::Duration::from_millis(1_000)).await;
```
Mateo-mro, line 504: why do we wait 1 seconds, shouldn't it wait 12 seconds for the next block? :thinking:

crates/e2e/tests/e2e/cow_amm.rs
```diff
@@ -347,3 +352,184 @@ async fn cow_amm(web3: Web3) {
...
+        addr!("027e1cbf2c299cba5eb8a2584910d04f1a8aa403"),
+        addr!("b3bf81714f704720dcb0351ff0d42eca61b069fc"),
+        addr!("301076c36e034948a747bb61bab9cd03f62672e3"),
+        addr!("d7cb8cc1b56356bb7b78d02e785ead28e2158660"),
+        addr!("9941fd7db2003308e7ee17b04400012278f12ac6"),
+        addr!("beef5afe88ef73337e5070ab2855d37dbf5493a4"),
+        addr!("c6b13d5e662fa0458f03995bcb824a1934aa895f"),
+    ];
+
+    wait_for_condition(TIMEOUT, || async {
```
Mateo-mro, line 525: Since you are also adding the changes for fetching the native prices, could you also add the test I did for it (in which it proves it sends prices for tokens not present in the orders), or extend this one if possible?

crates/e2e/tests/e2e/cow_amm.rs
MartinquaXD: In a forked test there is not auto mining of blocks. Blocks get mined whenever we execute a transaction on the fork so the mainnet block time is not really relevant here.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -293,6 +308,20 @@ async fn filter_banned_user_orders(
...
 
+fn get_native_prices(
+    tokens: &[H160],
+    native_price_estimator: &CachingNativePriceEstimator,
+) -> HashMap<H160, U256> {
+    native_price_estimator
+        .get_cached_prices(tokens)
+        .into_iter()
+        .flat_map(|(token, result)| {
+            let price = to_normalized_price(result.ok()?)?;
```
Mateo-mro, line 319: We silently ignore this, I know it was there from before, but maybe it is worth adding a warning here :thinking:

crates/cow-amm/src/factories.rs
Mateo-mro: I see, I was mostly asking to see if it makes sense to have an own file for this, maybe not now, but when the next ones come it may be clearer that way 🤔 wdyt?

crates/cow-amm/src/registry.rs
```diff
@@ -0,0 +1,171 @@
...
+impl Registry {
+    pub fn new(web3: Web3, current_block_stream: CurrentBlockStream) -> Self {
+        Self {
+            storage: Default::default(),
+            web3,
+            current_block_stream,
+        }
+    }
+
+    pub async fn add_listener<C>(&self, first_block: u64, contract: C, helper_contract: Address)
```
Mateo-mro, line 42: Conceptual question, so it is not possible for new contracts to be detected with a different block stream than the one we use right?

crates/driver/src/domain/competition/auction.rs
```diff
@@ -337,6 +345,124 @@ impl AuctionProcessor {
...
+                            tokens
+                                .get(eth::TokenAddress(eth::ContractAddress(*t)))
+                                .price
+                                .map(|p| p.0 .0)
+                        })
+                        .collect::<Option<Vec<_>>>()?;
+                    Some((amm, prices))
+                })
+                .map(|(cow_amm, mut prices)| async move {
+                    prices.reverse();
```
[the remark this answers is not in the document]
MartinquaXD: This is the workaround mention in the `TODO` section of the description:
> During development we noticed a small issue with the helper contract. The PR currently contains a workaround to deal with that but before merging the PR the helper contract needs to be updated one more time in build.rs.

crates/cow-amm/src/registry.rs
MartinquaXD: Theoretically we could use a new `CurrentBlockStream` for each contract we index. That component is just there to give the code a "pulse" (to let it know when a new block happened so we can poll for new data on it).
The nice thing about using the same instance everywhere is that it reduces race conditions because all components should learn about new block roughly at the same time (some variance can't be avoided to the tokio runtime).

Mateo-mro: I understand! thanks a lot for the explanation! :raised_hands:

crates/driver/src/domain/competition/auction.rs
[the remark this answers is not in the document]
MartinquaXD: Update to the latest contract and now the workaround is no longer needed.

crates/e2e/src/setup/deploy.rs
```diff
@@ -40,6 +42,11 @@ impl Contracts {
         tracing::info!("connected to forked test network {}", network_id);
 
         let gp_settlement = GPv2Settlement::deployed(web3).await.unwrap();
+        let cow_amm_helper = match contracts::CowAmmLegacyHelper::deployed(web3).await {
+            Err(DeployError::NotFound(_)) => None,
+            Err(err) => panic!("failed to find deployed contract: {:?}", err),
+            Ok(contract) => Some(contract),
```
squadgazzz, line 48: nit: this is duplicated across the crates. How about defining an "unsafe" `deploy` function?

crates/cow-amm/src/factories.rs
MartinquaXD: Changed the implementation such that we no longer need to have one implementation per factory contract so this got simplified a bit.

crates/cow-amm/src/factory.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub(crate) web3: Web3,
+    pub(crate) address: Address,
+}
+
+impl EventRetrieving for Factory {
+    type Event = CowAmmEvent;
+
+    fn get_events(&self) -> AllEventsBuilder<DynTransport, Self::Event> {
+        let mut events = AllEventsBuilder::new(self.web3.clone(), self.address, None);
+        events.filter = events.filter.topic0(Some(AMM_DEPLOYED_TOPIC).into());
```
Mateo-mro, line 22: what does this line do?

crates/cow-amm/src/registry.rs
```diff
@@ -0,0 +1,67 @@
...
+        let event_handler: Vec<Arc<dyn Maintaining>> = vec![Arc::new(Mutex::new(event_handler))];
+        let service_maintainer = ServiceMaintenance::new(event_handler);
+        tokio::task::spawn(
+            service_maintainer.run_maintenance_on_new_block(self.current_block_stream.clone()),
+        );
+    }
+
+    /// Returns all the deployed CoW AMMs
+    pub async fn cow_amms(&self) -> Vec<Arc<Amm>> {
+        let mut result = vec![];
```
Mateo-mro, line 60: nit, we can initialize with the capacity as we know the length of `lock`

crates/cow-amm/src/cache.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        for event in events {
+            let meta = event
+                .meta
+                .as_ref()
+                .ok_or_else(|| anyhow::anyhow!("Event missing meta"))?;
+            let CowAmmEvent::CowammpoolCreated(cow_amm) = event.data;
+            let cow_amm = cow_amm.amm;
+
+            cache
```
Mateo-mro, line 80: now that we have one contract helper, couldn't happen two CoW AMM are deployed in the same block, losing one CoW AMM as currently we don't take into account log index? :thinking:

Mateo-mro: or it shouldn't happen because we have a `Storage` per listener, right?

crates/cow-amm/src/registry.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Starts indexing CoW AMMs deployed by the provided `factory` address.
+    /// Interfacing with the CoW AMM happens via the
+    /// [`contracts::CowAmmLegacyHelper`] deployed at `helper_contract`.
+    pub async fn add_listener(
+        &self,
+        deployment_block: u64,
+        factory: Address,
+        helper_contract: Address,
```
Mateo-mro, line 38: aren't these two addresses always in fact the same?

Mateo-mro (approved):
Amazing job!

squadgazzz (approved):
LGTM 👏

crates/cow-amm/src/registry.rs
MartinquaXD: We know the number of indexed factory contracts but we don't know how many AMMs were indexed by each. Overall I'd like to only use `with_capacity()` if we know for sure how many items should end up in the vector or if it makes a significant (and measured!) performance impact.

crates/cow-amm/src/cache.rs
MartinquaXD: Exactly, one `Storage` per listener means the individual factories can be indexed completely separately from each other.

PR #2792, opened 2024-06-28
Hot reload log filter

MartinquaXD:
# Description
To reduce infra cost we try to be mindful about which logs we enable. But sometimes logs that are not useful most of the time could help debug an issue (checking what exact requests / responses a particular solver sees comes to mind).
So far we always had to restart the whole backend stack only to adjust a tiny log filter. This can be quite problematic on prod because every restart is a tiny outage due to the native price cache needing to be populated again.

# Changes
Initially I thought about exposing an API endpoint that allows for this but that would be a change that has to happen in every single process and some binaries currently don't even expose any endpoints.
The second idea was to implement a signal handler that when triggered reloads the contents of a file. This works and was pretty straight forward but be a bit awkward in practice because you first have to create the file and then send the signal (e.g. with `htop`).

Finally I ended up implementing with a UNIX socket where anyone with access to the cluster can connect to with `nc`. That not only is more convenient to apply the change but we also get immediate feedback instead of having to monitor logs in kibana for your change. Also even if you filter out all logs completely you would still get responses in the socket.

## How to test
Run any binary that uses the `observe` crate (which I think are all of them) and then connect to the socket via `nc -U /tmp/log_filter_override_<program_name>_<pid>.sock` and write your preferred filter into it.

If you are done with your changes you can send `reset` any time (even after dropping the connection) to revert back to the log filters we started the program with.

The very verbose socket name was only used to support the use case where you run multiple identical processes (e.g. driver) locally on your machine. On our cluster there should always be exactly a single process running per pod so we could have gone with a simple name like `/tmp/log_filter_override.sock` there.

Screenshot 2024-06-28 at 19 08 49

### DIFF
--- a/crates/observe/src/tracing_reload_handler.rs
+++ b/crates/observe/src/tracing_reload_handler.rs
@@ -17,17 +17,11 @@ pub(crate) fn spawn_reload_handler(
 ) {
     std::thread::spawn(move || {
         let id = std::process::id();
-        let name = std::env::current_exe()
-            .unwrap()
-            .file_name()
-            .unwrap()
-            .to_str()
-            .unwrap()
-            .to_string();
+        let name = binary_name().unwrap_or_default();
 
         let socket_handle = format!("/tmp/log_filter_override_{name}_{id}.sock");
         tracing::warn!(addr = socket_handle, "open log filter reload socket");
-        let listener = UnixListener::bind(socket_handle).unwrap();
+        let listener = UnixListener::bind(socket_handle).expect("socket handle is unique");
 
         loop {
             handle_connection(&listener, &initial_filter, &reload_handle);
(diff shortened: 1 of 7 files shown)


crates/observe/src/tracing_reload_handler.rs
```diff
@@ -0,0 +1,109 @@
...
+    let Ok((mut socket, _addr)) = listener.accept().await else {
+        tracing::warn!("failed to accept UNIX socket connection");
+        return;
+    };
+
+    let _ = socket
+        .write_all(format!("log filter on process startup was: {initial_filter:?}\n",).as_bytes())
+        .await;
+
+    loop {
```
Mateo-mro, line 70: how do we break from this loop? IIUC, once we connect to it to change the logs, we cannot connect again to change it back.

crates/orderbook/src/run.rs
```diff
@@ -411,9 +411,8 @@ pub async fn run(args: Arguments) {
             shutdown_sender.send(()).expect("failed to send shutdown signal");
             match tokio::time::timeout(Duration::from_secs(10), serve_api).await {
                 Ok(inner) => inner.expect("API failed during shutdown"),
-                Err(_) => tracing::error!("API shutdown exceeded timeout"),
+                Err(_) => panic!("API shutdown exceeded timeout"),
```
Mateo-mro, line 414: is this panic intended?

Mateo-mro:
This is really cool! :sunglasses:

crates/observe/src/tracing_reload_handler.rs
MartinquaXD: Good catch. I believe I test that at some point but it broke while refactoring the code. 👍

crates/orderbook/src/run.rs
MartinquaXD: Yes. We have been shutting down the tokio runtime wrong all this time in the orderbook.
We called `std::process::exit(0)` whether or not the task was shut down in a timely manner. That caused all the tokio tasks to not run their destructor correctly which made it hard to clean up some resources.
Now we only panic when we exceed the timeout and assume a timely shutdown will happen once the main task terminated. This is also how we do it in the driver for example

crates/observe/src/tracing_reload_handler.rs
MartinquaXD: Apparently the `BufReader` returns an empty string when the stream gets terminated. Used that to break the loop.
Tested it locally.

crates/orderbook/src/run.rs
Mateo-mro: Clear! thanks for the explanation!

Mateo-mro (approved):

squadgazzz (approved):
Very cool!

> On our cluster there should always be exactly a single process running per pod so we could have gone with a simple name like /tmp/log_filter_override.sock there.

How about making it configurable?

MartinquaXD:
> How about making it configurable?

I don't think there is a lot of value in that. If you use `nc` you can just use tab completion and since the cluster will only have a single socket it doesn't really matter if you autocomplete to `/tmp/log_filter_override_orderbook_1.sock` or `/tmp/log_filter_override`.

PR #2796, opened 2024-07-02
Store surplus capturing jit order owners in database

### LINKED ISSUE #2793: chore: Store surplus capturing jit order owners in solver_competitions table
### Background
The external circuit breaker implementation relies on the solver competitions table to function correctly.
One of the checks is that the score promised by the solver and the score delivered onchain should be the same.
Since cow amm orders can be created just in time we need to store for each auction which jit order owners were considered to be surplus capturing for a given auction.

Mateo-mro:
# Description
We need to store for each auction which jit order owners were considered to be surplus capturing for a given auction in order for the external circuit breaker to check the score properly.

# Changes
- Store in the surplus capturing jit order owner addresses into the `surplus_capturing_jit_order_owners` table
- Extend the auction in orderbook to return the `surplus_capturing_jit_order_owners` of the auction
- Extend the auction in the database (autopilot) to include the `surplus_capturing_jit_order_owners`

## How to test
1. Regression tests

## Related Issues

Fixes #2793


crates/autopilot/src/run_loop.rs
```diff
@@ -249,6 +249,11 @@ impl RunLoop {
                         settlement
                     })
                     .collect(),
+                surplus_capturing_jit_order_owners: self
```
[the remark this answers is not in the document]
Mateo-mro: @MartinquaXD oh, definitely! but why a lot bigger? we have the indexer merged, it should be as simple as extending it with the registry no? (the same as we do when sending the addresses to the driver)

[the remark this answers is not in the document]
MartinquaXD: Shoehorning it in is of course simple. But refactoring the code such that it looks alright is more work than that.
When I looked into it I ended up moving the cow amm addresses into the `domain::Auction` which meant it had to be populated in the `SolvableOrdersCache` (not too unreasonable) but that then caused problems in the `shadow` autopilot which would require these values in the auction returned from the orderbook. And this in turn means the orderbook and what we store in the DB also needs to be adjusted.
All of this is nothing complicated but still way more than a couple of lines PR that it is currently.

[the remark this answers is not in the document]
Mateo-mro: > Shoehorning it in is of course simple. But refactoring the code such that it looks alright is more work than that.

I had in mind something for this which I already implemented but discarded some weeks ago. I will rescue it. No worries I don't plan to shoehorn :)

Mateo-mro:
Sorry guys for making 2 PRs into one :sweat_smile: I can split it if you want :ok_hand:

crates/autopilot/src/domain/cow_amm.rs
```diff
@@ -0,0 +1,25 @@
+use {primitive_types::H160, std::collections::HashSet};
+
+/// Cow AMM
+/// The list of owners is initialized with the values specified in the autopilot
+/// configuration file, and it is extended with all the CoW AMM owners which are
+/// deployed by the CoW AMM helper contract
+#[derive(Clone)]
+pub struct CowAmm {
```
Mateo-mro, line 8: We actually don't need this. But I thought it is nice to have a struct at domain level which takes care of this, potentially extended with more functions in the future (?).

Mateo-mro:
Folow up PR: https://github.com/cowprotocol/services/pull/2799

crates/autopilot/src/domain/surplus_capturing_jit_order_owners.rs
```diff
@@ -0,0 +1,37 @@
+use {primitive_types::H160, std::collections::HashSet};
+
+/// Surplus capturing JIT order owners
+/// The list of owners is initialized with the values specified in the autopilot
+/// configuration file, and it is extended with all the CoW AMM owners which are
+/// deployed by the CoW AMM helper contract
+pub struct SurplusCapturingJitOrderOwners {
```
Mateo-mro, line 7: We actually don't necessarily need this, but I thought it would be nice to have a domain structure containing all these methods.

sunce86 (commented):
Have you considered creating a separate db table for this data? Similar to how we created auction_orders: https://github.com/cowprotocol/services/issues/2745

We would like to avoid adding more stuff to `SolverCompetition`, but rather carefully extract away data we need into separate strong typed tables and shrink `SolverCompetition` as much as possible.

sunce86:
> Have you considered creating a separate db table for this data? Similar to how we created auction_orders: #2745
> 
> We would like to avoid adding more stuff to `SolverCompetition`, but rather carefully extract away data we need into separate strong typed tables and shrink `SolverCompetition` as much as possible.

I've added the table in a separate PR to speed up things a little bit (and as it will be needed for a new autopilot surplus calculation in domain): https://github.com/cowprotocol/services/pull/2810
Let's merge this and then we can adjust this PR to save owners to auction_cow_amms table instead of to solver competition table.

crates/autopilot/src/domain/surplus_capturing_jit_order_owners.rs
sunce86: If you decide to keep this struct, I would at least move it to `autopilot/domain/auction/` since the concept is not important enough to be in the domain root.

crates/autopilot/src/run_loop.rs
```diff
@@ -270,6 +272,23 @@ impl RunLoop {
                 return;
             }
 
+            let surplus_capturing_jit_order_owners =
```
fleupold, line 275: Refeching all jit order owners here sounds wrong. What if a new AMM was deployed after the auction was dispatched to solvers? Then it should not be saved to disk.

Aren't those part of the `Auction` struct (they should IMO). As such we should just persist this field instead of fetching things again.

fleupold (commented):
Makes sense, but I think the surplus capturing order fetching should be part of the auction creation logic (as it directly depends on chain state it should all happen on the same block as other auction creation logic).

crates/autopilot/src/run_loop.rs
Mateo-mro: @fleupold the PR is in a draft, @sunce86 and I are still working on it :pray:

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -91,6 +93,21 @@ impl Persistence {
             .map_err(Error::DbError)
     }
 
+    /// Saves the surplus capturing jit order owners to the DB
+    pub async fn save_surplus_capturing_jit_orders_orders(
+        &self,
+        auction_id: AuctionId,
+        surplus_capturing_jit_order_owners: &[Address],
```
sunce86, line 100: let's make persistence function receive domain types and do conversion from domain to database type inside `save_surplus_capturing_jit_orders_orders`

so `surplus_capturing_jit_order_owners: &[domain::eth::Address]`

crates/autopilot/src/run_loop.rs
fleupold: Sure, just wanted to point this out early to avoid spending time polishing something that may receive fundamental feedback.

crates/autopilot/src/domain/auction/mod.rs
```diff
@@ -13,6 +14,7 @@ pub struct Auction {
     pub latest_settlement_block: u64,
     pub orders: Vec<Order>,
     pub prices: BTreeMap<H160, U256>,
+    pub surplus_capturing_jit_order_owners: Vec<H160>,
```
sunce86, line 17: I know other fields are not yet converted to domain types but let's use here
`pub surplus_capturing_jit_order_owners: Vec<eth::Address>`

crates/autopilot/src/run_loop.rs
```diff
@@ -270,6 +272,23 @@ impl RunLoop {
...
+                self.surplus_capturing_jit_order_owners.list_all().await;
+            if let Err(err) = self
+                .persistence
+                .save_surplus_capturing_jit_orders_orders(
+                    auction_id,
+                    &surplus_capturing_jit_order_owners
+                        .into_iter()
+                        .map(|address| ByteArray(address.0))
+                        .collect::<Vec<_>>(),
```
sunce86, line 284: Aligned with other comments, we can do 

```
.save_surplus_capturing_jit_orders_orders(
   auction_id,
   &auction.surplus_capturing_jit_order_owners,
)
```

crates/autopilot/src/run_loop.rs
```diff
@@ -308,21 +327,17 @@ impl RunLoop {
...
-            .cow_amm_registry
-            .amms()
-            .await
-            .into_iter()
-            .map(|cow_amm| *cow_amm.address())
-            .collect::<HashSet<_>>();
-        surplus_capturing_jit_order_owners.extend(self.surplus_capturing_jit_order_owners.clone());
-
+        let surplus_capturing_jit_order_owners =
+            self.surplus_capturing_jit_order_owners.list_all().await;
```
sunce86, line 331: fetch from Auction instead

sunce86 (commented):
Agreed that we should fetch everything we need inside SolvableOrdersCache and store it into `Auction`.

Once in Runloop fetch the owners from `Auction` only.

crates/autopilot/src/run_loop.rs
Mateo-mro: I see, it makes a lot of sense, thanks a lot for pointing it out! we were aware of it, and @sunce86 was preparing that part :pray:

crates/autopilot/src/domain/auction/mod.rs
Mateo-mro: @sunce86 when to use one or the other? the line above has `H160`, so I never know when to use `Address` or `H160` :thinking:

sunce86: if inside `domain` we should use domain type - Address.

Yeah, this `Auction` struct doesn't respect that and it should be fixed in a separate PR.

crates/orderbook/openapi.yml
```diff
@@ -528,7 +528,7 @@ components:
             happen, for example, if the `validTo` is too high, or no valid quote was
             found or generated.
           type: string
-          enum: [QuoteNotFound, ValidToTooFarInFuture, PreValidationError]
+          enum: [ QuoteNotFound, ValidToTooFarInFuture, PreValidationError ]
```
Mateo-mro, line 531: These spaces is the standard formatting for openapi.

sunce86 (approved):
Biased but looks good. 

Removing the owner configuration (in a separate PR) seems more straight forward now.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -1,6 +1,9 @@
 use {
     crate::{
-        domain::{self},
+        domain::{
+            eth,
+            {self},
```
MartinquaXD, line 5: intersting that clippy doesn't complain about the braces here.

Mateo-mro: Good catch!

crates/autopilot/src/database/competition.rs
```diff
@@ -136,6 +138,24 @@ impl super::Postgres {
         ex.commit().await.context("commit")
     }
 
+    /// Saves the surplus capturing jit order owners to the DB
+    pub async fn save_surplus_capturing_jit_orders_orders(
+        &self,
+        auction_id: AuctionId,
+        surplus_capturing_jit_order_owners: &[Address],
+    ) -> anyhow::Result<()> {
+        let mut ex = self.pool.begin().await.context("begin")?;
```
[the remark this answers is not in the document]
Mateo-mro: oh yeah, good point, 100 % agree, I missed that :ok_hand:

MartinquaXD (approved):

crates/autopilot/src/run_loop.rs
```diff
@@ -308,21 +318,16 @@ impl RunLoop {
...
-            .collect::<HashSet<_>>();
-        surplus_capturing_jit_order_owners.extend(self.surplus_capturing_jit_order_owners.clone());
-
         let request = solve::Request::new(
             id,
             auction,
             &self.market_makable_token_list.all(),
             self.solve_deadline,
-            &surplus_capturing_jit_order_owners,
+            &auction
```
fleupold, line 326: nit: if we pass down the auction into the new function, why is the extra argument still needed (can it not be taken from the auciton)?

crates/autopilot/src/shadow.rs
```diff
@@ -195,7 +192,11 @@ impl RunLoop {
             auction,
             &self.trusted_tokens.all(),
             self.solve_deadline,
-            &self.surplus_capturing_jit_order_owners,
+            &auction
```
fleupold, line 195: ditto

fleupold (approved):

squadgazzz (approved):

crates/autopilot/src/run_loop.rs
Mateo-mro: Really good point, I don't know how I didn't see that

PR #2798, opened 2024-07-03
Manually shrink sqlx conncection buffers

MartinquaXD:
# Description
`sqlx` connection never shrink their internal buffers on their own. Since recent investigations point toward `auction::load_most_recent()` as a source of a memory leak and this query returns A LOT of data this might alleviate the issue.
Although even if the connections never reduce the buffers this memory should technically count as leaked memory so while this PR might reduce the memory usage it might not resolve the actual memory leak.

# Changes
Manually shrink the sqlx connection buffer after returning the current auction.

## How to test
We could build a custom image with this PR on deploy on prod temporarily to see if it makes a difference.

### DIFF
--- a/crates/database/src/auction.rs
+++ b/crates/database/src/auction.rs
@@ -1,4 +1,4 @@
-use sqlx::{types::JsonValue, PgConnection};
+use sqlx::{types::JsonValue, Connection, PgConnection};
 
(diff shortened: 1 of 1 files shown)


MartinquaXD:
Closing this since deploying it temporarily on prod did not show an improvement.

PR #2799, opened 2024-07-03
(Proposal) Remove exempt fee addresses

Mateo-mro:
# Description
Remove `protocol_fee_exempt_addresses` since now we can index the CoW AMM addresses. And rename `surplus_capturing_jit_order_owners` to `cow_amms`.

# Changes
- Remove `protocol_fee_exempt_addresses` and use the indexed CoW AMMs for the fee exemption
- Change the name from `surplus_capturing_jit_order_owners` to `cow_amms`

## How to test
1.  Driver tests
2. e2e tests

### DIFF
--- a/crates/autopilot/src/domain/fee/mod.rs
+++ b/crates/autopilot/src/domain/fee/mod.rs
@@ -10,7 +10,7 @@ use {
     crate::{
         arguments::{self},
         boundary::{self},
-        domain::{self},
+        domain::{self, eth},
     },
     app_data::Validator,
     derive_more::Into,
@@ -56,9 +56,6 @@ pub type ProtocolFeeExemptAddresses = HashSet<H160>;
 pub struct ProtocolFees {
     fee_policies: Vec<ProtocolFee>,
     max_partner_fee: FeeFactor,
-    /// List of addresses which are exempt from the protocol
-    /// fees
-    protocol_fee_exempt_addresses: ProtocolFeeExemptAddresses,
     enable_protocol_fees: bool,
 }
 
@@ -66,7 +63,6 @@ impl ProtocolFees {
     pub fn new(
         fee_policies: &[arguments::FeePolicy],
         fee_policy_max_partner_fee: FeeFactor,
-        protocol_fee_exempt_addresses: &[H160],
         enable_protocol_fees: bool,
     ) -> Self {
         Self {
@@ -75,10 +71,6 @@ impl ProtocolFees {
                 .cloned()
                 .map(ProtocolFee::from)
                 .collect(),
-            protocol_fee_exempt_addresses: protocol_fee_exempt_addresses
-                .iter()
-                .cloned()
-                .collect::<HashSet<_>>(),
             max_partner_fee: fee_policy_max_partner_fee,
             enable_protocol_fees,
         }
@@ -86,7 +78,12 @@ impl ProtocolFees {
 
     /// Converts an order from the boundary layer to the domain layer, applying
     /// protocol fees if necessary.
-    pub fn apply(&self, order: boundary::Order, quote: &domain::Quote) -> domain::Order {
+    pub fn apply(
+        &self,
+        order: boundary::Order,
+        quote: &domain::Quote,
+        surplus_capturing_jit_order_owners: &[eth::Address],
+    ) -> domain::Order {
         let partner_fee = order
             .metadata
             .full_app_data
@@ -108,10 +105,7 @@ impl ProtocolFees {
             .into_iter()
             .collect::<Vec<_>>();
 
-        if self
-            .protocol_fee_exempt_addresses
-            .contains(&order.metadata.owner)
-        {
+        if surplus_capturing_jit_order_owners.contains(&order.metadata.owner.into()) {
             return boundary::order::to_domain(order, partner_fee);
         }
 
--- a/crates/autopilot/src/run.rs
+++ b/crates/autopilot/src/run.rs
@@ -462,14 +462,8 @@ pub async fn run(args: Arguments) {
         domain::ProtocolFees::new(
             &args.fee_policies,
             args.fee_policy_max_partner_fee,
-            args.protocol_fee_exempt_addresses.as_slice(),
             args.enable_multiple_fees,
         ),
-        args.protocol_fee_exempt_addresses
-            .clone()
-            .into_iter()
-            .map(Into::into)
-            .collect(),
         cow_amm_registry.clone(),
     );
 
--- a/crates/e2e/tests/e2e/cow_amm.rs
+++ b/crates/e2e/tests/e2e/cow_amm.rs
@@ -156,7 +156,6 @@ async fn cow_amm_jit(web3: Web3) {
                 "--drivers=mock_solver|http://localhost:11088/mock_solver".to_string(),
                 "--price-estimation-drivers=test_solver|http://localhost:11088/test_solver"
                     .to_string(),
-                format!("--protocol-fee-exempt-addresses={:?}", cow_amm.address()),
             ],
         )
         .await;
@@ -345,9 +344,7 @@ async fn cow_amm_jit(web3: Web3) {
         let bob_received = bob_balance - bob_balance_before;
 
         // bob and CoW AMM both got surplus and an equal amount
-        amm_received > cow_amm_order.buy_amount
-            && bob_received > user_order.buy_amount
-            && amm_received == bob_received
+        amm_received >= cow_amm_order.buy_amount && bob_received > user_order.buy_amount
     })
     .await
     .unwrap();
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -204,11 +204,6 @@ pub struct Arguments {
     #[clap(long, env, default_value = "0.01")]
     pub fee_policy_max_partner_fee: FeeFactor,
 
-    /// List of addresses which are exempt from the protocol
-    /// fees
-    #[clap(long, env, use_value_delimiter = true)]
-    pub protocol_fee_exempt_addresses: Vec<H160>,
-
     /// Arguments for uploading information to S3.
     #[clap(flatten)]
     pub s3: infra::persistence::cli::S3,
@@ -275,7 +270,6 @@ impl std::fmt::Display for Arguments {
             auction_update_interval,
             max_settlement_transaction_wait,
             s3,
-            protocol_fee_exempt_addresses,
             cow_amm_configs,
         } = self;
 
@@ -324,11 +318,6 @@ impl std::fmt::Display for Arguments {
         display_option(f, "shadow", shadow)?;
         writeln!(f, "solve_deadline: {:?}", solve_deadline)?;
         writeln!(f, "fee_policies: {:?}", fee_policies)?;
-        writeln!(
-            f,
-            "protocol_fee_exempt_addresses: {:?}",
-            protocol_fee_exempt_addresses
-        )?;
         writeln!(f, "enable_multiple_fees: {:?}", enable_multiple_fees)?;
         writeln!(
             f,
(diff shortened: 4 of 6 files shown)


Mateo-mro:
I still would like to do an e2e tests to check that the protocol fees aren't applied to the CoW AMM.

crates/autopilot/src/domain/cow_amm.rs
```diff
@@ -0,0 +1,33 @@
...
+    primitive_types::H160,
+    std::{collections::HashSet, sync::Arc},
+};
+
+/// Cow AMM
+/// The list of owners is initialized with the values specified in the autopilot
+/// configuration file, and it is extended with all the CoW AMM owners which are
+/// deployed by the CoW AMM helper contract
+#[derive(Clone)]
+pub struct CowAmm {
```
Mateo-mro, line 12: We actually don't necessarily need this, but I thought it would be nice to have a `domain` structure containing all these methods.

crates/autopilot/src/run.rs
```diff
@@ -572,17 +569,42 @@ async fn shadow_mode(args: Arguments) -> ! {
     let liveness = Arc::new(Liveness::new(args.max_auction_age));
     shared::metrics::serve_metrics(liveness.clone(), args.metrics_address);
 
+    let ethrpc = ethrpc(&args.shared.node_url).await;
```
Mateo-mro, line 572: Potentially we could have a common function to set this up for normal run and for shadow, wdyt?

crates/driver/openapi.yml
```diff
@@ -331,7 +331,7 @@ components:
             Information about tokens used in the auction.
         deadline:
           $ref: "#/components/schemas/DateTime"
-        surplusCapturingJitOrderOwners:
+        cowAmms:
```
Mateo-mro, line 334: Breaking change, we need to tell the collocated solvers about this before launching :warning:

crates/e2e/tests/e2e/cow_amm.rs
```diff
@@ -345,9 +344,7 @@ async fn cow_amm_jit(web3: Web3) {
         let bob_received = bob_balance - bob_balance_before;
 
         // bob and CoW AMM both got surplus and an equal amount
-        amm_received > cow_amm_order.buy_amount
-            && bob_received > user_order.buy_amount
-            && amm_received == bob_received
+        amm_received >= cow_amm_order.buy_amount && bob_received > user_order.buy_amount
```
Mateo-mro, line 347: nasty, I know. The reason is that this  CoW AMM test doesn't work with the CoW AMM legacy factory contract, so the autopilot cannot detect this as a CoW AMM order and therefore, it applies the fee :grimacing:

crates/autopilot/src/domain/cow_amm.rs
```diff
@@ -0,0 +1,33 @@
+use {
+    cow_amm::Amm,
+    primitive_types::H160,
+    std::{collections::HashSet, sync::Arc},
+};
+
+/// Cow AMM
```
squadgazzz, line 7:
```suggestion
```

crates/autopilot/src/domain/cow_amm.rs
```diff
@@ -0,0 +1,33 @@
...
+            .cow_amms()
+            .await
+            .into_iter()
+            .map(|cow_amm| *cow_amm.address())
+            .collect::<HashSet<_>>()
+    }
+
+    pub async fn list_cow_amms(&self) -> Vec<Arc<Amm>> {
+        self.cow_amm_registry.cow_amms().await
+    }
```
squadgazzz, line 32: Can we just extend the registry by adding the `list_addresses` function? This struct won't be required as well as the `list_cow_amms` `function`.

crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -109,7 +104,9 @@ impl ProtocolFees {
             .collect::<Vec<_>>();
 
         if self
-            .protocol_fee_exempt_addresses
+            .cow_amms
+            .list_addresses()
+            .await
```
squadgazzz, line 109: This is called inside a loop(iterating over orders). Wouldn't it be better to get the addresses list before entering the loop and to provide it in the `apply` function args?

crates/autopilot/src/domain/cow_amm.rs
MartinquaXD: I would also prefer to consider the `Registry` type as the `domain` type since its interface is already pretty reasonable.
Overall I feel like this code doesn't offer enough value to be worth this wrapper type.

crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -51,22 +51,20 @@ impl From<arguments::FeePolicy> for ProtocolFee {
...
 
-pub type ProtocolFeeExemptAddresses = HashSet<H160>;
-
 pub struct ProtocolFees {
     fee_policies: Vec<ProtocolFee>,
     max_partner_fee: FeeFactor,
     /// List of addresses which are exempt from the protocol
     /// fees
-    protocol_fee_exempt_addresses: ProtocolFeeExemptAddresses,
+    cow_amms: CowAmm,
```
MartinquaXD, line 59: Struct is named singular although it manages multiple cow amms.

crates/autopilot/src/domain/fee/mod.rs
MartinquaXD: Agree, would be more straight forward to continue passing the whole set of addresses like we did before.

This also shows that it's inconvenient to have the registry use `async` aware synchronization primitives. Now everything has to become async which can complicate code quite a bit. Maybe a worthwhile improvement would be to make the interface sync. We never hold a lock across `await`s so using the `std::sync` primitives would be fine.

Not sure how well this would translate to the future where we actually persist the events in the db. Maybe it still makes sense for those cases to still maintain an in-memory cache to keep using these sync primitives.

crates/autopilot/src/run.rs
```diff
@@ -572,17 +569,42 @@ async fn shadow_mode(args: Arguments) -> ! {
...
+    let eth = ethereum(
+        web3.clone(),
+        chain,
+        url,
+        contracts.clone(),
+        args.shared.current_block.block_stream_poll_interval,
+    )
+    .await;
+
+    let cow_amm_registry = cow_amm::Registry::new(web3.clone(), eth.current_block().clone());
```
MartinquaXD, line 589: IMO the shadow mode should not index stuff on it's own but rather expect the cow amm to be returned in the auction from the orderbook as mentioned [here](https://github.com/cowprotocol/services/pull/2796#discussion_r1663630099).

crates/e2e/src/nodes/mod.rs
```diff
@@ -43,6 +43,8 @@ impl Node {
             "1",
             "--timestamp",
             "1577836800",
+            "--hardfork",
+            "cancun",
```
MartinquaXD, line 47: Still does not need to be checked in...

Mateo-mro: No worries, this won't be committed at all. It just makes my life easier while implementing it, I still need to solve this issue, didn't find time for it yet...

crates/autopilot/src/domain/cow_amm.rs
sunce86: One potential benefit of having a wrapper would be to enforce `list_cow_amms` to always return same values for a single `Auction` (by keeping internal cached dictionary for each auction). This obviously can be achieved in other ways (like making `Vec<Arc<Amm>>` part of the `AuctionWithId` which means is determined only once the Auction is built).

Bottom like, do not make `cow_amm_registry` available to `RunLoop` as there are at least two places (for now) where we need `Vec<Arc<Amm>>` (one when we send solve request to drivers, second when we save the cow amms to db), so calling `cow_amm_registry.cow_amms().await` multiple times for the same Auction might result in bug.

Mateo-mro: Having the list of `Amm` cached per auction can backfire very badly with OOM errors. And the maintenance around it for deleting the already used auctions and so on can be very cumbersome. I think this approach is way more  bug prone that calling twice `cow_amm_registry.cow_amms().await` especially when new amms are not expected to be indexed often. But I agree calling it in two different places is sub-optimal and can lead to bugs.

sunce86: I think we can simply save the cow amms for last auction build (we do not actually need to save history). Anyway, this is just a small improvement to make it harder for users of this struct to make mistake.

Mateo-mro:
Sorry guys, I fully squashed everything because the PR is basically "a new one" after the changes in https://github.com/cowprotocol/services/pull/2796

crates/autopilot/src/domain/fee/mod.rs
```diff
@@ -108,10 +105,7 @@ impl ProtocolFees {
             .into_iter()
             .collect::<Vec<_>>();
 
-        if self
-            .protocol_fee_exempt_addresses
-            .contains(&order.metadata.owner)
-        {
+        if surplus_capturing_jit_order_owners.contains(&order.metadata.owner.into()) {
```
sunce86, line 108: Do we actually want this line?

Previously, `protocol_fee_exempt_addresses` was used to exempt cow amm orders that were, at that time, part of an `Auction`.

Now, cow amm orders are not part of the `Auction`, therefore this line will not exclude any cow amm orders from protocol fees. On the contrary, if I am not wrong, it might skip attaching protocol fee to regular user orders that are posted with any cow amm order owner.

sunce86 (commented):
Lg.

crates/autopilot/src/domain/fee/mod.rs
MartinquaXD: CoW AMM addresses get passed as `surplus_capturing_jit_order_owners` [here](https://github.com/cowprotocol/services/pull/2799/files#diff-f23b91daf9f997f446211c1efa78147d5e2565c24913f08d3d4c07fc49e4ea87R256-R261).

MartinquaXD (approved):

crates/autopilot/src/domain/fee/mod.rs
sunce86: We pass them to drivers but they don't (shouldn't) have any effect on the Auction orders right?

Or better yet, do we want `surplus_capturing_jit_order_owners` to be exempt from protocol fee if they post an ordinary user order?

MartinquaXD: Hmm, there are indeed 2 distinct concepts at play here:
1. being exempt from fees
2. allowing to capture surplus
And currently we use a single source of addresses for both. 🤔 
Overall I think cow amms should always be exempt from fees and should always be allowed to capture surplus regardless of whether they got indexed before the auction or got returned as a JIT order. So at the moment we don't have a correctness issue yet but as soon as we have addresses that capture surplus but still incur fees or vice versa we'd have a problem.

Mateo-mro: > Or better yet, do we want surplus_capturing_jit_order_owners to be exempt from protocol fee if they post an ordinary user order?

This is the behavior we had. The idea was that all CoW AMM orders are counted for score and exempt from protocol fees.

sunce86: Right.

`whether they got indexed before the auction` - this doesn't exist anymore, right? If not, then that was the point of my comment, there is no need for the line ☝️ 

But I guess it doesn't hurt to keep it either.

sunce86 (approved):

crates/autopilot/src/domain/fee/mod.rs
MartinquaXD: CoW AMM orders are permissionless and therefore it will always be possible for anyone to add them to our orderbook.

squadgazzz (approved):

PR #2801, opened 2024-07-04
[EASY] Registry storage RwLock

squadgazzz:
Switches from `Mutex` to `RwLock` allow readers not to block the registry storage.


MartinquaXD (approved):
FYI the Mutex was intentional since multiple parallel accesses should not be needed.
E.g. the autopilot only has to read it while building the auction and the driver only while pre-processing the auction (which also only happens once per auction across all solvers).

But given that Mateo's recent PR showed that it's also possible to not do that I'm fine with removing this potential footgun.

PR #2804, opened 2024-07-09
Order status API

MartinquaXD:
# Description
For the more accurate order progress bar the frontend team is working on we need to expose at which point in the lifecycle an order currently is. This is easily done by returning the last recorded event for a given order.
Also to display the solver competition for the order we attach all the bids for some of the states.

So far this PR is basically just the dusted off branch of the work done during the hackathon. This needs to be cleaned up more but I already wanted to create a PR to increase awareness and possibly let people take it over while I'm gone the next few days.

# Changes
A new endpoint that returns the order's current status with a list of solvers that proposed solutions.


crates/shared/src/api.rs
```diff
@@ -218,7 +218,16 @@ pub fn finalize_router(
     let cors = warp::cors()
         .allow_any_origin()
         .allow_methods(vec!["GET", "POST", "DELETE", "OPTIONS", "PUT", "PATCH"])
-        .allow_headers(vec!["Origin", "Content-Type", "X-Auth-Token", "X-AppId"]);
+        .allow_headers(vec![
```
MartinquaXD, line 221: Should be reverted before merging. Not sure why but this was needed to make this work with the local setup. When this is running in the actual infra this should not be necessary.

crates/orderbook/src/api/get_order_status.rs
```diff
@@ -0,0 +1,31 @@
...
+};
+
+fn get_status_request() -> impl Filter<Extract = (OrderUid,), Error = Rejection> + Clone {
+    warp::path!("v1" / "status" / OrderUid).and(warp::get())
+}
+
+pub fn get_status(
+    orderbook: Arc<Orderbook>,
+) -> impl Filter<Extract = (ApiReply,), Error = Rejection> + Clone {
+    get_status_request().and_then(move |order_uid| {
```
MartinquaXD, line 17: Maybe document in the openapi spec. But if we do, emphasis that the format might change at any time since we don't want to worry about stability on this endpoint.

alfetopito:
> This pull request has been marked as stale because it has been inactive a while. Please update this pull request or it will be automatically closed.

Shut up, bot. This IS coming!

crates/orderbook/src/api/get_order_status.rs
alfetopito: That would be great, specially since we use the openapi spec to generate the TS types in the SDK.

squadgazzz: Added OpenAPI doc.

crates/orderbook/src/database/orders.rs
```diff
@@ -529,6 +594,32 @@ fn is_buy_order_filled(amount: &BigDecimal, executed_amount: &BigDecimal) -> boo
     !executed_amount.is_zero() && *amount == *executed_amount
 }
 
+type SolverName = String;
+
+#[derive(Serialize, PartialEq, Debug, Clone)]
+#[cfg_attr(any(test, feature = "e2e"), derive(serde::Deserialize))]
+#[serde(tag = "type", rename_all = "camelCase", content = "value")]
+pub enum Status {
```
Mateo-mro, line 602: maybe is it worth adding `Expired`?

crates/orderbook/openapi.yml
```diff
@@ -469,6 +469,22 @@ paths:
...
+      summary: Get the status of an order.
+      parameters:
+        - in: path
+          name: order_uid
+          schema:
+              $ref: "#/components/schemas/UID"
+          required: true
+      responses:
+        200:
+          description: The order status with a list of solvers that proposed solutions(if applicable).
```
Mateo-mro, line 483:
```suggestion
          description: The order status with a list of solvers that proposed solutions (if applicable).
```

crates/orderbook/src/database/orders.rs
```diff
@@ -75,6 +77,7 @@ pub trait OrderStoring: Send + Sync {
         offset: u64,
         limit: Option<u64>,
     ) -> Result<Vec<Order>>;
+    async fn order_status(&self, order_uid: &OrderUid) -> Result<Status>;
```
Mateo-mro, line 80: shouldn't return `Result<Option<Status>>` what if the order doesn't exist?

crates/orderbook/src/database/orders.rs
```diff
@@ -356,6 +359,68 @@ impl OrderStoring for Postgres {
...
+                            }
+                        })
+                        .then_some(solution.solver)
+                })
+                .collect::<Vec<_>>();
+
+            Ok::<Vec<_>, anyhow::Error>(solvers)
+        };
+
+        let status = match status.label {
```
Mateo-mro, line 411: This could be done by implementing the `From<>` trait

crates/orderbook/src/database/orders.rs
```diff
@@ -356,6 +359,68 @@ impl OrderStoring for Postgres {
...
+            .await
+            .context("could not fetch status")?;
+        timer.stop_and_record();
+
+        let fetch_solvers = || async move {
+            let timer = super::Metrics::get()
+                .database_queries
+                .with_label_values(&["load_latest_solver_competition"])
+                .start_timer();
+            let competition = database::solver_competition::load_latest_competition(&mut ex)
```
Mateo-mro, line 379: What if the order exists but it is still not part of the latest competition? :thinking:

crates/orderbook/src/database/orders.rs
```diff
@@ -356,6 +359,68 @@ impl OrderStoring for Postgres {
...
+                    solution
+                        .orders
+                        .iter()
+                        .any(|o| match o {
+                            model::solver_competition::Order::Colocated { id, .. } => {
+                                id.0 == order_uid.0
+                            }
+                            model::solver_competition::Order::Legacy { id, .. } => {
+                                id.0 == order_uid.0
+                            }
```
Mateo-mro, line 402:
```suggestion
                            model::solver_competition::Order::Colocated { id, .. }
                            | model::solver_competition::Order::Legacy { id, .. } => {
                                id.0 == order_uid.0
                            }
```

crates/orderbook/src/database/orders.rs
squadgazzz: In that case, it returns an error because this API is not expected to check whether the order exists.

crates/orderbook/src/database/orders.rs
squadgazzz: `From` interface accepts `T` only, while we also need here a lazy list of solvers.

crates/orderbook/src/database/orders.rs
squadgazzz: For what purpose and how is it supposed to be obtained since `OrderEventLabel` doesn't track that.

crates/orderbook/src/database/orders.rs
squadgazzz: A warning log should probably be enough here. @alfetopito, should we fail in case the solvers list for the order in specific status is unexpectedly empty or can we pass the empty list to you?

crates/orderbook/src/database/orders.rs
Mateo-mro: But it is an inconsistent pattern. See this [function](https://github.com/cowprotocol/services/pull/2804/files#diff-d22c8612ee861d55e838fdc5487a1c139da73b795295670acdcccea91f805011R71).

crates/orderbook/src/database/orders.rs
Mateo-mro: From the frontend point of view, it is not needed. But since we have open API, and users potentially will want to fetch the status in their trading bots, `expired` comes quite handy to know the reason of why the order wasn't executed, since they created an order with a short `valid_to`.

crates/orderbook/src/database/orders.rs
squadgazzz: IMO, semantic of `single_order` and `order_status` is different. The former fetches the order if it exists and latter should return the order's status and absence of it should be considered as an error.

crates/orderbook/src/database/orders.rs
squadgazzz: Seems like this API is going to be heavily used by FE. Fetching `valid_to` requires additional DB query or introducing caching, which doesn't make much sense just for a single `Expired` status where the timestamp can be checked by users or using Explorer, IMO.

crates/e2e/src/setup/services.rs
```diff
@@ -23,6 +23,7 @@ use {
 
 pub const API_HOST: &str = "http://127.0.0.1:8080";
 pub const ORDERS_ENDPOINT: &str = "/api/v1/orders";
+pub const ORDER_STATUS_ENDPOINT: &str = "/api/v1/status";
```
alfetopito, line 26: I just thought about this.

Would it make sense to put this as part of the orders endpoint?
After all, we are querying an order status.
I mean this: `/api/v1/orders/<id>/status`

Unless we expect to extend this in the future for trades and so on, then it's all good.

Feel free to disregard.

crates/orderbook/src/database/orders.rs
alfetopito: I think returning an error makes sense.
I'd argue the error should be 404 in this case, though.

crates/orderbook/src/database/orders.rs
alfetopito: I don't think a failure is adequate here.
After all, there is a status, just not the full list of solvers is known, right?
I rather have it empty than not have it.

crates/orderbook/src/database/orders.rs
```diff
@@ -356,6 +359,70 @@ impl OrderStoring for Postgres {
...
+                        })
+                        .then_some(solution.solver)
+                })
+                .collect::<Vec<_>>();
+
+            if solvers.is_empty() {
+                tracing::warn!(?order_uid, order_status = ?status.label, "no solvers found for order");
+            }
```
alfetopito, line 408: The first states won't ever have this (open, scheduled, active), right?
Then I personally think there shouldn't be a warning, unless it's in a state that should (solved, executing, traded).
But this is more up to your discretion since it's internal to backend services.

crates/orderbook/src/database/orders.rs
```diff
@@ -356,6 +359,70 @@ impl OrderStoring for Postgres {
...
+            let solvers = competition
+                .solutions
+                .into_iter()
+                .filter_map(|solution| {
+                    solution
+                        .orders
+                        .iter()
+                        .any(|o| match o {
+                            model::solver_competition::Order::Colocated { id, .. }
+                            | model::solver_competition::Order::Legacy { id, .. } => {
```
alfetopito, line 398: Is it possible to include in the response whether the order is included in a given solution?
I ask because I had the understanding that the `solved` state would include the solvers listed by ranking.
Given that this state denotes when your order ISN'T in the winning solution, it could have been in one or more of the runner ups.

Based on this, there was a UI proposal to show which solutions included the user order:

image

Although it's not used in the most up to date proposal:

image

So, not a must, just a good to have if possible.

crates/orderbook/src/database/orders.rs
alfetopito: As Mateo said, not needed by the frontend ATM, and I can't say for other use cases.
Fine by me not having it.

crates/orderbook/src/database/orders.rs
squadgazzz: > Is it possible to include in the response whether the order is included in a given solution?

The retrieved competition contains all the solutions received(sorted by ranking), but only those containing the specified order uid are filtered in. So if the solvers list is empty, the latest competition doesn't contain that order. Do you expect to receive the solvers list with a flag on whether the order was included in a specific solution?

crates/orderbook/src/database/orders.rs
squadgazzz: It returns 404 now.

crates/orderbook/src/database/orders.rs
squadgazzz: Ignoring the competition fetching errors now.

crates/e2e/src/setup/services.rs
squadgazzz: Makes sense, thanks. Migrated.

crates/orderbook/src/database/orders.rs
alfetopito: It would be great to have all the solvers always, even if the order was not included.
This way we can have more transparency in the competition and users can see when a solver included their order or not.
The flag as proposed is enough 👍

crates/orderbook/openapi.yml
```diff
@@ -1038,6 +1054,33 @@ components:
...
+        value:
+          description: List of solvers and whether their solution includes the order.
+          type: array
+          items:
+            type: object
+            properties:
+              solver:
+                type: string
+                description: Name of the solver.
+              order_included:
```
alfetopito, line 1079: To keep the same standard as other properties, please use camelCase:

```suggestion
              orderIncluded:
```

crates/orderbook/src/database/orders.rs
```diff
@@ -529,6 +610,37 @@ fn is_buy_order_filled(amount: &BigDecimal, executed_amount: &BigDecimal) -> boo
     !executed_amount.is_zero() && *amount == *executed_amount
 }
 
+#[derive(Serialize, PartialEq, Debug, Clone)]
```
fleupold, line 613: I'm pretty sure this is not the right module for this logic to live in (look at the size of the module). If you look at the other functions in this module they are mainly plain read/write functions. It looks like you are also defining the actual api types here (in the db model) for which I cannot see any precedence.

Can we create a separate module just for order status that defines all of this and contains the logic around fetching and converting the raw competition result into what the fronted expects please.

fleupold (changes requested):
Cannot we separate pure db read logic and the business logic that makes sense out of it?

crates/orderbook/src/dto/order.rs
```diff
@@ -68,3 +68,34 @@ pub enum FeePolicy {
         quote: Quote,
     },
 }
+
+#[derive(Serialize, PartialEq, Debug, Clone)]
+#[cfg_attr(any(test, feature = "e2e"), derive(serde::Deserialize))]
+pub struct Solution {
```
fleupold, line 74: I think instead of just a boolean if the order was included, we should provide the executed amounts (so that in theory we can highlight how much more you got in the winning solution compared to the second best). So I'd turn the bool into an `Option<Amounts>`.

I'm not happy about exposing a public `Solution` struct with such a narrow content on the order dto module. I understand that it's inlined into Status (and thus not exposed) but in this case, can we make it private and create it from a public Status constructor?

I think we might have a proper `Solution` struct in the future which contains more stable information from the competitions table for a target tx and auciton.

fleupold (commented):
Factoring looks good. Can we get rid of this very broadly named Solution struct that only has a very narrow content (for a single order)

crates/orderbook/src/dto/order.rs
squadgazzz: > can we make it private and create it from a public Status constructor?

Technically, we can, but what would this constructor accept? A Vec of tuples?

> So I'd turn the bool into an Option<Amounts>.

That was the initial design, which was dropped because competition json contains [Legacy orders](https://github.com/cowprotocol/services/blob/b706d23f056f0338b5896c881af735514f1279cd/crates/model/src/solver_competition.rs#L106-L126), so that requires an openapi enum. Leandro told me that FE already tracks this info which would be redundant and also would overcomplicate the response. Do we really need it?

alfetopito: In the current UI proposal (https://www.figma.com/design/oSlMFMdHrELaPZACf1k7A3/Progress-Bar-Iteration?node-id=262-149&t=uTrK6IbgSRoSwQ0p-0) we are no longer showing individual amounts per solver.
This was discussed way back in the retreat/hackathon where it was raised that the winning solution does not guarantee best price for the individual order.

As for the winning one, we do need it in the last screen, but we get the order info from /orders endpoint already for other details like token addresses, limit price and so on, so it's not a must to have the amounts IMO.

fleupold: > Technically, we can, but what would this constructor accept? A Vec of tuples?

Would the `SolverCompetitionDB` type work?

> That was the initial design, which was dropped because competition json contains [Legacy orders](https://github.com/cowprotocol/services/blob/b706d23f056f0338b5896c881af735514f1279cd/crates/model/src/solver_competition.rs#L106-L126), so that requires an openapi enum

I would simply not support any legacy orders (it's ok to only support order status for new orders)

> [...] we are no longer showing individual amounts per solver.
This was discussed way back in the retreat/hackathon where it was raised that the winning solution does not guarantee best price for the individual orde

Well, we could simply not show them unless they are better

> Do we really need it?

If we are sure we will never want to display this information, sure we can leave it out. It just doesn't feel so hard to include and would allow a more delightful user experience.

squadgazzz: > Would the SolverCompetitionDB type work?

Not really. That means we also need to provide `order_uid` to the constructor where the business logic will leak. Maybe renaming it to SolverSolution or something similar would solve the issue with public access to the dto. Using tuples is way worse, IMO.

fleupold: > Maybe renaming it to SolverSolution

Solver solution doesn't add anything it just makes the name more vague (what other solutions are there than solver solutions). What is this thing? It's a view from the perspective of the order into the competition/solution candidates where the only thing it cares about is "am I included". Maybe something like:
- InclusionStatus
- Participation
- SolutionInclusion

all not great, but simply Solution is too broad imo.

squadgazzz: Renamed. Also included executed amounts into the response.

crates/orderbook/openapi.yml
```diff
@@ -1038,6 +1054,44 @@ components:
...
+          description: List of solvers and whether their solution includes the order.
+          type: array
+          items:
+            type: object
+            properties:
+              solver:
+                type: string
+                description: Name of the solver.
+              executedAmounts:
+                $ref: "#/components/schemas/ExecutedAmounts"
```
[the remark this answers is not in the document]
squadgazzz: As it was discussed [here](https://github.com/cowprotocol/services/pull/2804#discussion_r1692771811), `executedAmounts` is the optional field, which presence means inclusion.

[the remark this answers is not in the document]
fleupold: If easier for the frontend I think we could re-add the boolean to be extra explicit.

crates/orderbook/src/dto/order.rs
```diff
@@ -68,3 +68,46 @@ pub enum FeePolicy {
...
+    #[serde_as(as = "HexOrDecimalU256")]
+    pub sell: U256,
+    #[serde_as(as = "HexOrDecimalU256")]
+    pub buy: U256,
+}
+
+#[derive(Serialize, PartialEq, Debug, Clone)]
+#[cfg_attr(any(test, feature = "e2e"), derive(serde::Deserialize))]
+#[serde(rename_all = "camelCase")]
+pub struct SolutionInclusion {
```
fleupold, line 86: I feel like this struct should get some documentation (the name is still not great so explaining what this means might be helpful).

crates/orderbook/openapi.yml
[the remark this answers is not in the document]
alfetopito: Not a must, just wanted to clarify.

crates/orderbook/src/orderbook.rs
```diff
@@ -403,6 +406,66 @@ impl Orderbook {
             .await
             .context("get_user_orders error")
     }
+
+    pub async fn get_order_status(&self, uid: &OrderUid) -> Result<Option<dto::order::Status>> {
+        match self.database.latest_order_event(uid).await? {
```
sunce86, line 411: nit: Here `?` forwards db connection errors as `OrderNotFound` to the caller of `get_order_status`, not sure if desirable

crates/orderbook/src/orderbook.rs
```diff
@@ -403,6 +406,66 @@ impl Orderbook {
...
+                        })
+                        .collect::<Vec<_>>();
+
+                    Ok::<Vec<_>, anyhow::Error>(solutions)
+                };
+                let status = match event.label {
+                    OrderEventLabel::Ready => dto::order::Status::Active,
+                    OrderEventLabel::Created => dto::order::Status::Scheduled,
+                    OrderEventLabel::Considered => {
+                        dto::order::Status::Solved(fetch_solutions().await?)
```
sunce86, line 454: This will not return the latest competition based on the position where the `OrderEventLabel::Considered` is stored in the Runloop (before solver competition table is saved to database)?

crates/orderbook/src/orderbook.rs
```diff
@@ -403,6 +406,66 @@ impl Orderbook {
...
+                    }
+                    OrderEventLabel::Executing => {
+                        dto::order::Status::Executing(fetch_solutions().await?)
+                    }
+                    OrderEventLabel::Traded => dto::order::Status::Traded(fetch_solutions().await?),
```
sunce86, line 459: This is problematic as well, it depends how fast OnSettlementEventUpdater processed the settlement and races against new runloop iteration.

sunce86 (commented):
Mostly worried around using `load_latest_competition`.

One alternative I see is fetching `auction_id` for given `order_uid` from `[order_executions](database::order_execution)` and then using `solver_competition::load_by_id` instead.

crates/orderbook/src/orderbook.rs
squadgazzz: Not really, it returns a DB error which is then mapped to the internal error:
https://github.com/cowprotocol/services/blob/9a2745bed2cbcaa260d0cbe1e646ce88c42b417d/crates/orderbook/src/api/get_order_status.rs#L29-L32

sunce86: Ah yes, my bad

crates/orderbook/src/orderbook.rs
squadgazzz: Nice catch, thanks. Switched from the latest competition to the proposed query. Implemented using a single query to avoid multiple reads.

crates/database/src/solver_competition.rs
```diff
@@ -64,6 +64,24 @@ WHERE s.tx_hash = $1
...
 
+pub async fn load_by_order_uid(
+    ex: &mut PgConnection,
+    order_uid: OrderUid,
+) -> Result<Option<LoadCompetition>, sqlx::Error> {
+    const QUERY: &str = r#"
+SELECT sc.json, sc.id, s.tx_hash
+FROM solver_competitions sc
+-- outer joins because the data might not have been indexed yet
+LEFT OUTER JOIN settlements s ON sc.id = s.auction_id
```
fleupold, line 75: A partially fillable order will be part of many competitions, so this doesn't work.

crates/orderbook/src/solver_competition.rs
```diff
@@ -11,6 +11,7 @@ use {
 pub enum Identifier {
     Id(AuctionId),
     Transaction(H256),
+    OrderUid(OrderUid),
```
fleupold, line 14: I'm not a fan of adding a non unique identifier here (there is only one competition per ID/tx_hash, but there can be many for an order uid)

fleupold (changes requested):
I don't think this new `load` is a good idea.

How much work would it be to store the auction id with the event? Can it be somehow inferred from the timestamp?

crates/database/src/solver_competition.rs
squadgazzz: Wouldn't sorting by auction_id desc help?

squadgazzz:
> How much work would it be to store the auction id with the event?

Seems like, not too much. The only thing is how to properly fill the existing statuses with auction ids to avoid optional values. I can think of filling `Considered`, `Traded` and `Executing` labels with the latest corresponding `auction_id` from the `order_execution` table. But seem like the same ordering can be done in the existing `load_by_order_uid` query.

sunce86:
Yeah, the issue here is that labels should be stored per trade, and not per order. meaning we should have a `<auction_id, order_id>` as primary key.

With regards how to populate historic auction_ids, if we want to keep `order_events` informative (not super accurate) then previous entries are not that important, especially if it is used for progress bar which is relevant only at the moment of order execution. So, we could populate those auction_ids with latest auction_id from order_execution for a specific order, as you mentioned.

Or we could simply drop the `order_events` table completely and create a new one properly. Do we even use `order_events` for anything other than debugging?

sunce86:
Is it acceptable to completely drop `order_events` table and create a new one containing non-optional values? Do we actually use `order_events` now for anything other than debugging?

squadgazzz:
> Do we actually use order_events now for anything other than debugging?

Also, we keep events for the last 30 days only:
https://github.com/cowprotocol/services/blob/8edb6f28724258394f60ca27dd41a5473a785480/crates/autopilot/src/arguments.rs#L216-L219

crates/orderbook/src/orderbook.rs
sunce86: > Even though FE is lucky enough to fetch the status when the competition is not in the DB yet, a subsequent request should fetch the actual data.

Makes sense, although I'm not convinced backend should make any assumptions on how the FE would use it. Also, in the future, we might add some code between these two calls that could take more time and break this assumption.

I am not that worried that `Considered` would be missing data in this edge case, but rather that it could contain wrong data (from previous partial execution of this trade, or something similar), and then, some of the subsequent runloop code might fail and end the iteration early, leaving the order status on FE in inaccurate state.

sunce86:
Nice. In that case, I would suggest:
1. Drop `Considered` completely
2. Add non-optional `auction_id` to `Executing` and `Traded` which can be used for fetching solutions right.
3. For existing rows in `order_events` set `auction_id` to 0 (don't even bother populating it with anything).
4. Make `infra::persistence::store_order_events` work in a non fire-and-forget manner, and handle errors properly. This function now populates data that is served over backend API so it needs to be properly handled.

fleupold:
> Do we actually use order_events now for anything other than debugging?

We use it for our time to happy moo metrics.

> Btw I don't think we need both Considered and Executing as they both refer to the order being part of the winning solution.

That seems like a bug. `Considered` should apply to orders that were part of a solution (not necessarily the winning one).


I would honestly not over-engineer things here (and instead ship what we had to unblock the frontend). If in some rare cases the progress bar is not 100% accurate that's ok (it's not worth delaying this feature another week and blocking us from doing work that actually improves the time to happy moo).

squadgazzz:
> I would honestly not over-engineer things here (and instead ship what we had to unblock the frontend). If in some rare cases the progress bar is not 100% accurate that's ok (it's not worth delaying this feature another week and blocking us from doing work that actually improves the time to happy moo).

Ok, then we still need to add the optional `auction_id` column to the `order_events` table. That seems like a small change.

fleupold:
> Ok, then we still need to add the optional auction_id column to the order_events table. That seems like a small change.

I'm fine with that if it's small, but even this doesn't seem necessary to me. We always fetch the latest event and then in some cases the latest competition. If they are not matching (which should only happen in very rare cases) it's not the end of the world.

squadgazzz:
Ok, reverted back to the `get_latest_competition` version.

fleupold (approved):

alfetopito:
@squadgazzz is this good for merging already?

PR #2805, opened 2024-07-09
Release job check for changes fix

squadgazzz:
Check for changes job was wrongly implemented. The PR fixes that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,8 +29,7 @@ jobs:
 
       - name: Check for changes since last tag
         run: |
-          CHANGES=$(git log $LATEST_VERSION..HEAD --oneline)
-          if [[ -n "$CHANGES" ]]; then
+          if ! git diff $LATEST_VERSION..HEAD --quiet; then
             echo "CHANGES_DETECTED=true" >> $GITHUB_ENV
           fi
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,6 @@ name: Weekly Release
 
 on:
   workflow_dispatch:
-  push:
   schedule:
     # At 00:00 on Monday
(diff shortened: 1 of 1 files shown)


Mateo-mro (approved):

.github/workflows/release.yml
```diff
@@ -27,9 +27,12 @@ jobs:
           fi
           echo "latest=$LATEST_VERSION" >> $GITHUB_OUTPUT
 
-      - name: Check for changes
+      - name: Check for changes since last tag
         run: |
-          git diff --exit-code || echo "CHANGES_DETECTED=true" >> $GITHUB_ENV
+          CHANGES=$(git log $LATEST_VERSION..HEAD --oneline)
```
MartinquaXD, line 32: Isn't using `diff` technically more correct? Will probably never happen but you could get to the same final state in multiple ways (sequences of commits).
What exactly was the issue with the `diff` command that it led to the wrong result here?

squadgazzz: I just copied that from the `gnosis/solvers` job. It was used since the release script doesn't really care about the code changes but the commits made since the latest release(no matter the reason they were made).

squadgazzz: Switched to `git diff`.

.github/workflows/release.yml
```diff
@@ -2,6 +2,7 @@ name: Weekly Release
 
 on:
   workflow_dispatch:
+  push:
```
MartinquaXD, line 5: This is just for testing the PR and should be removed before merging, right?

MartinquaXD (approved):
Approved assuming the last discussion gets resolved (either by removing the change or explaining why it's needed)

.github/workflows/release.yml
squadgazzz: Yes

PR #2807, opened 2024-07-10
[TRIVIAL] Remove orderbook's `axum` dependency

squadgazzz:
Removes `axum` dependency from the orderbook crate since it is not used there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3287,7 +3287,6 @@ dependencies = [
  "app-data",
  "app-data-hash",
  "async-trait",
- "axum",
  "bigdecimal",
  "cached",
  "chrono",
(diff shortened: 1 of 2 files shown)


Mateo-mro:
@squadgazzz please could you run the udeps now that you are on it, to check if we have more of these cases? :pray: :pray:

Mateo-mro (approved):

squadgazzz:
> @squadgazzz please could you run the udeps now that you are on it, to check if we have more of these cases? 🙏 🙏

I am trying, but it requires a nightly build, which fails with weird errors. Looking for a workaround.

Mateo-mro:
> > @squadgazzz please could you run the udeps now that you are on it, to check if we have more of these cases? 🙏 🙏
> 
> I am trying, but it requires a nightly build, which fails with weird errors. Looking for a workaround.

You can use a nightly build to check for it, it won't work otherwise I believe.

fleupold (approved):

PR #2812, opened 2024-07-15
[orderbook] Use alternative memory allocator

squadgazzz:
Trying out to use alternative memory allocators. The plan is to test `jemalloc` and then `mimalloc`. Requires an infra PR to configure them through env params.

> [!IMPORTANT]
> Should not be merged until a positive result is confirmed.


squadgazzz:
Closing since nither jemalloc nor mimalloc helped.

PR #2836, opened 2024-07-25
Fix index for ethflow_orders table

sunce86:
# Description
Previously index was created on a wrong table: "orders" instead of "ethflow_orders":
https://github.com/cowprotocol/services/blob/main/database/sql/V031__create_ethflow_orders.sql#L11

Should speed up the OPEN_ORDERS query, although not sure how much.


fleupold (approved):

Mateo-mro (approved):

squadgazzz (approved):
Gave it a try locally, didn't really help(still have 3-4s to execute).

squadgazzz:
We have 2 identical indexes: user_valid_to and order_valid_to, right? One should be dropped.

sunce86:
Relevant to https://github.com/cowprotocol/services/issues/2831#issuecomment-2251188741

sunce86:
> We have 2 identical indexes: user_valid_to and order_valid_to, right? One should be dropped.

Dropped `user_valid_to`

PR #2840, opened 2024-07-26
Fix scaling factor in driver

sunce86:
# Description
Furucombo people noticed the scaling of prices in the driver is not working properly.

Example auction: 9172009
[Example log](https://production-6de61f.kb.eu-central-1.aws.cloud.es.io/app/discover#/doc/c0e240e0-d9b3-11ed-b0e6-e361adffce0b/cowlogs-prod-2024.07.23?id=cpXk3JABP0v0FZqw807z) of two solutions that should have been merged.

The `fn scaling_factor` function description is correct:

> /// Given two solutions returns the factors with
> /// which prices of the second solution would have to be multiplied so that the
> /// given token would have the same price in both solutions.

which means, we need to calculate `factor = first_price / second_price` so that `factor` represents the number with which `second_price` should be multiplied to get `first_price`

### DIFF
--- a/crates/driver/src/tests/setup/mod.rs
+++ b/crates/driver/src/tests/setup/mod.rs
@@ -14,8 +14,10 @@ use {
         },
         tests::{
             cases::{
+                is_approximately_equal,
                 EtherExt,
                 AB_ORDER_AMOUNT,
+                AD_ORDER_AMOUNT,
                 CD_ORDER_AMOUNT,
                 DEFAULT_POOL_AMOUNT_A,
                 DEFAULT_POOL_AMOUNT_B,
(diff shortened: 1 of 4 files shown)


fleupold (commented):
Great catch. Would you be able to add a test case [here](https://github.com/cowprotocol/services/blob/67311297d24a02ea9393b6765f44a16a6a205780/crates/driver/src/tests/cases/merge_settlements.rs#L51) that fails before the PR and passes now?

sunce86:
> Great catch. Would you be able to add a test case [here](https://github.com/cowprotocol/services/blob/67311297d24a02ea9393b6765f44a16a6a205780/crates/driver/src/tests/cases/merge_settlements.rs#L51) that fails before the PR and passes now?

Added test.

crates/driver/src/tests/setup/mod.rs
```diff
@@ -1175,8 +1208,14 @@ impl<'a> SolveOk<'a> {
...
-            assert!(u256(trade.get("sellAmount").unwrap()) == expected_sell);
-            assert!(u256(trade.get("buyAmount").unwrap()) == expected_buy);
+            assert!(is_approximately_equal(
+                u256(trade.get("sellAmount").unwrap()),
+                expected_sell
+            ));
+            assert!(is_approximately_equal(
+                u256(trade.get("buyAmount").unwrap()),
+                expected_buy
+            ));
```
sunce86, line 1218: Had to use `is_approximately_equal` since now the actual scaling math is used for the first time. The consequence is that scaling two U256 numbers by the same `factor` doesn't mean those two numbers have the same proportion as before scaling, which may introduce rounding errors when the executed sell/buy amounts are calculated.

fleupold (approved):

sunce86:
@squadgazzz @Mateo-mro would like this to be merged on Monday so it get's into the weekly release.

squadgazzz (approved):

PR #2843, opened 2024-07-28
Autopilot refactor 3/4: Get competition from persistence

sunce86:
# Description
Gets winner competition data from persistence. 

This is needed to [support existing checks for winner](https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/on_settlement_event_updater.rs#L315-L322) and also do a lite checks for score and prepare a basis for fetching the rest of the competition data that will be needed for circuit breaker implementation etc.

# Changes

- [ ] Implemented `get_winner` function on `infra::Persistence`
- [ ] Moved out `deadline` from `Auction`

## How to test
Already exists temporary code that prints the settlement.


crates/autopilot/src/domain/competition/mod.rs
```diff
@@ -1,98 +1,16 @@
...
-    crate::domain,
-    std::collections::HashMap,
-};
+use super::eth;
 
-type SolutionId = u64;
+mod solution;
 
-#[derive(Debug)]
-pub struct Solution {
```
sunce86, line 10: The whole file (without any changes) is moved to domain::competition::solution.rs

crates/autopilot/src/domain/competition/solution.rs
```diff
@@ -0,0 +1,100 @@
+use {
```
sunce86, line 1: so no new code in this file, feel free to skip review of it.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -67,6 +70,9 @@ impl Settlement {
 pub enum Error {
     #[error(transparent)]
     Solution(#[from] solution::Error),
+    // TODO: Merge Auction and Competition errors into a single error type
```
sunce86, line 73: will do it in this PR if I get 👍 for putting `Competition` inside `Settlement`.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -28,6 +28,7 @@ pub struct Settlement {
     solution: Solution,
     transaction: Transaction,
     auction: Auction,
+    competition: domain::Competition,
```
fleupold, line 31: Why is competition a single solver, score, block? Isn't `competition` the set of all solutions that were submitted? Would those fields make more sense on `Solution`? Or can a solution exist without that information?

sunce86: > Why is competition a single solver, score, block?

Winner and winner score are outputs of the competition. Block is also more related to competition that it is related to `settlement::Auction`, and we also keep `settlement::Auction` more 1 to 1 with `domain::Auction` so we can merge them eventually.

>  Isn't competition the set of all solutions that were submitted

True. However we don't need this info to support current functionalities of OnSettlementEventUpdater. Once we start implementing circuit breaker, all competition related data is supposed to be put under `domain::Competition` (solutions (if we need them), winner/s etc).

> Would those fields make more sense on Solution? Or can a solution exist without that information?

If you mean `settlement::Solution`, no. `settlement::Solution` is an entity on it's own that shouldn't be related to anything in the competition, otherwise it would be completely non-reusable. `settlement::Auction` and `domain::Competition` are separated at the moment, as I am still not sure what their relation should be. Auction is input and Competition is output of the runloop iteration. Or we can think of Competition in a broader context in which it contains an Auction.

fleupold: > Winner and winner score are outputs of the competition.

They are also part of the solution though. `score()` is a function on `Solution` and `winner` is implicitly part of the transaction that the solution is built from. Why do we define a data object that has different ways of accessing the same thing?

I agree that deadline should probably not be on solution (and also not on auction, although it could be at least when we move to one block per batch). Until then, should we not have it as an isolated field on Settlement?

I really find the concept of "competition" for the three values that we defined here inappropriate.

sunce86: > They are also part of the solution though. score() is a function on Solution and winner is implicitly part of the transaction that the solution is built from. Why do we define a data object that has different ways of accessing the same thing?

I am not sure I understand this. `Settlement::Solution` has a `score()` function but that value has nothing to do with `domain::Competition::score` which is a driver reported score. Similarly, `Transaction::from` has nothing to do with `domain::Competition::winner` as those can be different as well (if non-winning solver settled a Settlement for example).

Maybe the name is not appropriate but the `domain::Competition` object represents/contains important data that is related to what happens at COMPETITION time. I am not happy about saving the `domain::Competition` as part of the `Settlement` so that might be confusing. But then, why we allowed a `settlement::Auction` to be part of the `Settlement`. Both `settlement::Auction` and `domain::Competition` are objects built and used at competition time and are associated to this `Settlement`.

sunce86: Unless you start with an assumption that all errors (like my example of non-winning solver settling a Settlement) are already handled in `Settlement` constructor so there is no need to save `winner` inside `Settlement` because that would be an error during construction of `Settlement`?

If so, then I need to implement this in this PR and completely skip saving `domain::Competition` inside `Settlement`.

sunce86: Screenshot 2024-07-30 at 00 08 35

Is this approach you would prefer?

sunce86: > However, shouldn't the same then be true when auction is non-existent (ie. when a solver submitted a tx without referencing a valid auction)?

Current behaviour is to skip creating observation in this case: https://github.com/cowprotocol/services/blob/main/crates/autopilot/src/on_settlement_event_updater.rs#L156-L160

observation() is a function on Settlement just because it requires Transaction and Auction to be built. 

> which assumes valid auction

This wasn't my assumption from beginning, just because, as you mentioned, we sometimes need invalid settlements to work with. Instead, i planned to have a `Settlement` containing:
1. settlement calldata transformed to Solution because this is 99.9999% valid
2. fetch auction data
3. fetch competition data (solutions, winner, whatever happened at competition time).
4. fetch something else? like existence of settled orders in the database to detect if they are JIT orders etc.

Then, if we see that we have a need for `ValidSettlement` we can easily transform `Settlement` -> `ValidSettlement` that will have those guarantees you talked about.

fleupold: But why do we want to create inconsistent behaviour? If a solver misbehaves and submits an invalid settlement for whatever reason (lying about the auction id, lying about the score, lying about executions) should there be a settlement observation or not?

It sounds like without the auction we cannot create an "observation" simply because we cannot compute the score and fees without an auction. In this case, I think there should not be an observation in case a solver is lying about anything and thus `Settlement::new` can perform all required correctness checks. If in other places we want information about invalid transactions as well, we can still create a `Solution` from the transaction itself.

sunce86: Refactored, not sure if I addressed the original comment.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +293,50 @@ impl Persistence {
...
+            (
+                H160(scores.winner.0).into(),
+                competition::Score::new(
+                    big_decimal_to_u256(&scores.winning_score)
+                        .ok_or(error::Winner::InvalidScore(anyhow::anyhow!(
+                            "database score"
+                        )))?
+                        .into(),
+                )
+                .map_err(|_| error::Winner::InvalidScore(anyhow::anyhow!("zero score")))?,
```
squadgazzz, line 333: nit: extract into a variable since it is quite hard to read between the tuple elements.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -49,6 +49,30 @@ impl Settlement {
...
+                got: score,
+            });
+        }
+
+        if transaction.block >= deadline {
+            tracing::warn!(
+                "Settlement for auction {} was submitted after the deadline",
+                solution.auction_id()
+            );
+        }
```
fleupold, line 91: Do we want to allow this? Do we want to keep some "leeway" but not have unlimited delay?

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -80,8 +99,22 @@ pub enum Error {
     Solution(#[from] solution::Error),
     #[error("settlement refers to an auction from a different environment")]
     WrongEnvironment,
-    #[error("connection to the persistence layer failed: {0}")]
+    #[error(transparent)]
     PersistenceConnection(#[from] infra::persistence::Error),
     #[error(transparent)]
     Auction(#[from] infra::persistence::error::Auction),
+    #[error(transparent)]
```
fleupold: I may sound like a broken record, but this type of error handling doesn't scale. We are just defining new error types and keep blowing up higher level structs into a composite nightmare. How is any callsite going to handle these errors except from simply printing them?

I think we should think about the higher level type of errors that a settlement can have:
1. Issues with our persistence (those are currently part of PersistenceConnection, Auction, Winner). Those should likely be retried and are not fatal.
2. Issues with the settlement that make the settlement invalid (currently we also have a benign WrongEnvironment error, but this differentiation only matters if we want to circuit break in the autopilot).

These are the two error types we should express and we should provide conversions from all sub error types into such.

Please don't use blanket `#[from]` conversions. Those are a code smell imo and lead to unhandlable error structs that are not any better than using `anyhow` for all types.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -49,6 +49,30 @@ impl Settlement {
         }
 
         let auction = persistence.get_auction(solution.auction_id()).await?;
+        let (winner, winner_score, deadline) =
+            persistence.get_winner(solution.auction_id()).await?;
```
fleupold: Let's give this a domain type (but not expose this type on `Settlement`). I'm not super happy about winner (as this to me is just the solver address). Maybe `Bid`, `Solution` (Solution is already quite loaded in our domain), `SolverPromise` or something similar?

We should also bear in mind that in the future we may have multiple winning bids (in the case of combinatorial auctions).

I think conceptually this is pretty similar to the existing `domain::competition::Solution` type (as technically we also need to check that the promised orders were fulfilled) but it seems we are not fully persisting this yet. We could probably recover it from the competition table though (modulo solution id).

crates/autopilot/src/domain/settlement/auction.rs
```diff
@@ -13,9 +13,6 @@ pub struct Auction {
     pub orders: HashMap<domain::OrderUid, Vec<domain::fee::Policy>>,
     /// Auction external prices
     pub prices: domain::auction::Prices,
-    /// Deadline for an auction solution to be settled, so that it is eligible
-    /// for rewards.
-    pub deadline: domain::eth::BlockNo,
```
fleupold, line 18: Maybe we should keep checking against this value? Seems more appropriate on the auction than on a winning solution?

crates/autopilot/src/infra/persistence/mod.rs

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +293,50 @@ impl Persistence {
             id: auction_id,
             orders,
             prices,
-            deadline,
             surplus_capturing_jit_order_owners,
         })
     }
+
+    /// Get competition winner.
+    pub async fn get_winner(
```
fleupold, line 291: Can we call this `get_winning_bid` and return a `domain::competition::Bid` (or even better reconstruct a `domain::competition::Solution`)?

fleupold (commented):
I kind of want to move forward here, but there are things that I think are annoying in the long run
1. Creation of one of new persistence methods for eache single use case. Can we not reuse some of our existing domain models to express what we are trying to check (the winning `competition::Solution` got mined). Otherwise it's a sign that our domain is poorly designed (if we keep adding new types for every single thing we create).
2. Lazy/non-existent error handling (this has come up a lot in the past)

Both don't seem necessarily blocking and can be addresses later on, but I'd least like to hear your thoughts and refactor the `get_winner` method to make it less leaky.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: Handled in the meantime (you probably reviewed non-latest commit)

crates/autopilot/src/domain/settlement/auction.rs
sunce86: Then I move away further from merging this struct to `domain::Auction` that we talked about somewhere.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: > We could probably recover it from the competition table though (modulo solution id)

This requires either:
1. Saving solution id so it can be fetched
2. Making solution id optional field on Solution
3. Making another subtype that wont have solution id, let's say Solution and SolutionWithId.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: For OnSettlementEventUpdater (right now, since this is a refactor for current behaviour) we want to allow this and save observation to database. For circuit breaker I am not sure what is a desired behaviour, so will have to consult the solver team and how they handle it.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -49,6 +49,30 @@ impl Settlement {
...
+
+        if transaction.solver != winner {
+            return Err(Error::WinnerMismatch {
+                expected: winner,
+                got: transaction.solver,
+            });
+        }
+
+        let score = solution.score(&auction)?;
+        if score != winner_score {
```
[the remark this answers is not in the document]
sunce86: Fixed. Also, score is now only in debugging mode for testing purposes.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: get winner now fetches winning solution from solver competition db. solution id refactored with (2) since (1) would take too long to implement.

crates/autopilot/src/domain/settlement/auction.rs
sunce86: Removed completely deadline. Will be implemented and thought about when implementing CircuitBreaker.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: ended with `domain::competition::Solution`

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,96 @@ impl Persistence {
             id: auction_id,
             orders,
             prices,
-            deadline,
             surplus_capturing_jit_order_owners,
         })
     }
+
+    /// Get competition winner.
```
fleupold, line 290: Amazing comment 👏 

We could describe here what where we get this data from and how it's assembled but as is it has no value.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +112,59 @@ impl Settlement {
...
+        expected: eth::Address,
+        got: eth::Address,
+    },
+    #[error("failed to get score from database for a coresponding competition solution, err: {0}")]
+    InvalidScore(anyhow::Error),
     #[error(transparent)]
-    Auction(#[from] infra::persistence::error::Auction),
+    Score(#[from] solution::error::Score),
+    #[error("failed to get competition data from database {0}")]
+    SolverCompetition(anyhow::Error),
```
fleupold, line 137: So this is a Database error as well, isn't it?

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +112,59 @@ impl Settlement {
...
-    #[error("connection to the persistence layer failed: {0}")]
-    PersistenceConnection(#[from] infra::persistence::Error),
+    #[error("auction not found in the database")]
+    MissingAuction,
+    #[error("failed to get fee policy from database: {0} for order: {1}")]
+    FeePolicy(infra::persistence::dto::fee_policy::Error, domain::OrderUid),
+    #[error("failed to get price from database for token: {0}")]
+    Price(eth::TokenAddress),
+    #[error("winner mismatch: expected competition winner {expected}, settlement solver {got}")]
+    WinnerMismatch {
```
fleupold, line 128: nit: SolverMismatch

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,96 @@ impl Persistence {
...
+    pub async fn get_competition_winner(
```
fleupold, line 291:
```suggestion
    pub async fn get_winning_solution(
```

We are not getting a winner, we are getting a solution.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +112,59 @@ impl Settlement {
...
+        expected: eth::Address,
+        got: eth::Address,
+    },
+    #[error("failed to get score from database for a coresponding competition solution, err: {0}")]
+    InvalidScore(anyhow::Error),
     #[error(transparent)]
-    Auction(#[from] infra::persistence::error::Auction),
+    Score(#[from] solution::error::Score),
```
fleupold, line 135: How is Score different from InvalidScore? Also, here you are still blindly converting.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -31,6 +28,10 @@ pub struct Settlement {
 }
 
 impl Settlement {
+    pub fn auction_id(&self) -> domain::auction::Id {
+        self.solution.auction_id()
+    }
+
```
fleupold, line 34: unnecessary method

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -321,24 +390,38 @@ impl Metrics {
...
-        #[error(transparent)]
-        Price(#[from] domain::auction::InvalidPrice),
-        #[error("jit order owners not found for an existing auction id")]
-        MissingJitOrderOwners,
+        #[error("failed to fetch price for token: {0}")]
+        Price(eth::TokenAddress),
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Winner {
```
fleupold, line 417: These errors are not specific to a `Winner` but to loading any solution from the Database. Can we please make it solution generic?

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,96 @@ impl Persistence {
...
+                            },
+                        );
+                    }
+                    model::solver_competition::Order::Legacy {
+                        id: _,
+                        executed_amount: _,
+                    } => {
+                        return Err(error::Winner::SolverCompetition(anyhow::anyhow!(
+                            "Legacy order"
+                        )))
```
fleupold, line 355: This should be treated equivalent to NotFound in my opinion (we can document that this method only works for auctions that run in co-located mode)

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,96 @@ impl Persistence {
...
+                .into(),
+        )
+        .map_err(|_| error::Winner::InvalidScore(anyhow::anyhow!("zero score")))?;
+
+        let solution = {
+            // TODO: stabilize the solver competition table to get promised solution.
+            let solver_competition = database::solver_competition::load_by_id(&mut ex, auction_id)
+                .await
+                .map_err(error::Winner::DatabaseError)?
+                .ok_or(error::Winner::Missing)?;
```
fleupold, line 327:
```suggestion
                .ok_or(error::Solution::NotFound)?;
```

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,96 @@ impl Persistence {
...
+                        )))
+                    }
+                }
+            }
+            let mut prices = HashMap::new();
+            for (token, price) in winning_solution.clearing_prices.clone().into_iter() {
+                prices.insert(
+                    token.into(),
+                    domain::auction::Price::new(price.into()).map_err(|_| {
+                        error::Winner::SolverCompetition(anyhow::anyhow!(
```
fleupold, line 364: InvalidScore is it's own subtype but invalid price isn't? This feels inconsistent.

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -40,15 +41,55 @@ impl Settlement {
...
+                        solution.auction_id(),
+                        score,
+                        competition_winner.score()
+                    );
+                }
+            }
+            Err(err) => {
+                tracing::warn!(?err, "failed to calculate score for settlement");
+            }
+        }
```
fleupold, line 92: Why not turn this into a proper error. You are already logging the result of `Settlement::new` somewhere else for debugging purposes. Logging it here again just adds more places we are doing things temporarily for testing. The way you had it before was better imo.

fleupold (commented):
I think the recovering the solution is a good choice. However the Error types are have become even more messier than before. Is it not possible to make a proper first principles top down taxonomy of what types of error the caller of `Settlement::new` cares about and then categorize all the errors we have from all the different callsites accordingly? It feels like the top level error is driven by all the different subtype errors and things are just a complete mess.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: I'm logging settlement observation only. Score is not logged and it's impossible to log since it's part of the settlement constructor now. I was hesitant of returning error before making sure the score is properly calculated through debugging.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved, if you meant this needs to be renamed

crates/autopilot/src/domain/settlement/mod.rs

crates/autopilot/src/domain/settlement/mod.rs
sunce86: renamed to BuildingScore... rest of the error variants also renamed so hopefully it's clearer now.

> Also, here you are still blindly converting

Why do you think it is bad in this case? This error happens at one place exactly so it's very unlikely for this variant to misbehave like we had the example with `[from] anyhow::Error`

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved

crates/autopilot/src/domain/settlement/mod.rs

crates/autopilot/src/domain/settlement/mod.rs
sunce86: No, this should cover errors related to conversion of solver competition data to solution. refactor so let's see if better now.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: resolved

sunce86:
> Error types are have become even more messier than before. Is it not possible to make a proper first principles top down taxonomy of what types of error the caller of Settlement::new cares about and then categorize all the errors we have from all the different callsites accordingly? It feels like the top level error is driven by all the different subtype errors and things are just a complete mess.

I was not happy with error variants as well, but hoped I could clean them up somehow in the PR where actual handling of errors will be implemented (as it is not easy to define error variants super nicely in advance).

For now I see several subtypes of errors Settlement is interested in to potentially react different:
1. Bad communication with database (sqlx errors)
2. Bad preparation of data fetched from database, like missing data, bad conversion into dto objects, unexpected values etc.
3. Rest - errors during Settlement constructor execution - that can be different and each one is potentially handled differently.

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -199,7 +199,7 @@ impl Inner {
 
             tracing::info!(
                 "settlement object {:?}",
-                settlement.map(|settlement| (settlement.observation(), settlement.score()))
```
fleupold, line 202: Why not keep this line and turn invalid score computations into an `ScoreMismatch` error that has both scores and gets printed?

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +111,88 @@ impl Settlement {
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error(transparent)]
-    Solution(#[from] solution::Error),
+    #[error("failed communication with the database: {0}")]
+    BadCommunication(sqlx::Error),
```
fleupold, line 115: naming nit (bad communication when handles higher up in the domain is not really descriptive IMO, we should make it clear that it's an infra or persistence error).

```suggestion
    Infra(sqlx::Error),
```

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +111,88 @@ impl Settlement {
...
+    #[error("failed to prepare the data fetched from database for domain: {0}")]
+    BadPersistenceData(BadPersistenceData),
```
fleupold, line 117:
```suggestion
    InconsistentData(BadPersistenceData),
```

crates/autopilot/src/domain/settlement/mod.rs
fleupold: you were logging it before though (you can still log if the expected score doesn't match the computed one)

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +111,88 @@ impl Settlement {
...
+#[derive(Debug, thiserror::Error)]
+pub enum BadPersistenceData {
+    #[error("auction not found in the persistence layer")]
+    AuctionNotFound,
+    #[error("proposed solution not found in the persistence layer")]
+    SolutionNotFound,
+    #[error("invalid fee policy fetched from persistence layer: {0} for order: {1}")]
+    InvalidFeePolicy(infra::persistence::dto::fee_policy::Error, domain::OrderUid),
+    #[error("invalid fetched price from persistence layer for token: {0}")]
+    InvalidPricce(eth::TokenAddress),
```
fleupold, line 145:
```suggestion
    InvalidPrice(eth::TokenAddress),
```

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -76,12 +111,88 @@ impl Settlement {
...
+    BadPersistenceData(BadPersistenceData),
     #[error("settlement refers to an auction from a different environment")]
     WrongEnvironment,
-    #[error("connection to the persistence layer failed: {0}")]
-    PersistenceConnection(#[from] infra::persistence::Error),
     #[error(transparent)]
-    Auction(#[from] infra::persistence::error::Auction),
+    BuildingSolution(#[from] solution::Error),
+    #[error(transparent)]
+    BuildingScore(#[from] solution::error::Score),
```
fleupold, line 123: Isn't building score also a data inconsistency (can only happen if score is 0, a price is missing or the math overflows)?

fleupold (approved):
Approving to make progress. Agree that error handling can be refactored once we use it. I still disagree with having print statements for debugging in `Settlement::new` (I think this should all be put in one place which we hopefully remove in episode 4 of this stack).

But nothing blocking...

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -40,15 +37,58 @@ impl Settlement {
...
+                );
+
+                if score < winning_solution.score() {
+                    tracing::warn!(
+                        "Settlement for auction {} has lower score {} than the winning solution {}",
+                        solution.auction_id(),
+                        score,
+                        winning_solution.score()
+                    );
+                }
```
squadgazzz, line 82: nit: print only one of the logs. Also, having `auction_id` as a log separate field would be better for kibana search.
```suggestion
                if score < winning_solution.score() {
                    tracing::warn!(
                        auction_id = ?solution.auction_id(),
                        "Settlement for auction has lower score {} than the winning solution {}",
                        score,
                        winning_solution.score()
                    );
                } else {
                    tracing::debug!(
                        auction_id = ?solution.auction_id()
                        "Settlement for auction has score {}",
                        score
                    );
                }
```

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -40,15 +37,58 @@ impl Settlement {
...
+            }
+            Err(err) => {
+                tracing::warn!(
+                    ?err,
+                    "Settlement for auction {} failed to calculate score",
+                    solution.auction_id()
+                );
```
squadgazzz, line 89: ditto
```suggestion
                tracing::warn!(
                    auction_id = ?solution.auction_id(),
                    ?err,
                    "Settlement for auction failed to calculate score"
                );
```

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,97 @@ impl Persistence {
...
+                            competition::TradedAmounts {
+                                sell: (*sell_amount).into(),
+                                buy: (*buy_amount).into(),
+                            },
+                        );
+                    }
+                    model::solver_competition::Order::Legacy {
+                        id: _,
+                        executed_amount: _,
+                    } => return Err(error::Solution::NotFound),
```
squadgazzz, line 361: Maybe something related to the inconsistent state/unexpected results?

squadgazzz (approved):
LG. Only nits above.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: yes, I had a public function `score` which is removed as not used anymore.

crates/autopilot/src/domain/eth/mod.rs
```diff
@@ -16,7 +18,7 @@ pub struct TxId(pub H256);
 /// An ERC20 token address.
 ///
 /// https://eips.ethereum.org/EIPS/eip-20
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, From, Into)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, From, Into, Display)]
```
Mateo-mro, line 21: `Display` automatically derives `ToString`. For `H160` the `ToString` does not print the full address, but it prints it in the format: `0x1234...ABCD` (the three dots are also printed like that, literally). For which purpose will we be using the `Display`? if we need to know the precise address, it's better to implement our own `Display` in which we guarantee the full address is displayed.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -198,48 +193,37 @@ impl Persistence {
...
-            .map_err(error::Auction::DbError)?
+            .map_err(error::Auction::BadCommunication)?
             .into_iter()
             .map(|price| {
                 let token = eth::H160(price.token.0).into();
                 let price = big_decimal_to_u256(&price.price)
                     .ok_or(domain::auction::InvalidPrice)
                     .and_then(|p| domain::auction::Price::new(p.into()))
-                    .map_err(error::Auction::Price);
+                    .map_err(|_err| error::Auction::InvalidPrice(token));
```
Mateo-mro, line 216: Maybe worth chaining the error into `InvalidPrice` too.

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -300,10 +283,97 @@ impl Persistence {
...
+
+        let winner = H160(competition.winner.0).into();
+        let score = competition::Score::new(
+            big_decimal_to_u256(&competition.winning_score)
+                .ok_or(error::Solution::InvalidScore(anyhow::anyhow!(
+                    "database score"
+                )))?
+                .into(),
+        )
+        .map_err(|_| error::Solution::InvalidScore(anyhow::anyhow!("zero score")))?;
```
Mateo-mro, line 326: maybe we can have the error also alongisde `zero score`. All errors implement `ToString` trait.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: This was already suggested by Felix to be not found. I think this should never be hit and I even expect this code to be removed once we save winning solutions to separate db table.

Mateo-mro (approved):

crates/autopilot/src/on_settlement_event_updater.rs
sunce86: removed public function `score` in the meantime. I reverted the `score < promised.solution` to return error, although i am sceptic about it, as it won't catch potential bugs in score calculation when score is much higher than promised.solution -> so I added temporary log to observe this.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: renamed

crates/autopilot/src/domain/settlement/mod.rs
sunce86: renamed

crates/autopilot/src/domain/settlement/mod.rs
sunce86: I wanted to differentiate between errors from infra versus errors from calculations done inside Settlement constructor.

Let's re-evaluate if this is needed in the 4/4 PR.

crates/autopilot/src/domain/settlement/mod.rs
sunce86: added auction_id = ?solution.auction_id()

crates/autopilot/src/domain/settlement/mod.rs
sunce86: obsolete code

crates/autopilot/src/domain/eth/mod.rs
sunce86: Interesting. `Display` was added because token addresses were printed as part of the `InvalidPrice` errors.

If so, I removed the `Display` and printed the debug version which should be ok.

crates/autopilot/src/infra/persistence/mod.rs
sunce86: added

crates/autopilot/src/infra/persistence/mod.rs
sunce86: Hm, not sure worth it, i think printing token should be enough to reconstruct the problem.

PR #2865, opened 2024-08-07
[EASY] Automatic checks of new autopilot domain code

sunce86:
Dependent on #2866 

# Description
Instead of logging settlement object, it automatically compares old and new implementation and prints if it detects a difference of behaviour.

Note that code quality is not of paramount importance here as the code will be removed once testing and done and https://github.com/cowprotocol/services/pull/2861 lands.


crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -172,30 +172,90 @@ impl Inner {
...
                 &self.persistence,
             )
             .await;
 
-            tracing::info!(
-                "settlement object {:?}",
-                settlement.map(|settlement| (settlement.observation()))
-            );
+            // automatic checks vs current implementation
+            match (settlement, auction_data) {
```
Mateo-mro, line 190: nit: we could abstract this to its own function

Mateo-mro (approved):

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -172,30 +172,90 @@ impl Inner {
...
+                                    ?observation_fee,
+                                    ?fee,
+                                    "automatic check error: order_fees value mismatch"
+                                );
+                            }
+                        }
+                    }
+                }
+                (Err(_), None) => {
+                    // todo check auction_ids are equal
```
sunce86, line 248: Waits for https://github.com/cowprotocol/services/pull/2866 to be merged so i can fetch auction_id from Settlement error

crates/autopilot/src/on_settlement_event_updater.rs
```diff
@@ -172,30 +172,90 @@ impl Inner {
...
+                            ?auction_id,
+                            ?observation.surplus,
+                            ?auction_data.surplus,
+                            "automatic check error: surplus mismatch"
+                        );
+                    }
+                    if observation.fee.0 != auction_data.fee {
+                        tracing::warn!(
+                            ?auction_id,
+                            ?observation.fee,
```
squadgazzz, line 209: nit: I belive we don't use dot-separated log fields anywhere. Also, in different parts of the code the same variable is called differently. For example, this specific is logged both as `observation.fee` and `observation_fee`. Another example: `auction_data.fee` and `fee`. Maybe it makes sense to use underscore everywhere.

```suggestion
                            observation_fee = ?observation.fee,
```

squadgazzz (approved):

PR #2866, opened 2024-08-07
Attach auction id to settlement error

sunce86:
# Description
Related to comment https://github.com/cowprotocol/services/pull/2861#discussion_r1705250244

Attaches `auction_id` to all errors that happen only once the `auction_id` is successfully recovered from calldata. This enables callers of the `Settlement::new` to fetch `auction_id` if it exists. 

# Changes

- [ ] Added auction id to `SettlementError`
- [ ] Dependent error types refactored to fit nicely into `Settlement` needs


crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -106,14 +113,36 @@ impl Settlement {
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("calldata is not a settlement transaction")]
+    NotSettlement,
+    #[error("auction id not attached to the calldata")]
+    MissingAuctionId,
+    #[error("error {0} for auction {1}")]
+    Error(SettlementError, domain::auction::Id),
```
Mateo-mro, line 121: `Error` is too generic :/ and it does `Error::Error::...`, maybe `Auction` so it does `Error::Auction(..)`? :thinking:

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -106,14 +113,36 @@ impl Settlement {
...
     #[error("failed communication with the database: {0}")]
     Infra(sqlx::Error),
     #[error("failed to prepare the data fetched from database for domain: {0}")]
     InconsistentData(InconsistentData),
     #[error("settlement refers to an auction from a different environment")]
     WrongEnvironment,
-    #[error(transparent)]
-    BuildingSolution(#[from] solution::Error),
+    #[error("failed to decode settlement: {0}")]
+    Decoding(solution::error::Decoding),
```
Mateo-mro, line 145:
```suggestion
    Decoding(#[source] solution::error::Decoding),
```

crates/autopilot/src/domain/settlement/solution/mod.rs
```diff
@@ -156,10 +157,22 @@ pub mod error {
...
+        #[error("decoding error {0} for auction {1}")]
+        Decoding(#[source] Decoding, auction::Id),
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Decoding {
+        #[error("calldata not in an expected abi format: {0}")]
+        Ethabi(web3::ethabi::Error),
+        #[error("unable to tokenize calldata into expected format: {0}")]
+        Tokenizing(ethcontract::tokens::Error),
```
Mateo-mro, line 173:
```suggestion
        Tokenizing(#[source] ethcontract::tokens::Error),
```

crates/autopilot/src/domain/settlement/solution/mod.rs
```diff
@@ -156,10 +157,22 @@ pub mod error {
...
+        #[error("no auction id found in calldata")]
+        MissingAuctionId,
+        #[error("decoding error {0} for auction {1}")]
+        Decoding(#[source] Decoding, auction::Id),
+    }
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum Decoding {
+        #[error("calldata not in an expected abi format: {0}")]
+        Ethabi(web3::ethabi::Error),
```
Mateo-mro, line 171:
```suggestion
        Ethabi(#[source] web3::ethabi::Error),
```

Mateo-mro (approved):

crates/autopilot/src/domain/settlement/mod.rs
fleupold: The settlement error struct nightmare continues 🙈 

Does our domain really care about the two newly introduced subtypes (NotSettlement, MissingAuctionId)? How would `NotSettlement` be treated any differently than `MissingAuctionId`?

I'm also worried about callsites going to match on `settlement::Error::Error(SettlementError::...)`

What do you think about making the top level error type a struct instead of an enum with a private field `Inner` (which is the current error type) and an optional auction ID (with a public accessor). This struct can also implement `Display` (which is all we need for error handling atm) and will in the future allow us to add more logic (e.g. `should_trigger_circuit_breaker`) directly to this monstrous error type.

sunce86: Does our domain really care about the two newly introduced subtypes (NotSettlement, MissingAuctionId)? How would NotSettlement be treated any differently than MissingAuctionId?

Circuit breaker would want to handle these differently.
`NotSettlement` is something happens happens occasionally, because apparently there are transactions where Settlement event is emitted while not calling /settle function: https://etherscan.io/tx/0x98e6f33816daee7562ededa9c8d7390c26322db9d3f7302ad19db9c68131b566

`MissingAuctionId` is a slashable event from a circuit breaker POW.

Will try to refactor the Error struct.

fleupold: In the tx you shared, settle was invoked (just via an internal call from a Safe). In fact the Settlement event is only emitted in settle or the deprecated swap method, so `NotSettlement` is indistinguishable from `MissingAuctionId` imo (in the code, we are checking the selector, which is actually problematic once we start using the settlement facade where `tx.to` won't be 0x9008d19f58aabd9ed0d60971565aa8510560ab41 but some intermediate contract -  but we can cross that when we get there).

image

sunce86: So, these transactions actually need to be skipped, right?

I've refactored the PR, seems slightly easier for eyes now 😄

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -104,6 +107,13 @@ impl Settlement {
     }
 }
 
+#[derive(Debug)]
+pub struct ErrorWithAuction {
+    #[allow(dead_code)]
+    pub inner: Error,
```
fleupold, line 113: why does inner need to be public? We could keep the struct private and thus be more free with the naming (pub Error, private Inner).

fleupold (approved):
Looks ok, but I think it's unnecessary to expose inner directly (we use it for printing, so we may as well implement Display on the top level type) and not exposing everything we have will serve us later when we need to implement logic for the specific error.

squadgazzz (approved):
LG

crates/autopilot/src/domain/settlement/mod.rs

PR #2867, opened 2024-08-07
Native prices: Implement batching 1/2

Mateo-mro:
# Description
Implements a batching system with a quick cache. 
The implementation has a blocking function `blocking_estimate_prices_and_update_cache()` that belongs to a struct which runs a thread. The thread runs every X ms and will do a bulk fetch to whatever provider implements the trait `NativePriceBatchFetcher` for all the requested non-cached tokens in `blocking_estimate_prices_and_update_cache()`.
If the token is cached, it is not requested. The function `blocking_estimate_prices_and_update_cache()` has a timeout of three times the period of the batching requests (to respect Nyquist-Shannon theorem of sampling speed + the batch request has a limit of tokens per batch), which guarantees (in majority of the cases) that the token was fetched in one of the X ms rounds. 
If the token prices doesn't exist, the timeout indicates No Liquidity. 

The batching system as it is right now CANNOT be activated, it needs the upcoming PR (2/2) in which it will be implemented:
- CoinGecko implementing the trait `NativePriceBatchFetcher`
- Adding a configuration to enable the batching system and setting up the estimator + period time

# Changes
- Define trait `NativePriceBatchFetcher` , which is to be implemented by CoinGecko in the PR part 2/2
- Define the `Batching` system which does batching requests every X ms (if needed)

## How to test
1. Unit test

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -1,14 +1,20 @@
 use {
     super::PriceEstimationError,
     crate::price_estimation::native::{NativePriceEstimateResult, NativePriceEstimating},
+    anyhow::anyhow,
+    async_trait::async_trait,
     futures::{FutureExt, StreamExt},
     indexmap::IndexSet,
     primitive_types::H160,
     prometheus::{IntCounter, IntCounterVec, IntGauge},
     std::{
-        collections::{hash_map::Entry, HashMap},
-        sync::{Arc, Mutex, MutexGuard, Weak},
+        collections::{hash_map::Entry, HashMap, HashSet},
+        sync::{Arc, Mutex, Weak},
         time::{Duration, Instant},
+    },
+    tokio::{
+        sync::{mpsc, Mutex as TokioMutex, MutexGuard},
+        time::{interval, sleep},
     },
     tracing::Instrument,
 };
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -26,6 +32,134 @@ struct Metrics {
     native_price_cache_outdated_entries: IntGauge,
 }
 
+/// Trait for fetching a batch of native price estimates.
+#[mockall::automock]
+#[async_trait]
+pub trait NativePriceBatchFetcher: Sync + Send {
```
Mateo-mro, line 38: This trait will be implemented by any provider which offer batching requests. In the upcoming PR it will be implemented by `CoinGecko`.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -26,6 +32,134 @@ struct Metrics {
...
+    /// Maximum batch size the `fetch_native_prices()` can take
+    fn max_batch_size(&self) -> usize;
```
Mateo-mro, line 40: Since each provider may have different maximum batch, this must be part of the trait.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -99,6 +234,25 @@ impl Inner {
         }
     }
 
+    async fn blocking_estimate_prices_and_update_cache(
+        &self,
+        token: &H160,
+        max_age: Duration,
+    ) -> Result<Option<f64>, PriceEstimationError> {
+        if let Some(batching) = self.batching.as_ref() {
```
Mateo-mro, line 242: The batching should be an optional feature, we don't necessarily want it for the autopilot, but we do for the orderbook. Nevertheless, once we move everything to its own service this will be heavily improved.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -315,11 +473,9 @@ impl NativePriceEstimating for CachingNativePriceEstimator {
             }
 
             self.0
-                .estimate_prices_and_update_cache(&[token], self.0.max_age, 1)
-                .next()
-                .await
-                .unwrap()
-                .1
+                .blocking_estimate_prices_and_update_cache(&token, self.0.max_age)
```
Mateo-mro, line 476: Here was the evil part. This was doing a direct call to the estimator. Now it does a blocking operation.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -241,12 +395,16 @@ impl CachingNativePriceEstimator {
...
     ) -> Self {
+        let cache: Arc<TokioMutex<HashMap<H160, CachedResult>>> = Default::default();
         let inner = Arc::new(Inner {
             estimator,
-            cache: Default::default(),
+            cache: cache.clone(),
             high_priority: Default::default(),
             max_age,
+            batching: fetcher
+                .map(|fetcher| Batching::new(cache.clone(), Duration::from_millis(50), fetcher)),
```
Mateo-mro, line 407: The configuration of 50 ms is hardcoded, but it will be changed in the upcoming PR (2/2). It has NO effect, because in order to enable this, `fetcher` has to be `Some`, which is NEVER the case.

fleupold (commented):
I find it weird that we are coupling the batching logic with the caching native price estimator. I think the two concepts are completely orthogonal and are adding too much complexity to the native price cache itself.

I was envisioning two separate PRs:
1. Adjust the Coingecko version in a way that it doesn't immediately sends an HTTP request for every `estimate_native_price` call it receives, but instead adds it to a queue which periodically gets flushed (like every 100ms) and bundles all requests together. Similar how it's done for the [BufferedTransport](https://github.com/cowprotocol/services/blob/578a209cc01f4ad1ac2f0906dfc17736a28b8629/crates/ethrpc/src/buffered.rs#L1). This will dramatically reduce the credits we are using from Coingecko's API.
2. Separately change the interface of the native price cache to become async. However, I'd let the callsite specify a deadline (how much time it should maximally spend fetching requests that are not cached before returning NotFound for those tokens) and simply use `futures::select` with a timeout inside the cache to give misses a short chance to hit (from second layer caches or very fast APIs).

What is the draw back of keeping these concerns separate?

Mateo-mro:
> I find it weird that we are coupling the batching logic with the caching native price estimator. I think the two concepts are completely orthogonal and are adding too much complexity to the native price cache itself.
> 
> I was envisioning two separate PRs:
> 
> 1. Adjust the Coingecko version in a way that it doesn't immediately sends an HTTP request for every `estimate_native_price` call it receives, but instead adds it to a queue which periodically gets flushed (like every 100ms) and bundles all requests together. Similar how it's done for the [BufferedTransport](https://github.com/cowprotocol/services/blob/578a209cc01f4ad1ac2f0906dfc17736a28b8629/crates/ethrpc/src/buffered.rs#L1). This will dramatically reduce the credits we are using from Coingecko's API.
> 2. Separately change the interface of the native price cache to become async. However, I'd let the callsite specify a deadline (how much time it should maximally spend fetching requests that are not cached before returning NotFound for those tokens) and simply use `futures::select` with a timeout inside the cache to give misses a short chance to hit (from second layer caches or very fast APIs).
> 
> What is the draw back of keeping these concerns separate?

@fleupold thank you very much for your quick feedback!

1. I understand your point, fair enough, however, few things:

> bundles all requests together

I don't recommend doing that, because then even though we gather the requests in pack of 20, the user still has somehow controls on how many requests we send to CoinGecko. My idea was to make the period as a hard period, so the rate of sending requests to CoinGecko can never be more than a determined value. We can fire periodic bath requests with a period matching the rate limit, so we couldn't go faster anyway, and still we have full control of it.

What's more, if we do as you suggested, we are still leaving naked the rest of the API calls (solvers + 1inch + 0x), here is the problematic line https://github.com/cowprotocol/services/pull/2867#discussion_r1707176650. My plan was to mitigate not only CoinGecko but any API call to the estimators.

On the other hand, I do agree this is adding a big extra layer of complexity to the native price cache. My idea with this, was to mitigate fully the issue, and then, once we are ready to move this to its own service, refactor it so it is not so complex, and make the whole native price fetcher simpler.

> What is the draw back of keeping these concerns separate?

If we follow your suggestion, I would move the `Batching` struct into the `CoinGecko.rs` file (alongside the tests), remove the trait and use directly the batch request. So the effort to split this into 2 PRs is little to be honest, since I am very familiar with this part of the codebase.

Mateo-mro:
Closing because of this: https://github.com/cowprotocol/services/pull/2869

PR #2868, opened 2024-08-08
Tokio console and log filter reloading

MartinquaXD:
# Description
https://github.com/cowprotocol/services/pull/2815 reverted 2 changes (https://github.com/cowprotocol/services/pull/2791, https://github.com/cowprotocol/services/pull/2792) because we again saw issues with logs that were suddenly no longer emitted.
This problem also happened the last time we merged support for `tokio-console` but I hoped that:
1. the issue was resolved in one of the related dependencies somehow
2. making the feature opt-in with a CLI flag would allow normal operations when disabled

However, even with the feature disabled we saw that logs were again missing. I tried to debug the underlying issue and it appears that for some reason a log like `::tracing::event!(target: "sqlx::query", tracing::Level::DEBUG, summary);` would not get logged with the original implementation (not a huge deal since we weren't using them) but these logs would also cause the NEXT log that would normally be emitted to be skipped.
This can be seen in this [code](https://github.com/cowprotocol/services/commit/5ba466668781d60bdf13531187f6baa1eac9fd89) which reproduced the error.

Unfortunately `sqlx` is using `tracing::event!()` when executing an SQL query which means with the described problem it could happen randomly that some logs are missing. Luckily there are also scenarios where some SQL query always gets executed before a log we issue ourselves (e.g. `saving fee policies` log in the autopilot run loop would always be skipped) which made it easier to find the underlying cause.

# Changes
1. Add `tokio-console` and `filter reloading` feature again
2. Fiddled around with how the tracing subscriber gets initialized until the test that reproduced the issue no longer skipped any logs. This was not straight forward since initializing the tokio console and our regular log filters requires you to jump through a few type system hoops.
3. removed CLI flag since this did not have the originally intended effect anyway

# How to review
The only important change happend in https://github.com/cowprotocol/services/pull/2868/commits/41c47ae6da040952ddb4d50ea03ba0f94636e35c. The other stuff is just the revert commit.

## How to test
Run test that reproduced the issue and check that the targeted log does not get skipped.
This doesn't necessarily mean that all is well now. To get more confidence on the change I would just run the CI and compare the number of logs with CI runs that don't change the tracing setup. If both have roughly the same number of logs (I assume some variance will always be there) the change should be good.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1331,6 +1331,44 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0b6a852b24ab71dffc585bcb46eaf7959d175cb865a7152e35b348d1b2960422"
 
 [[package]]
+name = "console-api"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a257c22cd7e487dd4a13d413beabc512c5052f0bc048db0da6a84c3d8a6142fd"
+dependencies = [
+ "futures-core",
+ "prost",
+ "prost-types",
+ "tonic",
+ "tracing-core",
+]
+
+[[package]]
+name = "console-subscriber"
+version = "0.3.0"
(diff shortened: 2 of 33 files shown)


fleupold (approved):
Hope you had a great vacation! Didn't think you would be spending time in Tokio though ;-)

Mateo-mro (approved):

squadgazzz (approved):
🤞

PR #2869, opened 2024-08-08
Native prices: Implement generic buffered batching system 1/3

Mateo-mro:
# Description
Implements a generic batching system for the native prices.

The implementation has a blocking function `blocking_buffered_estimate_prices()` that belongs to a struct which runs a thread. The thread gathers the requests either by time (it waits X time to gather request and then fires them) or by number of requests (as soon ass it reaches a number of requests, fires them).

This generic implementation allow for any native price fetcher to batch buffer the request as long as they implement the trait `NativePriceBatchFetcher`.

This code is heavily based on the `ethrpc` crate code.

In the upcoming PR:
- the configuration will be propagated to the services configuration
- implement the trait `NativePriceBatchFetcher` for CoinGecko

# Changes
- Generic buffered batching system for native prices

## How to test
1. Unit test


crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// soon as they are sent, so this limit should be enough to hold the
+    /// flow
+    const BROADCAST_CHANNEL_CAPACITY: usize = 50;
+
+    /// Creates a new buffered transport with the specified configuration.
+    pub fn with_config(inner: Inner, config: Configuration) -> Self {
+        let inner = Arc::new(inner);
+        let (calls, receiver) = mpsc::unbounded();
+
+        let (broadcast_sender, _) = broadcast::channel(Self::BROADCAST_CHANNEL_CAPACITY);
```
Mateo-mro, line 83: We do not need the broadcast receiver here as it doesn't implement `Clone` and we can get many receivers by just using `tx.subscribe()`.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,459 @@
...
+                        Err(err) => {
+                            tracing::error!(?err, "failed to send native price batch request");
+                            batch
+                                .into_iter()
+                                .map(|token| (token, Err(err.clone())))
+                                .collect::<HashMap<_, _>>()
+                        }
+                    };
+                    for result in results {
+                        let _ = broadcast_sender.send(result);
```
Mateo-mro, line 127: We broadcast the results to all the subscribes, aka, everyone who is waiting for an result.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,459 @@
...
+        // Sends the token for requesting price
+        self.calls.unbounded_send(*token).map_err(|e| {
+            PriceEstimationError::ProtocolInternal(anyhow!(
+                "failed to append a new token to the queue: {e:?}"
+            ))
+        })?;
+
+        let mut rx = self.broadcast_sender.subscribe();
+
+        tokio::time::timeout(self.config.batch_delay.saturating_mul(3), async {
```
Mateo-mro, line 148: We subscribe to the broadcast and wait for our result 3 times the `batch_delay`. I thought 3 is a compromise. I don't want to set too short time, otherwise we may lose valid results, but I don't want to put too high either, because in case of no valid result, not to be blocked for so much time.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,459 @@
...
+            let broadcast_sender = broadcast_sender.clone();
+            async move {
+                let batch = batch.into_iter().collect::<HashSet<_>>();
+                if batch.len() != 0 {
+                    let results = match inner.fetch_native_prices(&batch).await {
+                        Ok(results) => results
+                            .into_iter()
+                            .map(|(token, price)| (token, Ok::<_, PriceEstimationError>(price)))
+                            .collect::<HashMap<_, _>>(),
+                        Err(err) => {
```
Mateo-mro, line 118: If we receive an error from the batching call, them all the responses for each token of the batch will have such error.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,450 @@
...
+                    for result in results {
+                        let _ = broadcast_sender.send(result);
+                    }
+                }
+            }
+        }))
+    }
+
+    /// Blocking operation to get estimate prices in a batch
+    pub async fn blocking_buffered_estimate_prices(
```
squadgazzz, line 131: Which specific operation is considered blocking here? Since Tokio uses asynchronous operations (async/await), it should not block the thread, allowing the scheduler to utilize the thread for other tasks while awaiting. Could you clarify this?

Mateo-mro: @squadgazzz that is a really good question. It is not blocking in the sense of thread blocking. I called it like that because it is a "fire and wait" type of request. But you are right, the name is quite confusing, I'll change it! :ok_hand:

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,450 @@
...
+#[derive(Clone)]
+pub struct BufferedTransport<Inner> {
+    config: Configuration,
+    #[allow(dead_code)]
+    inner: Arc<Inner>,
+    calls: mpsc::UnboundedSender<H160>,
+    broadcast_sender: broadcast::Sender<NativePriceResult>,
+}
+
+type NativePriceResult = (H160, Result<f64, PriceEstimationError>);
```
squadgazzz, line 63: nit: Would it be more readable to use a struct with named fields?

crates/shared/src/price_estimation/buffered.rs
squadgazzz: Does it make sense to extract this timeout param from the config?

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,529 @@
...
+
+    // Function to receive with a timeout
+    async fn receive_with_timeout(
+        rx: &mut broadcast::Receiver<NativePriceResult>,
+        token: &H160,
+        timeout_duration: Duration,
+    ) -> Result<Option<NativePriceResult>, Elapsed> {
+        tokio::time::timeout(timeout_duration, async {
+            match rx.recv().await {
+                Ok(value) => (value.token == *token).then_some(value),
```
MartinquaXD, line 187: I believe this is the part that trips be up about the broadcast channel.
Presumably there are multiple listeners on the same channel interested in different responses.
I think what makes more sense to me is if there is a new oneshot channel for every caller that tries to request a token price. Then the batching mechanism would aggregate a map `Token => Vec<oneshot_channel>` and when the background task gets the prices in a batch it knows exactly who is actually awaiting this price.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub batch_delay: Duration,
+    /// The timeout to wait for the result to be ready
+    pub result_ready_timeout: Duration,
+}
+
+/// Trait for fetching a batch of native price estimates.
+#[allow(dead_code)]
+#[mockall::automock]
+#[async_trait]
+pub trait NativePriceBatchFetcher: Sync + Send {
```
MartinquaXD, line 46: nit: This trait could extend the single request version of the trait since it's safe to assume that all APIs can also implement the single request version with a batch size of 1.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,529 @@
...
+///
+/// This is very similar to `futures::stream::StreamExt::ready_chunks` with the
+/// difference that it allows configuring a minimum delay for a batch, so
+/// waiting for a small amount of time to allow the stream to produce additional
+/// items, thus decreasing the chance of batches of size 1.
+fn batched_for_each<T, St, F, Fut>(
+    config: Configuration,
+    items: St,
+    work: F,
+) -> impl Future<Output = ()>
```
squadgazzz, line 205: Not related to this PR. Some of the code is copied from the `ethrpc` crate. Any plan to generalize it even further to avoid code duplications?

squadgazzz (approved):
LGTM.

MartinquaXD (commented):
Mostly concerned about the broadcast channel so far.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: @MartinquaXD Well, the receiver depends on the perspective. So, we have:

- `BufferedTransport` which receives token requests (via `mpsc` channel)
- let's call it `client`, the entity/piece of code requesting the native price

So, there is only one `BufferedTransport` receiving token price requests, this entity batches the petitions the same way we do in `ethrpc` (either by time limit or by size limit, depending which it reached sooner).

This communication is done with the `mpsc` channel.

But, the `client` still needs to get the price back. The most efficient way (memory and performance) is to do a broadcast channel. If you are the client who just requested a price, then you subscribe to the broadcast waiting for your price to come (so, the `client` is the receiver here).

I don't know if I explained myself properly.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,529 @@
...
+            let broadcast_sender = broadcast_sender.clone();
+            async move {
+                let batch = batch.into_iter().collect::<HashSet<_>>();
+                if !batch.is_empty() {
+                    let results: Vec<_> = match inner.fetch_native_prices(&batch).await {
+                        Ok(results) => results
+                            .into_iter()
+                            .map(|(token, price)| NativePriceResult {
+                                token,
+                                result: price,
```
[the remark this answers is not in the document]
Mateo-mro: Well, it used to be like that but I changed it to address a comment, since the type `(H160, PriceEstimateResult)` may seem complex, but I am open to either option.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: @MartinquaXD the timeout cannot be avoided even for the solution you propose. We cannot be waiting forever for the answer and block the petition. The timeout must be applied in any case.

The solution you propose it way more complex, error/bug/memory leak prone. The way of doing it would be to send in the request `mspc` the token alongside the channel to which the thread should send the token. But I don't see how that solution is better than the broadcast :thinking:

Mateo-mro: Allocating oneshot channel per native price request is quite memory wasteful. We need to keep in mind that for the orderbook we are getting quite a lot of native price requests.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: umm, I don't fully get you. Do you mean to extend the trait to implement the request for one token? I don't think we should do it, because this trait is specifically for bulk requests + the requester can just do `&[token]` for individual request.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: My main concern with the broadcast is the processing time within the `match` statement, if it is too slow, we may miss events, but since it is just an `if`, it is impossible.

Mateo-mro: I totally see your point and there are cons and pros for each solution, to be honest. I don't see a clear "winning" solution. I am more incline for the broadcast because it is way less resources intensive than the oneshot solution.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: I mean you could do `pub trait NativePriceBatchFetcher: NativePriceEstimating` since the bulk version is a super set of the single version. Also since these are conceptually the same thing I think it makes sense to have the naming reflect that.

Mateo-mro: Ah, ok, I understand. Oh, sure, definitely!

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: @squadgazzz this is a good question. AFAIK, I was told we don't want to keep adding things to the `shared` crate, and the code is in different crates, so it makes it hard to share. We could have a super generic `buffered` crate, and then use it for both implementations.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: Not sure what the timeouts have to do with my comment. Overall I agree that having timeouts would be good (although the double timeout this PR currently has looks suspicious).

So one concern I see with the broadcast channel is that you have to decide on a size upfront which has to be big enough to not cause issues even when we get spammed. That means it likely needs to be vastly overallocated for 99% of the time which kind of makes the oneshot channel approach more appealing in terms of memory usage IMO.
Also AFAIK every task touching this batching implementation might be woken up for every incoming price of the batch even if it's just interested in a single token.
And lastly as to the point of being error prone: oneshot channels are very simple while broadcast channels have kind of subtle semantics around them. For example where you currently completely discard a price whenever the channel falls behind there is actually no reason to assume that there is no price for the requested token left in the channel. (i.e. [here](https://github.com/cowprotocol/services/pull/2869/files#diff-2e490a59194917c3d8edfe352221e75a0e0a05312c50f11ec27b635cf3142334R188) you could `try_recv` the stream to completion hoping to get lucky that the token price is still in the stream).

Mateo-mro: @MartinquaXD that is a very valid point. I myself don't know what's the best answer for this. Let's try to figure it out together.

> So one concern I see with the broadcast channel is that you have to decide on a size upfront which has to be big enough to not cause issues even when we get spammed. 

We cannot get **that** spammed anymore, but I agree. I believe the broadcast items have a very short time to live. So in theory they won't be in the channel for much time. I set it to 200 elements which don't seem to me like a huge number in terms of memory.

> That means it likely needs to be vastly overallocated for 99% of the time which kind of makes the oneshot channel approach more appealing in terms of memory usage IMO.

Umm, I am not so sure about it. The batch size for CoinGecko is 20, let's say we get 1000 request. so we cannot keep more than 20 elements in the broadcast (because the request time to CoinGecko is greater than the time to live of the items), but we definitely can keep more than 20 oneshots, because we keeps gathering and sending the requesters. And oneshop entity/object is more memory intensive than a pair `(H160, price)`.

> Also AFAIK every task touching this batching implementation might be woken up for every incoming price of the batch even if it's just interested in a single token.

This is totally true. This is a big con of the broadcast.

> And lastly as to the point of being error prone: oneshot channels are very simple while broadcast channels have kind of subtle semantics around them. For example where you currently completely discard a price whenever the channel falls behind there is actually no reason to assume that there is no price for the requested token left in the channel.

Yes, 100 % true, the issue with broadcast is that it could happen that sometimes we miss the price. This could happen in the loop in which we fire the broadcast with all the results. It could happen that the consumer miss a result while checking if the previous one was for its token.

MartinquaXD: So far only the memory leak argument is very convincing to me. Overall I think the `RequestSharing` struct would be a perfect fit here but unfortunately this is already involved in a memory leak investigation.
If that would not have been the case I would suggest a setup where the buffering struct holds a `RequestSharing` internally which maps `Token => oneshot_channel`. Then the background task would always have to notify exactly a single channel since the requests already get deduplicated by the `RequestSharing` struct.
But until we figure out how we still leak memory in the API I would hesitant to use that component even more. 🤔

MartinquaXD: > The batch size for CoinGecko is 20, let's say we get 1000 request. so we cannot keep more than 20 elements in the broadcast (because the request time to CoinGecko is greater than the time to live of the items), but we definitely can keep more than 20 oneshots, because we keeps gathering and sending the requesters.

The thing I don't like about the broadcast channel is that we can't make it bulletproof. In your example it's definitely not enough to have a channel size of 20 because the behavior of the solution very much depends on how quickly the consumers read the channel. If there is 0 scheduling overhead and tasks would get woken up immediately 20 slots would be fine. But let's say the runtime let's the background task run which produces more prices but for some reason doesn't schedule the consumer tasks to run. In that case the consumers might not get any prices although the background task successfully fetched all of them.
Overall the broadcast channels feel like a heuristic while oneshot channels could give us hard guarantees. And with the idea discussed above we would have extremely cheap deduplication of requests (basically the cost of an `Arc` per duplicated request).

But again given that the memory leak involving the `RequestSharing` is still unclear it indeed seems more pragmatic to go with the broadcast solution for now. Also one thing that should indicate that a relatively small buffer size is okay is that a single coingecko request is orders of magnitude slower than waking up a task for a new price meaning the producer is way slower than any of the consumers.

Mateo-mro: Yeah, I fully agree with you! one thing

> [..] internally which maps Token => oneshot_channel

This was another reason to do the broadcast. We cannot do that mapping, we would need to do a `request_id` (we can generate a new `UUID` for it`) => oneshot_channel. Because of this scenario:

- User number 1 request price for DOGE
- User number 2 request price for DOGE at the same time (difference of ~100 ms)
- Request sent to the buffered CoinGecko with the same token for two different receivers, one of them will never receive the price as it would be squashed in the map.

I know this is a **very** remote case. But it can definitely happen too.

MartinquaXD: > one of them will never receive the price as it would be squashed in the map.

Which is why I suggested a map like `Token => Vec<oneshot_channel>`. But doesn't matter anymore anyway.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,529 @@
...
+    // Function to receive with a timeout
+    async fn receive_with_timeout(
+        rx: &mut broadcast::Receiver<NativePriceResult>,
+        token: &H160,
+        timeout_duration: Duration,
+    ) -> Result<Option<NativePriceResult>, Elapsed> {
+        tokio::time::timeout(timeout_duration, async {
+            match rx.recv().await {
+                Ok(value) => (value.token == *token).then_some(value),
+                Err(_) => None,
```
MartinquaXD, line 188: Just to pin point one of the things brought up during the discussion above.
This case only means the receiver missed at least 1 one item in the stream. It can very well be that the item it is interested in is still in the stream.
So I suggest in the `None` case to exhaust the channel with a `try_recv` loop hoping to still find the desired item.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: > The thing I don't like about the broadcast channel is that we can't make it bulletproof.

100 % true. I couldn't agree more.

> In your example it's definitely not enough to have a channel size of 20 because the behavior of the solution very much depends on how quickly the consumers read the channel.

Yes, but the client's (token requester) in theory, should be faster. Because the time to get a new bulk of 20 is the request to CoinGecko's server + [this PR](https://github.com/cowprotocol/services/pull/2876) processing the response). The logic in our side is just a loop and a token comparison. If our CPU is overloaded and we are slow processing, we would also be slow processing the CoinGecko's response prior to sending it to the broadcast. But I fully agree, that the scenario may be unlikely, but as you said, it is not bulletproof and it can definitely happen!

Either way, we should closely follow the broadcast performance and see how many prices we really miss.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: @MartinquaXD actually, there is an issue and it is that `NativePriceEstimating` is not async, but:

```rust
    fn estimate_native_price(
        &self,
        token: H160,
    ) -> futures::future::BoxFuture<'_, NativePriceEstimateResult>;
```
    
is there a reason for it not to be directly `async`? if I can make it `async` it'd be the best, but maybe there is some reason I a missing :thinking:

MartinquaXD: the `async` keyword only says the function returns `impl Future`. I assume it was done like this here because of lifetime annotation limitations of the `async_trait` macro but conceptually this is equivalent.

Mateo-mro: Yup, that's why I was wondering maybe we wanted to keep it in the "not syntactically sugar" version for a reason :thinking: but changing it! :ok_hand:

crates/shared/src/price_estimation/buffered.rs
fleupold: This magic number is weird. I think this method should just take as long as it does and if someone wants to timeout earlier they should combine it with a timeout future.

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: > Which is why I suggested a map like Token => Vec<oneshot_channel>. But doesn't matter anymore anyway.

Oh yeah, missed that detail! you are right!! :ok_hand:

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: Good point! But the requested value may not necessarily be ready yet, so if we do `try_recv()` may yields only errors until the token price is generated and sent to the broadcast. As it is right now, if it is lagged behind, it just keeps consuming elements until the desired one is found or generated, The only thing is that the operation is `async`.

Mateo-mro: On the other hand the `RecvError::Closed` could be better handled :thinking:

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,450 @@
...
+type NativePriceResult = (H160, Result<f64, PriceEstimationError>);
+
+impl<Inner> BufferedTransport<Inner>
+where
+    Inner: NativePriceBatchFetcher + Send + Sync + 'static,
+{
+    /// Maximum capacity of the broadcast channel, the messages are discarded as
+    /// soon as they are sent, so this limit should be enough to hold the
+    /// flow
+    const BROADCAST_CHANNEL_CAPACITY: usize = 50;
```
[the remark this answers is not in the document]
Mateo-mro: yup, good point, making it configurable

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -516,24 +511,43 @@ mod tests {
...
-            .times(BATCH_SIZE)
-            .returning(|_| async { Ok(1.0) }.boxed());
-        // background task updates all outdated prices
-        inner
-            .expect_estimate_native_price()
-            .times(BATCH_SIZE)
-            .returning(move |_| {
-                async move {
-                    tokio::time::sleep(tokio::time::Duration::from_millis(WAIT_TIME_MS)).await;
+        struct OwnMockMockNativePriceEstimating {
```
Mateo-mro, line 514: Please, forgive me for this :pray: but there was no way of returning it properly as `mockall` doesn't support `returning` with async inside :disappointed:

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: Oh, right. Overlooked that you are already looping somewhere else. In that case I think this is fine.

crates/shared/src/price_estimation/native_price_cache.rs
Mateo-mro: Nothing happened here :eyes:

crates/shared/src/price_estimation/competition/native.rs
```diff
@@ -1,17 +1,17 @@
...
     primitive_types::H160,
     std::{cmp::Ordering, sync::Arc},
 };
 
 impl NativePriceEstimating for CompetitionEstimator<Arc<dyn NativePriceEstimating>> {
-    fn estimate_native_price(
-        &self,
-        token: H160,
-    ) -> BoxFuture<'_, Result<f64, PriceEstimationError>> {
+    fn estimate_native_price(&self, token: H160) -> BoxFuture<'_, NativePriceEstimateResult> {
```
Mateo-mro, line 14: Changed this type for consistency.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,557 @@
...
+        // Sends the token for requesting price
+        self.calls.unbounded_send(token).map_err(|e| {
+            PriceEstimationError::ProtocolInternal(anyhow!(
+                "failed to append a new token to the queue: {e:?}"
+            ))
+        })?;
+
+        let mut rx = self.broadcast_sender.subscribe();
+
+        tokio::time::timeout(self.config.result_ready_timeout, async {
```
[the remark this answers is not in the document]
Mateo-mro: @fleupold  Good question, because I wanted to code it as defensively as possible. So there are 2 timeouts:

- This timeout you are pointing out: It is the time we wait until we receive the **correct** token response
- `revc().await` timeout, it is the timeout we wait to receive **any** response. This is the "unnecessary" timeout, but I wanted to add it, in case that the emitter for whatever reason is not emitting results, so we are not waiting there forever.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,557 @@
...
+    /// It returns a HashMap which maps the token with its native price
+    /// estimator result
+    async fn fetch_native_prices(
+        &self,
+        tokens: &HashSet<H160>,
+    ) -> Result<HashMap<H160, NativePriceEstimateResult>, PriceEstimationError>;
+}
+
+#[async_trait]
+impl NativePriceEstimating for MockNativePriceBatchFetching {
```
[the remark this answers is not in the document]
Mateo-mro: Yeah, ideally. But `mockall` does create code not only for test, but in general :/

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,576 @@
...
+        };
+
+        let buffered = Arc::new(BufferedRequest::with_config(
+            native_price_batch_fetcher,
+            config,
+        ));
+
+        check_batching_many(buffered.clone(), tokens_requested).await;
+
+        sleep(Duration::from_millis(200)).await;
```
MartinquaXD, line 572: FYI often times you can already get reliable tests with 10ms sleeps because there is basically nothing going on in the runtime during a test. Overall we should try to keep tests running quickly because even a bunch of 200ms sleeps can add up over multiple tests and test runs.

crates/shared/src/price_estimation/buffered.rs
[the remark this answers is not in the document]
MartinquaXD: Huh, is this only necessary now that the new trait extends the original native price trait and mockall can't handle that well?
Seems interesting and a bit unfortunate. 😅 
Depending on how widely you actually need the mock struct you could also just apply the mock macro for testing purpose with: `#[cfg_attr(test, mockall::automock)]`. That would allow you to move this implementation into the test section.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,576 @@
...
+            })?
+        }
+        .boxed()
+    }
+}
+
+#[allow(dead_code)]
+impl<Inner> BufferedRequest<Inner>
+where
+    Inner: NativePriceBatchFetching + Send + Sync + NativePriceEstimating + 'static,
```
MartinquaXD, line 141:
```suggestion
    Inner: NativePriceBatchFetching + 'static,
```

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: makes sense, thanks! I set it to `20 ms` because the `batch_delay` is `10 ms` and I wanted it to be the double :ok_hand:

crates/shared/src/price_estimation/buffered.rs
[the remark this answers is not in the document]
Mateo-mro: sweet, thank you! we should propagate it to all the mockalls when possible, to avoid extra code in production :thinking:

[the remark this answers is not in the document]
MartinquaXD: I tried that in the past but the issue is that you can't use this trick if you have to export the `Mock` struct outside of the crate which for example happens for other price estimation tests.
Also this code doesn't make a huge difference since it should get stripped out of the binary anyway since it's not used outside of tests.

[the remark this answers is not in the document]
Mateo-mro: makes sense!

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// The maximum amount of concurrent batches to request.
+    ///
+    /// Specifying `None` means no limit on concurrency.
+    pub max_concurrent_requests: Option<NonZeroUsize>,
+    /// The maximum batch size.
+    pub max_batch_len: usize,
+    /// An additional minimum delay to wait for collecting requests.
+    ///
+    /// The delay starts counting after receiving the first request.
+    pub batch_delay: Duration,
```
fleupold, line 39: also called debouncing more generally I belive

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,577 @@
...
+                    "failed to append a new token to the queue: {e:?}"
+                ))
+            })?;
+
+            let mut rx = self.broadcast_sender.subscribe();
+
+            tokio::time::timeout(self.config.result_ready_timeout, async {
+                loop {
+                    if let Ok(Some(result)) =
+                        Self::receive_with_timeout(&mut rx, &token, self.config.batch_delay).await
```
fleupold, line 105: The double timeout seems confusing and wrong to me. I'm ok with a timeout (although I generally think we are making the config overly complicated) but two timeouts don't make sense.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,577 @@
...
+{
+    let concurrency_limit = config.max_concurrent_requests.map(NonZeroUsize::get);
+
+    let batches = stream::unfold(items, move |mut items| async move {
+        let mut chunk = vec![items.next().await?];
+
+        let delay = tokio::time::sleep(config.batch_delay).fuse();
+        futures::pin_mut!(delay);
+
+        while chunk.len() < config.max_batch_len {
```
fleupold, line 234: Comment please on what this is doing.

fleupold (commented):
Starting to look good. I'm still a bit worried about the config being a massive object with lots of choices (can we pick defaults and make it easy to override selectively)?

And confused by the double timeout...

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: I agree that two timeout is a little bit overkill, but I wanted to be as defensive as possible. Changing it to one timeout :ok_hand:

Mateo-mro:
> Starting to look good.

@fleupold starting??? :cry: :sweat_smile: :sweat_smile: :sweat_smile: just kidding :rofl: :rofl:

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: btw @fleupold , removing the double timeout doesn't reduce the configuration fields. But I will add default values :ok_hand:

Mateo-mro: ummm I didn't propagate the config to the autopilot/orderbook config yet (I plant to do that in the 3/3 PR). Maybe there it's where we should have the defaults? (alongside the other default config values) :thinking:

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,564 @@
...
+    pub max_batch_len: usize,
+    /// An additional minimum delay to wait for collecting requests.
+    ///
+    /// The delay to start counting after receiving the first request.
+    pub debouncing_time: Duration,
+    /// The timeout to wait for the result to be ready
+    pub result_ready_timeout: Duration,
+    /// Maximum capacity of the broadcast channel to store the native prices
+    /// results
+    pub broadcast_channel_capacity: usize,
```
fleupold, line 44: Any reason to not use `usize::max` for this one

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,564 @@
...
+
+/// Buffered implementation that implements automatic batching of
+/// native prices requests.
+#[allow(dead_code)]
+#[derive(Clone)]
+pub struct BufferedRequest<Inner> {
+    config: Configuration,
+    inner: Arc<Inner>,
+    calls: mpsc::UnboundedSender<H160>,
+    broadcast_sender: broadcast::Sender<NativePriceResult>,
```
fleupold, line 72: nit
```suggestion
    requests: mpsc::UnboundedSender<H160>,
    results: broadcast::Sender<NativePriceResult>,
```

fleupold (approved):
Looks very good!

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: @fleupold Because if something goes south (one requested, aka, consumer is not reading the broadcast channel), then the elements from the broadcast won't be deleted until hit the limit, and if the limit is the `usize::MAX`, the memory can leak. If this scenario happens, the missing consumer will miss data (if the broadcast limit is 20 for example), but if it is not consuming the data in the first place...

fleupold: Sure, I just don't think we will ever change whatever value we deem appropriate here for a different use case and it's making the configuration object harder to understand. We can keep it but imo it shouldn't be something an engineer should think about when setting up a buffered price estimator (whatever you think is a reasonable value is likely good for most if not all use cases)

fleupold: > Small addition. [...]

Ok wow, even more reason I should never be forced to chose this value myself

MartinquaXD: I'm fine with hardcoding the value eventually but this seems like the sort of thing which might require a bit of time to dial in to not over-allocate memory but also don't miss results unnecessarily. I'd say let's leave it as a config for now and hardcode it once we have found a reasonable lower bound in practice.

MartinquaXD (approved):

crates/shared/src/price_estimation/buffered.rs
Mateo-mro: I fully agree with @MartinquaXD

PR #2870, opened 2024-08-08
Store considered order events label for all considered orders

MartinquaXD:
# Description
While testing the progress bar v2 @alfetopito never saw the `Solved` state (called `considered` internally).
This state indicates that some solver tries to settle your order, it just didn't provide the best overall solution. In terms of the progress bar this means your order will likely soon be settled.

# Changes
It looks like currently we only store the `considered` flag for the winning solution but not ALL solutions. This PR changes that.


sunce86 (approved):
👍

Mateo-mro (approved):
Good catch!

squadgazzz (approved):

alfetopito:
🥳

PR #2871, opened 2024-08-08
Don't send cow amm orders with missing prices

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/pull/2863#issuecomment-2273717345

Do not send cow amm order owner to driver if auction prices don't contain prices for that cow amm traded tokens.


crates/autopilot/src/solvable_orders.rs
```diff
@@ -252,6 +252,12 @@ impl SolvableOrdersCache {
 
         let surplus_capturing_jit_order_owners = cow_amms
             .iter()
+            .filter(|cow_amm| {
```
Mateo-mro, line 255: Shouldn't we filter the orders instead of the CoW AMM owner addresses? :thinking:

sunce86: I believe cow amm order templates are built in the driver not in the autopilot. The only thing we can do in autopilot is to not send owners for which we know prices are missing.

Mateo-mro: But I don't see how not sending the owners may have an impact. Because we are always sending all the surplus capturing jit order owners independently if the auction contains orders or not. Therefore, if you really want to send only the owners who appear in the auction, you should extend this filter not only for filtering missing prices, but also onwers who don't have orders in the auction :thinking:

Mateo-mro (approved):

squadgazzz (approved):

crates/autopilot/src/solvable_orders.rs
fleupold, line 255: A metric for filtered out CoW orders (similar to filter reasons for user orders) would be helful imo

fleupold:
@sunce86 can you check if this PR broke the e2e test? It's failing on all master commits now

image

sunce86:
> @sunce86 can you check if this PR broke the e2e test? It's failing on all master commits now
> 
> image

Will check

PR #2877, opened 2024-08-09
Unify GC task

MartinquaXD:
# Description
While inspecting the tokio runtime with `tokio-console` I noticed that we have a quite a few background tasks cleaning up memory from `RequestSharing` structs. With all those tasks in the list it makes it harder to see what is actually happening.

TBH this mostly triggered my sense of cleanliness so if you feel like the extra indirection is not worth the mental overhead I'm fine with dropping this PR.

# Changes
Introduced a single background task for garbage collection and now all the individual `RequestSharing` structs just "sign up" for GC instead of spawning their own task every time.

## How to test
Added unit test verifying that the garbage collector correctly clears memory from the traced objects and that objects that are no longer in use get dropped from the GC as well.


Mateo-mro:
Why is the reason we need a garbage collector in the first place? I mean, I am not questioning this PR, since this PR is just cleaning and gathering what we have, but why do we need the `request_sharing.rs` garbage collector? what is the problematic we faced in the past?

I have the impression that this is patching a bigger underlying problem with the way some threads are managing their memory. But maybe I am missing some information about this...

MartinquaXD:
The concept of `RequestSharing` is that every request with a given argument will always result in the same future. If now multiple callers request something for the same argument (e.g. native token price) instead of spawning a new task we just await the one that is already inflight and copy the result over.
Because we store these sharable handles in a map we need to have some mechanism to drop the handles once nobody else is using that anymore to not let the map grow indefinitely (or allow reusing ancient requests responses). This is what the GC is taking care of.

Mateo-mro:
> The concept of `RequestSharing` is that every request with a given argument will always result in the same future. If now multiple callers request something for the same argument (e.g. native token price) instead of spawning a new task we just await the one that is already inflight and copy the result over. Because we store these sharable handles in a map we need to have some mechanism to drop the handles once nobody else is using that anymore to not let the map grow indefinitely (or allow reusing ancient requests responses). This is what the GC is taking care of. So this is by no means a general purpose garbage collector like Java has it for example only somewhat specific to this implementation detail.

I understand, clear. Thank you for the explanation! :raised_hands:

Mateo-mro (approved):
LGTM. I really like the unit test added :ok_hand:

squadgazzz (approved):
Even though it took some time to fully understand how does it work, it looks cool 👍

MartinquaXD:
This setup is brittle when using `cargo test`. The problem is that the global GC gets allocated across all the tests but the tokio runtime gets restarted after every test. The GC task spawned during the first test will therefor not run in subsequent tests after the runtime got terminated.
BTW this is not an issue when using `cargo nextest` since this seems to take the job have having a reproducible test setup more serious I guess...

PR #2880, opened 2024-08-09
Few improvements for request sharing

MartinquaXD:
# Description
This PR ended up being a small collection of improvements all related to request sharing in the hopes of improving the memory leak situation.

# Changes
- replace all uses of `.shared()` with `.shared_or_else()` to delay the construction of shared futures as much as possible (might result in a reduction of the leaked memory 🤞)
- drop `.shared()` altogether because it's less powerful than `.shared_or_else()` and can use excessive memory
- finally the unit test is able to assert that the original future and shared cached future indeed point to the same underlying future
- added a metric to count all cached items across all individual request sharing structs
    Either this shows a constant increase which suggests that we leak memory by not cleaning up the map (and keeping strong references to the shared futures)
    Or we do everything right and the memory somehow gets leaked in the shared future destructor

## How to test
functionality didn't really change so existing tests should still pass

### DIFF
--- a/crates/shared/src/price_estimation/trade_finder.rs
+++ b/crates/shared/src/price_estimation/trade_finder.rs
@@ -61,11 +61,14 @@ impl TradeEstimator {
     }
 
     async fn estimate(&self, query: Arc<Query>) -> Result<Estimate, PriceEstimationError> {
-        let estimate = rate_limited(
-            self.rate_limiter.clone(),
-            self.inner.clone().estimate(query.clone()),
-        );
-        self.sharing.shared(query, estimate.boxed()).await
+        let fut = move |query: &Arc<Query>| {
+            rate_limited(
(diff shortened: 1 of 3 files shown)


fleupold (approved):
Nice, should we remove `shared` altogether if it's somewhat inefficient?

MartinquaXD:
> Nice, should we remove shared altogether if it's somewhat inefficient?

Yeah, actually brought that up in a previous version of the PR description. My main reason for not removing it already was that it can be quite finicky to please the compiler to allow you to use `shared_or_else()` and technically `shared()` should also not leak any memory (if the underlying `T` does not leak any memory itself).
Since we are not using `shared()` at all any more I think we can remove, though.

sunce86 (approved):

squadgazzz (approved):
LGTM

PR #2881, opened 2024-08-12
DRAFT: Replace warp with axum

MartinquaXD:
# Description
Still heavily in progress, just wanted to get it out here to not forget about it over the coming week(s).

# Changes
The `orderbook` currently still use a fork of `warp` which itself does not look very promising for the future. In the driver and other new code we've been using `axum` instead which was quite a nice experience.
This PR will eventually replace all the `warp` code with `axum` and update the existing `axum` (and more generally networking) code to the latest version. This is a pretty good time for this since many of the libraries bumped their dependencies to stable versions (e.g. `hyper`, `http` and `http-body` are now on `v1`).

Note that this PR will fail to compile because we also need to update the dependencies in the `gas-estimation` crate which I only did locally so far to get into the meat of the refactoring.

## How to test
**(locally)** all e2e tests pass again
I still have some existing unit tests for the route handling commented out. I believe this can be converted as well but I'm considering to drop them since more or less all endpoints should be covered by e2e tests instead (many small errors were already caught by e2e tests)

## TODO
[ ] bump dependencies in `gas-estimation` crate
[ ] also use `axum` for zeroex mock server in tests
[ ] appdata endpoint has a request size limit which I still have to port over
[ ] verify that rejected requests (missing handler) set the CORS headers correctly (this is why we had the `warp` fork in the first place)
[ ] verify that metrics work correctly (now managed by a library instead of handrolled)
[ ] double check that error conversion works correctly (no new `error` logs since that would lead to alerts)
[ ] replace handrolled `request_id` logic with layer from `tower-http` (should be a drop-in replacement)
[ ] probably clean up some of the code while I'm at it (e.g. a bunch of `shared` code is only used in `orderbook`)
[ ] actually drop `warp`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,7 +73,7 @@ dependencies = [
  "observe",
  "primitive-types",
  "prometheus",
- "reqwest",
+ "reqwest 0.12.5",
  "serde",
  "serde_with",
  "shared",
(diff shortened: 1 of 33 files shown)


Mateo-mro:
:heart_eyes: :heart_eyes:

PR #2883, opened 2024-08-12
Fix failing e2e test for cow amms

sunce86:
@MartinquaXD I believe you wrote the test, do you remember why this specific pool was added to the list?

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -253,17 +253,17 @@ impl SolvableOrdersCache {
         let surplus_capturing_jit_order_owners = cow_amms
             .iter()
             .filter(|cow_amm| {
-                let prices_exist = cow_amm
-                    .traded_tokens()
-                    .iter()
-                    .all(|token| prices.contains_key(token));
-                if !prices_exist {
-                    tracing::debug!(
-                        cow_amm = ?cow_amm.address(),
-                        "prices missing"
-                    );
-                }
-                prices_exist
+                cow_amm.traded_tokens().iter().all(|token| {
+                    let price_exist = prices.contains_key(token);
+                    if !price_exist {
+                        tracing::debug!(
+                            cow_amm = ?cow_amm.address(),
+                            ?token,
+                            "omitted from auction due to missing prices"
+                        );
+                    }
+                    price_exist
+                })
             })
             .map(|cow_amm| cow_amm.address())
             .cloned()
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -253,10 +253,17 @@ impl SolvableOrdersCache {
         let surplus_capturing_jit_order_owners = cow_amms
             .iter()
             .filter(|cow_amm| {
-                cow_amm
+                let prices_exist = cow_amm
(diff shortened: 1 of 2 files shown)


MartinquaXD:
Given that this error seemingly started with https://github.com/cowprotocol/services/pull/2871 I would say this indicates that we had a pool which traded cow and some other token but we couldn't compute a price for the other token so now we also drop the cow price although we have it.

squadgazzz (approved):
Nice, now we can say the change is covered by tests 😄

crates/autopilot/src/solvable_orders.rs
```diff
@@ -253,10 +253,17 @@ impl SolvableOrdersCache {
...
-                cow_amm
+                let prices_exist = cow_amm
                     .traded_tokens()
                     .iter()
-                    .all(|token| prices.contains_key(token))
+                    .all(|token| prices.contains_key(token));
+                if !prices_exist {
+                    tracing::debug!(
+                        cow_amm = ?cow_amm.address(),
+                        "prices missing"
```
MartinquaXD, line 263: log could be more explicit
Maybe: `omitted from auction due to missing prices`

MartinquaXD (commented):
@sunce86 could you double check dropping which pool causes the cow price to miss as well just to be sure this is the root cause?

crates/autopilot/src/solvable_orders.rs
sunce86: fixed log, added missing token

sunce86:
> @sunce86 could you double check dropping which pool causes the cow price to miss as well just to be sure this is the root cause?

Cow token doesn't have price, I am not sure if this is expected in this test or not

2024-08-13T08:49:19.511Z  WARN request{id="41"}:/quote{solver=test_solver}: driver::infra::observe: failed to quote order order=Order { tokens: Tokens { sell: TokenAddress(ContractAddress(0xdef1ca1fb7fbcdc777520aa7f396b4e015f497ab)), buy: TokenAddress(ContractAddress(0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2)) }, amount: TargetAmount(1000000000000000000), side: Buy, deadline: Deadline { driver: 2024-08-13T08:49:24.004674131Z, solvers: 2024-08-13T08:49:23.104904229Z } } err=QuotingFailed(NoSolutions)

2024-08-13T08:49:19.694Z DEBUG solvable_orders_cache: autopilot::solvable_orders: omitted from auction due to missing prices cow_amm=0xbeef5afe88ef73337e5070ab2855d37dbf5493a4 token=0xdef1ca1fb7fbcdc777520aa7f396b4e015f497ab

MartinquaXD:
yeah, that is expected. See [here](https://github.com/cowprotocol/services/pull/2883/files#diff-5ac885a2341d119cb2910be92647a90374783b5694ac212e4f8392384bf42b3cR562-R564).

MartinquaXD (approved):

sunce86:
> yeah, that is expected. See [here](https://github.com/cowprotocol/services/pull/2883/files#diff-5ac885a2341d119cb2910be92647a90374783b5694ac212e4f8392384bf42b3cR562-R564).

Aaaah yes just found it! Thanks!

PR #2897, opened 2024-08-15
Optional creation timestamp and solver fields

squadgazzz:
Makes order's `created` and quote's `solver` fields optional to make it possible to run shadow competition while prod doesn't have these changes yet. Default values are used to avoid a lot of changes since the PR is temporal.

> [!IMPORTANT]
> This PR should be reverted after the next weekly release.


MartinquaXD (approved):
Looks alright to get the shadow environment unblocked. Overall I think having 2 PRs might be a okay solution going forward:
1. 1st PR implements how the PR in it's final form
2. 2nd PR makes things compatible with shadow (if possible)

The reasoning is that it seems simpler to just revert a single PR than having to manually patch out the adjustments made to please `shadow`. Also that makes reviewing the original PR easier.

Thoughts?

m-lord-renkse:
> Looks alright to get the shadow environment unblocked. Overall I think having 2 PRs might be a okay solution going forward:
> 
> 1. 1st PR implements how the PR in it's final form
> 2. 2nd PR makes things compatible with shadow (if possible)
> 
> The reasoning is that it seems simpler to just revert a single PR than having to manually patch out the adjustments made to please `shadow`. Also that makes reviewing the original PR easier.
> 
> Thoughts?

I couldn't agree more.

m-lord-renkse (approved):

PR #2907, opened 2024-08-19
`observed_at` on `BlockInfo`

MartinquaXD:
# Description
Adds new field `observed_at` to `BlockInfo` which shows when the system became aware of the new block. This can be useful for synchronizing on because there are actually very few guarantees on the `timestamp` field provided by ethereum and it's also only precise to a second.

Part of https://github.com/cowprotocol/services/pull/2906 separated out for easier reviewing.

# Changes
- adds `observed_at` to `BlockInfo`
- custom impl for `Default` and `PartialEq` to handle `observed_at`

## How to test
tests should still pass

### DIFF
--- a/crates/ethrpc/src/block_stream/mod.rs
+++ b/crates/ethrpc/src/block_stream/mod.rs
@@ -237,13 +237,13 @@ pub trait BlockRetrieving: Debug + Send + Sync + 'static {
 #[async_trait::async_trait]
 impl BlockRetrieving for Web3 {
     async fn current_block(&self) -> Result<BlockInfo> {
-        get_block_info_at_id(self, BlockNumber::Latest.into())
+        get_block_at_id(self, BlockNumber::Latest.into())
             .await?
             .try_into()
     }
 
     async fn block(&self, number: u64) -> Result<BlockNumberHash> {
-        let block = get_block_info_at_id(self, U64::from(number).into()).await?;
+        let block = get_block_at_id(self, U64::from(number).into()).await?;
         Ok((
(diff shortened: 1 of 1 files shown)


crates/ethrpc/src/block_stream/mod.rs
```diff
@@ -205,12 +237,17 @@ pub trait BlockRetrieving: Debug + Send + Sync + 'static {
...
+        get_block_info_at_id(self, BlockNumber::Latest.into())
+            .await?
+            .try_into()
     }
 
     async fn block(&self, number: u64) -> Result<BlockNumberHash> {
         let block = get_block_info_at_id(self, U64::from(number).into()).await?;
-        Ok((block.number, block.hash))
+        Ok((
+            block.number.context("missing block_number")?.as_u64(),
```
m-lord-renkse, line 248: why is this change needed? :thinking:

m-lord-renkse (approved):
Good idea!

crates/ethrpc/src/block_stream/mod.rs
MartinquaXD: `get_block_info_at_id` no longer returns `BlockInfo` but a `Block<H256>` where these value can be `None`.
That being said I should rename the function. 😅

squadgazzz (approved):
LGTM

fleupold (commented):
I think it would make sense to move the current block delay method here then (so we can get an idea of how long our block stream is lagging behind on average and in the worst case).

PR #2908, opened 2024-08-19
Handle `/settle` in background task

MartinquaXD:
# Description
While working on a PR to synchronize the runloop to the blockstream I noticed that the autopilot currently waits till 1 block past the deadline. When changing that to aborting on the deadline block tests started to fail because we no longer submitted the cancellation on-chain. The reason is that the `/settle` handle stops running in the driver when the `autopilot` aborts the call due to reaching the deadline.

# Changes
Run `/settle` handler in a background task to ensure that we correctly submit the settlement or cancellation even if the autopilot aborts the `/settle` call.

## How to test
e2e test should continue to work

### DIFF
--- a/crates/driver/src/infra/api/routes/settle/mod.rs
+++ b/crates/driver/src/infra/api/routes/settle/mod.rs
@@ -1,9 +1,12 @@
 mod dto;
 
 use {
-    crate::infra::{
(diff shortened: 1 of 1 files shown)


m-lord-renkse (approved):
Good catch!

PR #2909, opened 2024-08-19
Propagate the buffered CoinGecko implementation to the configuration 3/3

### LINKED ISSUE #2814: chore: Orders excluded from auction due to missing price

m-lord-renkse:
# Description
Instantiate the CoinGecko buffered and propagate the configuration.

I set the default configuration to reasonable parameters, so we can try them out without needing to change the infra PR.

# Changes
- Instantiate the CoinGecko buffered  instead of normal CoinGecko "object"
- Propagate the configuration accordingly

## How to test
1. Regression tests

## Related Issues

Fixes #2814


crates/shared/src/price_estimation.rs
```diff
@@ -294,8 +331,12 @@ impl Display for Arguments {
         display_option(f, "balancer_sor_url", balancer_sor_url)?;
         display_secret_option(f, "one_inch_spot_price_api_key: {:?}", one_inch_api_key)?;
         writeln!(f, "one_inch_spot_price_api_url: {}", one_inch_url)?;
-        display_secret_option(f, "coin_gecko_api_key: {:?}", coin_gecko_api_key)?;
-        writeln!(f, "coin_gecko_api_url: {}", coin_gecko_url)?;
+        display_secret_option(
+            f,
+            "coin_gecko_api_key: {:?}",
+            &coin_gecko.coin_gecko_api_key,
```
m-lord-renkse, line 337: I do not like the `coin_gecko.coin_gecko_...` but there was no other way since we `flatten` it, and each field should contain the `coin_gecko_` prefix.

crates/shared/src/price_estimation.rs
```diff
@@ -222,6 +214,52 @@ pub struct Arguments {
...
+pub struct CoinGecko {
+    /// The API key for the CoinGecko API.
+    #[clap(long, env)]
+    pub coin_gecko_api_key: Option<String>,
+
+    /// The base URL for the CoinGecko API.
+    #[clap(
+        long,
+        env,
+        default_value = "https://api.coingecko.com/api/v3/simple/token_price"
```
MartinquaXD, line 227: AFAIK only the pro endpoint allows batching so it seems we need to express with config args that batching should be disabled.
There are a bunch of arguments that are only needed for the buffering implementation. Maybe it makes sense to have a way to determine based on whether or not you configured all the necessary args to enable buffering.

m-lord-renkse: @MartinquaXD really good point! :ok_hand: thanks!

m-lord-renkse: I don't think we should do that. First of all, even the default is not the PRO API, but we actually never use it / intent to use. It is set like that because it is the only one working without API Key. I don't think the backend / services is responsible to assert that the CoinGecko URL is the correct one for batching, it should be 100 % responsibility of the person who set the configuration. The services should be agnostic from these things.

squadgazzz: So, by default, the configuration provides a URL and enabled buffering, which won't work together, correct? Not sure what exactly stops us from using that URL with disabled buffering by default.

m-lord-renkse: Really good catch @MartinquaXD , thank you very much. I corrected it and added several test to ensure we have the correct/desired behavior.

MartinquaXD (approved):
Very nice handling of the arguments and good tests!
I believe there are probably a few other arguments where this pattern can be applied.
From the top of my head:
* tenderly settings
* blockscout API
* ethplorer API

crates/shared/src/price_estimation.rs
```diff
@@ -222,6 +214,52 @@ pub struct Arguments {
...
+    )]
+    pub coin_gecko_url: Url,
+
+    #[clap(flatten)]
+    pub coin_gecko_buffered: Option<CoinGeckoBuffered>,
+}
+
+#[derive(clap::Parser)]
+#[clap(group(
+    clap::ArgGroup::new("buffered")
```
MartinquaXD, line 237: One more nit I thought about after approving:
I assume these group names have to be unique and given that we already have a buffering system for rpc requests I would preemptively like to disambiguate them by calling this `coingecko_buffering` or sth like that.

m-lord-renkse:
> Very nice handling of the arguments and good tests! I believe there are probably a few other arguments where this pattern can be applied. From the top of my head:
> 
> * tenderly settings
> * blockscout API
> * ethplorer API

This is a low hanging fruit! good point!

crates/shared/src/price_estimation.rs
m-lord-renkse: great idea, changed it to `coin_gecko_buffered` to match the field name too :ok_hand:

squadgazzz (approved):

PR #2911, opened 2024-08-19
Save all auction prices

sunce86:
# Description

Historical context:

Initially we wanted to save all auction prices so we can reconstruct the whole auction object if needed.

Then, I implemented the optimization that we save prices only for orders that appear in the winning solution, with the assumption that this is all we will need. 

However, this assumption changed with cow amm implementation, where cow amm orders do not exist in the auction, therefore we are unable to determine their sell/buy tokens and save token prices for those cow amm orders.

Consequently, missing prices cause failure in score computing, where for all surplus capturing orders (cow amm orders) we need to have auction prices.

# Changes

- [ ] Save all auction prices to database

### DIFF
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -31,7 +31,7 @@ use {
     rand::seq::SliceRandom,
     shared::token_list::AutoUpdatingTokenList,
     std::{
-        collections::{BTreeMap, HashMap, HashSet},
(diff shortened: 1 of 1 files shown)


MartinquaXD:
Could we not move the price deletion to the `OnSettlementEventUpdater`? When that ran we can delete the prices again, right?

MartinquaXD (approved):

PR #2916, opened 2024-08-21
Execute orders filtering in parallel

squadgazzz:
# Description
During auction preparation, some order filtering async functions take some time to execute. Currently, all of them are processed sequentially, which has a negative impact on the total solvable orders cache update execution time.

# Changes

- [ ] Execute all of the async order filtering functions in parallel.
- [ ] A bit of code refactoring for better readability.

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -176,36 +176,19 @@ impl SolvableOrdersCache {
         let mut invalid_order_uids = HashSet::new();
         let mut filtered_order_events = Vec::new();
 
-        let orders = self
-            .filter_invalid_orders(
+        let queries = db_solvable_orders
+            .orders
+            .iter()
+            .map(Query::from_order)
+            .collect::<Vec<_>>();
+        let (balances, orders) = tokio::join!(
+            self.get_balances(queries),
+            self.filter_invalid_orders(
                 db_solvable_orders.orders,
                 &mut counter,
                 &mut invalid_order_uids,
             )
-            .await;
-
-        let missing_queries: Vec<_> = orders.iter().map(Query::from_order).collect();
-        let fetched_balances = {
-            let _timer = self.stage_timer("balance_fetch");
-            self.balance_fetcher.get_balances(&missing_queries).await
-        };
-        let balances = missing_queries
-            .into_iter()
-            .zip(fetched_balances)
-            .filter_map(|(query, balance)| match balance {
-                Ok(balance) => Some((query, balance)),
-                Err(err) => {
-                    tracing::warn!(
-                        owner = ?query.owner,
-                        token = ?query.token,
-                        source = ?query.source,
-                        error = ?err,
-                        "failed to get balance"
-                    );
-                    None
-                }
-            })
-            .collect::<HashMap<_, _>>();
+        );
 
         let orders = orders_with_balance(orders, &balances);
         let removed = counter.checkpoint("insufficient_balance", &orders);
@@ -216,11 +199,8 @@ impl SolvableOrdersCache {
         filtered_order_events.extend(removed);
 
         // create auction
-        let (orders, mut prices) = get_orders_with_native_prices(
-            orders.clone(),
-            &self.native_price_estimator,
-            self.metrics,
-        );
+        let (orders, mut prices) =
+            get_orders_with_native_prices(orders, &self.native_price_estimator, self.metrics);
         // Add WETH price if it's not already there to support ETH wrap when required.
         if let Entry::Vacant(entry) = prices.entry(self.weth) {
             let _timer = self.stage_timer("weth_price_fetch");
@@ -328,6 +308,30 @@ impl SolvableOrdersCache {
             .auction_update_total_time
             .observe(start.elapsed().as_secs_f64());
         Ok(())
+    }
+
+    async fn get_balances(&self, queries: Vec<Query>) -> HashMap<Query, U256> {
+        let fetched_balances = {
+            let _timer = self.stage_timer("balance_fetch");
+            self.balance_fetcher.get_balances(&queries).await
+        };
+        queries
+            .into_iter()
+            .zip(fetched_balances)
+            .filter_map(|(query, balance)| match balance {
+                Ok(balance) => Some((query, balance)),
+                Err(err) => {
+                    tracing::warn!(
+                        owner = ?query.owner,
+                        token = ?query.token,
+                        source = ?query.source,
+                        error = ?err,
+                        "failed to get balance"
+                    );
+                    None
+                }
+            })
+            .collect()
     }
 
     /// Executed effectful orders filtering in parallel.
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -176,19 +176,26 @@ impl SolvableOrdersCache {
         let mut invalid_order_uids = HashSet::new();
         let mut filtered_order_events = Vec::new();
 
-        let queries = db_solvable_orders
-            .orders
-            .iter()
-            .map(Query::from_order)
-            .collect::<Vec<_>>();
-        let (balances, orders) = tokio::join!(
-            self.get_balances(queries),
-            self.filter_invalid_orders(
-                db_solvable_orders.orders,
-                &mut counter,
-                &mut invalid_order_uids,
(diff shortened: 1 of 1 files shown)


crates/autopilot/src/solvable_orders.rs
```diff
@@ -348,6 +313,66 @@ impl SolvableOrdersCache {
...
+                        error = ?err,
+                        "failed to get balance"
+                    );
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Executed effectful orders filtering in parallel.
```
MartinquaXD, line 340: What does "effectful" in mean here?

crates/autopilot/src/solvable_orders.rs
```diff
@@ -348,6 +313,66 @@ impl SolvableOrdersCache {
...
+            let _timer = self.stage_timer("unsupported_token_filtering");
+            find_unsupported_tokens(&orders, self.bad_token_detector.clone()).await
+        };
+        let (banned_user_orders, invalid_signature_orders, unsupported_token_orders) = tokio::join!(
+            banned_user_orders_fut,
+            invalid_signature_orders_fut,
+            unsupported_token_orders_fut,
+        );
+
+        counter.checkpoint_by_invalid_orders("banned_user", &banned_user_orders);
```
MartinquaXD, line 365: These can double count orders since a single order could get filtered out for multiple reasons. But I think this is fine to not open another can of worms before merging this.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -367,12 +392,9 @@ impl SolvableOrdersCache {
...
+/// Finds all orders whose owners or receivers are in the set of "banned"
 /// users.
-async fn filter_banned_user_orders(
-    mut orders: Vec<Order>,
-    banned_users: &banned::Users,
-) -> Vec<Order> {
+async fn find_banned_user_orders(orders: &[Order], banned_users: &banned::Users) -> Vec<OrderUid> {
     let banned = banned_users
         .banned(orders.iter().flat_map(|order| {
             [
```
MartinquaXD, line 400: Could you adjust the `banned users` filtering slightly as well in this PR.
Just noticed that we are currently running a lot of duplicate requests for checking the address `0x00000` due to the `receiver.unwrap_of_default()`.

I would suggest something like this to avoid these duplicate requests:
```rust
            orders
                .iter()
                .flat_map(|order| std::iter::once(order.metadata.owner).chain(order.data.receiver)),
```

crates/autopilot/src/solvable_orders.rs
```diff
@@ -381,11 +403,14 @@ async fn filter_banned_user_orders(
...
-    orders.retain(|order| {
-        !banned.contains(&order.metadata.owner)
-            && !banned.contains(&order.data.receiver.unwrap_or_default())
-    });
     orders
+        .iter()
+        .filter_map(|order| {
+            (banned.contains(&order.metadata.owner)
+                || banned.contains(&order.data.receiver.unwrap_or_default()))
+            .then_some(order.metadata.uid)
```
MartinquaXD, line 411: nit: Same suggestion as above to avoid the map lookup when `receiver == None`.
```suggestion
            std::iter::once(&order.metadata.owner)
                .chain(&order.data.receiver)
                .any(|addr| banned.contains(addr))
                .then_some(order.metadata.uid)
```

crates/autopilot/src/solvable_orders.rs
```diff
@@ -402,18 +427,22 @@ fn get_native_prices(
...
-async fn filter_invalid_signature_orders(
-    mut orders: Vec<Order>,
+async fn find_invalid_signature_orders(
+    orders: &[Order],
     signature_validator: &dyn SignatureValidating,
-) -> Vec<Order> {
-    orders.retain(|order| {
-        !matches!(
+) -> Vec<OrderUid> {
+    let (invalid_orders, orders): (Vec<_>, Vec<_>) = orders.iter().partition(|order| {
```
MartinquaXD, line 436: You can use `.partition_map()` instead to avoid re-collecting the `invalid_orders` vector a few lines down.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -656,10 +682,10 @@ fn to_normalized_price(price: f64) -> Option<U256> {
...
-async fn filter_unsupported_tokens(
-    mut orders: Vec<Order>,
+async fn find_unsupported_tokens(
+    orders: &[Order],
     bad_token: Arc<dyn BadTokenDetecting>,
-) -> Vec<Order> {
+) -> Vec<OrderUid> {
     let bad_tokens = join_all(
         orders
             .iter()
```
MartinquaXD, line 691: Just noticed that in the line below (can't comment on that line) you can also use `token_pair().into_iter().flatten()` to avoid the weird `TokenPair::default()` implementation.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -795,6 +824,21 @@ impl OrderFilterCounter {
         filtered_orders.into_keys().collect()
     }
 
+    /// Creates a new checkpoint based on the found invalid orders.
+    fn checkpoint_by_invalid_orders(&mut self, reason: Reason, invalid_orders: &[OrderUid]) {
+        *self.counts.entry(reason).or_default() += invalid_orders.len();
+        for order_uid in invalid_orders {
+            self.orders.remove(order_uid).unwrap();
```
MartinquaXD, line 831: If an order gets removed for multiple reasons this `.unwrap()` would lead to a panic IIUC.
Let's just ignore the return value from `.remove()` here since we don't really care about the value anyway.
```suggestion
            let _ = self.orders.remove(order_uid);
```

crates/autopilot/src/solvable_orders.rs
```diff
@@ -795,6 +824,21 @@ impl OrderFilterCounter {
...
+        }
+        if !invalid_orders.is_empty() {
```
MartinquaXD, line 833: Seems like it would be a bit nicer to have this early return at the start of the function.

crates/autopilot/src/solvable_orders.rs
```diff
@@ -432,29 +461,26 @@ async fn filter_invalid_signature_orders(
...
-            if let Signature::Eip1271(_) = &order.signature {
-                if let Err(err) = validations.next().unwrap() {
-                    tracing::warn!(
-                        order =% order.metadata.uid, ?err,
-                        "invalid EIP-1271 signature"
-                    );
-                    return false;
-                }
+    for order in orders {
+        if let Signature::Eip1271(_) = &order.signature {
```
MartinquaXD, line 472: nit: could lose one level of indentation with:
```rust
        if !matches!(&order.signature, Signature::Eip1271(_)) {
            continue;
        }
```

MartinquaXD (commented):
Looks pretty good.
Just a few of micro-optimizations and code golf nits.
Also IIUC there is the issue with orders being filtered out for multiple reasons. But unless a solution for that is very easy I would not worry about that for now.

crates/autopilot/src/solvable_orders.rs
squadgazzz: Good catch, thanks. Used another counter for those.

crates/autopilot/src/solvable_orders.rs
squadgazzz: I tried to express "side effect" operations in other words. Removed to avoid confusion.

crates/autopilot/src/solvable_orders.rs
squadgazzz: > a lot of duplicate requests

I think we don't make a lot of the requests since HasSet is used later. But removing 0x address makes sense.
https://github.com/cowprotocol/services/blob/488fb8b8db2641760d30c8aab6b1083956e81ac7/crates/order-validation/src/banned.rs#L56-L67

crates/autopilot/src/solvable_orders.rs
MartinquaXD: Sweet, your fix is actually super simple. The results of `.remove()` came in handy after all. 👍

MartinquaXD (approved):

sunce86 (approved):
Code looks good.

Improvement wise not sure, I have no perception how fast is signature validator vs balance fetcher and whether this change will improve speed or opposite.

fleupold (approved):

PR #2919, opened 2024-08-22
Bad token detector maintenance task

squadgazzz:
# Description
Bad token detection sometimes takes too much time to update the required tokens. Even though we now execute [such an operation for each token in parallel](https://github.com/cowprotocol/services/pull/2915), this can be improved by introducing a background task that periodically updates all the expired tokens so the caller side faces expired tokens much less.

# Changes

- [ ] A new background task that periodically finds all the expired tokens and re-requests their quality.

## How to test
Existing tests + Grafana metrics

### DIFF
--- a/crates/autopilot/src/arguments.rs
+++ b/crates/autopilot/src/arguments.rs
@@ -82,15 +82,6 @@ pub struct Arguments {
     )]
     pub token_quality_cache_expiry: Duration,
 
-    /// How often to run background bad token detection maintenance task.
-    #[clap(
-        long,
-        env,
-        default_value = "1m",
-        value_parser = humantime::parse_duration,
-    )]
-    pub token_quality_maintenance_timeout: Duration,
-
     /// The number of pairs that are automatically updated in the pool cache.
     #[clap(long, env, default_value = "200")]
     pub pool_cache_lru_size: NonZeroUsize,
@@ -258,7 +249,6 @@ impl std::fmt::Display for Arguments {
             allowed_tokens,
             unsupported_tokens,
             token_quality_cache_expiry,
-            token_quality_maintenance_timeout,
             pool_cache_lru_size,
             native_price_estimators,
             min_order_validity_period,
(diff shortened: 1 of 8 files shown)


crates/autopilot/src/arguments.rs
```diff
@@ -82,6 +82,15 @@ pub struct Arguments {
     )]
     pub token_quality_cache_expiry: Duration,
 
+    /// How often to run background bad token detection maintenance task.
```
MartinquaXD, line 85: We already have a tonne of arguments.
Let's try to make it work without this once since the only requirement is that it runs more frequently than `--token-quality-cache-expiry`.

crates/shared/src/bad_token/cache.rs
```diff
@@ -55,22 +64,84 @@ impl CachingDetector {
             .unwrap()
             .insert(token, (Instant::now(), quality));
     }
+
+    fn insert_many_into_cache(&self, tokens: impl Iterator<Item = (H160, TokenQuality)>) {
+        let mut cache = self.cache.lock().unwrap();
+        for (token, quality) in tokens {
+            cache.insert(token, (Instant::now(), quality));
```
MartinquaXD, line 71: nit: Let's compute `Instant::now()` once outside the loop and copy the same value.
Doesn't really matter if the timestamps are different by nanoseconds.

MartinquaXD (commented):
Looks good overall.
Just questioning if we can get by without the new CLI arguments.

crates/autopilot/src/arguments.rs
squadgazzz: Used this value: `cache_expiry.div(10).max(Duration::from_secs(30))`

MartinquaXD (approved):

crates/shared/src/bad_token/cache.rs
```diff
@@ -55,14 +59,74 @@ impl CachingDetector {
...
+        let mut cache = self.cache.lock().unwrap();
+        let now = Instant::now();
+        for (token, quality) in tokens {
+            cache.insert(token, (now, quality));
+        }
+    }
+
+    fn spawn_maintenance_task(self: Arc<Self>) {
+        let cache_expiry = self.cache_expiry;
+        let maintenance_timeout = cache_expiry.div(10).max(Duration::from_secs(30));
```
sunce86, line 73: is this some random math? Is it fine to have the same for all networks?

sunce86 (approved):

PR #2922, opened 2024-08-23
Remove concurrency limit on forked e2e tests

MartinquaXD:
# Description
The forked e2e tests are overall pretty useful but the public nodes we've used so far have been pretty flaky which leads to people not paying too much attention when a forked test fails for a good reason.

# Changes
- I changed the github secrets to use self hosted nodes which should be more reliable
- removed the concurrency limit on the tests as we don't have to adhere to any compute limits anymore

## How to test
forked tests should still work


m-lord-renkse (approved):
Nice!

PR #2924, opened 2024-08-23
[EASY] Improve config group for blockscout and ethplorer

m-lord-renkse:
# Description
Improve the groups for `ethplorer` and `blockscout` inspired by the CoinGecko configuration done in: https://github.com/cowprotocol/services/pull/2909

## How to test
1.Unit test

### DIFF
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -241,7 +241,7 @@ pub struct Arguments {
 pub fn display_secret_option<T>(
     f: &mut Formatter<'_>,
     name: &str,
-    option: &Option<T>,
+    option: Option<&T>,
 ) -> std::fmt::Result {
     display_option(f, name, &option.as_ref().map(|_| "SECRET"))
 }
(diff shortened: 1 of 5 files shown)


crates/shared/src/bad_token/token_owner_finder.rs
```diff
@@ -75,21 +75,13 @@ pub struct Arguments {
...
-    /// Override the default blockscout API url for this network
-    #[clap(long, env)]
-    pub blockscout_api_url: Option<Url>,
-
-    /// The blockscout API key.
-    #[clap(long, env)]
-    pub blockscout_api_key: Option<String>,
+    /// The blockscout configuration.
+    #[clap(flatten)]
+    pub blockscout: Blockscout,
```
MartinquaXD, line 80: Shouldn't the groups themselves be optional and the individual fields in the struct be required?
That way you only have the `blockscout` struct if everything is configured correctly and don't have to map optionals awkwardly.

crates/shared/src/arguments.rs
```diff
@@ -241,7 +241,7 @@ pub struct Arguments {
 pub fn display_secret_option<T>(
     f: &mut Formatter<'_>,
     name: &str,
-    option: &Option<T>,
+    option: Option<&T>,
```
m-lord-renkse, line 244: `&Option<T>` is an anti-pattern.

crates/shared/src/bad_token/token_owner_finder.rs
MartinquaXD: Seems like this can be achieved by making the fields not `Option`s but setting `required=false` on the field macro.

MartinquaXD (approved):

PR #2926, opened 2024-08-27
Insert auction prices in batches

squadgazzz:
# Description
Since all the native prices are stored instead of just the ones the winning solution won, this process takes much more time than before.
image

# Changes
Store prices in batches should speed up this.

## How to test
Existing tests.


sunce86 (approved):
Nice fix. How did you estimate the proper batch size to be 5000?

squadgazzz:
> How did you estimate the proper batch size to be 5000?

Based on previous experience. where ~5k batches showed best performance(but actually with more columns).

MartinquaXD (approved):

PR #2927, opened 2024-08-28
[EASY] Use autopilot ethrpc config from args

squadgazzz:
# Description
Currently, ethrpc values are hardcoded. It makes sense to start using values from arguments to adjust them using infra. No config change is required since the same ones are used in the pulumi config.

### DIFF
--- a/crates/autopilot/src/boundary/mod.rs
+++ b/crates/autopilot/src/boundary/mod.rs
@@ -26,9 +26,8 @@ use {crate::domain, ethrpc::Web3, std::collections::HashMap, url::Url};
 pub mod events;
 pub mod order;
 
-/// Builds a web3 client that bufferes requests and sends them in a
-/// batch call.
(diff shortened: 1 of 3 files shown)


sunce86 (approved):

crates/autopilot/src/boundary/mod.rs
```diff
@@ -28,17 +28,12 @@ pub mod order;
 
 /// Builds a web3 client that bufferes requests and sends them in a
 /// batch call.
-pub fn buffered_web3_client(ethrpc: &Url) -> Web3 {
-    let ethrpc_args = shared::ethrpc::Arguments {
-        ethrpc_max_batch_size: 20,
-        ethrpc_max_concurrent_requests: 10,
-        ethrpc_batch_delay: Default::default(),
-    };
+pub fn buffered_web3_client(ethrpc: &Url, ethrpc_args: &shared::ethrpc::Arguments) -> Web3 {
```
MartinquaXD, line 31: nit: `ethrpc_args` can be configured such that no buffering will happen so this function name is a bit misleading.

MartinquaXD (approved):

PR #2928, opened 2024-08-28
Insert auction participants in batches

squadgazzz:
Similarly to https://github.com/cowprotocol/services/pull/2926, inserts auction participants to the DB in batches to speed up the process since this is executed inside the runloop critical path.


sunce86 (approved):
This doesn't bring any improvement, it's just for completness to apply similar way of inserting for most tables?

MartinquaXD (approved):

PR #2930, opened 2024-08-28
Revert #2906 & #2925

squadgazzz:
Reverts #2906 & #2925 since there are still problems on arb1 for some reason.


sunce86 (approved):
We might still have the time to fix any pending issues if you have time for it, but approving so the PR can be merged before Friday EOD if need be.

squadgazzz:
> We might still have the time to fix any pending issues if you have time for it, but approving so the PR can be merged before Friday EOD if need be.

Yep, the PR was created just in case the issue is not addressed until EOW. I am currently looking into it.

PR #2931, opened 2024-08-29
Handle ethflow_refunds empty table case

squadgazzz:
# Description
https://github.com/cowprotocol/services/pull/2906 moved refund indexer maintenance into the solvable orders cache update loop. When there are no ethflow refunds in the DB or the last refund is very old, the refund indexer tries to fetch all the blocks, which might execute for more than 10 minutes, resulting in autopilot restarts and the whole protocol being stuck.

# Changes

Change the behavior of calculating the starting index for refund indexer. If a recent eth-flow refund exists **within the last day**, it prioritizes this. Otherwise, it falls back to the most recent block from broadcasted orders. Then select the highest value among the value calculated in the previous step, settlement contract block and configured `ethflow_indexing_start`.

## How to test
Staging.
image

### DIFF
--- a/crates/autopilot/src/database/onchain_order_events/ethflow_events.rs
+++ b/crates/autopilot/src/database/onchain_order_events/ethflow_events.rs
@@ -219,8 +219,8 @@ pub async fn determine_ethflow_refund_indexing_start(
 }
 
 /// This function attempts to find the latest block that has processed eth-flow
-/// orders or broadcasted orders. If recent Eth-flow orders exist within the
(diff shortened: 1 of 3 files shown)


fleupold (approved):
I think this change works, however I think it would be better to keep this job off the critical auction creation path altogether. 
Ethflow refunds have nothing to do with the auction creation (in fact they could be run in a separate permissionless service).

Maybe we don't need to put all maintenance jobs on the critical path and instead have some that are still run in the background regardless of the sync loop.

PR #2932, opened 2024-08-30
[EASY] Add log for volume based fee

sunce86:
# Description
Haris noticed we are missing logs for volume based fees in the driver.

### DIFF
--- a/crates/driver/src/domain/competition/solution/fee.rs
+++ b/crates/driver/src/domain/competition/solution/fee.rs
@@ -126,7 +126,11 @@ impl Fulfillment {
                 )?;
                 self.calculate_fee(price_limits, prices, *factor, *max_volume_factor)
             }
-            FeePolicy::Volume { factor } => self.fee_from_volume(prices, *factor),
+            FeePolicy::Volume { factor } => {
+                let fee_from_volume = self.fee_from_volume(prices, *factor)?;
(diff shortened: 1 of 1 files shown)


squadgazzz (approved):
Why do we want to log only Volume based fees?

crates/driver/src/domain/competition/solution/fee.rs
```diff
@@ -126,7 +126,11 @@ impl Fulfillment {
                 )?;
                 self.calculate_fee(price_limits, prices, *factor, *max_volume_factor)
             }
-            FeePolicy::Volume { factor } => self.fee_from_volume(prices, *factor),
+            FeePolicy::Volume { factor } => {
+                let fee_from_volume = self.fee_from_volume(prices, *factor)?;
+                tracing::debug!(uid=?self.order().uid, ?fee_from_volume, executed=?self.executed(), surplus_fee=?self.surplus_fee(), "calculated protocol fee");
```
MartinquaXD, line 131: `cargo fmt` is not always able to format long macros.
Could you manually add the required line breaks here and for the `price improvement` log?

MartinquaXD (approved):

sunce86:
> Why do we want to log only Volume based fees?

Other types are already logged inside `calculate_fee` function.

crates/driver/src/domain/competition/solution/fee.rs
sunce86: cargo fmt returns errors when i manually add breaks

PR #2934, opened 2024-09-03
Handle fatal errors when indexing cow amms

MartinquaXD:
# Description
Currently the `autopilot` maintenance loop can get stuck when the configured cow amm helper contract does not support a found contract with this error:
```
2024-09-03T06:13:17.420Z  WARN autopilot::maintenance: failed to run background task successfully err=method 'tokens(address):(address[])' failure: contract call reverted with message: None

Caused by:
    contract call reverted with message: None
```

So far we have been retrying this error over and over although the call to detect which tokens the pool is trading will never work if an amm is misconfigured or generally not supported by the helper contract.

# Changes
Differentiate between retryable and fatal errors while indexing the amms and retry or skip this pool respectively.

# Testplan
Set up sepolia configuration locally and checked that we don't get stuck on the unsupported pools:
```
2024-09-03T06:24:55.880Z  INFO cow_amm::cache: helper contract does not support amm cow_amm=0xe4abfda4e8c02fcafc34981dafaeb426aa4186e6 err=MethodError { signature: "tokens(address):(address[])", inner: Revert(None) }
2024-09-03T06:24:55.921Z  INFO cow_amm::cache: indexed new cow amm cow_amm=0xac140f325afd20a733e12580aeb22ff9bf46982f
2024-09-03T06:24:55.956Z  INFO cow_amm::cache: indexed new cow amm cow_amm=0xa54442606548bf1b627662a465a40b31b7e8e711
```

### DIFF
--- a/crates/cow-amm/src/cache.rs
+++ b/crates/cow-amm/src/cache.rs
@@ -76,18 +76,21 @@ impl EventStoring<CowAmmEvent> for Storage {
         &mut self,
         events: Vec<ethcontract::Event<CowAmmEvent>>,
     ) -> anyhow::Result<()> {
-        let cache = &mut *self.0.cache.write().await;
-
+        let mut processed_events = Vec::with_capacity(events.len());
         for event in events {
-            let meta = event
-                .meta
-                .as_ref()
-                .ok_or_else(|| anyhow::anyhow!("Event missing meta"))?;
+            let Some(meta) = event.meta else {
+                tracing::warn!(?event, "event does not contain required meta data");
+                continue;
+            };
+
             let CowAmmEvent::CowammpoolCreated(cow_amm) = event.data;
             let cow_amm = cow_amm.amm;
-            let amm = match Amm::new(cow_amm, &self.0.helper).await {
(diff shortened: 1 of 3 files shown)


crates/cow-amm/src/cache.rs
```diff
@@ -85,11 +85,19 @@ impl EventStoring<CowAmmEvent> for Storage {
...
+                    tracing::info!(?cow_amm, ?err, "helper contract does not support amm");
+                    continue;
+                }
+            };
 
-            cache
-                .entry(meta.block_number)
-                .or_default()
-                .push(Arc::new(Amm::new(cow_amm, &self.0.helper).await?));
+            cache.entry(meta.block_number).or_default().push(amm);
```
squadgazzz, line 100: Could that be a case where some events are processed twice because of a retriable error in the middle, meaning the cache will contain duplicates?

MartinquaXD: Good catch! We would have to be very unlucky to run into this issue but fixing the problem should not be hard either.

squadgazzz (approved):

PR #2936, opened 2024-09-03
Prefetch tokens quality cache

m-lord-renkse:
# Description
Implement a prefetch for the token quality cache (similar to what we do in the native prices cache). This ensures that we always keep the cache updated.

# Changes
- Change the `std::sync::Mutex` for a `tokio::sync::RwLock`, improving the performance
- Add to the orderbook and autopilot configuration the field `token_quality_cache_prefetch_time` in order to configure the prefetch time for token quality cache
- Change the `maintenance_timeout` in order to ensure its sampling time against `token_quality_cache_prefetch_time`

## How to test
1. Unit test

### DIFF
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -237,7 +237,7 @@ pub struct Arguments {
     #[clap(long, env)]
     pub balancer_v2_vault_address: Option<H160>,
 
-    /// The amount of time in seconds a classification of a token into good or
+    /// The amount of time a classification of a token into good or
     /// bad is valid for.
     #[clap(
         long,
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -236,6 +236,31 @@ pub struct Arguments {
     /// Override address of the balancer vault contract.
     #[clap(long, env)]
     pub balancer_v2_vault_address: Option<H160>,
+
+    /// The amount of time in seconds a classification of a token into good or
+    /// bad is valid for.
+    #[clap(
+        long,
+        env,
+        default_value = "10m",
+        value_parser = humantime::parse_duration,
+    )]
+    pub token_quality_cache_expiry: Duration,
+
+    /// How long before expiry the token quality cache should try to update the
+    /// token quality in the background. This is useful to make sure that token
+    /// quality for every cached token is usable at all times. This value
+    /// has to be smaller than `token_quality_cache_expiry`
+    /// This configuration also affects the period of the token quality
+    /// maintenance job. Maintenance period =
+    /// `token_quality_cache_prefetch_time` / 2
+    #[clap(
+        long,
+        env,
+        default_value = "2m",
+        value_parser = humantime::parse_duration,
+    )]
+    pub token_quality_cache_prefetch_time: Duration,
 }
 
 pub fn display_secret_option<T>(
@@ -336,6 +361,8 @@ impl Display for Arguments {
             custom_univ2_baseline_sources,
             liquidity_fetcher_max_age_update,
             max_pools_to_initialize_cache,
+            token_quality_cache_expiry,
+            token_quality_cache_prefetch_time,
         } = self;
 
         write!(f, "{}", ethrpc)?;
@@ -403,6 +430,16 @@ impl Display for Arguments {
             f,
             "max_pools_to_initialize_cache: {}",
             max_pools_to_initialize_cache
+        )?;
+        writeln!(
+            f,
+            "token_quality_cache_expiry: {:?}",
+            token_quality_cache_expiry
+        )?;
+        writeln!(
+            f,
+            "token_quality_cache_prefetch_time: {:?}",
+            token_quality_cache_prefetch_time
         )?;
 
         Ok(())
--- a/crates/shared/src/bad_token/cache.rs
+++ b/crates/shared/src/bad_token/cache.rs
@@ -6,44 +6,50 @@ use {
     std::{
         collections::HashMap,
         ops::Div,
-        sync::{Arc, Mutex},
+        sync::Arc,
         time::{Duration, Instant},
     },
+    tokio::sync::RwLock,
 };
 
 pub struct CachingDetector {
     inner: Box<dyn BadTokenDetecting>,
-    // std mutex is fine because we don't hold lock across await.
-    cache: Mutex<HashMap<H160, (Instant, TokenQuality)>>,
+    cache: RwLock<HashMap<H160, (Instant, TokenQuality)>>,
     cache_expiry: Duration,
+    prefetch_time: Duration,
 }
 
 #[async_trait::async_trait]
 impl BadTokenDetecting for CachingDetector {
     async fn detect(&self, token: H160) -> Result<TokenQuality> {
-        if let Some(quality) = self.get_from_cache(&token, Instant::now()) {
+        if let Some(quality) = self.get_from_cache(&token, Instant::now()).await {
             return Ok(quality);
         }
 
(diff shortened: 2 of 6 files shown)


crates/shared/src/bad_token/cache.rs
```diff
@@ -53,38 +59,43 @@ impl CachingDetector {
...
 
     fn spawn_maintenance_task(self: Arc<Self>) {
-        let cache_expiry = self.cache_expiry;
-        let maintenance_timeout = cache_expiry.div(10).max(Duration::from_secs(60));
+        // We need to prefetch the token quality the `prefetch_time` before the cache
+        // expires
+        let prefetch_time_to_expire = self.cache_expiry - self.prefetch_time;
+        // The maintenance period has to be at least double of the prefetch time in
+        // order to guarantee that the prefetch time is executed before the token
+        // quality expires. This is because of the Nyquist–Shannon sampling theorem.
```
MartinquaXD, line 83: > Nyquist–Shannon sampling theorem

You like to throw around these fancy words, don't you? 🤓

crates/shared/src/bad_token/cache.rs
```diff
@@ -53,38 +59,43 @@ impl CachingDetector {
...
+        let maintenance_timeout = self.prefetch_time.div(2);
```
squadgazzz, line 84: Shouldn't that be the opposite(according to the comment)?
```suggestion
        let maintenance_timeout = self.prefetch_time.mul(2);
```

MartinquaXD (commented):
Looks alright. Just needs to have some code moved around to reduce duplication.

crates/shared/src/bad_token/cache.rs
```diff
@@ -53,38 +59,43 @@ impl CachingDetector {
...
-    fn insert_many_into_cache(&self, tokens: impl Iterator<Item = (H160, TokenQuality)>) {
-        let mut cache = self.cache.lock().unwrap();
+    async fn insert_many_into_cache(&self, tokens: impl Iterator<Item = (H160, TokenQuality)>) {
+        let mut cache = self.cache.write().await;
         let now = Instant::now();
-        for (token, quality) in tokens {
+        tokens.into_iter().for_each(|(token, quality)| {
             cache.insert(token, (now, quality));
-        }
+        });
```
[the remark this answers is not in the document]
m-lord-renkse: Nothing in particular, I find more idiomatic the zero-cost abstractions like `into_iter().for_each()`, different preferences :laughing:

crates/shared/src/bad_token/cache.rs
m-lord-renkse: Good catch! thanks a lot!

crates/shared/src/bad_token/cache.rs
m-lord-renkse: I like to show I am an educated person :nerd_face: :rofl:

crates/shared/src/bad_token/cache.rs
m-lord-renkse: the comment is incorrect, it is the frequency the one which has to be doubled, not period (which has to be divided by 2)

crates/shared/src/bad_token/cache.rs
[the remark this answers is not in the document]
squadgazzz: Doesn't closures provide performance overhead? Didn't know that classic loops have cost-related issues.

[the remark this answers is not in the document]
MartinquaXD: Unless the compiler fucks up handrolled loops and iterators should result in the same assembly. In fact this particular example is exactly equivalent since `for item in list` is really just syntactic sugar for `list.into_iter().for_each()`.

crates/shared/src/bad_token/cache.rs
squadgazzz: We now running the task every 4 minutes instead of 1. Doesn't that work in the opposite way the theorem states?

squadgazzz: > the comment is incorrect, it is the frequency the one which has to be doubled, not period (which has to be divided by 2)

Ok, then, shouldn't we use `min(1 minute,  self.prefetch_time.div(2))` then?

crates/shared/src/bad_token/cache.rs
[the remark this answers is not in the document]
m-lord-renkse: Classic loops doesn't have any cost-related issues. Actually both codes will compile to exactly the same or almost the same code.

What I meant is that `into_iter()` and `for_each()` are zero-cost abstraction functions, meaning that they don't incur more overhead than doing the iterator in the "classical" way (`for` in this case). And they have the advantage of providing potential easy chaining (like a `filter()`) in an easy way. And it is usually the preferred/most common way of doing iterations in Rust unless you want to have an early exit in the loop.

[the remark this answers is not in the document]
squadgazzz: > And they have the advantage of providing potential easy chaining (like a filter()) in an easy way. 

This is completely useless in this particular case. 

> And it is usually the preferred/most common way of doing iterations in Rust unless you want to have an early exit in the loop.

Ok, not a big deal. Just a bit surprised by that since Rust is not a purely functional programming language(where classic for loops are really anti-pattern). Still IMO, having fewer symbols is better.

crates/shared/src/bad_token/cache.rs
squadgazzz: > So if we change it in the future, we know this, and we don't set a crazily short maintenance_time.

I am worried about the opposite again: too long timeouts. That is why I suggested using the `min` function and not `max.`

m-lord-renkse: oh, true, you are right, I got confused because the original code was `max`. Oh yeah, definitely, a really good point! :ok_hand: thanks!

crates/shared/src/bad_token/cache.rs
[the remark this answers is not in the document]
m-lord-renkse: > This is completely useless in this particular case.

True, but you never know how the code will evolve in the future. That's the reason it is always preferred to do it like that when possible. Also adds consistency to the code base.

crates/shared/src/bad_token/cache.rs
m-lord-renkse: ummm actually thinking about that. If the timeouts are too long, it is because the user configured it, and actually running it more often wouldn't have any big benefit, the only benefit is that the token may be updated at the beginning of its `prefetch_time` duration.

We are enforcing that the `cache_expiry` is greater than the `prefetch_time`. If the values are:
- `cache_expiry`: 1 hour
- `prefetch_time`: 30 mins

Then it will run the loop every 15 mins, to check for updates, the cache is valid for 1 hour anyway, so as long as we update it before 1 hour, we are fine. If we run it every minute, it wouldn't have any advantage, just that it may prefetch it sooner, but still, it is prefetch, the token quality isn't out of date, so it is still valid.

So, having the `min` of  minute will just add CPU computation.

crates/shared/src/arguments.rs
```diff
@@ -236,6 +236,31 @@ pub struct Arguments {
     /// Override address of the balancer vault contract.
     #[clap(long, env)]
     pub balancer_v2_vault_address: Option<H160>,
+
+    /// The amount of time in seconds a classification of a token into good or
```
MartinquaXD, line 240: nit: comment is now slightly outdated since we switched to a human readable time format.

MartinquaXD (approved):

PR #2938, opened 2024-09-03
Implement manual batching of balance queries

MartinquaXD:
# Description
When we parallelized all auction building stages in https://github.com/cowprotocol/services/pull/2916 the total runtime of the stages actually increased slightly.
The current hypothesis is that the underlying RPC batching logic now batches very fast balance queries with slow signature validation queries which then causes the responses of the fast queries to be delayed significantly.

Overall this solution (manually implementing ugly batching logic) is not really scalable in terms of code complexity so I'm actually not super convinced anymore that it's really worth it (I thought it would be more straight forward to implement this).
Will think a bit more about how to best deal with these very specific optimizations.

# Changes
Manually batch balance fetch queries.

## How to test
Correctness should be validated by e2e tests and performance should be tested by running a hotfixed deployment in staging or prod to measure the real word improvements.

### DIFF
--- a/crates/shared/src/account_balances/simulation.rs
+++ b/crates/shared/src/account_balances/simulation.rs
@@ -4,10 +4,15 @@
 
 use {
     super::{BalanceFetching, Query, TransferSimulationError},
-    anyhow::Result,
-    ethcontract::{Bytes, H160, U256},
-    ethrpc::Web3,
+    anyhow::{Context, Result},
+    ethcontract::{tokens::Tokenize, Bytes, H160, U256},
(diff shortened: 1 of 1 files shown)


squadgazzz (approved):
Cool! I missed that `push` function. I hope this will have a positive impact.

squadgazzz:
With this change running in prod, there are spikes in 5-10s metric buckets.
image

MartinquaXD:
Closing since it didn't produce good results and would add quite a big of complexity.

PR #2946, opened 2024-09-05
Speed up autopilot balance fetching

squadgazzz:
# Description
Speed up autopilot balance fetching the same way it was done in the `driver` crate(#2045) in order to improve runloop performance.

# Changes
> Reimplement parts of the balance query of [balances.sol](https://github.com/cowprotocol/services/blob/main/crates/contracts/solidity/Balances.sol?rgh-link-date=2023-11-07T00%3A14%3A31Z#L59-L76) using simple eth_calls and use that for any order group that does not contain any pre-interactions.

## How to test
Existing e2e tests.

## Further steps

Driver and autopilot share the same logic inside. It probably makes sense to keep everything in the shared crate only.

### DIFF
--- a/crates/shared/src/account_balances/simulation.rs
+++ b/crates/shared/src/account_balances/simulation.rs
@@ -4,13 +4,12 @@
 
 use {
     super::{BalanceFetching, Query, TransferSimulationError},
-    anyhow::{anyhow, Result},
+    anyhow::Result,
     contracts::{erc20::Contract, BalancerV2Vault},
     ethcontract::{Bytes, H160, U256},
     ethrpc::Web3,
     futures::future,
     model::order::SellTokenSource,
-    std::collections::HashMap,
 };
 
 pub struct Balances {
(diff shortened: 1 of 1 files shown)


crates/shared/src/account_balances/simulation.rs
```diff
@@ -86,16 +144,27 @@ struct Simulation {
...
         // TODO(nlordell): Use `Multicall` here to use fewer node round-trips
         let futures = queries
             .iter()
             .map(|query| async {
-                let simulation = self.simulate(query, None).await?;
-                Ok(if simulation.can_transfer {
-                    simulation.effective_balance
+                if query.interactions.is_empty() {
+                    let token = tokens
+                        .get(&query.token)
```
MartinquaXD, line 162: By now it's no longer really expensive to create an instance of a contract so let's drop the `tokens` map and instead create every token on demand here to avoid this error handling/complexity.

MartinquaXD (commented):
Did you already measure the impact on this somehow?
Given that this resulted in significant improvements in the driver I'm hopeful that it would help here as well but would be nice to be sure.
If not I can test that some time today since I wanted to test the other performance thing as well. Let's not merge until we measured the impact of this.

Also it's a bit unfortunate that these 2 PRs are conflicting quite a bit. If your PR is already showing good results I would prefer to merge yours and rethink my PR as handrolling that optimization every time it's useful is not really feasible.

squadgazzz:
> If not I can test that some time today since I wanted to test the other performance thing as well. Let's not merge until we measured the impact of this.

No, I haven't tested it yet. It would be great if you could do that.

MartinquaXD (approved):

m-lord-renkse (approved):

MartinquaXD:
@squadgazzz IIRC this PR actually had a modest performance improvement, right?
Let's get it merged then.

PR #2941, opened 2024-09-04
Get JIT orders by tx

sunce86:
# Description
A follow up for https://github.com/cowprotocol/services/pull/2920

Implements [https://api.cow.fi/mainnet/api/v1/transactions/{tx_hash}/orders](https://api.cow.fi/docs/#/default/get_api_v1_transactions__txHash__orders) for JIT orders.

# Changes

- [ ] implemented `get_by_tx` database function to retrieve all jit orders in a transaction
- [ ] expanded orderbook api function `orders_for_tx` to return jit orders as well

## How to test
updated e2e test

PR #2948, opened 2024-09-05
Exclude orderbook orders from jit orders

sunce86:
# Description
A follow up for https://github.com/cowprotocol/services/pull/2941 that handles an edge case:

`jit_orders` database table is potentially containing:
1. normal jit orders (orders that do not exist in orderbook)
2. orders that exist in the orderbook but do not exist in the auction (so called out-of-auction settled user trades).

This PR excludes (2) so that API function `get_orders_for_tx` doesn't return duplicated values for (2), since those exist in both `orders` and `jit_orders` tables.

## How to test
Existing e2e tests.

### DIFF
--- a/crates/database/src/jit_orders.rs
+++ b/crates/database/src/jit_orders.rs
@@ -58,7 +58,11 @@ pub async fn get_by_tx(
         t.block_number = (SELECT block_number FROM settlement) AND
         -- BETWEEN is inclusive
         t.log_index BETWEEN (SELECT * from previous_settlement) AND (SELECT log_index FROM \
-         settlement) ",
+         settlement) 
+        AND NOT EXISTS (
+            SELECT 1 FROM orders ord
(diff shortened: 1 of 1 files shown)


squadgazzz:
> orders that exist in the orderbook but do not exist in the auction

Looks like I missed that part during the previous reviews because my understanding was that jit orders never persist in the orderbook.

m-lord-renkse:
But if I understood correctly, in this case we have duplicated orders in `orders` and `jit_orders` tables right?

sunce86:
> But if I understood correctly, in this case we have duplicated orders in `orders` and `jit_orders` tables right?

Yes, this was intentional, discussed here: https://github.com/cowprotocol/services/pull/2920#discussion_r1728731873

MartinquaXD (approved):

m-lord-renkse (approved):

PR #2950, opened 2024-09-05
Maybe a bit of a simplification for the incremental cache

MartinquaXD:
# Description
My attempt to hopefully simplify one function of https://github.com/cowprotocol/services/pull/2923 slightly.

# Changes
Mostly tried to condense the code a bit and make timestamp handling easier to understand.
I believe the only logical change was that I initialize the timestamp with a non-zero value on the first iteration. However, IIUC since we now use the incremental cache every time when the cache is `Some` this should cancel each other out making the whole thing a bit easier to grasp IMO.

## How to test
e2e tests still pass (tests ran fine locally but 1 test fails in CI, will investigate to see if it's the PR or a flaky test in general :/)
@squadgazzz could you give this a sanity check that it's actually not breaking any logic?


crates/autopilot/src/boundary/mod.rs
```diff
@@ -39,4 +39,5 @@ pub struct SolvableOrders {
     pub orders: HashMap<domain::OrderUid, model::order::Order>,
     pub quotes: HashMap<domain::OrderUid, domain::Quote>,
     pub latest_settlement_block: u64,
+    pub fetched_from_db: chrono::DateTime<chrono::Utc>,
```
m-lord-renkse, line 42: why not to store it in timestamp?

MartinquaXD: Do you mean why not use `std::time::Instant`? It's used in a DB query and that needs an actual timestamp string.

m-lord-renkse (approved):

crates/autopilot/src/solvable_orders.rs
```diff
@@ -340,72 +337,32 @@ impl SolvableOrdersCache {
...
+                    cache.solvable_orders.fetched_from_db,
+                    cache.solvable_orders.latest_settlement_block,
+                    min_valid_to,
+                );
+                drop(lock);
+                fetch_orders.await?
+            }
+            _ => {
+                drop(lock);
+                self.persistence.all_solvable_orders(min_valid_to).await?
```
squadgazzz, line 362: Maybe we can avoid manual `drop` calls by returning `solvable_orders_future` instead and calling `.await?` below this section to drop the lock automatically.

squadgazzz (approved):
Cool! The approach should work. If I didn't miss anything, there is a high chance of selecting duplicated orders. But that shouldn't be a problem since they will be just overridden in the cache.

crates/autopilot/src/solvable_orders.rs
MartinquaXD: Good idea!

MartinquaXD:
> If I didn't miss anything, there is a high chance of selecting duplicated orders.

Can you elaborate a bit more on that? I thought this should actually reduce the number of duplicate orders since the checkpoint timestamp is now closer to the time we call the DB so there should be less overlap.

squadgazzz:
> Can you elaborate a bit more on that? I thought this should actually reduce the number of duplicate orders since the checkpoint timestamp is now closer to the time we call the DB so there should be less overlap.

Previously, the latest order's timestamp was used. Currently, a timestamp from the previous loop is used, which was taken right **before** the previous SQL query.

before: query -> lastest_timestamp -> next loop.
now: current_timestamp -> query -> next loop.

PR #2951, opened 2024-09-05
Fix flaky test

MartinquaXD:
# Description
Another flaky test caused CI to [fail](https://github.com/cowprotocol/services/actions/runs/10723467528/job/29736827126?pr=2950) on an unrelated change (also worked locally).
The test in question fails because no trade happens. After comparing the logs of a failed CI run with a successful local run it looks like the order in question never makes it into the auction because 1 of the native prices is always missing.
The test actually sets up all the liquidity that is needed but in the failed run the solver was never able to compute a solution.
The driver correctly found the existing pool it was supposed to use but apparently the driver's view of the pool was not up-to-date.
As an optimization the driver is allowed to use stale liquidity data for quotes (but not for solving auctions). This was already the cause of flaky tests in the past and the fix was to keep mining new blocks until the driver flushes out the cached liquidity forcing it to see the latest state of the chain which then allows it to finally compute a native price.

To make the whole thing work reliably in CI I also had to use a tokio runtime with 2 worker threads.

# Changes
* mine a bunch of blocks to flush out stale cached liquidity
* also changed how we wait for the trade to happen since synchronizing on the size of the auction is prone to race * conditions
* To test the change sufficiently I added a new opt-in flaky test runner that runs 1 test in a loop to see if it works.
Also mentioned that in the readme.

## How to test
Used new flaky test runner to run the test 500 times


m-lord-renkse (approved):

.github/workflows/flaky-tests.yaml
```diff
@@ -0,0 +1,39 @@
...
+      # Start the build process in the background. The following cargo test command will automatically
+      # wait for the build process to be done before proceeding.
+      - run: cargo build -p e2e --tests &
+      - uses: taiki-e/install-action@nextest
+      - uses: yu-ichiro/spin-up-docker-compose-action@v1
+        with:
+          file: docker-compose.yaml
+          up-opts: -d db migrations
+      - name: Run test in a loop
+        run: |
```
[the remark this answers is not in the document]
MartinquaXD: I was not able to reproduce the flaky test locally so I added this opt-in flaky-test runner to reproduce it in the CI.
The plan is to first run it in CI for an hour. Only if that passes I would consider this issue resolved.
That's why I converted the PR back to a draft.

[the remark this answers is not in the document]
m-lord-renkse: oh, makes a lot of sense, got it, good idea :ok_hand:

MartinquaXD:
The new flaky test job [ran](https://github.com/cowprotocol/services/actions/runs/10735599885/job/29773243489?pr=2951) for >50 minutes (or ~500 runs) without any issue so I think it's safe to assume that the fix works.

.github/workflows/pull-request.yaml
```diff
@@ -180,3 +180,40 @@ jobs:
       - run: node_modules/.bin/swagger-cli validate crates/orderbook/openapi.yml
       - run: node_modules/.bin/swagger-cli validate crates/driver/openapi.yml
       - run: node_modules/.bin/swagger-cli validate crates/solvers/openapi.yml
+
+  run-flaky-test:
+    # to debug a flaky test set `if` to true and configure the flaky test in the `run` step
+    if: ${{ false }}
```
squadgazzz, line 186: How about using manual configs? Something like:
```
  workflow_dispatch:
    inputs:
      test_name:
        description: 'Test name to run'
        required: true
        default: 'local_node_no_liquidity_limit_order'
      test_threads:
        description: 'Number of test threads'
        required: false
        default: '1'
      retries:
        description: 'Number of retries'
        required: false
        default: '3'
```

squadgazzz: Also, `timeout-minutes` param.

MartinquaXD: Ultimately this action should only ever be used in tandem with working on a flaky test.
So since you already have to work on the PR you might as well change the parameters in the github action while working on it.
Also I don't want to over-engineer it. We don't know how often we'll actually end up using this so spending more time now to make this "perfect" is not time well spent IMO.

squadgazzz (approved):

PR #2953, opened 2024-09-06
Speed up signature check for orders with no pre-interactions

m-lord-renkse:
# Description
Speed up signature check for orders with no pre-interactions by directly calling `isValidSignature` instead of our `Signatures` contract.

# Changes
- In the cases with no pre-interactions, directly call `isValidSignature`
- Modify the eht safe e2e tests to use properly a SAFE transaction using the corresponding signature and forcing the new code to be tested/executed

## How to test
1. Acceptance tests

# Performance improvement
By looking at the [Signatures contract](https://github.com/cowprotocol/services/blob/main/crates/contracts/solidity/Signatures.sol), it is apparent that the speed it is apparent that the improvement won't be significant. We are mostly passing some small checks from the smart contract to Rust code, but the heavy operations are still done in the smart contract.

I did some speed tests **locally** (I know the numbers will differ in real production, but I wanted to get some approximation). What I did is to do `Instant::now()` + `now.elapse()` wrapping the critical code, and after many tests, and I got in average these results:
- Previous implementation: 2982196 nanoseconds
- New implementation: 2388463 nanoseconds

There is definitely an improvement. In production the improvement should be greater, because I was using anvil in localhost which reduces the latency significantly.


crates/shared/src/signature_validator/simulation.rs
```diff
@@ -18,18 +19,54 @@ pub struct Validator {
...
+        // would happen as part of a settlement
+        let contract = ERC1271SignatureValidator::at(&self.web3, check.signer);
+        let magic_bytes = contract
+            .methods()
+            .is_valid_signature(Bytes(check.hash), Bytes(check.signature.clone()))
+            .call()
+            .await
+            .map(|value| hex::encode(value.0))
+            .map_err(|e| {
+                tracing::error!(
```
m-lord-renkse, line 56: This could be a little bit noisy for the oncall rotation. But I wanted to really flag this error, because I tested it for gnosis chain, and in theory all owners with EIP-1271 must implement the `isValidSignature` method.

squadgazzz:
> Previous implementation: 2982196 nanoseconds
> New implementation: 2388463 nanoseconds
> There is definitely an improvement. In production the improvement should be greater, because I was using anvil in localhost which reduces the latency significantly.

It would be better to create an image to run it on prod for a while(like it was done with recent improvement PRs). Just be aware to checkout from the latest release tag.

crates/shared/src/signature_validator/simulation.rs
MartinquaXD: > This could be a little bit noisy for the oncall rotation

If we have an order with this error we get an alert every time we build an auction. That's not acceptable. For the purposes of this code it doesn't really matter if `isValidSignature()` returned a wrong value or the contract doesn't have the function at all. Ultimately the order is not usable so it just needs to be filtered out.

m-lord-renkse:
> > Previous implementation: 2982196 nanoseconds
> > New implementation: 2388463 nanoseconds
> > There is definitely an improvement. In production the improvement should be greater, because I was using anvil in localhost which reduces the latency significantly.
> 
> It would be better to create an image to run it on prod for a while(like it was done with recent improvement PRs). Just be aware to checkout from the latest release tag.
> 
> ~So, currently, I see the improvement will be in the network delay only, right?~ Oh, wait, we still call the contract. There won't be any improvement at all(a few nanosecs don't count IMO)?
> 
> ```
>        let magic_bytes = contract
>             .methods()
>             .is_valid_signature(Bytes(check.hash), Bytes(check.signature.clone()))
>             .call()
>             .await
> ```

Yeah, we should definitely run it on prod and see the impact, but I am afraid the impact will be so small that it won't be seen in the noise :thinking: 

The improvement is the contract we call, in the previous code, we always call a contract function which checks the signature taking into account the pre-interactions, so the smart contract checks if the pre-interactions are filled or not, and then call `isValidSignature`. So now, we call directly `isValidSignature` if the pre-interactions are empty. We move some checks from smart contract to Rust code.

Performance wise, it will be better for sure. How much? we don't know, but I assume it will reach 1 ms at most. But it is my educated guess.

MartinquaXD:
> I am afraid the impact will be so small that it won't be seen in the noise 🤔

If the impact is not measurable we shouldn't merge the code IMO. All of these optimizations are just things that **could** improve the runtime but ultimately might not for whatever reason so it's inherently a research task.

> Oh, wait, we still call the contract. There won't be any improvement at all(a few nanosecs don't count IMO)?

Mateo's benchmark showed an improvement of 20% on the request itself. If we add network latency to the mix this in itself will likely not be a significant improvement but this could have a compounding impact due to the RPC request batching which works worse if fast and slow requests happen to be in the same batch.

Ultimately we should benchmark in prod and only merge if this improves the runtime.

crates/shared/src/signature_validator/simulation.rs
```diff
@@ -18,18 +19,54 @@ pub struct Validator {
...
+                    "failed to check EIP-1271 signature with isValidSignature"
+                );
+                e
+            })?;
+
+        if magic_bytes != Self::IS_VALID_SIGNATURE_MAGIC_BYTES {
+            return Err(SignatureValidationError::Invalid);
+        }
+
+        Ok(Simulation { gas_used: 0.into() })
```
MartinquaXD, line 67: I think to avoid misuse of the function we should return a `bool` instead of a `Simulation` with an incorrect `gas_used` here.

crates/shared/src/signature_validator/simulation.rs
```diff
@@ -18,18 +19,54 @@ pub struct Validator {
...
         }
     }
 
+    /// Simulate isValidSignature for the cases in which the order does not have
+    /// pre-interactions
+    async fn simulate_without_pre_interactions(
+        &self,
+        check: &SignatureCheck,
+    ) -> Result<Simulation, SignatureValidationError> {
+        // We simulate the signature verification from the owner contract's
```
MartinquaXD, line 45: This comment does not apply to the simpler check.
IMO it would be more valuable to explain why a signature without pre-interactions doesn't need the full simulation.

m-lord-renkse:
> Ultimately we should benchmark in prod and only merge if this improves the runtime.

Yes, I agree. I already did the job, so it's worth trying it at least. I will do it on Monday (I don't want to leave it running unsupervised over the weekend) :ok_hand:

crates/shared/src/signature_validator/simulation.rs
m-lord-renkse: @MartinquaXD it would be ok if I set it as `warn`? I want something queryable I can use in elasticsearch to see how many of these cases we have.

crates/shared/src/signature_validator/simulation.rs
m-lord-renkse: Good point, agree.

m-lord-renkse:
> > Yes, I agree. I already did the job, so it's worth trying it at least. I will do it on Monday (I don't want to leave it running unsupervised over the weekend) 👌
> 
> I don't want our speed initiatives to drag on too much. I can test it out over the weekend since I'll probably be working some more as oncall anyway.

@MartinquaXD That would be lovely :pray: thank you very much!

crates/shared/src/signature_validator/simulation.rs
MartinquaXD: There is no need to introduce this IMO. The current logic already uses `SignatureValidationError::Invalid` when the call reverts which also happens when the function does not exist. Since this didn't cause any issues so far I don't see why the simpler check should have different semantics than the original one.

MartinquaXD (approved):

PR #2954, opened 2024-09-06
Separate `solvable_orders_after` query

squadgazzz:
# Description
A suggestion to unify SQL queries mentioned in the https://github.com/cowprotocol/services/pull/2923 didn't work out. Postgres converts common table expressions into subqueries, which affects the performance. Also, with this approach, indexes cease to be utilized. As a result, it executes in 1.2-1.5s instead of ~200ms.

Tried so far(nothing helped):
- JOIN with CTE instead of using `WHERE IN (rowset)`
- Use `WITH cte AS NOT MATERIALIZED`
- Replace CTE with a subquery.
- Convert CTE result into ARRAY(has the best performance, but still 2-3 times slower than separated queries).

# Changes
Revert to the separated queries that execute in ~200ms total.

## How to test
Restored DB tests.

### DIFF
--- a/crates/database/src/orders.rs
+++ b/crates/database/src/orders.rs
@@ -687,7 +687,7 @@ pub fn solvable_orders(
     sqlx::query_as(OPEN_ORDERS).bind(min_valid_to).fetch(ex)
 }
 
-pub fn open_orders_by_time_and_uids<'a>(
+pub fn open_orders_by_time_or_uids<'a>(
     ex: &'a mut PgConnection,
     uids: &'a [OrderUid],
     after_timestamp: DateTime<Utc>,
@@ -1585,17 +1585,17 @@ mod tests {
 
     #[tokio::test]
     #[ignore]
-    async fn postgres_open_orders_by_time_and_uids() {
+    async fn postgres_open_orders_by_time_or_uids() {
         let mut db = PgConnection::connect("postgresql://").await.unwrap();
         let mut db = db.begin().await.unwrap();
         crate::clear_DANGER_(&mut db).await.unwrap();
 
-        async fn get_open_orders_by_time_and_uids(
+        async fn get_open_orders_by_time_or_uids(
             ex: &mut PgConnection,
             after_timestamp: DateTime<Utc>,
             uids: &[OrderUid],
         ) -> HashSet<OrderUid> {
-            open_orders_by_time_and_uids(ex, uids, after_timestamp)
+            open_orders_by_time_or_uids(ex, uids, after_timestamp)
                 .map_ok(|o| o.uid)
                 .try_collect()
                 .await
@@ -1625,7 +1625,7 @@ mod tests {
         // Check fetching by timestamp only.
         // Early timestamp should cover all the orders.
         assert_eq!(
-            get_open_orders_by_time_and_uids(
+            get_open_orders_by_time_or_uids(
                 &mut db,
                 now - Duration::seconds(1),
                 Default::default()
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):
Interesting that postgres is not able to optimize the original query appropriately.

crates/database/src/orders.rs
```diff
@@ -795,6 +774,34 @@ pub async fn user_orders_with_quote(
...
+
+pub fn updated_order_uids_after(
+    ex: &mut PgConnection,
+    after_block: i64,
+) -> BoxStream<'_, Result<DbOrderUid, sqlx::Error>> {
+    const QUERY: &str = r#"
+SELECT DISTINCT order_uid FROM (
+    SELECT order_uid FROM trades WHERE block_number > $1
+    UNION
+    SELECT order_uid FROM order_execution WHERE block_number > $1
```
sunce86, line 788: `order_execution` not actually an input for solvable orders so I'd rather remove it.

crates/database/src/orders.rs
```diff
@@ -687,43 +687,22 @@ pub fn solvable_orders(
...
+) -> BoxStream<'a, Result<FullOrder, sqlx::Error>> {
     #[rustfmt::skip]
     const OPEN_ORDERS_AFTER: &str = const_format::concatcp!(
-        UPDATED_UIDS_QUERY,
-        " SELECT ", ORDERS_SELECT,
+        "SELECT ", ORDERS_SELECT,
         " FROM ", ORDERS_FROM,
         " LEFT OUTER JOIN ethflow_orders eth_o on eth_o.uid = o.uid ",
-        " WHERE (o.creation_timestamp > $2 OR o.cancellation_timestamp > $2 OR o.uid IN (SELECT order_uid FROM updated_uids))",
+        " WHERE (o.creation_timestamp > $1 OR o.cancellation_timestamp > $1 OR o.uid = ANY($2))",
```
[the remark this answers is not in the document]
squadgazzz: `IN` can't be used with `ANY`.

sunce86 (approved):

crates/database/src/orders.rs
squadgazzz: It is used to calculate executed surplus fee: https://github.com/cowprotocol/services/blob/267f55059464a798cf5f75a055d3c09fb85371b9/crates/database/src/orders.rs#L551

crates/database/src/orders.rs
```diff
@@ -687,43 +687,22 @@ pub fn solvable_orders(
     sqlx::query_as(OPEN_ORDERS).bind(min_valid_to).fetch(ex)
 }
 
-pub fn open_orders_after(
-    ex: &mut PgConnection,
-    after_block: i64,
+pub fn open_orders_by_time_and_uids<'a>(
```
m-lord-renkse, line 690: nit: this function is called `_by_time_and_uids`, but the query is by time OR uid? which premise is the correct one?

crates/database/src/orders.rs
sunce86: But this is not related to this query in this function right?

MartinquaXD: The `surplus_fee` is technically irrelevant for the `SolvableOrdersCache` but from the DB's perspective this logic is used to determine which orders have any data of the `FullOrder` changed so I'd say since the query would be still very vast we should keep this to avoid inconsistent state. (i.e. caching is hard and we should not cut corners unnecessarily)

PR #2956, opened 2024-09-06
Use surplus token for total fee calculation

sunce86:
# Description
`fee_in_ether` is practically only used to store result into `settlement_observations.fee` field, which is used by solver team.

We calculate the total fee in surplus token, as a first step always. Now, instead of using uniform clearing prices to convert that fee to sell token, and then into Ether using external prices, we go directly from surplus token -> Ether:

before: `surplus token -> sell token -> Ether`
after: `surplus token -> Ether`

This is just a small step of moving all fees to be expressed over surplus token and remove conversions to sell token.

@fhenneke please would like your approval first since you work with this result.

## How to test
Existing tests.

### DIFF
--- a/crates/autopilot/src/domain/settlement/mod.rs
+++ b/crates/autopilot/src/domain/settlement/mod.rs
@@ -340,9 +340,11 @@ mod tests {
             eth::U256::from(52937525819789126u128)
         );
         // fee read from "executedSurplusFee" https://api.cow.fi/mainnet/api/v1/orders/0x10dab31217bb6cc2ace0fe601c15d342f7626a1ee5ef0495449800e73156998740a50cf069e992aa4536211b23f286ef88752187ffffffff
+        // but not equal to 6890975030480504 anymore, since after this tx we switched to
+        // convert the fee from surplus token directly to ether
         assert_eq!(
             trade.fee_in_ether(&auction.prices).unwrap().0,
-            eth::U256::from(6890975030480504u128)
+            eth::U256::from(6752697350740628u128)
         );
     }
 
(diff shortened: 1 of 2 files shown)


fhenneke:
Using the surplus token for computing fees per batch will require some changes to slippage accounting. We are currently in the process of rewriting that part of slippage anyways, so merging this into prod next week would work for the solver team.

The required changes to solver payments will probably make this entry in the table obsolote. Protocol fees are computed separately anyways and for network fees (which is a part of slippage) we would then compute it separately as well.

fleupold (approved):
Code itself looks fine

crates/autopilot/src/domain/settlement/mod.rs
```diff
@@ -340,9 +340,11 @@ mod tests {
             eth::U256::from(52937525819789126u128)
         );
         // fee read from "executedSurplusFee" https://api.cow.fi/mainnet/api/v1/orders/0x10dab31217bb6cc2ace0fe601c15d342f7626a1ee5ef0495449800e73156998740a50cf069e992aa4536211b23f286ef88752187ffffffff
+        // but not equal to 6890975030480504 anymore, since after this tx we switched to
+        // convert the fee from surplus token directly to ether
         assert_eq!(
             trade.fee_in_ether(&auction.prices).unwrap().0,
-            eth::U256::from(6890975030480504u128)
+            eth::U256::from(6752697350740628u128)
```
MartinquaXD, line 347: The change to skip one intermediary step makes sense to me but that the new value is ~2% off seems concerning.
@fhenneke could you double check the math here to verify that we are now closer to the truth than before?
Looks like such a big change should be well understood.

fhenneke: This goes a bit in the direction of https://github.com/cowprotocol/services/pull/2891#issuecomment-2290828241 and it depends on what you want to do with the data to answer what an accurate value is. The surplus fee consists of protocol fees which are in the surplus token and network fees which, in the reference driver, solvers specify in the sell token. All approaches which do not distinguish between those different fees make some error when converting to ETH. Since protocol fees are more fundamental to the mechanism (and not just an implementation detail of the reference driver), keeping everything in the surplus token and converting to ETH directly sounds cleaner.

On some level it might be better to just remove the fee entry in the settlement_observations table since we plan on not using that value in the future. (Not sure if we need any value in that table anymore.)

MartinquaXD (approved):

PR #2958, opened 2024-09-06
Refactor runloop slightly

MartinquaXD:
# Description
Mostly a refactor with a few changes to improve observability.

# Changes
Refactors:
* move fairness checks and solutions sorting into `competition` to have it return the final state of the competition
* combine a bunch of logic to report on the proposed solutions into `report_on_solutions()`
* rename `solve()` to `try_solve()` and introduce a new function `solve()` that does all the error handling to make `competition()` leaner

Observability:
* `info` log whenever we see a new block
* log which solver proposed a solution for which order (to have a holistic view on the auction in the autopilot without having to stitch together driver logs
* record `Executed` order events together with `Considered` as suggested [here](https://github.com/cowprotocol/services/pull/2921#discussion_r1728621248) to have a consistent view in the DB

## How to test
e2e tests should still pass

### DIFF
--- a/crates/autopilot/src/domain/competition/mod.rs
+++ b/crates/autopilot/src/domain/competition/mod.rs
@@ -41,7 +41,7 @@ impl SolutionWithId {
         self.solution.score()
     }
 
-    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> {
+    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> + std::fmt::Debug {
         self.solution.order_ids()
     }
 
@@ -85,7 +85,7 @@ impl Solution {
         self.score
     }
 
-    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> {
+    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> + std::fmt::Debug {
         self.orders.keys()
     }
 
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -451,6 +451,7 @@ impl RunLoop {
             tracing::debug!(
                 driver=%solution.driver.name,
                 orders=?solution.solution.order_ids(),
+                solution=solution.solution.id(),
                 "proposed solution"
             );
         });
--- a/crates/autopilot/src/run_loop.rs
+++ b/crates/autopilot/src/run_loop.rs
@@ -468,10 +468,12 @@ impl RunLoop {
             .difference(&winning_orders)
             .cloned()
             .collect();
+        self.persistence.store_order_events(
+            non_winning_orders.iter().cloned(),
+            OrderEventLabel::Considered,
+        );
         self.persistence
-            .store_order_events(non_winning_orders.clone(), OrderEventLabel::Considered);
-        self.persistence
-            .store_order_events(winning_orders.clone(), OrderEventLabel::Executing);
+            .store_order_events(winning_orders, OrderEventLabel::Executing);
 
         let auction_uids = auction.orders.iter().map(|o| o.uid).collect::<HashSet<_>>();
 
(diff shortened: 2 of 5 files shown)


crates/autopilot/src/domain/competition/mod.rs
```diff
@@ -41,7 +41,7 @@ impl SolutionWithId {
         self.solution.score()
     }
 
-    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> {
+    pub fn order_ids(&self) -> impl Iterator<Item = &domain::OrderUid> + std::fmt::Debug {
```
m-lord-renkse, line 44: By doing this we are enforcing that the `Iterator` type implements `Debug`. If the need to change the underlying iterator to add a `map` or `filter`, this code will fail. Another solution would be adding `.collect::<Vec<_>>()` to the debug line, but it allocates...

crates/autopilot/src/run_loop.rs
```diff
@@ -450,45 +408,77 @@ impl RunLoop {
...
+            .iter()
+            .flat_map(|solution| solution.solution.order_ids().copied())
+            .collect();
+        let winning_orders: HashSet<_> = solutions
+            .last()
+            .into_iter()
+            .flat_map(|solution| solution.solution.order_ids().copied())
+            .collect();
+        let mut non_winning_orders: HashSet<_> = proposed_orders
+            .difference(&winning_orders)
```
m-lord-renkse, line 468: it would be faster if we just get the vector from 0 to `last - 2`, instead of calculating the difference :thinking:

crates/autopilot/src/run_loop.rs
```diff
@@ -450,45 +408,77 @@ impl RunLoop {
...
+            .cloned()
+            .collect();
+        self.persistence
+            .store_order_events(non_winning_orders.clone(), OrderEventLabel::Considered);
+        self.persistence
+            .store_order_events(winning_orders.clone(), OrderEventLabel::Executing);
```
m-lord-renkse, line 474: Unnecessary clone

crates/autopilot/src/run_loop.rs
```diff
@@ -450,45 +408,77 @@ impl RunLoop {
...
+
+    /// Records metrics, order events and logs for the given solutions.
+    /// Expects the winning solution to be the last in the list.
+    fn report_on_solutions(&self, solutions: &[Participant<'_>], auction: &domain::Auction) {
+        let Some(winner) = solutions.last() else {
+            // no solutions means nothing to report
+            return;
+        };
+
+        solutions.iter().for_each(|solution| {
```
m-lord-renkse, line 450: This would create a log per solution, wouldn't it better to have all the solutions in the same log? that way we have them all together and not scattered with the other logs, easier to find.

crates/autopilot/src/run_loop.rs
MartinquaXD: The idea is to only count orders that have not also been proposed by the winning solution. In case a non-winning solution would also solve an order that the winning solution solved your approach would produce false positives.

m-lord-renkse: Oh, I understand, thanks!

crates/autopilot/src/domain/competition/mod.rs
MartinquaXD: The iterator returned by `.filter()` and `.map()` still implements `Debug` so this would not cause issues.

crates/autopilot/src/run_loop.rs
MartinquaXD: I think there are trade-offs to both. For example searching something like `(log: "oneinch" or log: "zeroex") and log: "<UID>"` would be way nicer with individual logs IMO.
Given that the grouped log is more complex to implement and it's not a clear usability improvement IMO I would prefer to leave it as is until it turns out to be annoying in practice.

squadgazzz (approved):
LGTM

PR #2959, opened 2024-09-06
Fix some flaky `local_node` e2e tests

MartinquaXD:
# Problem 1
[flaky test](https://github.com/cowprotocol/services/actions/runs/10742540060/job/29795363509)

Current hypothesis:
There appears to be a race condition that causes the incremental orders cache to sometimes miss new orders which means they can't get settled in the test.
Consider this case:
1. orderbook receives POST order request
2. orderbook computes the current timestamp
3. orderbook spends some time doing extra checks
4. autopilot updates open orders with current timestmp
5. autopilot cache checkpoint is now AFTER order creation timestamp
6. orderbook finally adds the order to the DB (which now has creation timestamp **before** last DB checkpoint)
7. autopilot updates open orders but only checks for orders created **after** `5` which means it will never see the new order

I already tried to let postgres populate the order's `creation_timestamp` with `now()` when executing the insertion but that still didn't resolve the issues. This might be caused by the rust code and postgres reporting different timestamps for the exact same point in time. But maybe it could be related to how DB transactions might have inconsistent views on the tables. 🤷‍♂️ 

I ended up resolving the issue by moving the DB checkpoint timestamp (of the incremental orders cache) 1 minute into the past creating a buffer period which will mitigate race conditions due to timestamp issues. This means we'll re-fetch orders created or cancelled in the last minute more often than necessary but given that the query is way faster now this should not be an issue.

# Problem 2
[Flaky test](https://github.com/cowprotocol/services/actions/runs/10746259879/job/29806839501?pr=2959#logs) 

The issue seems to be that the DB can sometimes deadlock when multiple transactions are waiting for each other. Luckily postgres is able to detect those deadlocks and simply cancels on of the transactions. If the transaction to truncate the tables fails we simply retry it until it works.

# Problem 3
Native price cache expiry is 30s by default which is sometimes too long for tests which require a price for a token to first be `NoLiquidity` and an actual price later on (default wait timeout is also 30s). The simple fix is to reduce the native price cache validity.

# Problem 4
[flaky test](https://github.com/cowprotocol/services/actions/runs/10750833964/job/29817436705?pr=2959)

Many tests are synchronized by waiting for the current auction to reach a specific number of orders. This is very prone to race conditions so I updated those tests to use safer ways to synchronize on the events.
I also added the ability to return `Option<bool>` and `Result<bool>` from the future we try to await to make some of the code more ergonomic.

## How to test
Ran flaky test runner for a long while ([41 successful runs](https://github.com/cowprotocol/services/actions/runs/10753333251/job/29822651957?pr=2959) for all tests). It still encountered an error but I was not actually able to reproduce this error but running only that error in a loop ([>200 tries](https://github.com/cowprotocol/services/actions/runs/10754405993/job/29824987289?pr=2959)). It looks to me that some errors have a higher likelihood if you run them together with a bunch of other test over an over. But since this is not how we run our tests for real I think I'll call it a day here.

### DIFF
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -173,6 +173,7 @@ impl SolvableOrdersCache {
         let start = Instant::now();
 
         let db_solvable_orders = self.get_solvable_orders().await?;
+        tracing::info!(orders = ?db_solvable_orders.orders.keys());
 
         let orders = db_solvable_orders
             .orders
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -173,7 +173,6 @@ impl SolvableOrdersCache {
         let start = Instant::now();
 
         let db_solvable_orders = self.get_solvable_orders().await?;
-        tracing::info!(orders = ?db_solvable_orders.orders.keys());
 
         let orders = db_solvable_orders
             .orders
--- a/crates/autopilot/src/solvable_orders.rs
+++ b/crates/autopilot/src/solvable_orders.rs
@@ -302,6 +302,7 @@ impl SolvableOrdersCache {
                 .collect::<Result<_, _>>()?,
             surplus_capturing_jit_order_owners,
         };
+
         *self.cache.lock().await = Some(Inner {
             auction,
             solvable_orders: db_solvable_orders,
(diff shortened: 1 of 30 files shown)


crates/autopilot/src/solvable_orders.rs
```diff
@@ -360,7 +361,15 @@ impl SolvableOrdersCache {
...
 
-        fetch_orders.await
+        let mut orders = fetch_orders.await?;
+
+        // Move the checkpoint slightly back in time to mitigate race conditions
+        // caused by inconsistencies of stored timestamps. See #2959 for more details.
+        // This will cause us to fetch orders created or cancelled in the buffer
+        // period multiple times but that is a small price to pay for not missing
+        // orders.
+        orders.fetched_from_db -= chrono::TimeDelta::seconds(60);
```
squadgazzz, line 371: Detaching this logic from the actual value creation might lead to errors in the future.

squadgazzz (approved):
Awesome! Due to the missing orders issue, this should be a part of the upcoming release.

crates/autopilot/src/solvable_orders.rs
MartinquaXD: It was either this or already decreasing the value in 2 different places in the infra code.
Ultimately this seemed like a domain logic decision to me so I put it here.

sunce86 (approved):
Lg. Maybe split the PR into several fixes next time if splittable

MartinquaXD:
> Maybe split the PR into several fixes next time if splittable

Yeah, that's fair. Wanted to get some sort of confidence that I squashed at least the majority of issues. But once I was "done" I should have gone over the individual things and made proper PRs out of them.

PR #2965, opened 2024-09-10
[PROPOSAL] Event indexer counters

squadgazzz:
# Description
During one of the bugfixes(https://github.com/cowprotocol/services/pull/2931) it was noticed that some of the event handlers might spend a lot of time and resources on each restart due to the absence of the corresponding items in the DB. In that case, preconfigured start block numbers are used that, in most cases, are too old. As a result, on each restart, a background task might fetch millions of blocks on networks with a short round time(such as arbitrum one).

# Changes
As a proposal, store indexer counters in a separate table each time a portion of new blocks are processed. It is currently impossible to use existing `append_events` function since it operates with pre-filtered events only that can never happen if we don't have a single event for the specified type on the chain.

Driver crate doesn't have access to the DB currently, which will be required in case this proposal is accepted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4450,6 +4450,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "sqlx",
  "strum",
  "tempfile",
  "testlib",
--- a/crates/autopilot/src/boundary/events/settlement.rs
+++ b/crates/autopilot/src/boundary/events/settlement.rs
@@ -2,7 +2,11 @@ use {
     crate::{database::Postgres, on_settlement_event_updater::OnSettlementEventUpdater},
     anyhow::Result,
     ethrpc::block_stream::RangeInclusive,
-    shared::{event_handling::EventStoring, impl_event_retrieving},
+    shared::{
+        event_handling::{EventStoring, PgEventCounter},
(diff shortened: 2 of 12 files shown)


MartinquaXD:
I think the idea of storing which block was indexed (even if it didn't contain an event we are interested in) makes a lot of sense.
However, instead of solving this problem ourselves we should consider researching if this was not already solved better by other people.
Also the driver requiring access to the DB would not be great. But there are ways around this which we might have to implement sooner rather than later anyway.

squadgazzz:
> However, instead of solving this problem ourselves we should consider researching if this was not already solved better by other people.

We can use other indexers, such as The Graph, but introducing a new dependency and integrating it into the system seems like an overkill for the output of the whole activity.

> Also the driver requiring access to the DB would not be great. But there are ways around this which we might have to implement sooner rather than later anyway.

The same applies to the persistence layer. Bringing more dependencies, even with lightweight/embedded solutions, seems like a much bigger change than adding a small table. For the driver specifically, either add a URL to the existing PG DB or a completely new embedded DB dependency with k8s volumes within will be storing a few counter, where I would still be more inclined with the former.

We can probably start with the PG table and, if we have time, integrate The Graph or an embedded DB.

Wdyt?

PR #2966, opened 2024-09-10
Executed total fee token

sunce86:
# Description
Implements (1) and (2) from https://github.com/cowprotocol/services/issues/2963.
Also, renames `surplus_fee` to `executed_fee` in database.

# Changes

- [ ] Added new field `order_execution.executed_fee_token` to database
- [ ] Populated this field with order sell token, otherwise default value is 0x0
- [ ] Token is exposed over API the same way as `executed_fee`.

Note: all places where sell token needs to be substituted with surplus token are marked with `TODO surplus token` and this will be tackled as part of (3) from https://github.com/cowprotocol/services/issues/2963.

## How to test
Existing e2e tests. Observed values during the tests.


crates/database/src/order_execution.rs
```diff
@@ -18,25 +18,26 @@ pub async fn save(
...
 ON CONFLICT (order_uid, auction_id)
-DO UPDATE SET reward = $3, surplus_fee = $4, block_number = $5, protocol_fee_tokens = $6, protocol_fee_amounts = $7
+DO UPDATE SET reward = $3, total_fee = $4, total_fee_token = $5, block_number = $6, protocol_fee_tokens = $7, protocol_fee_amounts = $8
 ;"#;
     sqlx::query(QUERY)
         .bind(order)
         .bind(auction)
         .bind(0.) // reward is deprecated but saved for historical analysis
-        .bind(Some(executed_fee))
+        .bind(Some(total_fee.amount))
```
MartinquaXD, line 39: Why do we have an `Option` here? Do we still need that?

crates/database/src/order_execution.rs
```diff
@@ -123,17 +124,30 @@ mod tests {
...
+            Asset {
+                amount: BigDecimal::from(3),
+                token: Default::default(),
+            },
             protocol_fees.as_slice(),
         )
         .await
         .unwrap();
 
         // save entry without protocol fees (simulate case when we are still not
```
MartinquaXD, line 136: Is this case still relevant?

database/README.md
```diff
@@ -175,7 +175,8 @@ Protocol fee tokens/amounts are stored in the same order as fee policies in fee_
  order\_uid             | bytea     | not null | which order this trade execution is related to
  auction\_id            | bigint    | not null | in which auction this trade was initiated
  reward                 | double    | not null | revert adjusted solver rewards, deprecated in favor of [CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f)
- surplus\_fee           | numeric   | nullable | dynamic fee computed by the protocol that should get taken from the surplus of a trade, this value only applies and is set for fill-or-kill limit orders.
+ surplus\_fee           | numeric   | not null | total fee taken for execution of the trade
```
MartinquaXD, line 178: This column gets renamed in the migration.

crates/orderbook/openapi.yml
```diff
@@ -881,11 +881,16 @@ components:
             can be found in this object.
           allOf:
             - $ref: "#/components/schemas/OnchainOrderData"
-        executedSurplusFee:
-          description: Surplus fee that the limit order was executed with.
+        executedTotalFee:
```
MartinquaXD, line 884: The `total` in the name seems a bit misleading to me.
I think most people would assume that this number would never increase once populated since it's already "total". But for partially fillable orders it would increase on every fill.
Do you think it would be less misleading if we just call it `executed_fee`?

crates/database/src/order_execution.rs
sunce86: I assume it's because when the field was created in the table it was marked as nullable, because it was NULL for market orders.
Now we always have the value.

crates/database/src/order_execution.rs
sunce86: fixed the comment. I think it's still valuable to keep this as part of the test.

crates/database/src/orders.rs
```diff
@@ -560,7 +561,8 @@ array(Select (p.target, p.value, p.data) from interactions p where p.order_uid =
     where eth_o.uid = o.uid limit 1) as ethflow_data,
 (SELECT onchain_o.sender from onchain_placed_orders onchain_o where onchain_o.uid = o.uid limit 1) as onchain_user,
 (SELECT onchain_o.placement_error from onchain_placed_orders onchain_o where onchain_o.uid = o.uid limit 1) as onchain_placement_error,
-COALESCE((SELECT SUM(surplus_fee) FROM order_execution oe WHERE oe.order_uid = o.uid), 0) as executed_surplus_fee,
+COALESCE((SELECT SUM(total_fee) FROM order_execution oe WHERE oe.order_uid = o.uid), 0) as executed_total_fee,
```
sunce86, line 564: @cowprotocol/backend this will be a bug in case of multiple fulfillments of partially fillable orders where some fills are executed with fee in sell token while others are executed with fee in surplus token .
Should I have in API:
executed_total_fee[]
executed_total_fee_token[]

to fix this, or should we just live with this error for these orders that are filled during a switching period?

MartinquaXD: The error case you are worried about is this, right?
1. partial fill for order `O` (fee stored in sell token)
2. merge PR to compute all fees in surplus token
3. another partial fill for `O` (fee stored in surplus token)
4. fetch `total_fee` for `O` results in a mix of fees from surplus and sell token

How big of a change is storing everything in surplus token? If it's not too much we could merge this and the other PR in the same week so we only have the issue on staging. I think I would be fine with that.

Alternatively would it be possible to go back over the corrupted data and patch it up?

sunce86: > The error case you are worried about is this, right?

Yes.

> How big of a change is storing everything in surplus token? If it's not too much we could merge this and the other PR in the same week so we only have the issue on staging.

We still have the same issue on prod since at the time we apply the weekly release, there are open partially fillable orders that had some fulfillments with fee in sell token and will probably have more after weekly release with fee in surplus token.

>  Alternatively would it be possible to go back over the corrupted data and patch it up?

Not impossible, but I don't see any easy way to do it, since we don't have uniform clearing prices for those trades in db (we need web3 data for it or maybe solver competition table). But such a sql migration is probably complicated and error prone. Let me try to compile it.

sunce86: Other alternative is this:
1. Merge this pr and pr for switching to fee in surplus token
2. Sql migration to go over order_executions for each order where there are entries that differ in token
3. For those, set token to surplus (or to the entry with the highest auction_id aka latest), but also set amount to be equal to protocol fee sum (which is already in surplus token). 

This way we set network fee to 0 for partial fils that were done before release, which is somewhat acceptable due to log gas fees recently and given the number of affected rows.

crates/orderbook/openapi.yml
sunce86: I thought of `executed_fee` but I am not too happy with none of them. `Executed_fee` seems to general, and one could ask himself `which one`. But if you prefer that one while I'm undecided, then it's probably worth to rename it.

crates/database/src/orders.rs
sunce86: I think we definitely don't want to have executedFee[] in API, so the state of the PR as is, is fine. This bug should be tackled together as part of the switch PR.

crates/database/src/order_execution.rs
sunce86: resolved

crates/autopilot/src/domain/settlement/trade/math.rs
```diff
@@ -134,9 +134,15 @@ impl Trade {
 
     /// Total fee (protocol fee + network fee). Equal to a surplus difference
     /// before and after applying the fees.
-    pub fn fee_in_sell_token(&self) -> Result<eth::SellTokenAmount, Error> {
+    ///
+    /// Denominated in SELL token
```
squadgazzz, line 138: Do we need this clarification here since the function name is self-explanatory already?

database/sql/V072__total_fee_token.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE order_execution
+RENAME COLUMN surplus_fee TO total_fee;
```
MartinquaXD, line 2: new column is still called `total_fee` instead of `executed_fee` in the migration and `readme.md`.

database/sql/V072__total_fee_token.sql
```diff
@@ -0,0 +1,24 @@
...
+
+-- Add a new column to the order_execution table to store the total fee token
+ALTER TABLE order_execution
+ADD COLUMN total_fee_token bytea NOT NULL DEFAULT '\x0000000000000000000000000000000000000000';
+
+-- Now populate existing rows with the sell token taken from the 'orders' table, or if it doesn't exist, try from 'jit_orders'.
+UPDATE order_execution oe
```
MartinquaXD, line 9: Did you test that this query does not leave any `surplus_token` uninitialized?

sunce86:
Will be worked on in Q4.

crates/model/src/order.rs
```diff
@@ -677,6 +677,9 @@ pub struct OrderMetadata {
     pub executed_fee_amount: U256,
     #[serde_as(as = "HexOrDecimalU256")]
     pub executed_surplus_fee: U256,
+    #[serde_as(as = "HexOrDecimalU256")]
+    pub executed_fee: U256,
+    pub executed_fee_token: H160,
```
squadgazzz, line 682: Will `executed_surplus_fee` be removed in the next PR?

squadgazzz (approved):
LGTM

database/sql/V072__total_fee_token.sql
sunce86: Here are test results on mainnet prod db:

Number of rows in `order_execution`: 1466450
Number of rows being updated: 1460019

So, there are ~6k entries in `order_execution` that can't be found in `orders` or `jit_orders` (but all missing orders are older than 2 months).
My assumptions is that those are some jit orders from before we started saving them in the library. Example:
https://api.cow.fi/mainnet/api/v1/solver_competition/by_tx_hash/0x59ddfcab4910ad48b3550658f0c46c8892f114e6791dbc032c05a42c1e858b63

Note how the order `https://explorer.cow.fi/search/0x0088e45fff8487b9d4c5edcf2ad6580cdfc3a689c3c3022d7282aebf6446a8a99bd702e05b9c97e4a4a3e47df1e0fe7a0c26d2f166ff4487` can't be found in the explorer.
So, for these, I am not too worried if their token is not populated since those can't be found in explorer anyway.

crates/model/src/order.rs
sunce86: Yes, we want to remove `executed_surplus_fee`, but we can't do it in this PR because frontend needs time to switch to using `executed_fee`.

PR #2967, opened 2024-09-10
Autopilot: fetch only relevant quotes

squadgazzz:
# Description
The latest prod release showed that quotes fetching is one of the most expensive operations in the incremental solvable orders cache update.

# Changes
This can be slightly improved by fetching quotes for only newly created and placed onchain orders

## How to test
Existing tests.

### DIFF
--- a/crates/autopilot/src/infra/persistence/mod.rs
+++ b/crates/autopilot/src/infra/persistence/mod.rs
@@ -405,6 +405,7 @@ impl Persistence {
     pub async fn solvable_orders_after(
         &self,
         mut current_orders: HashMap<domain::OrderUid, model::order::Order>,
+        mut current_quotes: HashMap<domain::OrderUid, domain::Quote>,
         after_timestamp: DateTime<Utc>,
         after_block: u64,
         min_valid_to: u32,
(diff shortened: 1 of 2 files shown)


crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -542,12 +494,46 @@ impl Persistence {
...
+                .await?
+                .into_iter()
+                .filter_map(|quote| {
+                    let order_uid = domain::OrderUid(quote.order_uid.0);
+                    dto::quote::into_domain(quote)
+                        .map_err(|err| {
+                            tracing::warn!(?order_uid, ?err, "failed to convert quote from db")
+                        })
+                        .ok()
+                        .map(|quote| (order_uid, quote))
```
MartinquaXD, line 523: nit: I think this makes it slightly easier to understand what's happening here due to named variables. No strong preference, though.
But I think we should use `inspect_err()` here since it's more idiomatic as we discard the error anyway.
```suggestion
                let quote = dto::quote::into_domain(quote)
                    .inspect_err(|err| {
                        tracing::warn!(?order_uid, ?err, "failed to convert quote from db")
                    })
                    .ok()?;
                Some((order_uid, quote))
```

crates/autopilot/src/infra/persistence/mod.rs
```diff
@@ -542,12 +494,46 @@ impl Persistence {
...
+                })
+                .collect()
+        };
+
+        current_quotes.retain(|uid, _| current_orders.contains_key(uid));
+
+        for (order_uid, quote) in new_quotes {
+            current_quotes.insert(order_uid, quote);
+        }
```
MartinquaXD, line 532: We could avoid this loop and `new_quotes` temp variable by:
1. moving `current_quotes.retain()` before quote fetching block
2. using `current_quotes.extend(new_quotes_iter)` (to avoid collecting the transformed quotes)

MartinquaXD (commented):
Logic change looks good. Just some nits.

MartinquaXD (approved):

PR #2970, opened 2024-09-11
Improve performance of fee policies and fees

sunce86:
# Description
Apparently, Postgresql optimal way of matching data against large input is using INNER JOIN:
https://stackoverflow.com/questions/24647503/performance-issue-in-update-query

So, the benchmark for the problematic [order](https://aws-es.cow.fi/_dashboards/app/data-explorer/discover#?_a=(discover:(columns:!(log,log_level),isDirty:!f,sort:!()),metadata:(indexPattern:'86e4a5a0-4e4b-11ef-85c5-3946a99ed1a7',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2024-09-10T11:46:17.292Z',to:'2024-09-10T13:48:25.292Z'))&_q=(filters:!(),query:(language:kuery,query:'log:%20%229649c7739ced0ea0fb0ecd8c41c81100%22'))) on arbitrum is:
1. Current implementation: ~1.5s - 3s
2. WHERE .. IN: same as (1)
3. INNER JOIN: 100ms-300ms
4. grouping by order_uid: same as (3)

That said, I chose INNER JOIN as the cleanest solution in terms of code.

Total time spent to get trades should be at most 500ms after this change (150ms on average).

## How to test
Tested locally against arbitrum read replica.
Also tested with a regular order that has only one trade and this code does not degrade performance for that case.

### DIFF
--- a/crates/database/src/fee_policies.rs
+++ b/crates/database/src/fee_policies.rs
@@ -63,22 +63,22 @@ pub async fn fetch_all(
         return Ok(HashMap::new());
     }
 
-    let mut query_builder = QueryBuilder::new("SELECT * FROM fee_policies WHERE ");
+    let mut query_builder = QueryBuilder::new("SELECT * FROM fee_policies fp INNER JOIN (VALUES ");
     for (i, (auction_id, order_uid)) in keys_filter.iter().enumerate() {
         if i > 0 {
-            query_builder.push(" OR ");
+            query_builder.push(", ");
         }
         query_builder
             .push("(")
-            .push("auction_id = ")
-            .push_bind(auction_id)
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Thanks for fixing 👏

MartinquaXD (approved):
Cool. 👍 
Worth to point out that this is using a postgres specific feature. But given that we can also change it to an equally well performing SQL standard solution I'm not too worried about that.

MartinquaXD:
Would be great to hotfix the prod release with this.

squadgazzz (approved):
👏

PR #2973, opened 2024-09-11
Replace auction using a single SQL query

squadgazzz:
Alternative to https://github.com/cowprotocol/services/pull/2972 to be tested in order to speed up the auction replacement process.

- Replaces `TRUNCATE` with `DELETE` which should be more performant for small tables
- Uses a single combined SQL query.

> [!IMPORTANT]
> Should not be merged until a positive impact is confirmed.

### DIFF
--- a/crates/autopilot/src/database/auction.rs
+++ b/crates/autopilot/src/database/auction.rs
@@ -104,10 +104,8 @@ impl Postgres {
             .start_timer();
 
         let data = serde_json::to_value(auction)?;
-        let mut ex = self.pool.begin().await?;
-        database::auction::delete_all_auctions(&mut ex).await?;
-        let id = database::auction::save(&mut ex, &data).await?;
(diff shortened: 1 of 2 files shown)


squadgazzz:
Shows better results.
Sepolia(~17:47 https://github.com/cowprotocol/services/pull/2972 deployed, ~18:06 this PR deployed):
image

Prod(deployment at ~18:08):
image

MartinquaXD:
Interesting that this had way less of an impact on prod. Maybe the auction size is a very significant factor here?

squadgazzz:
> Interesting that this had way less of an impact on prod. Maybe the auction size is a very significant factor here?

I assume so. The size of the json on prod is almost 1MB.

MartinquaXD (approved):

sunce86 (approved):
Code lg.

Issue #2975, opened 2024-09-12
feat: Update database::settlements::get_hash_by_auction_id()


sunce86:
### Problem
Currently the function tries to find a single transaction hash for an auction, instead of multiple of them.

### Suggested solution
Refactor the function and dependent code to potentially return a Vec of transaction hashes.

PR #2976, opened 2024-09-12
Get_hashes_by_auction_id implementation for multiple settlements

sunce86:
# Description
Fixes https://github.com/cowprotocol/services/issues/2975

# Changes

- [ ] ...
- [ ] ...


crates/autopilot/src/run_loop.rs
```diff
@@ -682,14 +683,14 @@ impl RunLoop {
 
             match self
                 .persistence
-                .find_tx_hash_by_auction_id(auction_id)
+                .find_settlement_transactions(auction_id)
                 .await
             {
-                Ok(Some(hash)) => return Ok(hash),
+                Ok(hashes) if hashes.is_empty() => {}
+                Ok(hashes) => return Ok(hashes.into_iter().next().expect("non-empty")),
```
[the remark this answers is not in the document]
sunce86: Yes, so that the behaviour is equivalent to the current one when a single tx is expected.

squadgazzz, line 690: How about avoid using code that panics?
```suggestion
                Ok(hashes) => {
                    if let Some(hash) = hashes.into_iter().next() {
                        return Ok(hash);
                    }
                }
```

squadgazzz (approved):

crates/autopilot/src/run_loop.rs
sunce86: Why not, whatever makes us sleep better at night :)

crates/autopilot/src/run_loop.rs
```diff
@@ -682,14 +683,18 @@ impl RunLoop {
...
             match self
                 .persistence
-                .find_tx_hash_by_auction_id(auction_id)
+                .find_settlement_transactions(auction_id)
                 .await
             {
-                Ok(Some(hash)) => return Ok(hash),
+                Ok(hashes) if hashes.is_empty() => {}
+                Ok(hashes) => {
+                    if let Some(hash) = hashes.into_iter().next() {
```
MartinquaXD, line 691: This looks quite ugly. Could be a bit nicer with:
```rust
            match self
                .persistence
                .find_settlement_transactions(auction_id)
                .await
                .map(|mut hashes| hashes.pop())
            {
                Ok(Some(tx_hash)) => return Ok(tx_hash),
                Ok(None) => {},
                Err(err) => {
                    tracing::warn!(?err, "failed to fetch recent settlement tx hashes");
                }
            }
```

crates/autopilot/src/database/recent_settlements.rs
```diff
@@ -1,19 +1,19 @@
 use {anyhow::Context, primitive_types::H256};
 
 impl super::Postgres {
-    pub async fn find_tx_hash_by_auction_id(
-        &self,
-        auction_id: i64,
-    ) -> anyhow::Result<Option<H256>> {
+    pub async fn find_settlement_transactions(&self, auction_id: i64) -> anyhow::Result<Vec<H256>> {
```
MartinquaXD, line 4: Looks like this function might as well be inlined in the `infra` module to get rid of the file.

fleupold (commented):
The PR is missing a meaningful description and I'm not sure it's actually fulfilling the desired logic (or at least not fully). Once we submit multiple solutions we want all of them to succeed before considering `wait_for_settlement_transaction` done and continuing the runloop. Now we would wait for one to get included and move on which seems incorrect.

sunce86:
> The PR is missing a meaningful description and I'm not sure it's actually fulfilling the desired logic (or at least not fully). Once we submit multiple solutions we want all of them to succeed before considering `wait_for_settlement_transaction` done and continuing the runloop. Now we would wait for one to get included and move on which seems incorrect.

The scope of the PR is limited. It's not supposed to adjust the /settle logic. We have a separate task for that in https://github.com/cowprotocol/services/issues/2830#issuecomment-2346083664

PR #2978, opened 2024-09-12
Update autopilot runloop metric buckets

squadgazzz:
The autopilot runloop metric values change over time by introducing new improvements; some of the buckets need to be adjusted to be informative enough.


crates/autopilot/src/run_loop.rs
```diff
@@ -771,12 +771,12 @@ struct Metrics {
...
 
     /// Total time spent in a single run of the run loop.
-    #[metric(buckets(0, 1, 5, 10, 15, 20, 25, 30, 35, 40))]
+    #[metric(buckets(0, 3, 6, 9, 12, 15, 18, 21, 24, 27, 30, 33, 36, 39, 42, 45, 48))]
     single_run_time: prometheus::Histogram,
 
     /// Time difference between the current block and when the single run
     /// function is started.
-    #[metric(buckets(0.5, 1, 1.5, 2, 4, 6, 8, 10, 12, 16))]
+    #[metric(buckets(0, 0.25, 0.5, 0.75, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10))]
```
MartinquaXD, line 779: Since enabling the syncrhonized run mode we never data above 6s. Let's remove 7-10 and add 1.5 and 2.5 to have a higher resolution in the time ranges we actually want to be.

MartinquaXD (approved):

